PR #88, opened 2026-02-19
Add guides for the fetchlater use cases

### LINKED ISSUE #29: fetchLater
The fetchLater() method requests a deferred fetch sent at an unknown time. The browser chooses a reliable time to send the request, ideally when the document is unloaded, and ignores the response. This API is useful for sending beacons to a server without expecting a particular response.

### MDN Documentation
- [Fetch API](https://developer.mozilla.org/docs/Web/API/Fetch_API)

### Chrome Documentation

<details>
<summary>View 7 articles</summary>

- https://developer.chrome.com/en/blog/chrome-80-deps-rems
- https://developer.chrome.com/en/blog/extended-lifetime-shared-workers-origin-trial
- https://developer.chrome.com/en/blog/fetch-later-api-origin-trial
- https://web.dev/en/articles/imperative-caching-guide
- https://web.dev/en/articles/network-error-logging
- https://web.dev/en/articles/sending-messages-with-web-push-libraries
- https://web.dev/en/learn/forms/javascript
</details>

---
Feature ID: fetchlater

**Use cases:**
- [x] #139
- [x] #141

philipwalton:
Fixes #29 

Progress on https://github.com/GoogleChrome/guidance/issues/139 and https://github.com/GoogleChrome/guidance/issues/141


guides/performance/batch-analytics-events/expectations.md
```diff
@@ -0,0 +1,9 @@
+# Expectations: `batch-analytics-events`
```
philipwalton: @paulirish @micahjo7 can you comment on whether these expectations are comprehensive enough? E.g. do we need an expectation that the API is called with the correct arguments for the use case?

guides/performance/full-session-analytics/expectations.md
```diff
@@ -0,0 +1,7 @@
+# Expectations: `full-session-analytics`
```
philipwalton: @paulirish @micahjo7 same question here...

guides/performance/batch-analytics-events/expectations.md
```diff
@@ -0,0 +1,9 @@
...
+
+- The `fetchLater()` API is invoked with a URL string as the first argument, and a `DeferredRequestInit` object as the second argument.
+- The `fetchLater()` API is the only API that should be used to send beacons. Other APIs like `fetch()`, `sendBeacon()`, `XMLHttpRequest`, or `new Image()` should not used.
+- `fetchLater()` should be invoked with the `activeAfter` option set.
+- Multiple invocations of `fetchLater()` withing the `activateAfter` time window should be batched into a single request (e.g. prior calls should be aborted).
+- Batching should be limited in some way to prevent starvation or quota overflow.
```
rviscomi, line 5: I'd be curious to see if this expectation is specific enough to auto-generate the corresponding eval

guides/performance/full-session-analytics/expectations.md
```diff
@@ -0,0 +1,7 @@
...
+
+- The `fetchLater()` API is invoked with a URL string as the first argument, and (optionally) a `DeferredRequestInit` object as the second argument.
```
rviscomi, line 1: If the second arg is optional and doesn't affect the eval, is it worth mentioning?

philipwalton: I'm not sure. This gets back to the discussion about whether or not the tools need access to reference docs.

It's possible that someone wanting to implement this use case will need to specify custom fetch params (e.g. custom headers or a specific request method), but that's not relevant to the use case.

So maybe that means we need to add another use case that covers custom fetch param? That use case wouldn't be specific to the `fetchLater()` use cases, though it would be relevant since there is a lot of API overlap.

guides/performance/batch-analytics-events/expectations.md
paulirish: It's gonna take me some more time to answer this question (and rick's below).

but yeah i don't want to block you guys.

so.. for now, i'm totally fine with landing these and i can file issues against already landed things as needed.

Issue #36, opened 2026-02-19
:user-valid and :user-invalid


rviscomi:
The :user-valid and :user-invalid pseudo-classes match form controls that have been marked as valid or invalid based on their validation constraints.

### MDN Documentation
- [:user-invalid](https://developer.mozilla.org/docs/Web/CSS/Reference/Selectors/:user-invalid)
- [:user-valid](https://developer.mozilla.org/docs/Web/CSS/Reference/Selectors/:user-valid)

### Chrome Documentation

<details>
<summary>View 7 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://web.dev/en/articles/codelab-sign-up-form-best-practices
- https://web.dev/en/articles/sign-in-form-best-practices
- https://web.dev/en/articles/user-valid-and-user-invalid-pseudo-classes
- https://web.dev/en/blog/new-in-the-web-io2024
- https://web.dev/en/learn/accessibility/forms
- https://web.dev/en/learn/forms/validation
- https://web.dev/en/shows/this-is-baseline/2iXLIlfpcVA
</details>

---
Feature ID: user-pseudos

**Use cases:**
- [ ] #313
- [ ] #316
- [x] #314
- [x] #315
- [x] #317

PR #92, opened 2026-02-20
feat: Add new user experience guides covering various psuedo selectors for form validation

PaulKinlan:
I spent some time with Gemini thinking about some use-cases, but I also did a heap of browsing some common issues on the likes of stack-overflow to find some examples that :user-valid and :user-invalid would work well with.

Note there is one about A11Y that I would like to do some more research on, but I think it makes sense.

This is for #36


philipwalton:
@PaulKinlan should we close this PR in favor of #97?

PaulKinlan:
Wait ... I'm not even sure why this one exist.... yes, close this.

Issue #72, opened 2026-02-19
scroll-initial-target


rviscomi:
The scroll-initial-target: nearest CSS declaration sets the initial scroll position of its scroll container to the top of the element, much like scrolling to a URL fragment.

### Chrome Documentation

<details>
<summary>View 3 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://github.com/DavMila/explainer-scroll-initial-target
- https://chromestatus.com/feature/6276178888097792
- https://drafts.csswg.org/css-scroll-snap-2/#propdef-scroll-initial-target
</details>

---
Feature ID: scroll-initial-target

**Use cases:**
- [x] #147
- [x] #300
- [x] #818
- [ ] #713

rviscomi:
Closing this in favor of tracking each use case separately:

https://github.com/GoogleChrome/guidance/issues/144
https://github.com/GoogleChrome/guidance/issues/145
https://github.com/GoogleChrome/guidance/issues/147

PR #94, opened 2026-02-20
Adding in scroll-initial-target use cases

pattishin:
Adds use cases for #72.


philipwalton:
@pattishin same comment as before, I know your started this before we formalized the new guidance structure, but can you update this to follow that new format?

guides/user-experience/scroll-to-middle-image-in-gallery/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: scroll-to-middle-image-in-gallery
+description: An image gallery that starts scrolled to the middle image
```
philipwalton: I would generalize this use case so it can match more than just image galleries (while still listing that as a primary use case). I'd also rephrase to make it more active.

How about something like: "Build a scrollable carousel of images (or other visual elements) where the scroll position can be set to display any one of the elements in view on the initial render."

guides/user-experience/scroll-to-selected-menu-list/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: scroll-to-selected-menu-list
+description: A menu list that starts scrolled to the selected option based on previous user selection.
```
philipwalton: I believe this use case was taken from [the explainer](https://github.com/DavMila/explainer-scroll-initial-target?tab=readme-ov-file#use-cases), but to be honest, I don't really understand this, do you? Perhaps it's worth checking with awogbemila@ (feature owner), but my sense is that this use case is not sufficiently different from the one you have above to justify a whole new guide. WDYT?

That said, I do think the second use case listed in the explainer (the "pull to reveal" pattern) is a good one to cover.

philipwalton (changes requested):
Thanks @pattishin for getting started on this. I left a few comments in the PR, also (FYI) I updated the issue with a new set resource links because the previous ones were not accurate.

guides/user-experience/scroll-to-selected-menu-list/guide.md
philipwalton: Another interesting use case that Gemini deep research help me discover is a search feature in a chat or messaging application, where you want to load the conversation scrolled to a particular message, but then still make it possible for the user to scroll up or down through the full conversation.

skills/web-dev-ui/guides/scroll-initial-target.md
bramus, line 1: This file was deleted in the third commit of this PR. Was that intentional?

bramus: Ah, I read in yesterday’s meeting notes just now that there should be 2 PRs _(one with the guidance, and one with the use-cases)_, so this seems intentional indeed :)

guides/user-experience/scroll-to-selected-menu-list/guide.md
pattishin: Oh interesting - will update with your suggestions! From the explainer, I suspected it would set the target based on last selected topic section or item if it's a shopping scenario, but I'll swap it out in favor of the above!

philipwalton: No, you have to manually apply the `scroll-initial-target` property to the item you want scrolled into view. There isn't anything auto-magic that happens AFAIK.

guides/user-experience/focus-chat-message/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: chat-message-search
+description: Build a chat message feature that would enable the user to land on a specific message in the conversation on initial render.
```
philipwalton: I would update this to make sure it's super clear that this use case involves having the message container *scrolled* to the message in question, rather than just visually displaying it to the user in some other way.

guides/user-experience/focus-chat-message/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: chat-message-search
+description: Build a chat messagefeature that automatically scrolls to a specific message on initial render.
```
philipwalton:
```suggestion
description: Build a chat message feature that can display a conversation automatically scrolled to a specific message (e.g. matching a search query) on initial render.
```

philipwalton (approved):
LGTM with one suggestion.

PR #97, opened 2026-02-20
User validation guides

PaulKinlan:
This is a WIP set of guides for the user-pseudos.

1. [accessible-error-announcement](https://github.com/GoogleChrome/guidance/commit/f7f4077761418de2f372a7481892b613d930adab) - I've tested this with the MacOS screenreader to validate if it works as expected. some other notes.
   1. I also took the decision to only check `:user-invalid` after every keypress after the first failure and while it's failing. The audio didn't get re-announced, and it's nice to get immediate feedback when it passes the validation.
   2. the polyfiling makes this messy, but it's complete.
   3. the intersection between a11y aria and :user-invalid I think makes :user-invalid less useful than people think...
   
2. Guides implemented so far:
+ [x] style-parent-with-has
+ [x] validate-email-after-interaction
+ [x] password-complexity-validation
+ [x] accessible-error-announcement
+ [x] DELETE prevent-validation-styling-on-draft-save - Phil was partially right that this wouldn't work. The demo correctly didn't style the element if it failed validation but because we are not submitting a form and refreshing the page any subsequent interaction would cause the :focus not to be true anymore and then the validation issues would appear in the UI.
+ [x] required-field-feedback
+ [x] select-menu-interaction
+ [x] DELETE web-component-validation - after extensive testing, it's not possible to know if the element is tainted by an interaction and thus :user-valid won't be set. Feels like a spec issue tbh


guides/user-experience/password-complexity-validation/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: password-complexity-validation
+description: Providing feedback on password pattern requirements only after the user has interacted with the field, preventing intimidating error states during initial entry.
```
philipwalton: I wonder if it makes sense to broaden this, and not limit it to just password fields? I agree that passwords is likely the most common occurance of this, but I think we'd want this guidance to also match other form-field use cases like email, phone number, etc.

@rviscomi WDYT?

guides/user-experience/prevent-validation-styling-on-draft-save/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: prevent-validation-styling-on-draft-save
+description: Use :user-invalid with :has() to conditionally style invalid fields, ensuring validation errors don't appear when a user opts to save a draft using formnovalidate.
```
philipwalton: TIL about `formnovalidate`! But I still don't understand how this would work...

Something like `:has([formnovalidate]) :user-invalid {...}` would match if the form contained a draft submit button, not just if the user pressed that.

Or were you imagining something else?

guides/user-experience/required-field-feedback/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: required-field-feedback
+description: Highlighting mandatory fields that were skipped or left empty after user focus and blur, ensuring feedback is contextually relevant to the user's flow.
```
philipwalton:
```suggestion
description: Highlighting required fields that were skipped or left empty after user focus and blur, ensuring feedback is contextually relevant to the user's flow.
```

Nit: I think "required" might be more likely to match a semantic search since there's a `required` attribute in HTML.

guides/user-experience/style-parent-with-has/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: style-parent-with-has
+description: Declaratively style parent elements like labels or fieldsets when a child input is in the :user-invalid state, eliminating the need for JavaScript state management.
```
philipwalton, line 3: Great use case. Also a good use case for `:has()`, so we should confirm that @pattishin (who is signed up for `:has()`) is not also writing this guide.

guides/user-experience/validate-email-after-interaction/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: validate-email-after-interaction
+description: Validation feedback for email inputs that only appears after the user has finished their initial interaction, avoiding premature errors on page load.
```
philipwalton: Same comment as above. I think it might make sense to combine this with the password one? WDYT?

guides/user-experience/password-complexity-validation/guide.md
rviscomi: Yeah, generalizing this to other form fields makes sense. It's hard to imagine substantive differences between the guides, but I'm interested to hear if there are any.

PaulKinlan: Yeah - I'm not opposed, I was thinking in this one that the use case is very clear vs a generic description that might get missed. In the guide I'm writing I also get to be quite clear about how to handle password fields.

Actually, that reminds me, if we do re-type password you don't want the error message to appear until you are ready. It constantly annoys me when I just started typing and it says it's not valid :D

guides/user-experience/prevent-validation-styling-on-draft-save/guide.md
PaulKinlan: You can have two input buttons on a form, one for draft and the other one to actually submit. I want to be able to click "save draft" and not have user-invalid trigger. You can do that by looking to see if the draft button is active or focused etc (iirc)

guides/user-experience/required-field-feedback/guide.md
PaulKinlan: fair - fixing on the branch

guides/user-experience/validate-email-after-interaction/guide.md
PaulKinlan: Yeah - this is fair. I still don't know how SEO (Skill Engine Optimisation.. heh) will work.

guides/user-experience/prevent-validation-styling-on-draft-save/guide.md
philipwalton: Have you tried that? I'm asking because I don't think it works. I think what will happen is you'll get the error message as soon as the user blurs the field, but then that error message will go away when you click the "Save draft" button — and I'm not sure that's actually a better experience.

Or are you thinking you'd always suppress errors if there's a "Save draft" button and then only show them once the "Submit" button matches `:active`?

Anyway, I guess TL;DR is it'd be good to see a demo of how you're expecting this to work.

guides/user-experience/style-parent-with-has/guide.md
PaulKinlan: pushed this one to the PR

guides/user-experience/prevent-validation-styling-on-draft-save/guide.md
PaulKinlan: It's actually the opposite - but eitherway, you can't get this to work in an SPA without JS, which means you might as well just ignore using CSS in this context. I'll remove this one for now.

guides/user-experience/password-complexity-validation/guide.md
PaulKinlan: Let me know what you want to do - I feel like naming the usecase is important because it's a feature I might ask for "add email validation etc"... that being said, I'm not precious.

philipwalton: I think we should consolidate for now, but make sure all of the sub-use-cases are enumerated in the description. We can always add more use case guides later if we determine that the tool isn't properly matching in real world usage.

Issue #40, opened 2026-02-19
blocking="render"


rviscomi:
The blocking="render" attribute for <link>, <script>, and <style> elements blocks rendering until the external script or stylesheet has been loaded. For <link rel="expect">, rendering is blocked until a specific element is in the DOM.

### MDN Documentation
- [HTMLLinkElement: blocking property](https://developer.mozilla.org/docs/Web/API/HTMLLinkElement/blocking)
- [HTMLScriptElement: blocking property](https://developer.mozilla.org/docs/Web/API/HTMLScriptElement/blocking)
- [HTMLStyleElement: blocking property](https://developer.mozilla.org/docs/Web/API/HTMLStyleElement/blocking)

### Chrome Documentation

<details>
<summary>View 31 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/google-ai-mode-view-transitions-case-study
- https://developer.chrome.com/en/blog/removing-document-write
- https://developer.chrome.com/en/blog/script-component
- https://developer.chrome.com/en/blog/view-transitions-misconceptions
- https://developer.chrome.com/en/docs/devtools/coverage
- https://developer.chrome.com/en/docs/devtools/network
- https://developer.chrome.com/en/docs/lighthouse/performance/render-blocking-resources
- https://developer.chrome.com/en/docs/lighthouse/performance/resource-summary
- https://developer.chrome.com/en/docs/lighthouse/performance/unused-css-rules
- https://developer.chrome.com/en/docs/performance/insights/render-blocking
- https://developer.chrome.com/en/docs/web-platform/view-transitions/cross-document
- https://web.dev/en/articles/critical-rendering-path/adding-interactivity-with-javascript
- https://web.dev/en/articles/critical-rendering-path/analyzing-crp
- https://web.dev/en/articles/critical-rendering-path
- https://web.dev/en/articles/critical-rendering-path/page-speed-rules-and-recommendations
- https://web.dev/en/articles/critical-rendering-path/render-blocking-css
- https://web.dev/en/articles/defer-non-critical-css
- https://web.dev/en/articles/howbrowserswork
- https://web.dev/en/articles/imports
- https://web.dev/en/articles/optimize-vitals-lighthouse
- https://web.dev/en/articles/preload-scanner
- https://web.dev/en/articles/speed-script-loading
- https://web.dev/en/articles/third-party-javascript
[...]

PR #100, opened 2026-02-23
Use cases and file structure for blocking="render"

rowan-m:
Use cases for client-side A/B testing and cross-document view transitions as per #40 for the `blocking="render"` attribute.

### DIFF
--- a/guides/user-experience/consistent-cross-document-transitions/guide.md
+++ b/guides/user-experience/consistent-cross-document-transitions/guide.md
@@ -1,0 +1,6 @@
+---
+name: consistent-cross-document-transitions
+description: Ensure critical page state is loaded and stable before initiating a cross-document view transition.
+web-feature-ids:
+  - blocking-render
+---
--- a/guides/user-experience/consistent-cross-document-transitions/guide.md
+++ b/guides/user-experience/consistent-cross-document-transitions/guide.md
@@ -1,6 +1,6 @@
 ---
 name: consistent-cross-document-transitions
-description: Ensure critical page state is loaded and stable before initiating a cross-document view transition.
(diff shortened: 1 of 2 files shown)


guides/user-experience/consistent-cross-document-transitions/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: consistent-cross-document-transitions
+description: Ensure critical page state is loaded and stable before initiating a cross-document view transition.
```
philipwalton: I think this use-case description is correct, but I wonder if it would be better to expand on it to make it more easily discoverable by the agent (given a potentially vague user prompt).

E.g. the phrase "loaded and stable" doesn't (to me at least) fully convey that the browser has done everything it needs to do in order to visually render the page in the expected state for the cross-document view transition to appear seamless to the user.

philipwalton (changes requested):

guides/user-experience/consistent-cross-document-transitions/guide.md
rowan-m: I've added the specifics from the [MDN reference](https://developer.mozilla.org/en-US/docs/Web/API/View_Transition_API/Using#stabilizing_page_state_to_make_cross-document_transitions_consistent) which calls out what is being waited on. I was erring on the side of being concise and not jumping too much into a solution, but I guess for some of these lower-level features the use case is generally referring to specific technical details versus just the user effect.

philipwalton: Yeah, it's a balancing act for sure. But the most important thing is that the MCP tool is able to match this guide against a user's prompt, so that's how I've been approaching the decision re: when to add more detail vs. when to be concise.

philipwalton (approved):

PR #102, opened 2026-02-23
Move before usecases

### LINKED ISSUE #66: moveBefore()
The moveBefore() DOM method relocates a node while preserving its state. For example, you can move the active element without losing focus, move an animated element without resetting the animation, or move an iframe without reloading its content.

### MDN Documentation
- [Element: moveBefore() method](https://developer.mozilla.org/docs/Web/API/Element/moveBefore)
- [Document: moveBefore() method](https://developer.mozilla.org/docs/Web/API/Document/moveBefore)
- [DocumentFragment: moveBefore() method](https://developer.mozilla.org/docs/Web/API/DocumentFragment/moveBefore)

### Chrome Documentation

<details>
<summary>View 2 articles</summary>

- https://developer.chrome.com/en/blog/movebefore-api
- https://web.dev/en/articles/a11y-for-teams
</details>

---
Feature ID: move-before

**Use cases:**
- [ ] #1226
- [x] #175
- [x] #176

PaulKinlan:
Fixes #66 

Open question. Should we have one use-case per type of element that isn't now no-longer impacted by a re-parenting?

### DIFF
--- a/guides/user-experience/keep-an-animation-running-while-moving-the-element/guide.md
+++ b/guides/user-experience/keep-an-animation-running-while-moving-the-element/guide.md
@@ -1,0 +1,6 @@
+---
+name: keep-an-animation-running-while-moving-the-element
+description: Move or reparent an element and keep in-flight CSS transitions and Web Animations API animations running.
+web-feature-ids:
+- move-before
+---
--- a/guides/user-experience/keep-an-animation-running-while-moving-the-element/guide.md
+++ b/guides/user-experience/keep-an-animation-running-while-moving-the-element/guide.md
@@ -1,6 +1,0 @@
----
-name: keep-an-animation-running-while-moving-the-element
-description: Move or reparent an element and keep in-flight CSS transitions and Web Animations API animations running.
-web-feature-ids:
-- move-before
----
--- a/guides/user-experience/keep-focus-while-moving-element/guide.md
+++ b/guides/user-experience/keep-focus-while-moving-element/guide.md
@@ -1,6 +1,0 @@
----
-name: keep-focus-while-moving-element
-description: Maintain the active element's visual and programmatic focus when moving elements around or reparenting them in the DOM (for example when dragging and dropping elements or reordering a list).
-web-feature-ids:
-- move-before
----
--- a/guides/user-experience/keep-focus-while-moving-element/guide.md
+++ b/guides/user-experience/keep-focus-while-moving-element/guide.md
@@ -1,0 +1,6 @@
+---
+name: keep-focus-while-moving-element
+description: Maintain the active element's visual and programmatic focus when moving elements around or reparenting them in the DOM (for example when dragging and dropping elements or reordering a list).
+web-feature-ids:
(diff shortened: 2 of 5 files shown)


philipwalton (commented):
> Open question. Should we have one use-case per type of element that isn't now no-longer impacted by a re-parenting?

I think it primarily comes down to:
1. How different do we expect the guides to be for each of these similar use cases
2. Whether or not we think we can combine them into a single use case where the AI tool would still reliably match each of the sub-use-cases.

In this case, I think a single use case like the following could potentially work: "Move or reparent a DOM element without losing important element state, such as interactivity states (:focus/:active), <iframe> loading state, animation/transition state, etc.". You could potentially even list all of the things mentioned [here in the MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/moveBefore#description).

Then another use case not mentioned in this PR that I think is important is the VDOM / JS framework reactivity use case. While the guidance for that is probably the same as the one I just listed, I think its description is different enough that it might match different prompts.

WDYT?

PaulKinlan:
> In this case, I think a single use case like the following could potentially work: "Move or reparent a DOM element without losing important element state, such as interactivity states (:focus/:active), <iframe> loading state, animation/transition state, etc.". 

yeah, I think I might go back to this because they're all essentially the same thing, that being said the code example for each could be very specific to the context (i.e, is showing the iframe or animation logic more important than the `moveBefore` alone)

> You could potentially even list all of the things mentioned [here in the MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/moveBefore#description).

I was doing my best to not copy and paste :D

> Then another use case not mentioned in this PR that I think is important is the VDOM / JS framework reactivity use case. While the guidance for that is probably the same as the one I just listed, I think its description is different enough that it might match different prompts.

Yeah - I spent some time thinking about this one last night and I erred on not including it because I felt it might be implied by the fact that we are moving elements.... but now I think about it, it's the tear down and re-create the world issue... Let me have a think about it because I wonder if the examples we might need here will be very React specific.

philipwalton:
> yeah, I think I might go back to this because they're all essentially the same thing, that being said the code example for each could be very specific to the context (i.e, is showing the iframe or animation logic more important than the `moveBefore` alone)

My assumption/hope is that the AI agent can figure that out on its own, but we'd have to validate that. I think for now it's fine to consolidate and we can expand if we determine that the agents need more help.

> I was doing my best to not copy and paste :D

Eh, it's only going to be read by the agent!
 
> ...but now I think about it, it's the tear down and re-create the world issue... Let me have a think about it because I wonder if the examples we might need here will be very React specific.

This was clearly the main use case for Chrome creating the API to begin with, but at the same time, there are only a handful of people in the world building VDOM diffing libraries, so it's probably not a super applicable use case for average developers using frameworks, and arguably all the framework authors already know this API exists...

PaulKinlan:
@philipwalton I've merged all the usecases in to one

philipwalton (approved):

Issue #31, opened 2026-02-19
hidden="until-found"


rviscomi:
The hidden="until-found" attribute hides an element until it is found using the browser's find-in-page search or it is directly navigated to by following a URL fragment.

### MDN Documentation
- [HTML hidden global attribute](https://developer.mozilla.org/docs/Web/HTML/Reference/Global_attributes/hidden)

### Chrome Documentation

<details>
<summary>View 6 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/focus-start-point
- https://developer.chrome.com/en/docs/css-ui/hidden-until-found
- https://developer.chrome.com/en/docs/devtools/dom
- https://web.dev/en/blog/interop-2025
- https://web.dev/en/blog/whats-new-in-web-io2025
- https://web.dev/en/learn/html/navigation
</details>

---
Feature ID: hidden-until-found

**Use cases:**
- [x] #148
- [ ] #1232
- [ ] #1251

rviscomi:
Closing this now that use case-specific issues have been created: https://github.com/GoogleChrome/guidance/issues/143 https://github.com/GoogleChrome/guidance/issues/148

PR #104, opened 2026-02-23
Add use cases for `hidden="until-found"`

taraojo:
Resolving issue #31 with use cases for performance and user experience improvement.

Notes:
1. The `search-hidden-content` use case could also be an a11y use case as a11y was commonly referenced in the documentation. I opted for UX as this didn't seem to be solely for a11y.
2. Expecting future cross-over with `content-visibility` feature for `reduce-rendering-time`.

### DIFF
--- a/guides/user-experience/deep-link-to-hidden-content/guide.md
+++ b/guides/user-experience/deep-link-to-hidden-content/guide.md
@@ -1,0 +1,6 @@
+---
+name: deep-link-to-hidden-content
+description: Deep-link to visually hidden content (for example in collapsed accordions, closed tabs, or below "Read more" sections) using URL fragments and "Scroll to Text Fragment" links. 
+web-feature-ids:
+  - hidden-until-found
+---
--- a/guides/user-experience/search-hidden-content/guide.md
+++ b/guides/user-experience/search-hidden-content/guide.md
@@ -1,0 +1,6 @@
+---
+name: search-hidden-content
+description: Hide content from view using patterns such as accordions, tabs, and "Read more" sections, and enable native "Find in page" search to show hidden regions, while ensuring content can be indexed for search and referenced by `aria` attributes for accessibility.
+web-feature-ids:
+  - hidden-until-found
+---


guides/user-experience/deep-link-to-hidden-content/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: deep-link-to-hidden-content
+description: Deep-link to hidden content using URL fragments and "Scroll to Text Fragment" links.
```
philipwalton: I think it might be worth expanding on what "hidden content" means here, in the hopes that it matches more prompts. Perhaps something like: "Deep-link to collapsed or visually hidden content using..."

WDYT?

guides/user-experience/search-hidden-content/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: search-hidden-content
+description: Hide content from view using patterns such as accordions, tabs, and "Read more" sections, and enable native "Find in page" search to show hidden regions. Ensure content can be indexed for search and referenced by `aria` attributes for accessibility.
```
philipwalton:
```suggestion
description: Hide content from view using patterns such as accordions, tabs, and "Read more" sections, and enable native "Find in page" search to show hidden regions, while ensuring content can be indexed for search and referenced by `aria` attributes for accessibility.
```

Nit: I'd make this one sentence, otherwise it reads like two use cases.

philipwalton (changes requested):

guides/user-experience/deep-link-to-hidden-content/guide.md
taraojo: I wasn't sure what level of detail to go into. I guess we don't want to reference the features in the description as that is the output. Perhaps providing examples of common front-end components that regularly hide code would match more prompts. Will update!

guides/user-experience/search-hidden-content/guide.md
taraojo: I had it like this initially but changed it to try and improve readability :)

guides/performance/reduce-rendering-time/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: reduce-rendering-time
+description: Reduce rendering times in content-heavy web pages, such as long feeds or dashboards. Improve critical Core Web Vitals, including Largest Contentful Paint (LCP) and Interaction to Next Paint (INP).
```
philipwalton: I think I'd remove this use case.

As worded, I think it's too generic and will likely match lots of prompts where the hidden-until-found feature probably isn't helpful or the best fit (especially given the generic mention of improving Core Web Vitals). I'm also not sure we'd recommend this API as a solution to improving Core Web Vitals, so I think it's fine to omit this from the tool.

philipwalton (approved):
LGTM if you remove the use case I mentioned in my comment.

guides/performance/reduce-rendering-time/guide.md
taraojo: I've removed from this PR, but I've tweaked the wording and re-added to the `content-visibility` follow up #113 as this seems to be the core use case for these features.

PR #105, opened 2026-02-24
Creating an AGENTS.md in the guidance folder

PaulKinlan:
The intent is to help the people creating the guidance create consistent guides, expectations and demos. It should also help us transform all the guides more easily if we change the structure or how they should work.


paulirish (approved):
love it

Issue #83, opened 2026-02-19
text-wrap: pretty


rviscomi:
The text-wrap: pretty CSS declaration prioritizes better layout over speed when text is broken into multiple lines.

### MDN Documentation
- [text-wrap](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/text-wrap)

### Chrome Documentation

<details>
<summary>View 9 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/css-text-wrap-pretty
- https://developer.chrome.com/en/blog/css-wrapped-2023
- https://developer.chrome.com/en/docs/devtools/css/reference
- https://developer.chrome.com/en/docs/devtools/settings/preferences
- https://web.dev/en/articles/5-css-snippets-every-front-end-developer-should-know-in-2024
- https://web.dev/en/articles/css-regions-exclusions
- https://web.dev/en/learn/accessibility/typography
- https://web.dev/en/learn/design/typography
- https://web.dev/en/learn/pwa/foundations
</details>

---
Feature ID: text-wrap-pretty

**Use cases:**
- [x] #772

Issue #82, opened 2026-02-19
text-wrap: balance


rviscomi:
The text-wrap: balance CSS declaration balances the length of each line when text is broken into multiple lines. Also known as headline balancing.

### MDN Documentation
- [text-wrap](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/text-wrap)

### Chrome Documentation

<details>
<summary>View 8 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/whats-new-css-ui-2023
- https://developer.chrome.com/en/docs/css-ui/css-text-wrap-balance
- https://web.dev/en/articles/5-css-snippets-every-front-end-developer-should-know-in-2024
- https://web.dev/en/blog/interop-2024-midyear
- https://web.dev/en/blog/interop-2024-wrapup
- https://web.dev/en/blog/interop-2024
- https://web.dev/en/case-studies/natgeo
- https://web.dev/en/learn/accessibility/typography
</details>

---
Feature ID: text-wrap-balance

**Use cases:**
- [x] #772

Issue #81, opened 2026-02-19
text-wrap


rviscomi:
The text-wrap CSS property sets how lines break in text that overflows the container. It is a shorthand for text-wrap-style and text-wrap-mode.

### MDN Documentation
- [text-wrap](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/text-wrap)

### Chrome Documentation

<details>
<summary>View 4 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/whats-new-css-ui-2023
- https://developer.chrome.com/en/docs/css-ui/css-text-wrap-balance
- https://web.dev/en/blog/interop-2024-wrapup
- https://web.dev/en/blog/interop-2024
</details>

---
Feature ID: text-wrap

**Use cases:**
- [x] #236

PR #106, opened 2026-02-24
Use cases for text-wrap values: nowrap, balance, pretty, and stable.

rowan-m:
Use cases for `text-wrap` values: `nowrap`, `balance`, `pretty`, and `stable` as per #81, #82, and #83.

### DIFF
--- a/guides/user-experience/improve-body-text-layout-and-legibility/guide.md
+++ b/guides/user-experience/improve-body-text-layout-and-legibility/guide.md
@@ -1,0 +1,7 @@
+---
+name: improve-body-text-layout-and-legibility
+description: Improve the layout and legibility of long text content by enabling the browser to manage line breaks to reduce line rag, orphaned text, and other visual artifacts.
+web-feature-ids:
+  - text-wrap
+  - text-wrap-pretty
+---
--- a/guides/user-experience/improve-heading-text-layout-and-legibility/guide.md
+++ b/guides/user-experience/improve-heading-text-layout-and-legibility/guide.md
@@ -1,6 +1,6 @@
 ---
 name: improve-heading-text-layout-and-legibility
-description: Improve the layout and legibility of short standalone text content, such as headings, by enabling the browser to apply evenly balanced line breaks when wrapping text.
+description: Improve the layout and legibility of short standalone text content, such as headings no longer than a few lines, by enabling the browser to apply evenly balanced line breaks when wrapping text.
 web-feature-ids:
   - text-wrap
   - text-wrap-balanced
--- a/guides/user-experience/improve-heading-text-layout-and-legibility/guide.md
+++ b/guides/user-experience/improve-heading-text-layout-and-legibility/guide.md
@@ -1,0 +1,7 @@
+---
+name: improve-heading-text-layout-and-legibility
+description: Improve the layout and legibility of short standalone text content, such as headings, by enabling the browser to apply evenly balanced line breaks when wrapping text.
+web-feature-ids:
+  - text-wrap
+  - text-wrap-balanced
+---
--- a/guides/user-experience/improve-heading-text-layout-and-legibility/guide.md
+++ b/guides/user-experience/improve-heading-text-layout-and-legibility/guide.md
@@ -3,5 +3,5 @@ name: improve-heading-text-layout-and-legibility
 description: Improve the layout and legibility of short standalone text content, such as headings no longer than a few lines, by enabling the browser to apply evenly balanced line breaks when wrapping text.
 web-feature-ids:
   - text-wrap
-  - text-wrap-balanced
+  - text-wrap-balance
 ---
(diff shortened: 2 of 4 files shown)


rowan-m:
Open questions:
1. I've worked on the assumption that `text-wrap` is a "parent" feature here, so creating use cases for its values (excluding the default) but explicitly tagged the features for `text-wrap-balanced` and `text-wrap-pretty`. Is that right?
2. I've kept this pretty concise as the use cases seem relatively simple. Should these be expanded?
3. Or, should this be cut down? Just a general `text-wrap` use case and remove the separate `nowrap` and `stable` ones?

guides/user-experience/improve-body-text-layout-and-legibility/guide.md
una, line 1: text-wrap balance and pretty are more for headings and short-form content rather than long text content. Long text doesn't suffer from the same typographic layout issues since it tends to be blockier to begin with. For long text, `pretty` would prevent single-line orphans. It doesn't really affect line rag in the rest of the block.

guides/user-experience/improve-heading-text-layout-and-legibility/guide.md
una, line 1: Note that `balance` is limited to 4 lines of text.

philipwalton:
> 1\. I've worked on the assumption that `text-wrap` is a "parent" feature here, so creating use cases for its values (excluding the default) but explicitly tagged the features for `text-wrap-balanced` and `text-wrap-pretty`. Is that right?

Given that these are related features, there's going to be a lot of overlap in the use cases, and that's totally fine. You can specify multiple `web-feature-ids` values in each use case guide. It's also fine to address all three of those features in this single PR.

But I wouldn't just consider `text-wrap` to be a parent feature and ignore use cases for it that aren't relevant to `balance` and `pretty`. I don't know the full history here, but IIUC the `text-wrap` CSS property was created to address limitations of the `white-space` property, so it might be good to identify if there are uses cases for when you'd want to use `text-wrap: nowrap` instead of `white-space: nowrap`.

> 2\. I've kept this pretty concise as the use cases seem relatively simple. Should these be expanded?

I think the length is good!

> 3\. Or, should this be cut down? Just a general `text-wrap` use case and remove the separate `nowrap` and `stable` ones?

If there are any meaningful use cases for `text-wrap` that don't involved `balance` or `pretty`, then definitely include them.

guides/user-experience/improve-body-text-layout-and-legibility/guide.md
philipwalton: BTW, on MDN it [specifically says](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-wrap#pretty) that `pretty` can be used for body copy

> Results in the same behavior as wrap, except that the user agent will use a slower algorithm that favors better layout over speed. This is intended for body copy where good typography is favored over performance (for example, when the number of orphans should be kept to a minimum).

guides/user-experience/improve-heading-text-layout-and-legibility/guide.md
philipwalton: Also BTW, on MDN [it says](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-wrap#balance) balance is limited to 6 (or less) lines in Chrome. Should this be updated to say 4?

guides/user-experience/improve-body-text-layout-and-legibility/guide.md
rowan-m: Yeah, I was unsure about this one as it feels like it also depends on the implementation we're trying to encourage. i.e. Safari has [their view](https://webkit.org/blog/16547/better-typography-with-text-wrap-pretty/) with a wider selection of visual improvements and a claim of minimal performance impact. I also see some instances of [`pretty` in CSS resets](https://www.joshwcomeau.com/css/custom-css-reset/).

So, **core question**: should I update this to include  the "appearance over performance" message or should this be encouraging `pretty` as a default for sensibly tagged body text?

If I'm including the performance messaging, at least for my own personal curiosity, it would be helpful to know how much of an impact it has. _(I mean, I can also just go and test this myself too.)_

guides/user-experience/improve-heading-text-layout-and-legibility/guide.md
rowan-m: I've updated this to include "no longer than a few lines" in the spirit of not linking things too closely to implementation details but giving sensible guard rails. For example, the existing "standalone" is trying to cover the [guidance about headings in containers with borders, etc.](https://developer.chrome.com/docs/css-ui/css-text-wrap-balance#balancing_won%E2%80%99t_change_the_inline-size_of_the_element)

una: I followed up about this this morning. We shipped with a 4 line limit. The spec says 10, and MDN says 6. I'll let you know when I hear back from the feature owner.

guides/user-experience/improve-body-text-layout-and-legibility/guide.md
philipwalton: It would probably be good to get the feature owner's perspective on this. My gut reaction is not recommend this as the default for body text, but to suggest it as a solution appropriate for body text on sites that care deeply about their typography and design aesthetics.

philipwalton (approved):
Overall this LGTM, modulo getting some clarity on the recommendation for `pretty`.

philipwalton:
I'm going to merge this as is. We can revisit the actual recommendations when reviewing the guides.

Issue #76, opened 2026-02-19
scrollbar-width


rviscomi:
The scrollbar-width CSS property sets the width of the scrollbar.

### MDN Documentation
- [scrollbar-width](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/scrollbar-width)

### Chrome Documentation

<details>
<summary>View 7 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/docs/css-ui/custom-scrollbar
- https://developer.chrome.com/en/docs/css-ui/scrollbar-styling
- https://web.dev/en/blog/baseline-scrollbar-props
- https://web.dev/en/blog/interop-2024-wrapup
- https://web.dev/en/case-studies/mathboard
- https://web.dev/en/learn/css/overflow
- https://web.dev/en/learn/design/intro
</details>

---
Feature ID: scrollbar-width

**Use cases:**
- [x] #818
- [ ] #174

Issue #75, opened 2026-02-19
scrollbar-gutter


rviscomi:
The scrollbar-gutter CSS property reserves space for the scrollbar, preventing unwanted layout changes as the scrollbar appears and disappears.

### MDN Documentation
- [scrollbar-gutter](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/scrollbar-gutter)

### Chrome Documentation

<details>
<summary>View 10 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/new-in-web-ui-io-2025-recap
- https://developer.chrome.com/en/docs/css-ui/custom-scrollbar
- https://developer.chrome.com/en/docs/devtools/rendering/performance
- https://web.dev/en/articles/building/a-settings-component
- https://web.dev/en/articles/building/a-tabs-component
- https://web.dev/en/blog/baseline-scrollbar-props
- https://web.dev/en/blog/interop-2024-wrapup
- https://web.dev/en/blog/next-gen-css-2019
- https://web.dev/en/blog/viewport-units
- https://web.dev/en/learn/css/grid
</details>

---
Feature ID: scrollbar-gutter

Issue #74, opened 2026-02-19
scrollbar-color


rviscomi:
The scrollbar-color CSS property sets the color of the scrollbar track and thumb.

### MDN Documentation
- [scrollbar-color](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/scrollbar-color)

### Chrome Documentation

<details>
<summary>View 5 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/docs/css-ui/custom-scrollbar
- https://developer.chrome.com/en/docs/css-ui/scrollbar-styling
- https://web.dev/en/articles/accent-color
- https://web.dev/en/blog/baseline-scrollbar-props
- https://web.dev/en/learn/css/overflow
</details>

---
Feature ID: scrollbar-color

**Use cases:**
- [x] #171
- [ ] #174

PR #108, opened 2026-02-24
Add new use cases,  guides and demos for scrollbar customization, animation and preference adaptation.

PaulKinlan:
#74 #75 #76 - Adding use-cases and also guides for scrollbar customization

I spent a heap of time thinking just past the basic "set the colour and set the width" (which I did anyway)

I came up with two new use-cases:

1. Adapting the scroll-bar colour to the theme (dark, light)
2. Animating the scroll-bar colour based on scroll position... tbh, I never seen anyone do it, but I thought it would be novel.

Because I really wanted to learn how it works I spent some time coding up a demo and by that point I thought I could structure some guides that handle the WebKit pseudo element.


guides/user-experience/adapt-scrollbar-to-contrast-preferences/guide.md
```diff
@@ -0,0 +1,68 @@
+---
+name: adapt-scrollbar-to-contrast-preferences
+description: Enhance scrollbar visibility for users who prefer high-contrast interfaces
+web-features:
+  - scrollbar-color
```
philipwalton, line 5:
```suggestion
  - scrollbar-color
  - prefers-contrast
```

If this is a key feature for this use case then it should be included here

guides/user-experience/adapt-scrollbar-to-contrast-preferences/guide.md
```diff
@@ -0,0 +1,68 @@
+---
+name: adapt-scrollbar-to-contrast-preferences
+description: Enhance scrollbar visibility for users who prefer high-contrast interfaces
```
philipwalton, line 3: I think this is a good as a use case, but perhaps this is also something that should be included in a top-level skill file. @rviscomi WDYT?

guides/user-experience/adapt-scrollbar-to-contrast-preferences/guide.md
```diff
@@ -0,0 +1,68 @@
...
+    /* Use extremely distinct colors like solid black against white */
+    --scrollbar-thumb: #000000;
+    --scrollbar-track: #ffffff;
+  }
+}
+```
+
+## Legacy WebKit Fallbacks
+
+If you are using legacy WebKit pseudo-elements to ensure custom colored scrollbars on older versions of Safari/Chrome, the variable assignments from the media query above will automatically cascade to the fallback.
```
philipwalton, line 48: Ideally we'd be specifying this in terms of a Baseline target, rather than just "legacy versions of Safari/Chrome", so that AI agent can understand whether or not it needs to include this in a given project.

Perhaps something like "If your Baseline target is older than Baseline 2025, then..."

@rviscomi we still need to figure out exactly how we phrase this, and also test that it works.

guides/user-experience/adapt-scrollbar-to-light-dark-preferences/guide.md
```diff
@@ -0,0 +1,77 @@
+---
+name: adapt-scrollbar-to-light-dark-preferences
+description: Ensure the scrollbar visually matches the user's operating system light/dark mode preference
+web-features:
+  - scrollbar-color
```
philipwalton, line 5:
```suggestion
  - scrollbar-color
  - color-scheme
  - prefers-color-scheme
```

guides/user-experience/adapt-scrollbar-to-light-dark-preferences/guide.md
```diff
@@ -0,0 +1,77 @@
...
+  color-scheme: light dark;
+}
+```
+
+## 2. Custom Color Adaptation
+
+If you are using `scrollbar-color` or the legacy `::-webkit-scrollbar` pseudo-elements to explicitly define custom scrollbar colors, you MUST ensure these colors are legible and appropriate in both light and dark modes.
+
+MANDATORY: Use CSS custom properties (variables) to define your colors and update them within a `prefers-color-scheme` media query to avoid repetition.
+MANDATORY: To prevent conflicts between standard properties and legacy WebKit selectors in browsers that support both natively (like modern Chrome), you MUST wrap legacy WebKit fallbacks in an `@supports not (scrollbar-color: auto)` block.
```
philipwalton: I wonder if it's better to say "non-standard" rather than "legacy", which could maybe side step the Baseline issue I mentioned in a comment above.

guides/user-experience/adapt-scrollbar-to-light-dark-preferences/expectations.md
```diff
@@ -0,0 +1,9 @@
+* The agent has defined `color-scheme: light dark` on the `:root` element.
+* The agent has created a scrollable element with `overflow: auto` or `overflow: scroll` or `overflow-y` set.
+* The agent has defined CSS variables (custom properties) for the scrollbar colors.
+* The agent has provided a `@media (prefers-color-scheme: dark)` block that updates the CSS variables to darker overrides to prevent color repetition.
+* The explicit scrollbar colors use the standard `scrollbar-color: var(--thumb) var(--track)` property applied directly to the scrollable element.
+* The explicit `scrollbar-width` property is also applied to the scrollable element to ensure the custom colors render on macOS.
+* The explicit `scrollbar-gutter: stable` property is applied to ensure the track background is visible on macOS.
+* The fallback `::-webkit-scrollbar-*` styling is conditionally applied within an `@supports not (scrollbar-color: auto)` or equivalent feature query to prevent overriding modern properties.
```
philipwalton: Repeating what I said above, it'd be great if we could test that this fallback isn't included if the project's Baseline target doesn't require it.

guides/user-experience/animate-scrollbar-color-on-scroll/guide.md
```diff
@@ -0,0 +1,78 @@
+---
+name: animate-scrollbar-color-on-scroll
+description: Animate the scrollbar color dynamically as the user scrolls down the page
+web-features:
+  - scrollbar-color
+  - scroll-driven-animations
+  - registered-custom-properties
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/animation-timeline
```
philipwalton, line 10:
```suggestion
  - https://developer.mozilla.org/en-US/docs/Web/CSS/animation-timeline
  - https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Properties_and_values_API/Registering_properties
  
```

guides/user-experience/customize-scrollbar-color/guide.md
```diff
@@ -0,0 +1,60 @@
+---
+name: customize-scrollbar-color
+description: Change the color scheme of the scrollbar
```
philipwalton: I'd combine this and the following use case into a single "Customize the color or thickness of a scrollbar" use case that shows how to do both.

Otherwise, if a user's prompt asks to do both then the MCP server would likely return both guides (which is probably unnecessary).

philipwalton (changes requested):

guides/user-experience/customize-scrollbar-color/guide.md
PaulKinlan: Yeah - I struggled with this one. I'll just merged them into one

guides/user-experience/adapt-scrollbar-to-contrast-preferences/guide.md
rviscomi: Maybe for now let's see how well agents are able to discover this use case with specific and vague prompts, and if it's not as discoverable as we'd like, we can add something to the skill file.

PaulKinlan: I'm not sure what you want me to do? Is it ok to keep this as is then?

guides/user-experience/adapt-scrollbar-to-contrast-preferences/guide.md
PaulKinlan: I can add this in, but given this was a WebKit-only feature that still works (and has no fallback for Firefox we are going to have to clarify that how to get it manage the fallback case specific to WebKit

guides/user-experience/adapt-scrollbar-to-light-dark-preferences/guide.md
PaulKinlan: Yep - I think that we can do that.

guides/user-experience/adapt-scrollbar-to-light-dark-preferences/expectations.md
PaulKinlan: Do you want me to do that now? or is that something the framework will manage. I can clear up the wording for Baseline broadly

guides/user-experience/adapt-scrollbar-to-contrast-preferences/guide.md
PaulKinlan: I think I've clarified this enough

guides/user-experience/adapt-scrollbar-to-light-dark-preferences/guide.md
PaulKinlan: Should be done now

guides/user-experience/customize-scrollbar-color/guide.md
PaulKinlan: done.

guides/user-experience/adapt-scrollbar-to-contrast-preferences/guide.md
philipwalton: Yeah, keep it, I was just saying we should have a plan for how to log these things as they come up.

guides/user-experience/adapt-scrollbar-to-light-dark-preferences/expectations.md
philipwalton: I'll update the expectation so we don't forget to enable it

philipwalton:
```suggestion
* If the legacy WebKit pseudo elements are needed to support the user's Baseline target, `::-webkit-scrollbar-*` styling is conditionally applied within an `@supports not (scrollbar-color: auto)` or equivalent feature query to prevent overriding modern properties.
```

guides/user-experience/adapt-scrollbar-to-light-dark-preferences/expectations.md
```diff
@@ -0,0 +1,9 @@
...
+* The fallback includes basic `::-webkit-scrollbar` dimensions (e.g., `width` or `height`) so the scrollbar renders its colors in webkit browsers.
```
philipwalton:
```suggestion
* If the legacy WebKit pseudo elements are needed to support the user's Baseline target, the fallback includes basic `::-webkit-scrollbar` dimensions (e.g., `width` or `height`) so the scrollbar renders its colors in webkit browsers.
```

philipwalton (approved):
LGTM with a few additional suggestions.

Issue #57, opened 2026-02-19
image-set()


rviscomi:
The image-set() CSS function provides a set of images at different resolutions or pixel densities, which the browser can pick from, depending on the device capabilities.

### MDN Documentation
- [image-set()](https://developer.mozilla.org/docs/Web/CSS/Reference/Values/image/image-set)

### Chrome Documentation

<details>
<summary>View 11 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/devtools-mobile
- https://web.dev/en/articles/easy-high-dpi-images
- https://web.dev/en/articles/high-dpi
- https://web.dev/en/articles/image-optimization-survey-2019
- https://web.dev/en/articles/optimizing-content-efficiency-save-data
- https://web.dev/en/articles/preload-responsive-images
- https://web.dev/en/articles/responsive-images
- https://web.dev/en/articles/web-performance-made-easy
- https://web.dev/en/case-studies/oz
- https://web.dev/en/learn/design/intro
- https://web.dev/en/learn/design/responsive-images
</details>

---
Feature ID: image-set

**Use cases:**
- [x] #642
- [x] #162
- [ ] #1222

rviscomi:
Guides and evals for these use cases are now being tracked in https://github.com/GoogleChrome/guidance/issues/162 and https://github.com/GoogleChrome/guidance/issues/167

PR #109, opened 2026-02-24
Use cases: image-set()

marcianoskate:
Use cases for `image-set()`. Addresses #57 .

### DIFF
--- a/guides/performance/inject-sharp-pseudo-element-icons/guide.md
+++ b/guides/performance/inject-sharp-pseudo-element-icons/guide.md
@@ -1,0 +1,7 @@
+---
+name: inject-sharp-pseudo-element-icons
+description: Inject sharp, resolution-optimized UI icons into ::before and ::after pseudo-elements without adding unnecessary DOM nodes using the image-set() CSS functional notation within the content property
+web-feature-ids: 
+  - image-set
+---
+
--- a/guides/performance/inject-sharp-pseudo-element-icons/guide.md
+++ b/guides/performance/inject-sharp-pseudo-element-icons/guide.md
@@ -1,6 +1,6 @@
 ---
-name: inject-sharp-pseudo-element-icons
-description: Inject sharp, resolution-optimized UI icons into ::before and ::after pseudo-elements without adding unnecessary DOM nodes using the image-set() CSS functional notation within the content property
+name: resolution-optimized-pseudo-elements
+description: "Use resolution-optimized images in CSS pseudo-elements (such as `::before` and `::after`) to reduce the number of DOM nodes."
 web-feature-ids: 
   - image-set
 ---
--- a/guides/user-experience/deliver-appropriate-image-resolutions/guide.md
+++ b/guides/user-experience/deliver-appropriate-image-resolutions/guide.md
@@ -1,0 +1,6 @@
+---
+name: deliver-appropriate-image-resolutions
+description: Deliver appropriate background image resolutions for varying device pixel densities and network bandwidth constraints using the image-set() CSS functional notation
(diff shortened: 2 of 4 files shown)


guides/performance/inject-sharp-pseudo-element-icons/guide.md
```diff
@@ -0,0 +1,7 @@
+---
+name: inject-sharp-pseudo-element-icons
+description: Inject sharp, resolution-optimized UI icons into ::before and ::after pseudo-elements without adding unnecessary DOM nodes using the image-set() CSS functional notation within the content property
```
philipwalton: Great use case. I'd recommend generalizing it a bit so that it matches more varied prompts. I think you can also drop the feature name from the description, since it'll be in the guide content.

Here's a suggestion, but feel free to tweak if you want.

```suggestion
description: Use resolution-optimized images in CSS  pseudo elements (such as `::before` and `::after`) to reduce the number of DOM nodes 
```

The guide description can then expand on why developers needed to use DOM nodes for this in the past.

guides/user-experience/deliver-appropriate-image-resolutions/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: deliver-appropriate-image-resolutions
+description: Deliver appropriate background image resolutions for varying device pixel densities and network bandwidth constraints using the image-set() CSS functional notation
```
philipwalton: I wonder if this and the following use case should be combined into a single use case that's essentially: "Deliver responsive images in CSS, supporting situations where modifying the HTML is difficult or impossible".

philipwalton (changes requested):

guides/user-experience/deliver-appropriate-image-resolutions/guide.md
marcianoskate: After clarifying something, this function is only meant to provide a set of images for decorative purposes (background, icons, or masks). `image-set()` enables the browser to pick the best image for the available formats and resolution for the given device.

I think the important aspect to highlight is that the browser is the one that selects the best combination of format and pixel density based on the current user characteristics (i.e. bandwidth, resolution, format support of browser).

> Deliver optimized decorative images (such as backgrounds, UI icons, or complex masks) by simultaneously providing next-generation image formats (like AVIF or WebP) alongside multiple pixel densities (like 1x and 2x) so the browser can dynamically negotiate the best combination of file size and visual quality for the user's device capabilities.

philipwalton: Agreed, though my point was the `` and `<picture>` already give developers this ability, and those APIs are widely available. So AFAICT the main reason to use `image-set()` is in situations where you can't use `` or `<picture>` for whatever reason, and I think we want to frame the use case in a way that will likely match developer prompts for those cases.

marcianoskate: Got it. Correct, this is only intended for decorative sources of images. If it is a content image, it should use `` or `<picture>`.

I believe this should be clearly explained in the guide. Should we make sure to add it to the summary of the use-case?

philipwalton: > I believe this should be clearly explained in the guide. Should we make sure to add it to the summary of the use-case?

I think it's fine to have it in the guide itself (no need to spell it out in the use case). We'll just need to make sure this requirement is stated in the use case issue so it's clear to the contractors to mention it. cc: @rviscomi

philipwalton (approved):

Issue #62, opened 2026-02-19
light-dark()


rviscomi:
The light-dark() CSS function accepts two colors and uses one depending on the current color scheme.

### MDN Documentation
- [light-dark()](https://developer.mozilla.org/docs/Web/CSS/Reference/Values/color_value/light-dark)

### Chrome Documentation

<details>
<summary>View 11 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/delaying-shipping-of-css-functions
- https://developer.chrome.com/en/docs/android/custom-tabs/guide-ui-customization
- https://developer.chrome.com/en/docs/devtools/css/reference
- https://web.dev/en/articles/baseline-in-action-color-theme
- https://web.dev/en/articles/building/a-theme-switch-component
- https://web.dev/en/articles/building/a-tooltip-component
- https://web.dev/en/articles/light-dark
- https://web.dev/en/articles/new-patterns-july-2022
- https://web.dev/en/learn/design/media-features
- https://web.dev/en/patterns/theming/theme-switch
- https://web.dev/en/shows/gui-challenges/kZiS1QStIWc
</details>

---
Feature ID: light-dark

**Use cases:**
- [x] #227
- [ ] #1241
- [x] #855

PR #111, opened 2026-02-24
Use cases: `light-dark()`

marcianoskate:
Describe the use cases for `light-dark()` as described in #62 

## Semantic color palettes:

image

image

## Component specific light-dark theme:

image


guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: component-specific-light-dark-theme
+description: Create component-specific themes by forcing explicit color schemes on individual UI elements, giving users theme choices that are decoupled from their global operating system preferences by combining localized color-scheme properties with the light-dark() CSS color function.
```
philipwalton:
```suggestion
description: Create component-specific themes by forcing explicit color schemes on individual UI elements, giving users theme choices that are decoupled from their global operating system preferences
```

I'd remove the mention of the solution (i.e. the feature) in the use case description.

guides/user-experience/light-dark-theme-semantic-color-palettes/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: light-dark-theme-semantic-color-palettes
+description: Implement semantic color palettes for both light and dark modes natively on a single line, eliminating the need to write bulky @media (prefers-color-scheme: dark) blocks to overwrite custom properties using the light-dark() CSS color function.
```
philipwalton:
```suggestion
description: Implement semantic color palettes for both light and dark modes natively on a single line, eliminating the need to write bulky @media (prefers-color-scheme: dark) blocks to overwrite custom properties
```

Same comment as above.

philipwalton (changes requested):

guides/user-experience/light-dark-theme-semantic-color-palettes/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: light-dark-theme-semantic-color-palettes
+description: "Implement semantic color palettes for both light and dark modes natively on a single line, eliminating the need to write bulky `@media (prefers-color-scheme: dark)` blocks."
```
marcianoskate, line 3: Adding quotation marks around the description might prevent potential confusing YAML interpretations of the text.

philipwalton (commented):
These use cases LGTM.

I think, for these in particular, a minimal demo showing them in action would be useful to make it clear to the folks writing the guide exactly what you mean.

marcianoskate:
I can do that. Let me work on it!

philipwalton:
@malchata can you review these demos?

guides/user-experience/component-specific-light-dark-theme/demo.html
```diff
@@ -0,0 +1,181 @@
...
+    By scoping <code>light-dark()</code> variables and the <code>color-scheme</code> property directly to a component
+    class rather than the <code>:root</code>, you can force individual instances of that component to render in a
+    specific theme—regardless of the surrounding page's system preference!
+  </p>
+
+  <button id="themeToggleBtn" class="theme-toggle">
+    Toggle Page Theme (System Default)
+  </button>
+
+  <div class="card-grid">
```
malchata: Consider using a `<main>` or `<section>` element here.

guides/user-experience/component-specific-light-dark-theme/demo.html
```diff
@@ -0,0 +1,181 @@
...
+    <!-- Card 1: Defaults to the system preference -->
+    <div class="themed-card">
+      <h2>System Preference</h2>
+      <p>This card organically inherits the <code>:root</code> color-scheme, toggling automatically between light and
+        dark modes.</p>
+      <button>Action</button>
+    </div>
+
+    <!-- Card 2: Exclusively forced to light mode -->
+    <div class="themed-card force-light">
```
malchata, line 163: Is a `<div>` here the appropriate approach? Wondering if a unordered list might be semantically better.

guides/user-experience/light-dark-theme-semantic-color-palettes/guide.md
malchata: Not sure what these are called, but would a `>-` and a line break before the property value also fix this?

guides/user-experience/component-specific-light-dark-theme/demo.html
```diff
@@ -0,0 +1,181 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Component-Specific Default Theme Demo</title>
+  <style>
```
malchata, line 13: Wondering if using `light-dark()` here in an inline style without something to block rendering until this is complete cause a potential "flash of color preference" here (I just made that up!) where the initial OS setting will appear first and suddenly shift? Just a thing to think about.

malchata (commented):
Just some quick questions/changes.

guides/user-experience/component-specific-light-dark-theme/demo.html
philipwalton: FYI: Inline <style> tags in the head will block rendering.

malchata: Thought so. Just had a nagging thought.

philipwalton:
> @malchata can you review these demos?

@malchata I should have clarified that the demos will not be used to train the AI, nor will they be published for developer consumption, so the only thing that needs to be reviewed is whether or not the demo properly implements the use case (and makes it clear exactly what the use case is). Aspects of the demo unrelated to the use case can be safely ignored.

guides/user-experience/component-specific-light-dark-theme/demo.html
marcianoskate: Good catch:

The best practice is to use the meta element `<meta name="color-scheme" content="light dark">` and the css root attribute.

The risk of FOUC is low with this two attributes because the  meta element instructs the browser to adapt even before downloading the CSS.

I understand that this was one of the reasons for the introduction of this functionality. Before it, people would have to style with some CSS that could produce the style flashing but this should reduce the risk of it happening.

guides/user-experience/component-specific-light-dark-theme/demo.html
marcianoskate: Fair. I'm not sure. I was intending two separate cards, so maybe that's why `<div />` made sense to me. Happy to accommodate :)

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: component-specific-light-dark-theme
+description: Create component-specific themes by forcing explicit color schemes on individual UI elements, giving users theme choices that are decoupled from their global operating system preferences
+web-feature-ids: 
+  - light-dark
```
philipwalton, line 6:
```suggestion
  - color-scheme
  - light-dark
```

guides/user-experience/component-specific-light-dark-theme/demo.html
philipwalton: Resolving since I don't think this matters for the purposes of this demo.

guides/user-experience/light-dark-theme-semantic-color-palettes/guide.md
```diff
@@ -0,0 +1,6 @@
...
+web-feature-ids: 
+  - light-dark
```
philipwalton, line 6:
```suggestion
  - color-scheme
  - light-dark
```

philipwalton (approved):

philipwalton:
I added the `color-scheme` feature to the list of web features since I believe that's required as well to use `light-dark()`

Issue #71, opened 2026-02-19
Scheduler API


rviscomi:
The scheduler API provides a way to prioritize all tasks belonging to an application.

### MDN Documentation
- [Prioritized Task Scheduling API](https://developer.mozilla.org/docs/Web/API/Prioritized_Task_Scheduling_API)

### Chrome Documentation

<details>
<summary>View 7 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/introducing-scheduler-yield-origin-trial
- https://developer.chrome.com/en/blog/use-scheduler-yield
- https://developer.chrome.com/en/docs/extensions/mv2/reference/events
- https://web.dev/en/articles/optimize-long-tasks
- https://web.dev/en/articles/top-cwv
- https://web.dev/en/case-studies/pubconsent-inp
- https://web.dev/en/case-studies/taboola-inp
</details>

---
Feature ID: scheduler

**Use cases:**
- [x] #625
- [x] #636

rviscomi:
Guides and evals for these use cases are now being tracked in:

- https://github.com/GoogleChrome/guidance/issues/159
- https://github.com/GoogleChrome/guidance/issues/160
- https://github.com/GoogleChrome/guidance/issues/161

PR #112, opened 2026-02-24
Use cases for Scheduler API

NourNabil:
Use cases for Scheduler API for issue #71

### DIFF
--- a/guides/performance/complex-computations-and-loops/guide.md
+++ b/guides/performance/complex-computations-and-loops/guide.md
@@ -1,0 +1,6 @@
+---
+name: complex-computations-and-loops
+description: Break up heavy synchronous processing or DOM updates, to let the browser handle user input and repaint the screen using the Scheduler API.
+web-feature-ids:
+    - scheduler
+---
--- a/guides/performance/complex-computations-and-loops/guide.md
+++ b/guides/performance/complex-computations-and-loops/guide.md
@@ -1,6 +1,6 @@
 ---
 name: complex-computations-and-loops
-description: Break up heavy synchronous processing or DOM updates, to let the browser handle user input and repaint the screen using the Scheduler API.
+description: Break up heavy synchronous processing or DOM updates, to let the browser handle user input and repaint the screen.
 web-feature-ids:
     - scheduler
 ---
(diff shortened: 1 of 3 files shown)


guides/performance/complex-computations-and-loops/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: complex-computations-and-loops
+description: Break up heavy synchronous processing or DOM updates, to let the browser handle user input and repaint the screen using the Scheduler API.
```
philipwalton:
```suggestion
description: Break up heavy synchronous processing or DOM updates, to let the browser handle user input and repaint the screen
```

guides/performance/prioritized-preloading/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: prioritized preloading
+description: Orchestrate background network requests without competing with critical rendering using the Scheduler API.
```
philipwalton: I'd generalize the "background network requests" to just something like "low priority work", in order to match more varied prompts.

philipwalton (changes requested):
I made a few suggestions inline.

In addition, I think one missing use case is the "yield + continuation" use case for `scheduler.yield()`, where you want to yield to the main thread but you don't want your continuation work to get put at the end of the queue: https://developer.chrome.com/blog/use-scheduler-yield#prioritized_continuations_after_yielding

guides/performance/complex-computations-and-loops/guide.md
NourNabil: Removed mention of the API in the second commit.

guides/performance/prioritized-preloading/guide.md
NourNabil: Modified the description to "Orchestrate low priority work, such as background network requests, without competing with critical rendering.". I wanted to keep the example but show it could be any kind of work.

NourNabil:
> I made a few suggestions inline.
> 
> In addition, I think one missing use case is the "yield + continuation" use case for `scheduler.yield()`, where you want to yield to the main thread but you don't want your continuation work to get put at the end of the queue: https://developer.chrome.com/blog/use-scheduler-yield#prioritized_continuations_after_yielding

I added the use case "prioritized-continuation-after-yielding", I thought this was an implicit effect of using the API but maybe I should've considered it as the most basic use case.

philipwalton (approved):
Updates look great, thanks!

Issue #49, opened 2026-02-19
content-visibility


rviscomi:
The content-visibility CSS property delays rendering an element, including layout and painting, until it is needed.

### MDN Documentation
- [content-visibility](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/content-visibility)

### Chrome Documentation

<details>
<summary>View 17 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/focus-start-point
- https://developer.chrome.com/en/blog/frameworks-may-2024
- https://developer.chrome.com/en/blog/styling-details
- https://developer.chrome.com/en/blog/view-transitions-misconceptions
- https://developer.chrome.com/en/docs/chromium/renderingng-architecture
- https://developer.chrome.com/en/docs/css-ui/hidden-until-found
- https://web.dev/en/articles/avoid-large-complex-layouts-and-layout-thrashing
- https://web.dev/en/articles/content-visibility
- https://web.dev/en/articles/critical-rendering-path/render-tree-construction
- https://web.dev/en/articles/dom-size-and-interactivity
- https://web.dev/en/articles/loading-ads-page-speed
- https://web.dev/en/articles/rendering-performance
- https://web.dev/en/articles/virtualize-long-lists-react-window
- https://web.dev/en/blog/2021-scroll-survey-report
- https://web.dev/en/blog/css-content-visibility-baseline
- https://web.dev/en/case-studies/rumvision
- https://web.dev/en/shows/http-203/FFA-v-CIxJQ
</details>

---
Feature ID: content-visibility

**Use cases:**
- [x] #208
- [x] #209
- [x] #210
- [x] #211

knowler:
Heads up if the `contentvisibilityautostatechange` event is ever used in any of these guides (perhaps #209): there’s an interop issue which probably will require a spec change (https://github.com/w3c/csswg-drafts/issues/11310). The event was not spec’d to bubble even though that’s what I believe the design intent was. Chrome bubbles. Safari and Firefox do not. So as a workaround until that’s sorted using a capturing listener or just a more direct event target works.

rviscomi:
Thanks @knowler. I do see that event mentioned in https://github.com/GoogleChrome/guidance/pull/643.

[...]

PR #113, opened 2026-02-24
Add use-cases for `content-visibility`

taraojo:
~Awaiting https://github.com/GoogleChrome/guidance/pull/104 to be merged (first commit) as `content-visibility` has crossover with `hidden=until-found`.~ Merged

This PR addresses #49 and adds use cases for `content-visibility`.

Made the assumption that the corresponding `contentvisibilityautostatechange` event is also part of this feature even though they were released in separate Chrome versions. They have the same baseline.

### DIFF
--- a/guides/performance/defer-rendering-heavy-content/guide.md
+++ b/guides/performance/defer-rendering-heavy-content/guide.md
@@ -1,0 +1,7 @@
+---
+name: defer-rendering-heavy-content
+description: Reduce rendering times in content-heavy web pages, such as pages with long feeds, lots of articles, or complex dashboards, by using native browser features to defer rendering for any content that is not immediately visible to the user.
+web-feature-ids:
+  - hidden-until-found
+  - content-visibility
+---
--- a/guides/performance/defer-rendering-heavy-content/guide.md
+++ b/guides/performance/defer-rendering-heavy-content/guide.md
@@ -1,6 +1,6 @@
 ---
 name: defer-rendering-heavy-content
-description: Reduce rendering times in content-heavy web pages, such as pages with long feeds, lots of articles, or complex dashboards, by using native browser features to defer rendering for any content that is not immediately visible to the user.
+description: Reduce rendering times in content-heavy web pages (e.g. pages with long feeds, lots of articles, or complex dashboards), by deferring rendering for any content that is not immediately visible to the user.
 web-feature-ids:
   - hidden-until-found
   - content-visibility
--- a/guides/performance/efficient-background-processing/guide.md
+++ b/guides/performance/efficient-background-processing/guide.md
@@ -1,0 +1,6 @@
+---
+name: efficient-background-processing
+description: Pause high-CPU tasks like <canvas> animations, WebSocket data streams, or heavy API polling, when they go out of view and then resume them just-in-time when they scroll back into view.
+web-feature-ids:
+  - content-visibility
+---
--- a/guides/performance/efficient-background-processing/guide.md
+++ b/guides/performance/efficient-background-processing/guide.md
@@ -1,6 +1,6 @@
 ---
 name: efficient-background-processing
-description: Pause high-CPU tasks like <canvas> animations, WebSocket data streams, or heavy API polling, when they go out of view and then resume them just-in-time when they scroll back into view.
+description: Pause high-CPU tasks like `<canvas>` animations, WebSocket data streams, or heavy API polling, when they go out of view to conserve device resources, and then resume them just-in-time when they scroll back into view.
 web-feature-ids:
   - content-visibility
 ---
--- a/guides/performance/faster-spa-view-transitions/guide.md
+++ b/guides/performance/faster-spa-view-transitions/guide.md
@@ -1,0 +1,6 @@
+---
+name: faster-spa-view-transitions
+description: Enable faster transitions between views in Single-Page Applications or complex modals, so when a user switches between views, the browser doesn't rebuild the view from scratch, instead it is shown instantly.
+web-feature-ids:
+  - content-visibility
+---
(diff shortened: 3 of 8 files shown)


guides/performance/faster-spa-view-transitions/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: faster-spa-view-transitions
+description: Faster transitions between views in Single-Page Applications or complex modals. When a user switches between views, the browser doesn't rebuild the view from scratch, instead it is shown instantly.
```
philipwalton:
```suggestion
description: Enable faster transitions between views in Single-Page Applications or complex modals, so when a user switches between views, the browser doesn't rebuild the view from scratch, instead it is shown instantly.
```

I'd try to make descriptions one sentence if possible.

guides/performance/interactions-in-complex-layouts/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: interactions-in-complex-layouts
+description: Make interactions snappier and more responsive by avoiding layout re-calculations in complex layouts, such as data-heavy dashboards or spreadsheet-style grids. This reduces Interaction to Next Paint (INP) scores.
```
philipwalton:
```suggestion
description: Make interactions snappier and more responsive (reducing Interaction to Next Paint (INP) scores) by avoiding layout re-calculations in complex layouts, such as data-heavy dashboards or spreadsheet-style grids.
```

Again, trying to phrase this as a single sentence.

guides/performance/efficient-background-processing/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: efficient-background-processing
+description: Pause high-CPU tasks like <canvas> animations, WebSocket data streams, or heavy API polling, when they go out of view and then resume them just-in-time when they scroll back into view.
```
philipwalton: TIL!

guides/performance/defer-rendering-off-screen-content/guide.md
```diff
@@ -0,0 +1,7 @@
+---
+name: defer-rendering-off-screen-content
+description: Reduce rendering times of off-screen content in content-heavy web pages, such as content with long feeds, lots of articles, or complex dashboards, by deferring rendering for any content below the fold.
+web-feature-ids:
+  - hidden-until-found
```
philipwalton: I think this use case can be implemented without `hidden-until-found`, right?

I suppose it'll depend on what the guide says, but I'd remove this unless your intent here is to ensure the person writing the guide includes this feature in the solution.

philipwalton (changes requested):

guides/performance/defer-rendering-off-screen-content/guide.md
taraojo: I understood this list as features that can be used for this use case. So my intention for the guide was that it explains when to use `hidden-until-found` vs `content-visibility`. Which is whether you want that content to be accessible via page search etc. If we separated these use-cases 99% of the guidelines would be the same. WDYT?

philipwalton: I guess it's just not clear to me how `hidden-until-found` fits into this use case, which is why I said "I suppose it'll depend on what the guide says". Were you planning on writing the guide for this? If not then we just need to make sure your implementation plan for the use case is clear in the issue.

taraojo: Got it, I was thinking of this use case as reducing first page load by hiding below the fold content.  `hidden=until-found` uses `content-visibility:hidden` under the hood so they have the same perf benefits. The main difference is should the below the fold content be searchable.

`content-visibility:auto` would also be relevant in this use case as it gives the browser responsibility of when to render the content.

I can write the guide if there is time, but I can add these details to the issue before then if this context makes sense to you?

philipwalton: Thanks for clarifying (and sorry for the slow reply, this thread accidentally got archived).

I *think* I now understand what you're describing, but it would be helpful to see a demo because my naive understanding of `hidden-until-found` is that it's a design consideration and not something you'd use for performance reasons. If you want to improve performance but keep off-screen text searchable, then you'd use `content-visibility: auto`. Right?

taraojo: Yes and no, yes to `content-visibility: auto`, but if you actually want to hide content, have it searchable and have performance benefits of `content-visibility: hidden` then you can use `hidden=until-found`. I've added demos, hopefully they help!

guides/performance/defer-rendering-heavy-content/guide.md
```diff
@@ -0,0 +1,7 @@
+---
+name: defer-rendering-heavy-content
+description: Reduce rendering times in content-heavy web pages, such as pages with long feeds, lots of articles, or complex dashboards, by using native browser features to defer rendering for any content that is not immediately visible to the user.
```
philipwalton:
```suggestion
description: Reduce rendering times in content-heavy web pages (e.g. pages with long feeds, lots of articles, or complex dashboards), by deferring rendering for any content that is not immediately visible to the user.
```

I removed the "native browser features" bit to keep this focused on the use case and not the solution. I also reworded a bit to make it easier to read (at least for me).

guides/performance/efficient-background-processing/guide.md
philipwalton:
```suggestion
description: Pause high-CPU tasks like `<canvas>` animations, WebSocket data streams, or heavy API polling, when they go out of view to conserve device resources, and then resume them just-in-time when they scroll back into view.
```

guides/performance/faster-spa-view-transitions/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: faster-spa-view-transitions
+description: Enable faster transitions between views in Single-Page Applications or complex modals, so when a user switches between views, the browser doesn't rebuild the view from scratch, instead it is shown instantly.
```
philipwalton:
```suggestion
description: Enable faster transitions between views in Single-Page Applications or other complex user interfaces with large DOM structures that can normally take a long time to render.
```
Remove the bit about "rebuilding from scratch", as I think that can be mentioned as part of the solution in the guide.

philipwalton (approved):
LGTM, after making a few suggestions.

Issue #42, opened 2026-02-19
color-scheme


rviscomi:
The color-scheme CSS property sets which color schemes (light or dark) an element uses and may prevent automatic dark mode adjustments by the browser.

### MDN Documentation
- [color-scheme](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/color-scheme)

### Chrome Documentation

<details>
<summary>View 34 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/auto-dark-theme
- https://developer.chrome.com/en/docs/android/custom-tabs/browser-support
- https://developer.chrome.com/en/docs/css-ui/css-color-mix
- https://developer.chrome.com/en/docs/devtools/css/reference
- https://developer.chrome.com/en/docs/devtools/settings/preferences
- https://web.dev/en/articles/accent-color
- https://web.dev/en/articles/app-like-pwas
- https://web.dev/en/articles/baseline-in-action-color-theme
- https://web.dev/en/articles/building/a-breadcrumbs-component
- https://web.dev/en/articles/building/a-button-component
- https://web.dev/en/articles/building/a-color-scheme
- https://web.dev/en/articles/building/a-dialog-component
- https://web.dev/en/articles/building/a-loading-bar-component
- https://web.dev/en/articles/building/a-media-scroller-component
- https://web.dev/en/articles/building/a-settings-component
- https://web.dev/en/articles/building/a-split-button-component
- https://web.dev/en/articles/building/a-theme-switch-component
- https://web.dev/en/articles/building/a-tooltip-component
- https://web.dev/en/articles/color-scheme
- https://web.dev/en/articles/light-dark
- https://web.dev/en/articles/new-patterns-july-2022
- https://web.dev/en/articles/new-responsive
- https://web.dev/en/articles/prefers-color-scheme
- https://web.dev/en/case-studies/terra-dark-mode
- https://web.dev/en/case-studies/zdf
- https://web.dev/en/learn/accessibility/color-contrast
- https://web.dev/en/learn/design/media-features
- https://web.dev/en/learn/design/theming
- https://web.dev/en/learn/pwa/app-design
- https://web.dev/en/patterns/theming/color-schemes
- https://web.dev/en/patterns/theming/hsl-starter-kit
- https://web.dev/en/patterns/theming/theme-switch
[...]

Issue #37, opened 2026-02-19
accent-color


rviscomi:
The accent-color CSS property sets a color for checkboxes, radio buttons, and other form controls.

### MDN Documentation
- [accent-color](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/accent-color)

### Chrome Documentation

<details>
<summary>View 12 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://web.dev/en/articles/accent-color
- https://web.dev/en/articles/building/a-button-component
- https://web.dev/en/articles/building/a-loading-bar-component
- https://web.dev/en/articles/building/a-settings-component
- https://web.dev/en/articles/building/a-switch-component
- https://web.dev/en/articles/cds2021-updates
- https://web.dev/en/articles/sign-in-form-best-practices
- https://web.dev/en/blog/state-of-css-2022
- https://web.dev/en/learn/design/theming
- https://web.dev/en/learn/forms/form-fields
- https://web.dev/en/learn/forms/styling-form-controls
- https://web.dev/en/learn/pwa/app-design
</details>

---
Feature ID: accent-color

**Use cases:**
- [x] #232
- [ ] #1217
- [ ] #1218
- [ ] #1220
- [ ] #1236
- [ ] #1246
- [x] #855

PR #115, opened 2026-02-24
Use cases for color-scheme

syntxerror:
This PR addresses #42 and adds use cases for color-scheme.

This PR also addresses #37 and adds a use case for accent-color.


guides/performance/incorrect-theme-flash/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: prevent-incorrect-theme-flash
+description: Leverage the color-scheme specification at the earliest possible stage: the HTML parsing layer, by placing <meta name="color-scheme" content="light dark"> high within the <head> of the HTML document. The browser is reliably informed of the document's supported color schemes well before it attempts its initial viewport paint.
```
philipwalton: The `name` field implies what the use case is here, but the `description` field focuses more on the solution. Can you rework this so the description just covers the use case?

guides/user-experience/stateful-theme-switching/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: decouple-site-and-os-color-schemes
+description: This feature ensures users can customize their visual experience for specific contexts, such as utilizing a light theme for improved readability within an app or site while maintaining a system-wide dark theme to conserve battery or protect their screen lifetime.
```
philipwalton: Same comments as above. I think I understand the use case from the name, but the `description` field could make it more clear.

guides/performance/incorrect-theme-flash/guide.md
```diff
@@ -1,6 +1,7 @@
 ---
 name: prevent-incorrect-theme-flash
-description: Leverage the color-scheme specification at the earliest possible stage: the HTML parsing layer, by placing <meta name="color-scheme" content="light dark"> high within the <head> of the HTML document. The browser is reliably informed of the document's supported color schemes well before it attempts its initial viewport paint.
+description: Leverage the color-scheme specification at the earliest possible stage, the HTML parsing layer, so the browser is reliably informed of the document's supported themeswell before it attempts its initial viewport paint.
```
philipwalton: This description is framed more as a technical solution than a use case.

I think the `name` property does describe the use case well. Can you expand on that for the description? Perhaps something like:

"Implement light/dark theme support in a way that ensures the user's chosen theme is always displayed on first render, without ever seeing a flash of the default theme"

guides/user-experience/brand-consistent-typographic-scaling-and-legibility/guide.md
```diff
@@ -0,0 +1,7 @@
+---
+name: brand-consistent-typographic-scaling-and-legibility
+description: Leverage advanced CSS properties alongside foundational design principles, ensuring web interfaces are resilient, readable, and perfectly aligned with brand visual standards.
```
philipwalton: Similar content to the above. This is framed as a solution. Can you reword to describe the use case that the user is trying to implement?

philipwalton (changes requested):

guides/performance/incorrect-theme-flash/guide.md
syntxerror: Will do!

philipwalton:
@malchata can you review this one, and also keep an eye out for any overlap with the use cases in #111?

philipwalton:
@syntxerror FYI, I added demos to these as well as slightly tweaking the use case wording.

PR #117, opened 2026-02-25
Add :has() use cases

pattishin:
Fixes #56
(Part 1 / 2)

* Adds new directory for utility APIs
* Adds use cases: conditional layouts, form input validation, sibling dimming

Issue #67, opened 2026-02-19
:not()


rviscomi:
The :not() functional pseudo-class matches elements that do not match the selectors in its argument.

### MDN Documentation
- [:not()](https://developer.mozilla.org/docs/Web/CSS/Reference/Selectors/:not)

### Chrome Documentation

<details>
<summary>View 7 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/has-with-cq-m105
- https://developer.chrome.com/en/docs/css-ui/css-nesting
- https://developer.chrome.com/en/docs/css-ui/css-nth-child-of-s
- https://web.dev/en/articles/components-howto-tabs
- https://web.dev/en/articles/css-is-and-where
- https://web.dev/en/learn/css/functions
- https://web.dev/en/learn/css/specificity
</details>

---
Feature ID: not

**Use cases:**
- [x] #274
- [x] #275

PR #118, opened 2026-02-25
Add `:not()` and `:has()` use cases

### LINKED ISSUE #56: :has()
The :has() CSS functional pseudo-class matches an element if any of the selectors passed as parameters would match at least one element.

### MDN Documentation
- [:has()](https://developer.mozilla.org/docs/Web/CSS/Reference/Selectors/:has)

### Chrome Documentation

<details>
<summary>View 18 articles</summary>

- https://developer.chrome.com/en/blog/css-ui-ecommerce-cq
- https://developer.chrome.com/en/blog/css-ui-ecommerce-has
- https://developer.chrome.com/en/blog/css-ui-ecommerce
- https://developer.chrome.com/en/blog/has-m105
- https://developer.chrome.com/en/blog/has-with-cq-m105
- https://developer.chrome.com/en/blog/help-css-nesting
- https://developer.chrome.com/en/blog/pop-ups-theyre-making-a-resurgence
- https://developer.chrome.com/en/blog/whats-new-css-ui-2023
- https://web.dev/en/articles/5-css-snippets-every-front-end-developer-should-know-in-2024
- https://web.dev/en/articles/building/a-tooltip-component
- https://web.dev/en/blog/baseline2023
- https://web.dev/en/blog/interop-2023-wrapup
- https://web.dev/en/blog/interop-2023
- https://web.dev/en/blog/new-in-the-web-io2024
- https://web.dev/en/blog/next-gen-css-2019
- https://web.dev/en/blog/state-of-css-2022
- https://web.dev/en/blog/state-of-css-html-2024
- https://web.dev/en/shows/gui-challenges/Y5EIC_UyPME
</details>

---
Feature ID: has

**Use cases:**
- [x] #274
- [x] #275
- [ ] #1234
- [ ] #1241

pattishin:
Fixes #56 and #67

* Includes use cases for both :not() and :has().
  * content based styling (:has())
  * child state based styling (:has())


philipwalton (commented):
These are good use cases, but I think they're a bit too specific. See this thread for more context as to why that matters: https://chat.google.com/room/AAQAPIjpZaI/d5uNBYMRDX0/d5uNBYMRDX0?cls=10

Also, I think your second use case does this, but I'd definitely make sure to include a use case that covers both `:not()` and `:has()` e.g. (doesn't have) because IMO `:not()` became 10x more useful once `:has()` landed :)

pattishin:
> These are good use cases, but I think they're a bit too specific. See this thread for more context as to why that matters: https://chat.google.com/room/AAQAPIjpZaI/d5uNBYMRDX0/d5uNBYMRDX0?cls=10
> 
> Also, I think your second use case does this, but I'd definitely make sure to include a use case that covers both `:not()` and `:has()` e.g. (doesn't have) because IMO `:not()` became 10x more useful once `:has()` landed :)

That makes sense! It'll be easier if I just add them together. I just closed https://github.com/GoogleChrome/guidance/pull/117 and just added/updated the use cases!

pattishin:
After feedback from @philipwalton, updated use cases to be a little less specific, moved them to `user-experience/`, and added `demo.html` for each use case! (Hat tip 🎩  @matthiasrohmer!)

guides/user-experience/conditional-layouts/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: conditional-layouts
+description: Build a component that changes its layout based on whether it contains an image. For example, if the component contains an image, use a multi-column layout, otherwise default to a single-column layout.
```
philipwalton:
```suggestion
description: Build a component that changes its layout based on whether it contains specific child elements. For example, if the component contains an image, use a multi-column layout, otherwise default to a single-column layout.
```

guides/user-experience/conditional-layouts/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: conditional-layouts
```
philipwalton:
```suggestion
name: content-based-styling
```
WDYT?

I worry "conditional-layouts" could apply to too many other use cases (e.g. container queries, style queries, etc.)

guides/user-experience/default-theme-styling/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: default-theme-styling
+description: Build a component that has a default theme that can be overridden by another theme (i.e. dark mode).
```
philipwalton: I like this demo and use case, but I don't think this description captures it. How about something like: "Build a component that changes its layout based on the state of one of its child elements. For example, a component that renders in light or dark mode based on whether a theme toggle is checked.

And then maybe the name is `child-state-based-styling`? WDYT?

guides/user-experience/sibling-dimming/demo.html
philipwalton, line 1: I know I suggested a `:has()` and `:not()` demo, but the effect in this particular demo could be easily achieved without `:has()` and `:not()`, so I'm not sure how useful this is.

I'm now thinking we should just make sure `:not(:has()` is present in the guide as an example for `:has()` usage in both your `content-based-styling` and `child-state-based-styling` use cases.

In other words, feel free to remove this use case as I don't think it's distinct enough.

philipwalton (changes requested):

guides/user-experience/sibling-dimming/demo.html
pattishin: Removed!

philipwalton (approved):
I tweaked the demo and use cases slightly to incorporate `:not()`.

PR #132, opened 2026-02-25
Add guide: `autofill-highlight-inputs`

samdutton:


guides/user-experience/autofill-highlight-inputs/demo.html
```diff
@@ -0,0 +1,606 @@
...
+        margin: 0 0 35px 0;
+      }
+
+      form section:last-of-type {
+        margin: 0 0 60px 0;
+      }
+    }
+
+  </style>
+  <script>
```
philipwalton: FYI, I'm getting an error when running the demo page because this script is run in the `<head>`, which means it doesn't find the `<form>` element.

Is this JS code required for the demo?

guides/user-experience/autofill-highlight-inputs/demo.html
```diff
@@ -0,0 +1,606 @@
...
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <title>:autofill highlight inputs</title>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+
+  <!-- <link rel="stylesheet" href="css/main.css"> -->
```
philipwalton: Should these comments be removed? (This and the next line.)

guides/user-experience/autofill-highlight-inputs/demo.html
```diff
@@ -0,0 +1,606 @@
...
+          <option value="ZW">Zimbabwe</option>
+        </select>
+      </section>
+
+      <section>
+        <label for="tel">Telephone</label>
+        <input id="tel" name="tel" autocomplete="tel" type="tel" maxlength="30" pattern="[\d \-\+]+" required>
+      </section>
+      <input type="submit" value="Submit" id="submit">
+      <!--       <button id="save-address" type="submit">Save address</button> -->
```
philipwalton: Should this be removed? If not then I'd recommend a comment explanation (similar to "" above).

guides/user-experience/autofill-highlight-inputs/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: autofill-highlight-inputs
+description: Use CSS to clearly highlight inputs that have been autofilled by the browsers and not edited by the user.
```
philipwalton:
```suggestion
description: Use CSS to clearly highlight form fields that have been autofilled by the browsers and not edited by the user.
```

Nit: should we say "form fields" instead of "input" since autofill works with `<select>` elements as well?

philipwalton (commented):
Overall this LGTM, but I did have a few questions before merging.

guides/user-experience/autofill-highlight-inputs/expectations.md
```diff
@@ -0,0 +1,8 @@
+# Expectations: `autofill-highlight-inputs`
+
+- In browsers that support the `:autofill` CSS pseudo-class, the `:autofill` CSS pseudo-class can be used to style autofilled `<input>` elements as required. JavaScript or other mechanisms SHOULD NOT be used to style autofilled `<input>` elements in browsers where the `:autofill` CSS pseudo-class is supported.
```
philipwalton: Both <input> and `<select>` right?

guides/user-experience/autofill-highlight-inputs/demo.html
```diff
@@ -0,0 +1,606 @@
...
+    /* Copyright 2026 Google LLC.
+    SPDX-License-Identifier: Apache-2.0 */
+
+    :root {
+      --small-mobile-font-size: 20px;
+      --large-mobile-font-size: 22px;
+      --desktop-font-size: 22px;
+    }
+
+    :-webkit-autofill {
```
philipwalton: Maybe add a comment explaining why this is here?

Also, is it reliable to assume that `:-webkit-autofill` will always match the same elements as `:autofill`? I think it'd be better to define all styles on both rather than define all `border` styles on the legacy selector and then only override the `border-style` property on `:autofill`.

guides/user-experience/autofill-highlight-inputs/demo.html
samdutton: Yeah... Removed in next commit.

guides/user-experience/autofill-highlight-inputs/demo.html
samdutton: Can't remember quite what the purpose was for this originally.

Changed in next commit as per your suggestion.

guides/user-experience/autofill-highlight-inputs/demo.html
samdutton: D'oh! Fixed in next commit.

samdutton: The JavaScript I've used here is best practice for forms in general, but I think it could be removed. In hindsight, I think the whole demo should be simplified, as per the autofill-background-color use case. I've done that on the next commit.

guides/user-experience/autofill-highlight-inputs/demo.html
samdutton: Removed.

guides/user-experience/autofill-highlight-inputs/expectations.md
samdutton: Good point — changed.

guides/user-experience/autofill-highlight-inputs/guide.md
samdutton: Thanks — yes!

guides/user-experience/autofill-highlight-inputs/demo.html
```diff
@@ -0,0 +1,68 @@
...
+    display: block;
+    font-weight: 500;
+    margin: 0 0 0.5rem 0;
+  }
+  input:autofill,
+  input:-webkit-autofill,
+  input:-webkit-autofill,
+  input:-webkit-autofill:hover,
+  input:-webkit-autofill:focus {
+    /*-webkit-box-shadow: 0 0 0px 40rem #efe inset;*/
```
philipwalton:
```suggestion
    /*box-shadow: 0 0 0 100vmax #efe inset;*/
```
Use `100vmax` instead just to be safe? Also switched just `box-shadow`.

philipwalton: Also, per may comment in https://github.com/GoogleChrome/guidance/pull/134#issuecomment-3987730598, I'd keep this here (commented out) and add a comment above it explaining that `box-shadow` can be used customize the background, given that `background` can't be overridden directly.

guides/user-experience/autofill-highlight-inputs/guide.md
```diff
@@ -0,0 +1,10 @@
+---
+name: autofill-highlight-inputs
+<<<<<<< HEAD
```
philipwalton: Looks like this merge conflict got left in.

guides/user-experience/autofill-highlight-inputs/prompts.md
```diff
@@ -0,0 +1,5 @@
+Add code to highlight autofilled form inputs.
+
+Use the `:autofill` CSS pseudo-class.
+
+Use CSS to highlight form fields that have been autofilled by the browser and not edited by the user.
```
philipwalton: Optional: you might even add another, more specific prompt asking to style both the border and background color of an autofilled-input, because if we want to recommend that then we should have an eval that tests that the AI tool can use `box-shadow` as an alternative to `background`.

philipwalton (changes requested):

guides/user-experience/autofill-highlight-inputs/guide.md
samdutton: Gah — fixed.

guides/user-experience/autofill-highlight-inputs/prompts.md
samdutton: Good idea — done.

guides/user-experience/autofill-highlight-inputs/demo.html
samdutton: Done.

philipwalton (approved):

PR #134, opened 2026-02-25
Add guide: `autofill-background-color`

samdutton:


philipwalton (commented):
This seems like it's basically the same use case as #132, just with background instead of border, correct? If so I'd recommend combining them (unless there's a good reason not to?).

samdutton:
@philipwalton 

> This seems like it's basically the same use case as #132, just with background instead of border, correct? If so I'd recommend combining them (unless there's a good reason not to?).

Yup - I could combine them. 

The rationale for having two use cases is that setting the background color for an autofilled input is a slightly different type of use case from setting the border (or whatever), since setting the background color cannot be done with `background-color`. Having said that, the background color workaround I've provided (from [CSS Tricks](https://css-tricks.com/almanac/pseudo-selectors/a/autofill/#aa-overriding-user-agent-auto-fill-styles) and elsewhere) is a bit of a hack, so maybe we should avoid it?

Also, of course, the core use case here is to demonstrate styling with `:autofill`.

philipwalton:
> Having said that, the background color workaround I've provided (from [CSS Tricks](https://css-tricks.com/almanac/pseudo-selectors/a/autofill/#aa-overriding-user-agent-auto-fill-styles) and elsewhere) is a bit of a hack, so maybe we should avoid it?

Ahh, TIL that you can't set `background-color` directly on an `:autofill` rule due to `!important` in the UA style sheet! That seems weird to me, do you know the context behind that?

Still, I think it makes sense to add this piece of info as an implementation note in the other use case, rather than having a separate use case. The reason is because, if someone is prompting an AI tool to update the styling of an autofilled input—and if they don't specifically say that they want to customize *just* the border or *just* the background color, then the tool would likely match both of these guides, and that would unnecessarily bloat their context window.

It would be better if only a single guide matched that contained all the info they need about how to style inputs matching `:autofill`.

samdutton:
> > Having said that, the background color workaround I've provided (from [CSS Tricks](https://css-tricks.com/almanac/pseudo-selectors/a/autofill/#aa-overriding-user-agent-auto-fill-styles) and elsewhere) is a bit of a hack, so maybe we should avoid it?
> 
> Ahh, TIL that you can't set `background-color` directly on an `:autofill` rule due to `!important` in the UA style sheet! That seems weird to me, do you know the context behind that?
> 
> Still, I think it makes sense to add this piece of info as an implementation note in the other use case, rather than having a separate use case. The reason is because, if someone is prompting an AI tool to update the styling of an autofilled input—and if they don't specifically say that they want to customize _just_ the border or _just_ the background color, then the tool would likely match both of these guides, and that would unnecessarily bloat their context window.
> 
> It would be better if only a single guide matched that contained all the info they need about how to style inputs matching `:autofill`.

OK — that makes sense. I'll close this PR and keep https://github.com/GoogleChrome/guidance/pull/132 (and https://github.com/GoogleChrome/guidance/pull/188).

samdutton:
See https://github.com/GoogleChrome/guidance/pull/132 (add border and background color to autofilled form fields) and https://github.com/GoogleChrome/guidance/pull/188 (general autofill guidance).

Issue #65, opened 2026-02-19
Masks


rviscomi:
The mask CSS property (and several longhand properties) partially or completely hides an element according to the shape and depth of an image.

### MDN Documentation
- [mask](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/mask)

### Chrome Documentation

<details>
<summary>View 9 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/paintapi
- https://web.dev/en/articles/building/a-theme-switch-component
- https://web.dev/en/articles/css-masking
- https://web.dev/en/articles/how-we-built-designcember
- https://web.dev/en/articles/maskable-icon
- https://web.dev/en/articles/shapes-getting-started
- https://web.dev/en/blog/interop-2023
- https://web.dev/en/case-studies/natgeo
- https://web.dev/en/learn/css/paths-shapes-clipping-masking
</details>

---
Feature ID: masks

**Use cases:**
- [ ] #1222
- [x] #190
- [ ] #1247
- [ ] #1249
- [x] #189
- [x] #856
- [x] #191
- [x] #193

rviscomi:
@jamesnw similar question: would your team be able to review https://github.com/GoogleChrome/guidance/pull/613 (4 guides)?

PR #135, opened 2026-02-25
Add use cases for `mask`

matthiasrohmer:
This adds five use cases for the CSS mask API as part of #65. The five use cases are:

#  Interactive content reveal

image

# Soft edge content fade

image

# Stencil cut outs

image

# Non-geometric shape clipping

image

# Visually texture content

image


guides/user-experience/non-geometric-shape-clipping/guide.md
```diff
@@ -0,0 +1,7 @@
+---
+name: non-geometric-shape-clipping
```
philipwalton: Nit: should this be "complex-shape" instead of "non-geometric"? I don't think it matters too much since this is just a name slug, but it stuck out to me because I think this use case also applies to geometric shapes.

guides/user-experience/soft-edge-content-fade/demo.html
```diff
@@ -0,0 +1,52 @@
...
+    .fade-container {
+      width: 300px;
+      max-height: 120px;
+      overflow: hidden;
+      line-height: 1.5;
+      background: white;
+      padding: 20px;
+      border-radius: 8px;
+      box-shadow: 0 2px 10px rgba(0, 0, 0, 0.1);
+      mask-image: linear-gradient(to bottom, black 60%, transparent 100%);
```
philipwalton, line 28: TIL you can use a linear-gradient as a mask image!

philipwalton (approved):
Great use cases! LGTM, with just one question about a use case name.

guides/user-experience/non-geometric-shape-clipping/guide.md
matthiasrohmer: Not really a nit, but my core pain with this one 😄 So what I want to say is: something else than a box. And more than what border-radius and corner-shape can do, which would be all sorts of geometrical (in the mathematical sense) shapes.

Clipping is also not great, because it suggests overlap with `clip-path`, which there is. But `mask-` is usually easier to handle for the same use cases.

I'll update to complex shape, I like that.

guides/user-experience/soft-edge-content-fade/demo.html
matthiasrohmer: Think of all `mask-` properties in their `background-` equivalents, than it's very easy to wrap your head around. The only thing you need to understand on top is then `mask-mode`.

rachelandrew: all kinds of gradients, the MDN page has a fun radial gradient example, which uses a gradient and an image https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/mask-image

PR #137, opened 2026-02-25
Add guidance for `search-hidden-content`

taraojo:
Add guide, expectations, demo and prompts for this use case for #31 and https://github.com/GoogleChrome/guidance/issues/148. 

Would appreciate feedback on:

- the level of detail of the expectations
- the framing of the prompts, I've aimed to frame them like a dev working on a specific feature, but maybe they should be more generic?

### DIFF
--- a/guides/user-experience/search-hidden-content/demo.html
+++ b/guides/user-experience/search-hidden-content/demo.html
@@ -1,0 +1,203 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Search Hidden Content Demo</title>
+  <style>
+    :root {
+      --primary: #3b82f6;
+      --bg: #f8fafc;
+      --text: #0f172a;
+      --border: #e2e8f0;
+    }
+
+    body {
+      font-family: system-ui, -apple-system, sans-serif;
+      background: var(--bg);
+      color: var(--text);
+      padding: 2rem;
+      max-width: 640px;
+      margin: 0 auto;
+      line-height: 1.6;
+    }
+
+    h1 {
+      font-size: 1.75rem;
+      margin-bottom: 0.5rem;
+    }
+
+    .instructions {
+      background: #dbeafe;
+      color: #1e40af;
+      padding: 1rem 1.25rem;
+      border-radius: 8px;
+      margin-bottom: 2rem;
+      font-size: 0.95rem;
+    }
+
+    .accordion {
+      border: 1px solid var(--border);
+      border-radius: 8px;
+      background: white;
+      overflow: hidden;
+      box-shadow: 0 1px 3px rgba(0, 0, 0, 0.05);
+    }
+
+    .accordion-item {
+      border-bottom: 1px solid var(--border);
+    }
+
+    .accordion-item:last-child {
+      border-bottom: none;
+    }
+
+    .accordion-header {
+      width: 100%;
+      text-align: left;
+      padding: 1.25rem;
+      background: transparent;
+      border: none;
+      font-size: 1.05rem;
+      font-weight: 500;
+      cursor: pointer;
+      display: flex;
+      justify-content: space-between;
+      align-items: center;
+      color: var(--text);
+    }
+
+    .accordion-header:hover {
+      background: #f1f5f9;
+    }
+
+    .icon {
+      font-size: 1.5rem;
+      font-weight: 300;
+      color: #64748b;
+      line-height: 1;
+      transition: transform 0.2s;
+    }
+
+    /* Rotate '+' icon to an 'x' when open */
+    .is-open .icon {
+      transform: rotate(45deg);
+    }
+
+    /* DO NOT apply display or visibility to elements with hidden="until-found" */
+    .accordion-pane {
+      margin: 0;
+    }
+
+    /* DO apply padding to nested wrappers instead */
+    .accordion-content {
+      padding: 0 1.25rem 1.25rem 1.25rem;
+      color: #334155;
+      font-size: 0.95rem;
+    }
+  </style>
+</head>
+
+<body>
+  <h1>Hidden Content Search</h1>
+  <div class="instructions">
+    <strong>Test it out:</strong> Press <code>Cmd/Ctrl + F</code> and search for words like <em>"strawberry"</em>,
+    <em>"mango"</em>, or <em>"hidden"</em>. The browser will automatically expand the right accordion section and jump
+    to the text!
+  </div>
+
+  <div class="accordion">
+    <div class="accordion-item is-open" id="item1">
+      <button class="accordion-header" aria-expanded="true" aria-controls="pane1">
+        Section 1: Apples & Oranges
+        <span class="icon" aria-hidden="true">+</span>
+      </button>
+      <div id="pane1" class="accordion-pane">
+        <div class="accordion-content">
+          <p>This section is open by default. You can see all the content right away without searching.</p>
+        </div>
+      </div>
+    </div>
+
+    <!-- hidden="until-found" is applied to the pane -->
+    <div class="accordion-item" id="item2">
+      <button class="accordion-header" aria-expanded="false" aria-controls="pane2">
+        Section 2: The Hidden Berries
+        <span class="icon" aria-hidden="true">+</span>
+      </button>
+      <div id="pane2" class="accordion-pane" hidden="until-found">
+        <div class="accordion-content">
+          <p>This section contains some <strong>hidden</strong> information about berries. If you searched for a term
+            like <em>blueberry</em> or <em>strawberry</em>, the browser automatically dispatched a
+            <code>beforematch</code> event, allowing our JavaScript to update the UI icons before revealing this panel.
+          </p>
+        </div>
+      </div>
+    </div>
+
+    <div class="accordion-item" id="item3">
+      <button class="accordion-header" aria-expanded="false" aria-controls="pane3">
+        Section 3: Tropical Fruits
+        <span class="icon" aria-hidden="true">+</span>
+      </button>
+      <div id="pane3" class="accordion-pane" hidden="until-found">
+        <div class="accordion-content">
+          <p>Here in the tropics, you will find a juicy <em>mango</em>! Even though this panel was initially collapsed,
+            it remained natively indexed by your browser's search capabilities.</p>
+        </div>
+      </div>
+    </div>
+  </div>
+
+  <script>
+    // 1. Map standard button toggling
+    document.querySelectorAll('.accordion-header').forEach(button => {
+      button.addEventListener('click', () => {
+        const paneId = button.getAttribute('aria-controls');
+        const pane = document.getElementById(paneId);
+        const item = button.closest('.accordion-item');
+        const isExpanded = button.getAttribute('aria-expanded') === 'true';
+
+        if (isExpanded) {
+          button.setAttribute('aria-expanded', 'false');
+          pane.setAttribute('hidden', 'until-found');
+          item.classList.remove('is-open');
+        } else {
+          button.setAttribute('aria-expanded', 'true');
+          pane.removeAttribute('hidden');
+          item.classList.add('is-open');
+        }
+      });
+    });
+
+    // 2. Synchronize UI upon browser search using 'beforematch'
+    document.querySelectorAll('.accordion-pane').forEach(pane => {
+      pane.addEventListener('beforematch', () => {
+        // The browser is about to remove the 'hidden' attribute.
+        // We just need to sync our UI classes and accessibility attributes.
+        const item = pane.closest('.accordion-item');
+        const button = item.querySelector('.accordion-header');
+        button.setAttribute('aria-expanded', 'true');
+        item.classList.add('is-open');
+      });
+    });
+
+    // 3. Fallback strategy for browsers without 'beforematch' support
+    // (We auto-expand all hidden content)
+    if (!('onbeforematch' in HTMLElement.prototype)) {
+      console.warn("hidden=until-found not supported. Executing fallback: expanding all content.");
+      document.querySelectorAll('[hidden="until-found"]').forEach(el => {
+        el.removeAttribute('hidden');
+        const paneId = el.id;
+        const button = document.querySelector(`.accordion-header[aria-controls="${paneId}"]`);
+        if (button) {
+          button.setAttribute('aria-expanded', 'true');
+          button.closest('.accordion-item').classList.add('is-open');
+        }
+      });
+    }
+  </script>
+</body>
+
+</html>
--- a/guides/user-experience/search-hidden-content/expectations.md
+++ b/guides/user-experience/search-hidden-content/expectations.md
@@ -1,7 +1,7 @@
 # Expectations: `search-hidden-content`
 
 - Any content intended to be visually hidden but remain searchable via the browser's native "Find in page" feature MUST use `<details>` or the `hidden="until-found"` attribute.
-- Elements utilizing the `hidden="until-found"` attribute MUST NOT have `display: none`, `visibility: hidden`, or any associated `display` or `visibility` CSS properties applied to them directly (unless as part of the fallback).
+- Elements utilizing the `hidden="until-found"` attribute MUST NOT have `display: none`, `visibility: hidden`, or any associated `display` or `visibility` CSS properties applied to them directly.
 - The `hidden="until-found"` attribute MUST NOT be used to hide sensitive information, internal data tokens, or "screen reader only" text.
 - If the hidden content has related UI state (e.g., updating ARIA attributes, toggling open/close classes, or managing accordion icons), that state MUST be synchronized using a beforematch event listener.
 - If the users Baseline target is older than Baseline 2025, a fallback strategy MUST be used. The implementation MUST include an explicit JavaScript feature detection check for native support (e.g., `if (!('onbeforematch' in HTMLElement.prototype))`) and execute a fallback UI strategy for unsupported browsers.
--- a/guides/user-experience/search-hidden-content/expectations.md
+++ b/guides/user-experience/search-hidden-content/expectations.md
@@ -1,7 +1,7 @@
 # Expectations: `search-hidden-content`
 
 - Any content intended to be visually hidden but remain searchable via the browser's native "Find in page" feature MUST use `<details>` or the `hidden="until-found"` attribute.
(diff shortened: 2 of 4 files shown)


guides/user-experience/search-hidden-content/guide.md
```diff
@@ -4,3 +4,94 @@ description: Hide content from view using patterns such as accordions, tabs, and
...
+Expand all hidden content for unsupported browsers, for example, for accordions, expand all sections.
+
+```javascript
+if (!('onbeforematch' in document.body)) {
+    // Expand all hidden content for unsupported browsers
+    document.querySelectorAll('[hidden="until-found"]').forEach(el => el.removeAttribute('hidden'));
+}
+```
+
+#### Use the `<details>` element
```
philipwalton: Is this supposed to be in the fallback section?

guides/user-experience/search-hidden-content/guide.md
```diff
@@ -4,3 +4,94 @@ description: Hide content from view using patterns such as accordions, tabs, and
...
+### Feature detection fallback strategy
+
+Use the `onbeforematch` feature detection to detect if the browser supports the `hidden="until-found"` attribute and `beforematch` event. If not, consider one of the following fallback strategies:
+
+#### Expand all hidden content
+
+Expand all hidden content for unsupported browsers, for example, for accordions, expand all sections.
+
+```javascript
+if (!('onbeforematch' in document.body)) {
```
philipwalton:
```suggestion
if (!('onbeforematch' in HTMLElement.prototype)) {
```

I know the DCC article has what you have here, but I think it's safer to check `HTMLElement.prototype` because it's less likely that someone modified that object.

guides/user-experience/search-hidden-content/guide.md
```diff
@@ -4,3 +4,94 @@ description: Hide content from view using patterns such as accordions, tabs, and
 web-feature-ids:
   - hidden-until-found
 ---
+
+# Search hidden content
+
+Web interfaces often hide content from view for reasons such as improving the user experience, saving space or to improve performance. Patterns such as accordions, tabs and collapsible sections are commonly used to achieve this using `display: none` or `visibility: hidden` making text invisible to the browser's native "Find in page" (Ctrl+F) tool.
+
+The hidden="until-found" attribute introduces a "matchable" state. It allows content to remain visually concealed while keeping its text nodes indexed for search.
```
philipwalton:
```suggestion
The `hidden="until-found"` attribute introduces a "matchable" state. It allows content to remain visually concealed while keeping its text nodes indexed for search.
```

guides/user-experience/search-hidden-content/guide.md
```diff
@@ -4,3 +4,94 @@ description: Hide content from view using patterns such as accordions, tabs, and
...
+
+## How to implement
+
+1. Apply the `hidden="until-found"` HTML attribute to the elements that contain content that should be hidden from view.
+2. Use the `beforematch` event to synchronise the state of related elements, for example, to close a tab before the hidden tab is opened or to rotate the icon for an accordion.
```
philipwalton: This step is optional, right? I.e. hidden-until-found will work on pages that don't register `beforematch` listeners, correct?

If so, then I'd clarify in these instructions.

philipwalton: I'd also clarify that the target for this event is the element with the `hidden="until-found"` attribute. That wasn't immediately obvious to me from these instructions, and I think the AI will need to know what element to register the event listener on.

guides/user-experience/search-hidden-content/guide.md
```diff
@@ -4,3 +4,94 @@ description: Hide content from view using patterns such as accordions, tabs, and
...
+
+## Example code
```
philipwalton, line 34: I think it would be good to show a very basic example that doesn't use any JavaScript, and then you can also include this accordion-style example. WDYT?

I'm concerned that most people using this API will not need the accordion behavior, so I'd hesitate to have that as the only example.

guides/user-experience/search-hidden-content/guide.md
```diff
@@ -4,3 +4,94 @@ description: Hide content from view using patterns such as accordions, tabs, and
...
+## Best Practices
+
+- **DO** apply borders, padding, and backgrounds to nested wrappers rather than the element with the hidden attribute. 
+- **DO NOT** apply `display` or `visibility` styles to elements with the `hidden="until-found"` attribute.
+- **DO NOT** use `hidden="until-found"` for sensitive or irrelevant data that should not be revealed through search, like a hidden form field or security token.
+- **DO NOT** use `hidden="until-found"` for "screen reader only" content.
+
+## Browser support and fallback strategies
+
+The `hidden="until-found"` attribute and `beforematch` event are not currently supported in all modern browsers (Baseline limited availability), thus a fallback strategy is typically required.
```
philipwalton: It looks like there's some complexity with this API, as `beforematch` [is Baseline](https://developer.mozilla.org/en-US/docs/Web/API/Element/beforematch_event#browser_compatibility), but there's [a bug in Safari](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/hidden#browser_compatibility) affecting the `until-found` attribute. Given that, feature detecting using `beforematch` is not going to be 100% reliable.

That said, I don't think this bug warrants avoiding this API. Especially for a site that is already hiding content, as this API would be a pure enhancement.

I think it's worth given a stronger recommendation to use this feature for sites that are already hiding content and want to unlock a better find-in-page experience. But you could also give the caveat that for site's were find-in-page is critical, they should still feature detect and expand the content when `beforematch` is not supported.

guides/user-experience/search-hidden-content/prompts.md
```diff
@@ -0,0 +1,5 @@
+# Prompts: `search-hidden-content`
```
philipwalton, line 1: If this is supposed to be a single prompt, then put it all on the same line. Multiple lines is intended for uses cases where we want to test multiple prompts.

guides/user-experience/search-hidden-content/expectations.md
```diff
@@ -0,0 +1,7 @@
+# Expectations: `search-hidden-content`
+
+- Any content that is meant to be visually hidden but remain searchable by the browser's native "Find in page" feature must use the `hidden="until-found"` attribute.
+- Elements utilizing the `hidden="until-found"` attribute must NOT have `display: none`, `visibility: hidden`, or any associated `display` or `visibility` CSS properties applied to them directly.
+- The `hidden="until-found"` attribute must NOT be used to hide sensitive information, internal data tokens, or "screen reader only" text.
+- If there is related UI state (e.g., updating ARIA attributes, toggling open/close classes, or managing accordion icons), this should be synchronized with the JavaScript event listener for the `beforematch` event. 
+- If targeting baseline widely available or baseline newly available, the implementation must include an explicit feature detection check for `hidden="until-found"` support (e.g., `if (!('onbeforematch' in document.body))`) with a fallback strategy.
```
philipwalton:
```suggestion
- If targeting baseline widely available or baseline newly available, the implementation must include an explicit feature detection check for `hidden="until-found"` support (e.g., `if (!('onbeforematch' in HTMLElement.prototype))`) with a fallback strategy.
```

guides/user-experience/search-hidden-content/demo.html
```diff
@@ -0,0 +1,203 @@
...
+        const item = pane.closest('.accordion-item');
+        const button = item.querySelector('.accordion-header');
+        button.setAttribute('aria-expanded', 'true');
+        item.classList.add('is-open');
+      });
+    });
+
+    // 3. Fallback strategy for browsers without 'beforematch' support
+    // (We auto-expand all hidden content)
+    if (!('onbeforematch' in document.body)) {
```
philipwalton:
```suggestion
    if (!('onbeforematch' in HTMLElement.prototype)) {
```

philipwalton (changes requested):
Overall I think this guide is looking pretty good and I definitely learned a few things reading it!

There were a few issues I noticed though, and I left comments/suggestions for those.

guides/user-experience/search-hidden-content/guide.md
taraojo: It is intentional, but maybe it's not really a fallback but a progressive enhancement? The `<details>` element provides the search-hidden-content behaviour by default. It collapses the content and the collapsed content can be searched. 

I was thinking for those looking for baseline widely available, this could be an option instead. What do you think?

guides/user-experience/search-hidden-content/prompts.md
taraojo: These are intended to be 3 separate prompts, the phrasing was intended to match what a developer could ask.

guides/user-experience/search-hidden-content/guide.md
philipwalton: Got it. It wasn't clear to me that this behavior worked natively inside a `<details>`.

So if this works natively inside of `<details>`, then why wouldn't that just be the default recommendation? Are there reasons not to use `<details>`?

guides/user-experience/search-hidden-content/guide.md
taraojo: Where would be a good place to give a refactoring type recommendation, perhaps best practices?

The difference in baseline versions makes the feature detection a bit tricky if targeting newly available as `beforematch` will exist, but there is a browser where `hidden=until-found` will not exist. Does it make sense to call this out in the guide? Or perhaps think of a different fallback?

guides/user-experience/search-hidden-content/guide.md
taraojo: `<details>` is good for native, accordion style behaviour. `hidden="until-found"` should be for custom, non-accordion components that need more control over the hiding functionality and styling. We could say this should be the default for accordion patterns?

philipwalton: You can use `<details>` for non-accordion content hide/show toggles as well.

I guess my point was, if using `<details>` fulfills the use case and is the simplest way to implement it, then I think that should be the primary recommendation. You can then offer `hidden=until-found` as a secondary recommendation for situations where you need more control than what's offered by `<details>`.

Note on the styling point that [details-content](https://webstatus.dev/features/details-content) (for styling `<details>` elements) is another newly available feature that wasn't on the list but potentially could be covered here as well, since it's related to this use case.

guides/user-experience/search-hidden-content/guide.md
philipwalton: To be clear, it's not that `hidden=until-found` doesn't it exist (it does), it's that it has a [bug in Safari](https://bugs.webkit.org/show_bug.cgi?id=304174) where the text content is revealed but the browser doesn't scroll to it. And actually, in my testing, Safari *does* scroll to it if you keep pressing "find next", it just doesn't scroll to it on the very first "find next" invocation.

Given that there's no way (AFAICT) to feature detect this bug (and the best we can do is feature detect `beforematch`), I think the fallback you already recommend is sufficient.

guides/user-experience/search-hidden-content/prompts.md
philipwalton: Got it. Re-reading them again I can see that they're distinct. I'd recommend rephrasing so they all start with the same "Create a..." intro, to make that a bit more clear.

guides/user-experience/search-hidden-content/prompts.md
```diff
@@ -0,0 +1,5 @@
...
+
+- Create an accordion component with 3 sections, with the second and third sections hidden by default. Text in the sections should be visible to the browser's native "Find in page" feature.
+- Use tabs to display comparisons of 2 products, I should be able to search for a product name and the tab with that product should be revealed.
```
philipwalton: I'm curious, did you test this prompt to see if it worked?

Specifically I'm wondering if the AI understands that the phrase "I should be able to search" refers to the "find in page" feature. (Same question for the next prompt.)

philipwalton (changes requested):

guides/user-experience/search-hidden-content/prompts.md
taraojo: Yes, I tried with and without the guides.  For each prompt the LLM didn't use `hidden=until-found` without the guide and with it found the guide and used the recommendation. `<details>` for the first prompt, and `hidden=until-found` for the second two prompts.

guides/user-experience/search-hidden-content/guide.md
taraojo: I've restructured all of the sections with a focus on `<details>` it does seem like this would be the best option in most of the simpler use cases. I didn't add details-content as it didn't seem to be required, we can revisit though. I'll resolve for now and any new comment can be made on the new version.

guides/user-experience/search-hidden-content/guide.md
taraojo: Ok, I misunderstood the meaning of limited availability in webstatus.dev. I'll leave the fallback as is.

guides/user-experience/search-hidden-content/expectations.md
```diff
@@ -0,0 +1,7 @@
+# Expectations: `search-hidden-content`
+
+- Any content intended to be visually hidden but remain searchable via the browser's native "Find in page" feature MUST use the `hidden="until-found"` attribute.
```
philipwalton: Add `<details>` here as another option.

guides/user-experience/search-hidden-content/expectations.md
```diff
@@ -0,0 +1,7 @@
...
+- Elements utilizing the `hidden="until-found"` attribute MUST NOT have `display: none`, `visibility: hidden`, or any associated `display` or `visibility` CSS properties applied to them directly.
+- The `hidden="until-found"` attribute MUST NOT be used to hide sensitive information, internal data tokens, or "screen reader only" text.
+- Only if there is related UI state (e.g., updating ARIA attributes, toggling open/close classes, or managing accordion icons), the implementation MUST synchronize this state using a JavaScript event listener for the `beforematch` event.
```
philipwalton: Nit: this phrase reads awkwardly to me. I think it would be clearer to phrase it something like:

"If the hidden content has related UI state ... that state MUST be synchronized using a `beforematch` event listener."

That said, these expectations aren't for human consumption, so it's not a huge deal either way.

guides/user-experience/search-hidden-content/expectations.md
```diff
@@ -0,0 +1,7 @@
+# Expectations: `search-hidden-content`
+
+- Any content intended to be visually hidden but remain searchable via the browser's native "Find in page" feature MUST use the `hidden="until-found"` attribute.
+- Elements utilizing the `hidden="until-found"` attribute MUST NOT have `display: none`, `visibility: hidden`, or any associated `display` or `visibility` CSS properties applied to them directly.
+- The `hidden="until-found"` attribute MUST NOT be used to hide sensitive information, internal data tokens, or "screen reader only" text.
+- Only if there is related UI state (e.g., updating ARIA attributes, toggling open/close classes, or managing accordion icons), the implementation MUST synchronize this state using a JavaScript event listener for the `beforematch` event. 
+- Only if the target is baseline widely available or baseline newly available, a fallback strategy MUST be present. The implementation MUST include an explicit JavaScript feature detection check for native support (e.g., `if (!('onbeforematch' in HTMLElement.prototype))`) and execute a fallback UI strategy for unsupported browsers.
```
philipwalton: Similar comment here as above. I'd rephrase this something like: "If the users Baseline target is older than Baseline 2025, a fallback strategy MUST be used..."

guides/user-experience/search-hidden-content/guide.md
```diff
@@ -4,3 +4,128 @@ description: Hide content from view using patterns such as accordions, tabs, and
 web-feature-ids:
   - hidden-until-found
 ---
+
+# Search hidden content
+
+Web interfaces often hide content from view to improve the user experience, save screen space, or increase page performance. Traditional methods like `display: none` or `visibility: hidden` completely remove text from the browser's search index.
+
+In most cases, you **SHOULD** use the native HTML `<details>` element to resolve this issue (for example, for accordions, tabs, and "Read more" sections). Modern browsers natively support "Find in page" for closed `<details>` elements.
```
philipwalton: This paragraph recommends using `<details>` before explaining that `<details>` is a solution to the use case. I'd recommend structuring this section a bit differently. Something like:

> Web interfaces often hide content from view to improve the user experience, save screen space, or increase page performance. Traditional methods like `display: none` or `visibility: hidden` work to hide content visually, but they also make that content completely inaccessible to screen readers and browser features like "Find in page".
>
> To hide content visually but still allow it to be searchable by users, you can use either the HTML `<details>` element or the `hidden="until-found"` attribute. The `<details>` element is generally recommended as it's simpler to implement and maintain, but there are some more complex cases where `<details>` is not sufficient and `hidden="until-found"` is required.
>
> For example:
>- If you want full control over the styling of the show/hide mechanism.
>- If the UI controls to show/hide the content are in another part of the DOM
>- If you don't want to support hiding the content after it's shown

And then the next sections can explain how to use each, though I'd start with `<details>` since that's the primary recommendation.

philipwalton (changes requested):

guides/user-experience/search-hidden-content/guide.md
taraojo: I initially didn't add a `How to implement <details>` because I assume models would already know how to implement features that have been around for a long time. WDYT?

philipwalton: As we discussed today, let's include a very brief example showing the `<details>` element, just to make sure its omission doesn't signal to the model that `hidden-until-found` should be preferred.

We can re-evaluate whether its needed better once we have the evals done.

guides/user-experience/search-hidden-content/guide.md
```diff
@@ -4,3 +4,131 @@ description: Hide content from view using patterns such as accordions, tabs, and
 web-feature-ids:
   - hidden-until-found
```
philipwalton, line 7:
```suggestion
  - details
  - details-name
  - hidden-until-found
```

Given that both of these features are mentioned in the guide.

philipwalton (changes requested):

guides/user-experience/search-hidden-content/guide.md
taraojo: I added a 1-liner since there is already a code example below.

philipwalton (approved):
@taraojo I pushed a few updates myself. If these LGTY then feel free to merge.

FYI: I tried switching the fallback to a CSS-based approach to make it more resilient, but then I ultimately reverted it once I saw your demo and realized that the fallback needed to also account for `aria` state. Another benefit of having a demo!

PR #149, opened 2026-02-25
gradergen fixes, enable run-agent, build mcp with guides/, remove ignored folders (harness/results/)

micahjo7:
1. Removes all of `harness/results/` since that is part of `.gitignore` now, except one `example-result` so that the eval-view playwright test that runs with GH checks still passes.
2. Remove `serving/mcp-server/guides/` because it was causing confusion with guide authors. ***Updates `pnpm build:mcp` to build all guide.md files in `guides/`***.
3. Remove files in `skills` dir since we are not using it anymore and can cause confusion with guide authors. If we want to test it in the future, the infrastructure is still there, just need to add the updated SKILL files.
4. Remove `tasksToComplete` and `expectedGuides` from the config. `tasksToComplete` was unused. Updated guide validation to use the task associated with the corresponding run instead of specifying `expectedGuides`.
5. Updates the agent prompts for grader-gen and negative-gen to be more compatible with generated tests.
6. Update addendum to prompt in `harness/run_suite` to force single HTML file output.
7. Restore `run-agent` functionality to test single guides.

cc @rviscomi


guides/negative-gen.ts
```diff
@@ -41,6 +41,7 @@ Then read the demo.html file, which represents a perfect working example of the
 With this information, create another example file that represents an anti-example of what is outlined in guide.md and expectations.md.
 Make sure that it does not fulfill anything that the expectations and guide suggest.
 
+Within the generated code, do not include any comments, and do not indicate in any way that this is a negative example.
```
paulirish, line 44: 👍

paulirish (approved):
that is a delightful PR
thank you

PR #150, opened 2026-02-26
GH action and script to manage issues for use cases

rviscomi:
This is the script I used to create https://github.com/GoogleChrome/guidance/issues?q=is%3Aissue%20state%3Aopen%20label%3Anew-use-case

It requires a `GH_PAT` secret in your `.env` to run locally. I've also added it as a repo secret. The token needs RW permissions for repo issues and org projects.

It will run in dry_run mode on PRs to act as a validator, causing errors if required frontmatter fields are missing or invalid. It replaces [guide.test.ts](https://github.com/GoogleChrome/guidance/blob/main/guides/guide.test.ts) as a way to test guide validity, with the added benefits of running in a GH action and previewing changes to the issues.

After PRs are merged, the script will run for real (not dry_run) and create, reopen, modify, and/or close issues for each use case as needed.


rviscomi:
@paulirish @matthiasrohmer you good for me to merge this as-is and we can iterate on it later, or do you want to review it more closely first?

matthiasrohmer:
I had a look! I didn't feel 100% qualified to actually speak on it 😅 

LGTM in general. Style wise I would have probably asked to break up guides/sync-use-cases.ts a little more but don't feel strongly at all. Also style wise I think I would have made DRY_RUN the default, and checked for any of the GITHUB_* env variables (or workflow names even) to make it a real run.

matthiasrohmer (approved):

paulirish (approved):

PR #152, opened 2026-02-26
Use case for accent-color

syntxerror:
Addresses #37 . Add use case for `accent-color`


guides/user-experience/pwa-native-parity/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: "PWA Native Parity"
+description: "Synchronize internal UI form styling with the external operating system's overall theme to create a deeply integrated, seamless application environment that mimics a true native mobile experience."
```
philipwalton: I think something about "brand colors" should be mentioned here, correct (IIUC)?

philipwalton:
@syntxerror FYI I added a demo and reworded the use case slightly.

Issue #87, opened 2026-02-19
View transitions


rviscomi:
View transitions allow you to create animated visual transitions between different states of a document.

### MDN Documentation
- [View Transition API](https://developer.mozilla.org/docs/Web/API/View_Transition_API)

### Chrome Documentation

<details>
<summary>View 9 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/new-in-web-ui-io-2024
- https://developer.chrome.com/en/blog/scoped-view-transitions-feedback
- https://developer.chrome.com/en/blog/spa-view-transitions-land
- https://developer.chrome.com/en/blog/view-transitions-in-2025
- https://developer.chrome.com/en/blog/view-transitions-update-io24
- https://developer.chrome.com/en/docs/web-platform/view-transitions/cross-document
- https://developer.chrome.com/en/docs/web-platform/view-transitions
- https://web.dev/en/articles/animating-between-views
- https://web.dev/en/learn/css/view-transitions-spas
</details>

---
Feature ID: view-transitions

**Use cases:**
- [ ] #1215
- [ ] #1353
- [x] #338
- [x] #334
- [ ] #1226
- [x] #335
- [x] #705
- [ ] #1230

Issue #86, opened 2026-02-19
view-transition-class


rviscomi:
The view-transition-class CSS property sets a name that can be used to apply styles to multiple named view transition pseudo-elements.

### MDN Documentation
- [view-transition-class](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/view-transition-class)

### Chrome Documentation

<details>
<summary>View 17 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/advanced-attr
- https://developer.chrome.com/en/blog/chrome-2024-recap
- https://developer.chrome.com/en/blog/spa-view-transitions-land
- https://developer.chrome.com/en/blog/view-transitions-in-2025
- https://developer.chrome.com/en/blog/view-transitions-update-io24
- https://developer.chrome.com/en/blog/web-at-io24
- https://developer.chrome.com/en/blog/whats-new-css-ui-2023
- https://developer.chrome.com/en/docs/css-ui/css-names
- https://developer.chrome.com/en/docs/css-ui/view-transitions/nested-view-transition-groups
- https://developer.chrome.com/en/docs/devtools/css/animations
- https://developer.chrome.com/en/docs/devtools/css/reference
- https://developer.chrome.com/en/docs/lighthouse/pwa/pwa-page-transitions
- https://developer.chrome.com/en/docs/web-platform/view-transitions
- https://developer.chrome.com/en/docs/web-platform/view-transitions/same-document
- https://web.dev/en/blog/googleio22-recap
- https://web.dev/en/blog/interop-2025
- https://web.dev/en/blog/same-document-view-transitions-are-now-baseline-newly-available
</details>

---
Feature ID: view-transition-class

**Use cases:**
- [x] #335

Issue #38, opened 2026-02-19
Active view transition


rviscomi:
The :active-view-transition CSS pseudo-class matches the root element when a view transition is active. The :active-view-transition-type() CSS pseudo-class matches only when the active view transition was started with the specified type.

### MDN Documentation
- [:active-view-transition](https://developer.mozilla.org/docs/Web/CSS/Reference/Selectors/:active-view-transition)

### Chrome Documentation

<details>
<summary>View 16 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/chrome-2024-recap
- https://developer.chrome.com/en/blog/css-ui-ecommerce-vt
- https://developer.chrome.com/en/blog/google-ai-mode-view-transitions-case-study
- https://developer.chrome.com/en/blog/new-in-web-ui-io-2024
- https://developer.chrome.com/en/blog/spa-view-transitions-land
- https://developer.chrome.com/en/blog/view-transitions-case-studies
- https://developer.chrome.com/en/blog/view-transitions-in-2025
- https://developer.chrome.com/en/blog/view-transitions-misconceptions
- https://developer.chrome.com/en/blog/view-transitions-update-io24
- https://developer.chrome.com/en/docs/css-ui/view-transitions/nested-view-transition-groups
- https://developer.chrome.com/en/docs/devtools/css/reference
- https://developer.chrome.com/en/docs/lighthouse/pwa/pwa-page-transitions
- https://developer.chrome.com/en/docs/web-platform/view-transitions/cross-document
- https://developer.chrome.com/en/docs/web-platform/view-transitions
- https://web.dev/en/blog/same-document-view-transitions-are-now-baseline-newly-available
- https://web.dev/en/learn/css/view-transitions-spas
</details>

---
Feature ID: active-view-transition

**Use cases:**
- [x] #334

PR #153, opened 2026-02-26
Add use cases related to view transition

kevinkiklee:
This PR is for the following issues: 
* View Transition API - https://github.com/GoogleChrome/guidance/issues/87
* view-transition-class - https://github.com/GoogleChrome/guidance/issues/86
* :active-view-transition - https://github.com/GoogleChrome/guidance/issues/38

---

## Use cases

| Use case  | Description |
| ------------- | ------------- |
| element-transition-on-cross-document-navigation  | Animate the entire screen state when a user navigates between distinct HTML documents, such as creating seamless cross-fades or custom page-level slide effects without relying on single-page application architectures. |
| element-transition-on-same-document-navigation  | Animate visual state changes when dynamically updating the DOM within a single-page application or interactive component, such as preventing abrupt visual jumps when filtering lists, switching tabs, or opening dialogs.  |
| shared-element-transition-on-navigation  | Visually connect persisting interface components across different navigation states by smoothly morphing their size, position, and styling, such as expanding a product thumbnail into a full-bleed hero image or uniformly animating an entire grid of items. |
| directional-element-transition-on-navigation  | Animate visual state changes to reflect the direction of a user's navigational flow, such as sliding new content in from the right when advancing forward or from the left when returning to a previous screen. |


kevinkiklee:
DO NOT REVIEW.  

I will be adding the demo.html files

rviscomi:
Converting to draft mode while the demo.html files are created

kevinkiklee:
* Added the demo files and the PR is ready for review.  
* Added @bramus as FYI (any insights/advices will be highly appreciated)

guides/user-experience/directional-element-transition-on-navigation/demo.html
```diff
@@ -0,0 +1,145 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="UTF-8">
+  <title>Direction-Aware Transitions</title>
+  <style>
+    /* CORE: View Transitions API */
+    html:active-view-transition-type(forward) ::view-transition-old(root) {
```
bramus: This code won’t work. The pseudo’s (ultimate) originating element is the `html` element, so you need to glue the pieces together:


```suggestion
    html:active-view-transition-type(forward)::view-transition-old(root) {
```

guides/user-experience/directional-element-transition-on-navigation/demo.html
```diff
@@ -0,0 +1,145 @@
...
+      animation: slide-left 0.4s ease-out forwards;
+    }
+
+    html:active-view-transition-type(forward) ::view-transition-new(root) {
```
bramus:
```suggestion
    html:active-view-transition-type(forward)::view-transition-new(root) {
```

guides/user-experience/directional-element-transition-on-navigation/demo.html
```diff
@@ -0,0 +1,145 @@
...
+      animation: slide-from-right 0.4s ease-out forwards;
+    }
+
+    html:active-view-transition-type(backward) ::view-transition-old(root) {
```
bramus:
```suggestion
    html:active-view-transition-type(backward)::view-transition-old(root) {
```

guides/user-experience/directional-element-transition-on-navigation/demo.html
```diff
@@ -0,0 +1,145 @@
...
+      animation: slide-right 0.4s ease-out forwards;
+    }
+
+    html:active-view-transition-type(backward) ::view-transition-new(root) {
```
bramus:
```suggestion
    html:active-view-transition-type(backward)::view-transition-new(root) {
```

bramus (commented):
Just a minor code issue in the demo. The rest looks good.

guides/user-experience/directional-element-transition-on-navigation/demo.html
kevinkiklee: yea, just tested everything out and you are absolutely right.  thanks a ton for catching this.

guides/user-experience/shared-element-transition-on-navigation/demo.html
philipwalton, line 1: When I run this demo locally, I don't see any transitions. Can you double-check?

guides/user-experience/directional-element-transition-on-navigation/guide.md
```diff
@@ -0,0 +1,7 @@
+---
+name: directional-element-transition-on-navigation
```
philipwalton:
```suggestion
name: directional-navigation-transitions
```
Nit, but this is shorter and I think reads clearer.

Also, I think the "on navigation" part is a bit confusing because these aren't necessarily "navigations" in the browser page, and the "on" prefix kinda makes it sound like you're listening to a navigate event (at least to me).

guides/user-experience/directional-navigation-transitions/guide.md
```diff
@@ -0,0 +1,7 @@
+---
+name: directional-element-transition-on-navigation
+description: Animate visual state changes to reflect the direction of a user's navigational flow, such as sliding new content in from the right when advancing forward or from the left when returning to a previous screen.
+web-feature-ids:
+  - view-transitions
+  - active-view-transition
```
philipwalton, line 6: Question: is the `active-view-transition` feature required to implement this use case (i.e. does it need to be part of the guide), or does it just happen to be used in your demo?

If it needs to be present in the guide then it's fine to have this feature in the list, otherwise take it out.

guides/user-experience/element-transition-on-cross-document-navigation/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: element-transition-on-cross-document-navigation
+description: Animate the entire screen state when a user navigates between distinct HTML documents, such as creating seamless cross-fades or custom page-level slide effects without relying on single-page application architectures.
+web-feature-ids:
+  - view-transitions
```
philipwalton:
```suggestion
  - view-transitions
  - cross-document-view-transitions
```

BTW, I think it makes sense to include both same-doc and cross-doc view transitions in this use case (since the choice of tech is really just an implementation detail).

Also, I noticed that @taraojo put her name down for `cross-document-view-transitions` in the feature sheet. Tara, have you started on this? If not, maybe we should just use this PR to cover both types of VT use cases? There's a lot of overlap in use cases.

guides/user-experience/element-transition-on-cross-document-navigation/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: element-transition-on-cross-document-navigation
```
philipwalton:
```suggestion
name: seamless-full-page-transitions
```

guides/user-experience/element-transition-on-cross-document-navigation/guide.md
```diff
@@ -0,0 +1,6 @@
...
+description: Animate the entire screen state when a user navigates between distinct HTML documents, such as creating seamless cross-fades or custom page-level slide effects without relying on single-page application architectures.
```
philipwalton:
```suggestion
description: Create smooth, seamless transitions between full page navigations, such as cross-fades, custom reveal effects, or morphing of content from one page to the next.
```

Cross document view transitions don't necessarily have to animate the entire page. Also, I'd remove the SPA mention, since it's kinda an implementation details.

guides/user-experience/shared-element-transition-on-navigation/guide.md
```diff
@@ -0,0 +1,7 @@
+---
+name: shared-element-transition-on-navigation
```
philipwalton:
```suggestion
name: shared-element-transitions
```

guides/user-experience/shared-element-transition-on-navigation/guide.md
```diff
@@ -0,0 +1,7 @@
...
+description: Visually connect persisting interface components across different navigation states by smoothly morphing their size, position, and styling, such as expanding a product thumbnail into a full-bleed hero image or uniformly animating an entire grid of items.
+web-feature-ids:
+  - view-transitions
+  - view-transition-class
```
philipwalton: Question: is `view-transition-class` the recommended way to implement this use case, or is this just an example of where you *can* use `view-transition-class`?

guides/user-experience/same-document-transitions/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: element-transition-on-same-document-navigation
+description: Animate visual state changes when dynamically updating the DOM within a single-page application or interactive component, such as preventing abrupt visual jumps when filtering lists, switching tabs, or opening dialogs.
```
philipwalton: Is this the same as the previous use case, but for same-doc instead of cross-doc? If so, then I'd combine them.

philipwalton (changes requested):
Thanks for putting this together @kevinkiklee. This is a really broad topic, but I think the use cases you've identified do a good job of capturing the core uses cases at the right granularity.

I've left a few comments to be addressed, but overall I think this is pretty good.

guides/user-experience/element-transition-on-cross-document-navigation/guide.md
taraojo: I didn't get a notification for this mention but just noticed there was a lot overlap, I'll reassign to you @kevinkiklee

guides/user-experience/directional-navigation-transitions/guide.md
```diff
@@ -0,0 +1,7 @@
+---
+name: directional-element-transition-on-navigation
+description: Animate visual state changes to reflect the direction of a user's navigational flow, such as sliding new content in from the right when advancing forward or from the left when returning to a previous screen.
+web-feature-ids:
+  - view-transitions
```
taraojo, line 5:
```suggestion
  - view-transitions
  - cross-document-view-transitions
```

I also had this example for cross-document, it looks like the guidance would be similar. It would just give direction on how to choose the right one.

guides/user-experience/shared-element-transition-on-navigation/guide.md
```diff
@@ -0,0 +1,7 @@
+---
+name: shared-element-transition-on-navigation
+description: Visually connect persisting interface components across different navigation states by smoothly morphing their size, position, and styling, such as expanding a product thumbnail into a full-bleed hero image or uniformly animating an entire grid of items.
+web-feature-ids:
+  - view-transitions
```
taraojo:
```suggestion
  - view-transitions
  - cross-document-view-transitions
```

taraojo (commented):
I didn't get very far with the `cross-document-view-transitions` but they are mostly the same as what you have. Feel free to take a look at the commit https://github.com/GoogleChrome/guidance/commit/a300a215067906e7278fa64db5423131120edd91 though it is completely AI generated I haven't edited it or checked if the demos work!

guides/user-experience/directional-navigation-transitions/guide.md
kevinkiklee: Yes, the main part of this use case is implement animation based on the **direction** the user is navigating, and the primary feature that enables that is active-view-transition.

Note that the actual feature being used is `active-view-transition-type` was mentioned in the `active-view-transition` Issue (https://github.com/GoogleChrome/guidance/issues/38) so I've grouped them together.  Let me know if I shouldn't.

guides/user-experience/shared-element-transition-on-navigation/demo.html
kevinkiklee: Thanks for catching it.  I ran the demo code through AI multiple times to clean it up and simplify it, and looks like it broke and I didn't catch it during QA.    

However, I am rethinking through this use case, and I will likely scrap this one, and create a new demo.

guides/user-experience/element-transition-on-cross-document-navigation/guide.md
kevinkiklee: didn't realize cross-document-view-transition was there.  thanks for bringing it to my attention.  i'll include it here.

guides/user-experience/same-document-transitions/guide.md
kevinkiklee: so, i'm still not sure exactly how to approach this.  there is the `cross-document-view-transitions` that tara brought to my attention, and it seems like the cross-document demo above (the `element-transition-on-cross-document-navigation` one) is a good place to add that feature.  and if we add that feature there, then i feel like grouping SPA code under that use case feels awkward.  

i'm still leaning towards having a separate demo for SPA.  let me know how you feel.

guides/user-experience/shared-element-transition-on-navigation/guide.md
kevinkiklee: This use case was meant to show something that can only be easily done with `view-transition-class`, however, thinking it through more, I don't think my demo is a good representation.  

I am rethinking through this use case, and I'll update this one.

kevinkiklee (commented):
I am rethinking through one of the use cases, and will update it.

kevinkiklee:
i've modified the use cases and demo slightly: 

| Use case  | Reasoning |
| ------------- | ------------- |
| same-document-transitions | This use case utilizes only the barebones of the VT API, and is the most basic use case |
| cross-document-page-transitions | This is the basic use case of VT + `@view-transition` |
| shared-element-transitions | This is the basic use case of VT + `view-transition-class` to showcase that a group of elements can share a view transition animation |
| directional-navigation-transitions | This showcases that VT + `active-view-transition-type()` can be used to add transition animation based on which direction the user is traversing |

guides/user-experience/directional-navigation-transitions/guide.md
philipwalton: Got it, thanks!

guides/user-experience/same-document-transitions/guide.md
philipwalton: I agree it's a bit complicated in the case of View Transitions. The problem is that many (though not all) use cases can be implemented with either cross-doc VTs or same-doc VTs, and I want to enable an AI tool to implement a given use case regardless of what architecture the project has set.

That said, I do think there will devs asking explicitly asking their AI tool for cross-document view transitions—and I also think there are use cases where, even though they *can* be implemented with cross-doc VTs, the better choice is sam-doc VT (e.g. animating the removal of an item from a list of items).

So why don't we just stick to those use cases that are more clear cut, and drop the vague ones, e.g. this one.

I'll make a separate suggestion for a new use case to add.

guides/user-experience/shared-element-transitions/demo.html
```diff
@@ -0,0 +1,92 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <meta charset="UTF-8">
+  <title>Shared Element Transitions</title>
+  <style>
+    /* CORE: All list items share the same transition */
+    .item {
+      view-transition-class: item;
```
philipwalton: This is not actually doing anything, because you're not defining any view transition styles referencing this class name.

philipwalton: There was also an issue where the items weren't actually transitioning positions because they didn't have a unique VT name. I got Gemini to fix it and I pushed https://github.com/GoogleChrome/guidance/pull/153/commits/2c55ca6b8cfed2e0b610adbd4e12dfcd8ea7451c

guides/user-experience/shared-element-transitions/guide.md
```diff
@@ -0,0 +1,7 @@
+---
+name: shared-element-transitions
+description: Apply uniform transition animations across multiple distinct elements, such as items in a dynamic list or grid, without needing to write or dynamically generate unique CSS selectors for every individual element's view transition name.
```
philipwalton: I think the multiple elements simultaneously transitioning is a good use case, (and I think we should keep this demo), but I actually think that's a separate use case from the classic "shared element" VT, which is transitioning something like an image thumbnail into the full-res hero image when going from a product in a grid view to the product details view.

Can you add a new use case so both are captured? Maybe with names:
- `shared-element-transitions`
- `multiple-element-tarnsitions`

WDYT?

kevinkiklee: That sounds good and I've added a demo/guidance for `multiple-element-transition`,

guides/user-experience/same-document-transitions/guide.md
kevinkiklee: Understood.  The new use case you suggest covers similar concepts in this use case.  I'll remove this one.

guides/user-experience/shared-element-transitions/demo.html
kevinkiklee: Apologies.  I realize that I wasn't QA-ing it properly.  I reverted it and fixed the code to continue to use `view-transition-class` since that's the main objective of this use case.  Thanks for catching it.

kevinkiklee (commented):
* Added `multiple-element-transitions` use case
* Fixed `shared-element-transitions` demo
* Removed `same-document-transitions` use case

philipwalton (approved):
LGTM. Tweaked the use case wording and demos slightly.

Issue #44, opened 2026-02-19
contain-intrinsic-size


rviscomi:
The contain-intrinsic-size CSS property sets the intrinsic size of an element. When using size containment, the browser will lay out the element as if it had a single child of this size.

### MDN Documentation
- [contain-intrinsic-size](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/contain-intrinsic-size)

### Chrome Documentation

<details>
<summary>View 8 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/css-containment
- https://developer.chrome.com/en/blog/layoutNg-2
- https://developer.chrome.com/en/blog/masonry
- https://developer.chrome.com/en/docs/devtools/css/reference
- https://web.dev/en/articles/carousel-best-practices
- https://web.dev/en/articles/content-visibility
- https://web.dev/en/learn/css/backgrounds
- https://web.dev/en/learn/css/box-model
</details>

---
Feature ID: contain-intrinsic-size

**Use cases:**
- [ ] #311

Issue #43, opened 2026-02-19
Inline-size containment


rviscomi:
The contain: inline-size CSS declaration prevents the element's inline dimension from being set by the element's contents. This permits the browser to avoid slower layout calculations.

### MDN Documentation
- [contain](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/contain)

### Chrome Documentation

<details>
<summary>View 15 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/cq-polyfill
- https://developer.chrome.com/en/blog/css-containment
- https://developer.chrome.com/en/blog/css-ui-ecommerce-cq
- https://developer.chrome.com/en/blog/inside-the-container-query-polyfill
- https://developer.chrome.com/en/docs/devtools/css/reference
- https://developer.chrome.com/en/docs/performance/insights/dom-size
- https://developer.chrome.com/en/docs/performance/insights/slow-css-selector
- https://web.dev/en/articles/avoid-large-complex-layouts-and-layout-thrashing
- https://web.dev/en/articles/building/a-multi-select-component
- https://web.dev/en/articles/howbrowserswork
- https://web.dev/en/blog/css-content-visibility-baseline
- https://web.dev/en/blog/how-to-use-container-queries-now
- https://web.dev/en/case-studies/netflix-cq
- https://web.dev/en/learn/css/box-model
- https://web.dev/en/learn/css/container-queries
</details>

---
Feature ID: contain-inline-size

PR #154, opened 2026-02-26
Use cases for contain-intrinsic-size and contain-inline-size

syntxerror:
Addresses #43 and #44. Add use cases for `contain-intrinsic-size` and `contain-inline-size`


rviscomi:
Looks like some CI tests are failing. Could you merge the latest main branch to trigger a rerun?

guides/performance/reserved-space-for-third-party-widgets/expectations.md
taraojo, line 1: Looks like this one was accidentally copied into this expectations file

taraojo (commented):
Thanks Jon, would you be able to add/generate a small demo for each use case? We've started doing this across use cases and it helps the review.

guides/performance/reserved-space-for-third-party-widgets/expectations.md
syntxerror: can fix. thanks for catching.

syntxerror: @taraojo added demos.

guides/user-experience/caption-wrapping-for-visual-media/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: caption-wrapping-for-visual-media
+description: Ensures that descriptive labels remain restricted to the same horizontal footprint as their primary images rather than stretching the container to fit a single line of text.
```
philipwalton: Hmmm, TBH, `contain-inline-size` doesn't seem like the right features to use to achieve this result. I get that it works, but setting a width of `min-content` on the `<figure>` in your demo would also work, and I think would be the right tool for the job.

I'd drop this use case.

guides/performance/reserved-space-for-third-party-widgets/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: reserved-space-for-third-party-widgets
+description: Create designated areas for third-party widgets that stay occupied by a default height and width even before external scripts or advertisements have finished downloading.
```
philipwalton: FYI, your demo uses `contain: strict` instead of `contain: inline-size`.

Also, would the best way to reserve space for 3P widget be to specify a width an a height (which your demo also does)?

If there isn't a good reason, then I think we should drop this one.

guides/performance/scrollbar-stabilization-for-infinite-feeds/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: scrollbar-stabilization-for-infinite-feeds
+description: Provides a temporary placeholder for off-screen components, ensuring the user's position remains steady as content is rendered.
```
philipwalton: This is a good use case, and I think you can merge the first one into this one. Can you update it to be the following, and also add a demo?

`optimize-scroll-stability`: Keep the page layout, scroll bar, and scroll position as stable as possible as users scroll through content-heavy pages (feeds, dashboards, long lists) where elements transition from hidden to visible.

Also, the more I think about it, the more I think that we don't need to create use cases for `contain-inline-size`, and just stick to `contain-intrinsic-size`.

I think the `contain` property in CSS would be good to mention in a general performance skill, but probably doesn't need its own use case.

guides/performance/rendering-complex-tables/guide.md
```diff
@@ -0,0 +1,7 @@
+---
+name: rendering-complex-tables
+description: Informs the browser it can calculate the width of a specific section independently, preventing a single update from forcing a recalculation of the entire document's horizontal layout.
```
philipwalton: I mention this in the comment below, but I'd merge this use case into the 3rd one.

philipwalton (changes requested):

guides/user-experience/caption-wrapping-for-visual-media/guide.md
syntxerror: Removed in latest commit

guides/performance/scrollbar-stabilization-for-infinite-feeds/guide.md
syntxerror: SG. Will update now

guides/performance/rendering-complex-tables/guide.md
syntxerror: Merged in latest commit

guides/performance/reserved-space-for-third-party-widgets/guide.md
syntxerror: I think given that because it can be accomplished with `contain: strict` its not explicitly niche enough to warrant the use case. I'll drop it.

philipwalton (approved):
LGTM. I like the demo!

guides/performance/optimize-scroll-stability/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: optimize-scroll-stability
+description: Keep the page layout, scroll bar, and scroll position as stable as possible as users scroll through content-heavy pages (feeds, dashboards, long lists) where elements transition from hidden to visible.
+web-feature-ids:
+  - contain-intrinsic-size
```
rviscomi, line 5: @syntxerror should this also include `contain-inline-size`? The [issue](https://github.com/GoogleChrome/guidance/issues/43) associated with that feature got closed, but it's unclear if it's because there are no use cases for it or if it was a mistake.

syntxerror: I was following a recommendation from @philipwalton we likely do not need to support `contain-inline-size` use cases. See [here](https://github.com/GoogleChrome/guidance/pull/154#discussion_r2932885774)

rviscomi: Ah I see, thanks!

PR #155, opened 2026-02-26
Update guide validation to use mcp tool call logging directly

micahjo7:
instead of the agent INSTRUCTIONS.md file

also removes `use-cases.gen.ts` and `harness/results/` and makes necessary changes to pass GitHub CI checks

removing these files allows them to be properly ignored by git so that GH checks don't fail whenever they are regenerated


paulirish:
image
micah's on fire.

paulirish (approved):

PR #156, opened 2026-02-27
mcp server error logging

paulirish:
if the mcp server fails i wanted it to be more obvious when running the eval suite.

now any console.errors (or exceptions) will end up in the output.  
eg

image

### DIFF
--- a/harness/agents/gemini-cli-agent.ts
+++ b/harness/agents/gemini-cli-agent.ts
@@ -5,7 +5,7 @@ import { spawn } from 'child_process';
 
 import config, { Agents } from '../config.ts';
 
-import { updateMcpConfig, createIsolatedHome, cleanupIsolatedHome, copyFileIfExists, parseAgentArgs, createWorkDir, copyResultsToTarget, copySkills } from '../lib/agent-shared.ts';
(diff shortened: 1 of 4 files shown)


paulirish:
@micahjo7 i now see your mcp_log_dir. do you want to just share a single log file for all of the server? and we can parse out logs/errors/toolcalls on the receiving end?  i'm cool with whatever.

paulirish:
> @micahjo7 i now see your mcp_log_dir. do you want to just share a single log file for all of the server? and we can parse out logs/errors/toolcalls on the receiving end? i'm cool with whatever.

resolved it i think.

harness/agents/gemini-cli-agent.ts
```diff
@@ -78,6 +78,10 @@ async function run() {
 
     console.log(`Executing: ${command} ${commandArgs.join(' ')}`);
 
+    process.env.MCP_LOG_DIR = targetDir;
+    const stopWatchingErr = watchLogFile(path.join(targetDir, 'mcp-server-error.log'));
+    const stopWatchingTool = watchLogFile(path.join(targetDir, 'mcp_tool_calls.log'));
+
```
micahjo7, line 84: should we do this for the other agents too?

micahjo7 (approved):

Issue #79, opened 2026-02-19
Temporal


rviscomi:
The Temporal API allows you to work with dates, times, time zones, and durations. It is more powerful than the Date API.

### MDN Documentation
- [Temporal](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Temporal)

---
Feature ID: temporal

**Use cases:**
- [ ] #375
- [ ] #376
- [x] #377
- [x] #378
- [x] #379
- [x] #380
- [x] #381
- [x] #382
- [ ] #374

sffc:
Just to check, did you reference the official TC39 Temporal cookbook when writing these guides?

https://tc39.es/proposal-temporal/docs/cookbook.html

NourNabil:
When creating the use cases I mainly depended on the MDN documentation.

PR #157, opened 2026-02-27
Use cases for Temporal API

NourNabil:
Use cases for Temporal API for issue #79. PTAL and let me know what you think. I am also open to changes regarding the high level disciplines for these use cases.


philipwalton:
@malchata can you review this one?

guides/performance/sequence-distributed-events/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: sequence-distributed-events
+description: Log and sequence operations in distributed microservices or high-throughput tracing environments. Utilize Temporal.Instant to record timestamps with nanosecond resolution, ensuring chronological accuracy and preventing sorting collisions in high-frequency telemetry.
```
taraojo: Remove references to the specific API functions and features as the description should focus on what the user is trying to achieve and not the implementation details. That will come later in the guide. Same comment for the other use cases too.

guides/user-experience/format-human-readable-durations/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: format-human-readable-durations
+description: Manage how time spans are displayed to users by controlling "balancing" logic. This allows developers to choose between unit-specific counts (e.g., "90 minutes") or wrapped human-friendly formats (e.g., "1 hour and 30 minutes") during formatting and display
```
taraojo: Similar comment, the description should be an overview of the use case, where as this feels like a description of the API feature.

taraojo (commented):
Thanks Nour, I've added comments that have been common across other PRs. Can you update to also add/generate a small demo for each use case? We've started doing this across use cases and it helps the review, the demos may also help us see where we can combine some of these use cases as there are quite a few here.

guides/user-experience/calculate-event-differentials/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: calculate-event-differentials
+description: Determine the exact time elapsed between two discrete events to calculate trial expirations or prorated costs. levarage the .since() and .until() methods to return balanced duration objects, replacing error-prone manual timestamp division.
```
malchata: Similar concern here to what @taraojo has mentioned in another use case. Additionally, words like "leverage" in this case inhibit clarity, where "use" would be more direct. In fact, the second sentence might be (mostly) unnecessary:

```suggestion
description: Determine the exact time elapsed between two discrete events to calculate trial expirations or prorated costs while avoiding error-prone manual timestamp division.
```

guides/user-experience/manage-financial-intervals/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: manage-financial-intervals
+description: Perform complex date mathematics for recurring subscriptions or payroll. Use Temporal.Duration to add or subtract months and years while utilizing native "overflow" clamping to safely handle edge cases like adding one month to January 31st.
```
malchata: Would prefer to avoid mentions of discrete methods and/or properties (`Temporal.Duration` here). I also believe that the specific example here ("edge cases like adding one month to January 31st"). I think the use case could be written to communicate the benefit of "overflow clamping" without adding too much to the overall specificity of the use case.

guides/user-experience/stabilize-reactive-state/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: stabilize-reactive-state
+description: Prevent reference-based bugs in UI frameworks (like React or Vue) by utilizing immutable chronological objects. Every update yields a brand-new memory reference, ensuring framework diffing algorithms correctly trigger re-renders without silent state mutations.
```
malchata: I _think_ we want to avoid mentioning specific frameworks—my concern is that generated "best practices" guidance could key in on mentions of them and possibly provide guidance that is too narrow (but maybe we're okay with that).

malchata (changes requested):
The overarching concern here, I think, is the mention of specific API methods and properties, and to keep the use cases general enough to focus on the problem without wading into implementation details.

NourNabil:
> Thanks Nour, I've added comments that have been common across other PRs. Can you update to also add/generate a small demo for each use case? We've started doing this across use cases and it helps the review, the demos may also help us see where we can combine some of these use cases as there are quite a few here.



> Thanks Nour, I've added comments that have been common across other PRs. Can you update to also add/generate a small demo for each use case? We've started doing this across use cases and it helps the review, the demos may also help us see where we can combine some of these use cases as there are quite a few here.

@taraojo I added demoes as request and adjust the descriptions to be more general. PTAL and let me know what you think.

NourNabil:
> The overarching concern here, I think, is the mention of specific API methods and properties, and to keep the use cases general enough to focus on the problem without wading into implementation details.



> The overarching concern here, I think, is the mention of specific API methods and properties, and to keep the use cases general enough to focus on the problem without wading into implementation details.

@malchata Hey Jeremy, I rewrote the descriptions of the use cases to be more general. I tried not to mention specific API methods. PTAL and let me know what you think.

philipwalton (approved):
Thanks @NourNabil! I made a few minor changes to wording, but other than that your latest changes LGTM.

PR #158, opened 2026-02-27
:user-valid and :user-invalid use-cases and guides

### LINKED ISSUE #36: :user-valid and :user-invalid

PaulKinlan:
This is a WIP set of guides for the user-pseudos.

1. [accessible-error-announcement](https://github.com/GoogleChrome/guidance/commit/f7f4077761418de2f372a7481892b613d930adab) - I've tested this with the MacOS screenreader to validate if it works as expected. some other notes.
   1. I also took the decision to only check `:user-invalid` after every keypress after the first failure and while it's failing. The audio didn't get re-announced, and it's nice to get immediate feedback when it passes the validation.
   2. the polyfiling makes this messy, but it's complete.
   3. the intersection between a11y aria and :user-invalid I think makes :user-invalid less useful than people think...
   
2. Guides implemented so far:
+ [x] style-parent-with-has
+ [x] MERGED validate-email-after-interaction with password-complexity-validation
+ [x] accessible-error-announcement
+ [x] required-field-feedback
+ [x] select-menu-interaction

Fixes #36


guides/user-experience/accessible-error-announcement/guide.md
```diff
@@ -0,0 +1,215 @@
...
+
+# Accessible Error Announcement
+
+## The Problem
+Standard HTML5 validation provides visual feedback (via `:invalid` or `:user-invalid`), but it doesn't automatically synchronize with accessibility attributes like `aria-invalid`. 
+
+If you use standard `:invalid` styling, screen readers might announce "Invalid entry" the moment a user tabs into a required field that is currently empty. This creates a disruptive experience for users using assistive technologies, as the error is announced before interaction has occurred.
+
+## The Solution
+We want the *programmatic* state (`aria-invalid="true"`) to be applied **at the exact same moment** the *visual* state (`:user-invalid`) applies. Since `:user-invalid` relies on the browser's internal "user-interacted" flag, we can use JavaScript to check for this selector matches during standard interaction events.
```
philipwalton:
```suggestion
We want the *programmatic* state (`aria-invalid="true"`) to be applied **at the exact same moment** the *visual* state (`:user-invalid`) applies. Since `:user-invalid` relies on the browser's internal "user-interacted" flag, we can use JavaScript to check that this selector matches during standard interaction events.
```

guides/user-experience/accessible-error-announcement/guide.md
```diff
@@ -0,0 +1,215 @@
...
+
+See [MDN aria-invalid](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-invalid) for more details.
+
+### Implementation Strategy
+
+1.  **Visual Layer**: Use CSS `:user-invalid` to show borders/icons.
+2.  **Accessibility Layer**: Use `aria-invalid` and `aria-errormessage` to communicate state to Assistive Technology (AT).
+3.  **Bridge Visual & Accessibility Layer**: A lightweight JavaScript utility that listens for `blur` and `input` events, checks if the element matches `:user-invalid`, and updates the ARIA attributes accordingly.
```
philipwalton:
```suggestion
3.  **Bridge Visual & Accessibility Layer**: Create a lightweight JavaScript utility that listens for `blur` and `input` events, checks if the element matches `:user-invalid`, and updates the ARIA attributes accordingly.
```

guides/user-experience/accessible-error-announcement/guide.md
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/* Optional: Visual cues on the input itself */
+input:user-invalid {
+  border-color: #d93025;
+}
+```
+
+### 3. JavaScript
+Since there is no "UserInvalidChanged" event, we hook into standard form events to check the state.
```
philipwalton:
```suggestion
Since there is no "UserInvalidChanged" event, hook into standard form events to check the state.
```

guides/user-experience/accessible-error-announcement/guide.md
```diff
@@ -0,0 +1,215 @@
...
+
+```javascript
+const inputs = document.querySelectorAll('input, textarea, select');
```
philipwalton: This won't apply to fields that are programmatically added to the DOM.

I think the recommendation should be to add capturing `focus` and `blur` event listeners to the window/document and then check if the target matches `input, textarea, select`

guides/accessibility/accessible-error-announcement/guide.md
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/* Show error message for both cases */
+input:user-invalid ~ .error-msg,
+input.user-invalid-fallback ~ .error-msg {
+  display: block;
+}
+```
+
+### Fallback Logic
```
philipwalton: I guess adding fallback logic is fine, but I do wonder how necessary it is and whether or not this should just be a pure progressive enhancement.

guides/user-experience/accessible-error-announcement/guide.md
```diff
@@ -0,0 +1,215 @@
...
+    root.addEventListener('input', handleEvent);
+    root.addEventListener('change', handleEvent);
+    root.addEventListener('reset', handleEvent, true); // Capture resets
+  };
+
+  return { init };
+})();
+
+// Initialize for a specific form
+const form = document.querySelector('#demo-form');
```
philipwalton: Similar to above, this would need to get re-called any time a form is added to the DOM. Why not just use capturing event listeners for all blur/input/change/reset events, so there's less manually wiring up required.

guides/user-experience/accessible-error-announcement/guide.md
```diff
@@ -0,0 +1,215 @@
...
+
+1.  **`aria-live` vs. `aria-errormessage`**: 
+    *   `aria-errormessage` connects the input to the text, but screen readers might not announce it immediately upon appearance (only when focusing the input).
+    *   If you need *immediate* announcement when the error appears (e.g., on blur), consider adding `role="alert"` or `aria-live="polite"` to the error message container, but test thoroughly to avoid "double announcement" when the user focuses the field to fix it.
+
+2.  **Internationalization**:
+    *   Ensure the text content of your error message (`#email-error`) is translated. The logic remains the same.
+
+3.  **Performance**:
+    *   `input.matches(':user-invalid')` is very fast. Attaching these listeners to hundreds of inputs is generally negligible, but event delegation (listening on the `<form>` element) is a good optimization for large forms.
```
philipwalton: I see you kinda answer my above comments here, but TBH I still think event delegation is a better solution since focus/blur fire relatively infrequently and input/change/reset would only be added in older browsers, so it won't impact most users.

guides/user-experience/required-field-feedback/guide.md
```diff
@@ -0,0 +1,166 @@
+---
+name: required-field-feedback
+description: Highlighting required fields that were skipped or left empty after user focus and blur, ensuring feedback is contextually relevant to the user's flow.
```
philipwalton:
```suggestion
description: Provide error message for required form fields that were skipped or left empty *only* after user interaction, to avoid preemptive errors and ensure feedback is timely and contextually relevant to the user's flow.
```

guides/user-experience/required-field-feedback/guide.md
```diff
@@ -0,0 +1,166 @@
...
+/* Optional: Subtle indicator for required fields that are valid */
+input:required:user-valid {
+  border-color: #188038;
+  border-width: 2px;
+}
+```
+
+## Fallbacking & Browser Support
+
+For older browsers, we need to manually track the "visited" or "dirty" state.
```
philipwalton: Given that these fallbacks add implementation complexity, I think we should frame them as option. How about something like:

If the user's Baseline target is older than Baseline 2023, consider adding a fallback strategy for browsers that don't support `:user-invalid`...

guides/user-experience/required-field-feedback/guide.md
```diff
@@ -0,0 +1,166 @@
...
+// Initialize for a specific form
+const form = document.querySelector('#demo-form');
+UserInvalidFallback.init(form);
+```
+
+## Other Considerations
+
+1.  **Asterisks**: It is still best practice to indicate required fields visually (e.g., with an asterisk `*`) in the label, so users know what to expect *before* they interact.
+2.  **Submit Buttons**: Unlike `disabled` buttons, keep your submit button enabled. If the user clicks it, the browser will automatically trigger `:user-invalid` on all empty required fields and focus the first one. This is excellent for accessibility and UX.
+3.  **Accessibility**: While the fallback script automatically toggles `aria-invalid`, native `:user-invalid` does not automatically sync with ARIA attributes. To ensure a consistent accessibility experience for all users, see the [Accessible Error Announcement](../accessible-error-announcement/guide.md) guide.
```
philipwalton: This reference to the other file likely won't work in the context of a MCP response.

guides/user-experience/select-menu-interaction/guide.md
```diff
@@ -0,0 +1,166 @@
+---
+name: select-menu-interaction
+description: Validating that a non-default option has been chosen in a select menu only after the user has interacted with the control.
```
philipwalton:
```suggestion
description: Validate that a non-default option has been chosen in a select menu only after the user has interacted with the control.
```

guides/user-experience/select-menu-interaction/guide.md
```diff
@@ -0,0 +1,166 @@
...
+}
+
+select:user-invalid + .error-msg,
+select.user-invalid-fallback + .error-msg {
+  display: block;
+}
+```
+
+### JavaScript Fallback
+If the browser doesn't support the selector, we can use a reusable utility that tracks interaction state using a `WeakMap`. This avoids polluting the DOM with "dirty" classes or data attributes.
```
philipwalton: Same comment here as the other use case. I think we should present this an optional fallback strategy, given it will only be needed by a small portion of users, and (presumably) the site already has other validation mechanism in place (e.g. server-side validation).

guides/user-experience/style-parent-with-has/guide.md
```diff
@@ -0,0 +1,182 @@
...
+
+## The Problem
+Often, an error state requires styling elements *outside* the input itself—for example, changing the color of a parent `fieldset` border, highlighting the `<label>`, or showing a global error icon in the card header. Historically, this required JavaScript to toggle classes on parent elements.
+
+## The Solution
+By combining `:has()` with `:user-invalid`, we can declaratively style any ancestor based on the validity state of a specific descendant. This keeps all presentation logic in CSS.
+
+### Implementation Strategy
+
+1.  **Selector**: Use `.parent:has(input:user-invalid)` to target the container.
```
philipwalton:
```suggestion
1.  **Selector**: Use `.parent:has(:user-invalid)` to target the container.
```
To include selects as well? I don't think this will have worse performance, but if so you could also do `:is(input,select):user-invalid`.

guides/user-experience/style-parent-with-has/guide.md
```diff
@@ -0,0 +1,182 @@
...
+  background-color: #fff8f8;
+}
+
+/* Change the icon too */
+.card-section:has(input:user-invalid) .status-icon::after {
+  content: "⚠️";
+}
+```
+
+## Fallbacking & Browser Support
```
philipwalton: Same comment here as for other use cases. IMO this one is a great candidate for pure progressive enhancement because presumably the input itself will have the error styling and the only thing that won't is the parent element.

I wonder if we should just recommend to make sure the input has some visual styling and then no fallback is ok?

guides/user-experience/validate-input-after-interaction/guide.md
```diff
@@ -0,0 +1,254 @@
...
+<form>
+  <div class="field">
+    <label for="password">New Password</label>
+    <!-- DO: Use pattern and minlength for complex password validation -->
+    <input 
+      type="password" 
+      id="password" 
+      autocomplete="new-password"
+      required
+      /*
```
philipwalton: Not sure if this will matter, but this is not a valid HTML comment...

guides/user-experience/validate-input-after-interaction/guide.md
```diff
@@ -0,0 +1,254 @@
...
+
+/* DO: Ensure the fallback highlights rule lists equally */
+input:user-invalid + .rules-list,
+input.user-invalid-fallback + .rules-list {
+  color: #d93025;
+}
+```
+
+### JavaScript Fallback
+MANDATORY: If the browser doesn't support the selector, we can use a reusable utility that tracks interaction state using a `WeakMap`. This avoids polluting the DOM with "dirty" classes or data attributes.
```
philipwalton: Same comment as before, I'm not sure if we should present this as "mandatory".

philipwalton (changes requested):

guides/user-experience/accessible-error-announcement/guide.md
PaulKinlan: re: programaitcally added, that's fair. Will sort that.

guides/accessibility/accessible-error-announcement/guide.md
PaulKinlan: If you want the same functionality the only way to do it is with a fallback. Without the fallback you end up having the announcement happening as soon as you start typing.

guides/user-experience/validate-input-after-interaction/guide.md
PaulKinlan: I'm not actually sure how I didn't pick this up. I'll clear this up.

guides/user-experience/validate-input-after-interaction/guide.md
PaulKinlan: I've rewritten these

guides/user-experience/select-menu-interaction/guide.md
PaulKinlan: I've clarified the fallback startegy a bit to say "if you want to maintain the same implementation across older browsers:

guides/user-experience/required-field-feedback/guide.md
PaulKinlan: What should we do?

guides/user-experience/style-parent-with-has/guide.md
PaulKinlan: That won't work, because the old :valid / :invalid will have the behaviour the developer doesn't want.

guides/user-experience/accessible-error-announcement/guide.md
PaulKinlan: I went with delegation

guides/user-experience/required-field-feedback/guide.md
philipwalton: @rviscomi suggestions?

guides/user-experience/style-parent-with-has/guide.md
```diff
@@ -0,0 +1,187 @@
+---
+name: style-parent-with-has
+description: Declaratively style parent elements like labels or fieldsets when a child input is in the :user-invalid state, eliminating the need for JavaScript state management.
```
philipwalton:
```suggestion
description: Style parent elements of a form field (e.g. labels or fieldsets) when the field is invalid.
```

I think we want avoid mentioning solutions or implementation details in the use case, otherwise I worry that a user prompt like the following won't match this use case: "write me some JavaScript code to automatically style a form label..."

guides/user-experience/validate-input-after-interaction/guide.md
```diff
@@ -0,0 +1,254 @@
+---
+name: validate-input-after-interaction
+description: Validation feedback for form inputs that only appears after the user has finished their initial interaction, avoiding premature errors on page load or while the user is typing. This consolidated guide covers sub-use-cases including password complexity validation and validating email after interaction.
```
philipwalton:
```suggestion
description: Show form field validation feedback (e.g. password complexity or email format requirements) only after the user has finished their initial interaction, avoiding premature errors on page load or while the user is typing.
```

philipwalton (approved):
LGTM with some suggestions.

guides/user-experience/required-field-feedback/guide.md
rviscomi: If we absolutely had to link two guides together, the way to do it would be something like `MANDATORY: Call get_best_practices("accessible-error-announcement") to ensure a consistent accessibility experience.`

However, that should be extremely rare if we ever need to do it at all. Ideally, in the interest of speeding up responses and conserving tokens, guides will be as self-contained as possible so that agents have everything they need to implement the use case correctly.

My suggestion would be to bake the a11y guidance directly into the implementation section of this guide. I think we can get it down to just a few lines of JS:

```js
// Sync aria-invalid with the CSS :user-invalid state
const syncAria = (el) => {
  el.toggleAttribute?.('aria-invalid', el.matches(':user-invalid'));
};

// Update on blur (to show error) and input (to clear it)
document.addEventListener('blur', (e) => syncAria(e.target), true);
document.addEventListener('input', (e) => {
  if (e.target.hasAttribute('aria-invalid')) syncAria(e.target);
});
```

We should probably also revisit the part about "You don't need `onBlur` handlers" to make clear that some JS is needed, even in the modern implementation.

guides/accessibility/accessible-error-announcement/guide.md
rviscomi, line 1: Should this be under `guides/accessibility`?

philipwalton: Yeah, good call. @PaulKinlan can you move this when you add the rest of your changes?

Issue #78, opened 2026-02-19
sibling-count() and sibling-index()


rviscomi:
The sibling-count() and sibling-index() CSS functions return integers that are useful to style elements based on their positions among siblings or on the number of siblings, for example as part of a calc() expression.

### MDN Documentation
- [sibling-count()](https://developer.mozilla.org/docs/Web/CSS/Reference/Values/sibling-count)
- [sibling-index()](https://developer.mozilla.org/docs/Web/CSS/Reference/Values/sibling-index)

---
Feature ID: sibling-count

**Use cases:**
- [x] #364
- [x] #363
- [ ] #393

PR #170, opened 2026-02-27
Use cases: sibling-count() and sibling-index()

marcianoskate:
Describe the use cases for `sibling-count()` and `sibling-index()` as described in #78 

## Complex math driven spatial patterns

Create complex geometric layouts that require math to arrange the elements. For example: fan out elements in a circle.

image

## Complex animations

Create staggered animations and cascading transition delays for a group of sibling elements

image


philipwalton:
@malchata can you review this one?

guides/user-experience/build-quantity-responsive-layouts/guide.md
```diff
@@ -0,0 +1,7 @@
+---
+name: build-quantity-responsive-layouts
+description: Build quantity-responsive layouts and adaptive typography components by calculating fluid widths or adjusting font sizes dynamically based on the total number of sibling elements present, eliminating the need for complex CSS quantity queries or JavaScript.
```
malchata: Wondering if this could be condensed a bit. It doesn't _need_ to change, but it feels a little long for a use case. Thoughts, @philipwalton?

guides/user-experience/calculate-math-driven-spatial-patterns/guide.md
```diff
@@ -0,0 +1,7 @@
+---
+name: calculate-math-driven-spatial-patterns
+description: Calculate math-driven spatial patterns and complex geometric layouts, such as fanning items out into a perfect circle or plotting boustrophedon (zigzag) grid rows, by feeding an element's positional index into trigonometric or arithmetic CSS functions.
```
malchata: Same here. For example, I'd avoid terminology like "boustrophedon" and just say what it means (which you have in parentheses, which I'm ironically using parentheses to talk about your parentheses).

malchata (changes requested):
Some suggestions here.

guides/user-experience/create-procedural-staggered-animations/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: create-procedural-staggered-animations
+description: Create procedural staggered animations and cascading transition delays for any group of sibling elements, including standard sequential fades and reverse staggers, by computing time values based on an element's index and the total count of its siblings.
```
philipwalton:
```suggestion
description: Create procedural staggered animations and cascading transition delays for any group of sibling elements, including standard sequential fades and reverse staggers.
```
The last part should be clarified in the guide itself, so I don't think it needs to be in the description.

guides/user-experience/create-procedural-staggered-animations/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: create-procedural-staggered-animations
```
philipwalton: Nit: for the use case slug, I think I'd drop "create". I don't think we need the slugs to use active phrasing, and I also worry that all the slugs would then be prefixed with  "create" or "build", which would make it harder to scan through them in a list.

And I guess this applies to all the use case slugs in this PR.
```suggestion
name: procedural-staggered-animations
```

guides/user-experience/calculate-math-driven-spatial-patterns/guide.md
philipwalton: I think it's ok to use an obscure term (the AI tool will know what it means), as long as its a core use case and not just a random example. In this case I'm not sure how common it would be for someone to want to do this.

Also, I'd recommend reframing the last part to keep the solution out of the use case description.

```suggestion
description: Create math-driven spatial patterns and complex geometric layouts based on how many children an element has, e.g. fanning items out into a perfect circle or plotting boustrophedon (zigzag) grid rows. 
```

guides/user-experience/build-quantity-responsive-layouts/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: build-quantity-responsive-layouts
+description: Build layouts and typography for content that varies in quantity (quantity-responsive) by dynamically calculating fluid widths or adjusting font sizes based on the total number of sibling items present, eliminating the need for complex CSS rules or JavaScript.
```
philipwalton: TBH, I'm not convinced that we'd want to always suggest sibling-count/index as the solution to this use case. In particular, the demo you shared could be more easily built with either flexbox or grid.

Unless we can come up with a specific case where sibling-count/index really unlocks some new potential, I'd suggest we drop this one.

philipwalton (changes requested):

guides/user-experience/build-quantity-responsive-layouts/guide.md
philipwalton: @marcianoskate ping on this. Any objections to dropping this use case?

philipwalton (approved):
@marcianoskate the 2nd and 3rd use case LGTM. Feel free to remove the first one and merge unless you feel strongly that we should keep it.

guides/user-experience/build-quantity-responsive-layouts/guide.md
philipwalton: Removed

PR #185, opened 2026-03-01
Baseline status macro

rviscomi:
Adds support for a `{{ BASELINE_STATUS("feature-id") }}` macro that can be used in markdown guides as placeholders for a feature's Baseline status. It takes an optional BCD key as the second parameter, in case the feature-level status is too coarse. The status is generated at build time, before vectorization. Feature IDs and BCD keys are validated as part of the sync-use-cases script.


serving/mcp-server/data/baseline.ts
```diff
@@ -113,6 +124,84 @@ export function resolveFeatureId(featureId: string): string[] {
...
+/**
+ * Internal helper to format status messages consistently.
+ */
+function formatStatusMessage(subject: string, status: { baseline?: string | boolean; baseline_low_date?: string }): string {
+  const { baseline, baseline_low_date } = status;
+
+  if ((baseline === 'low' || baseline === 'high') && baseline_low_date) {
+    const isWidely = baseline === 'high' || (baseline_low_date <= subtractMonths(new Date().toISOString().split('T')[0], 30));
+    const statusName = isWidely ? 'Widely available' : 'Newly available';
+    return `${subject} has been Baseline since ${baseline_low_date} (${statusName})`;
```
paulirish: bikeshedding.. i'd expect something like 

    {Feature} is Widely Available. It's been Baseline since {low_date}

only bring this up cuz right now the status name in parens feels disconnected and hard to interpret how it connects to the rest.

paulirish (commented):
good idea

serving/mcp-server/data/baseline.ts
rviscomi: Updated

Issue #41, opened 2026-02-19
calc-size()


rviscomi:
The calc-size() CSS function computes mathematical expressions that include height and width keyword values, for example calc-size(fit-content, size / 2). Not to be confused with the calc() CSS function, which cannot use keyword values.

### MDN Documentation
- [calc-size()](https://developer.mozilla.org/docs/Web/CSS/Reference/Values/calc-size)

### Chrome Documentation

<details>
<summary>View 2 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/docs/css-ui/anchor-positioning-api
- https://developer.chrome.com/en/docs/css-ui/animate-to-height-auto
</details>

---
Feature ID: calc-size

**Use cases:**
- [x] #226
- [x] #242

PR #187, opened 2026-03-02
Use cases for calc-size

devnook:
This PR adds 3 usa case for the #41 calc-size feature.

**animate-to-height-auto**

Screenshot 2026-03-02 at 11 29 33

**add-fixed-offset-to-dynamically-sized-container**

Screenshot 2026-03-02 at 11 28 47

**apply-math-boundaries-to-intrinsically-sized-content**

Screenshot 2026-03-02 at 11 28 57


guides/user-experience/animate-to-height-auto/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: animate-to-height-auto
+description: Animate smoothly between two different content-aware states, e.g. from the most compact single-line state to the fully expanded state that is defined by the content length and described by height auto.
```
philipwalton: I think this description captures the gist of the use case, but I wonder if "content-aware states" is too generic and in this case it might be better to start with the primary use case and then generalize? For example:

```suggestion
description: Animate smoothly to and from `height: auto`, or  other intrinsic sizes that depend on the element's content.
```

guides/user-experience/animate-to-height-auto/expectations.md
```diff
@@ -0,0 +1,6 @@
+* The agent has defined a CSS `@property` with `syntax: '<color>'` to register an interpolatable variable.
```
philipwalton: Same question here as in the other expectations file.

guides/user-experience/apply-math-boundaries-to-intrinsically-sized-content/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: apply-math-boundaries-to-intrinsically-sized-content
+description: Create highly resilient components that default to their intrinsic size but enforce strict mathematical guardrails without needing extra lines of CSS or media queries.
```
philipwalton:
```suggestion
description: Create highly resilient components that default to their intrinsic size but enforce strict mathematical guardrails.
```

I'd drop the "without needing extra lines of CSS" because it may limit what prompts this guide matches.

Also WDYT about saying "enforce strict size constraints" instead of "enforce strict mathematical guardrails"? I get that the "math" part relates to "calc", but I'm not sure how relevant that part actually is to this use case.

guides/user-experience/apply-math-boundaries-to-intrinsically-sized-content/expectations.md
```diff
@@ -0,0 +1,6 @@
+* The agent has defined a CSS `@property` with `syntax: '<color>'` to register an interpolatable variable.
```
philipwalton: Same question here as in the other expectations file.

guides/user-experience/add-fixed-offset-to-dynamically-sized-container/expectations.md
```diff
@@ -0,0 +1,6 @@
+* The agent has defined a CSS `@property` with `syntax: '<color>'` to register an interpolatable variable.
```
philipwalton: Are these expectations from another use case?

guides/user-experience/add-fixed-offset-to-dynamically-sized-container/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: add-fixed-offset-to-dynamically-sized-container
+description: Create elements like tooltips, buttons, modals, or dropdowns that need to fit their content plus a specific mathematical offset by adding fixed offsets, padding allowances, or viewport units directly to a intrisically sized element.
```
philipwalton, line 3: After looking a bit at this use case and the `apply-math-boundaries-to-intrinsically-sized-content` use case, I'm not convinced they need to be separate use cases. WDYT about combining them?

philipwalton (changes requested):

guides/user-experience/add-fixed-offset-to-dynamically-sized-container/guide.md
devnook: I'm not sure they are the same, one is about adding something to an intrinsic size, while other is about using a css function with an intrinsic size. I updated the description to reflect that - wdyt.

devnook:
PTAL

guides/user-experience/add-fixed-offset-to-dynamically-sized-container/guide.md
philipwalton: Ok, yeah, I see you're point. Though I think I will need to update the describes a bit more to make it more clear (to me at least).

I'll do that at merge, so we can get this closed out.

philipwalton: Actually, I changed my mind again. I'm thinking about what the guides will look like for these use cases, and I think they'd be almost identical, at least in regards to the `calc-size()` usage instructions.

Also, for the "adding something to an intrinsic size" use case, there are lots of ways to solve that outside of `calc-size()` (e.g. using flexbox or grid), and I'm not sure if we want to always recommend `calc-size()` (which could happen if that was the only guide that matched the user's prompt).

Anyway, in the interest of time I opened #241 to close out this feature, but feel free to keep discussing this if you disagree. We can always add more use cases later.

devnook: Let's close this one then in favor of #241

PR #198, opened 2026-03-02
Add use cases for the interest-invokers API

### LINKED ISSUE #59: Interest invokers
Interest invokers, registered by the interestfor HTML attribute, trigger events and actions on a target element when a user shows interest in the element, through behaviors such as hover, focus, or long-press. Pseudo-elements apply styles to sources and targets of interest.

### Chrome Documentation

<details>
<summary>View 3 articles</summary>

- https://developer.chrome.com/en/blog/new-in-web-ui-io-2025-recap
- https://developer.chrome.com/en/blog/popover-hint
- https://developer.chrome.com/en/blog/web-at-io25
</details>

---
Feature ID: interest-invokers

**Use cases:**
- [x] #389
- [x] #390
- [ ] #1243

ewewraw:
Add 3 common use cases for the Interest Invoker API:

- assign-custom-interest-based-behavior
- create-hint-popover
- create-hovercards
- provide-interest-based-popover-behavior


philipwalton:
@ewewraw thanks for submitting these, but I wonder if they're already going to be sufficiently covered in @una's PR for popover (#138) (which also includes interest invokers). Una WDYT?

una:
One of the 4 use cases I included for `popover`  (tooltip) mentions interest invokers, but I think its worth exploring more. For example, hover cards tend to be much more complex and include more complex information and interactions inside of them. It's probably worth a deeper guide here since popover is an attribute that has a lot of different effects and interestfor is am invoking behavior. I'm not sure what the difference is between the first and last item here. Additionally, there should be mention of the mobile interaction for interest invokers as well in this one (I don't get into that in the popover guidance much)

philipwalton:
Since there's been no movement on this, I think we can close this PR and focus on covering this feature in the popover PR, since AFAICT the primary use case for interest invokers is showing and hiding popovers.

@una

> One of the 4 use cases I included for popover (tooltip) mentions interest invokers, but I think its worth exploring more. For example, hover cards tend to be much more complex and include more complex information and interactions inside of them.

But building a hovercard is also a use case for popover, correct? And the hovercard use cases has a lot of overlap with building a tooltip, so I think they can likely be covered in the same guide.

> It's probably worth a deeper guide here since popover is an attribute that has a lot of different effects and interestfor is am invoking behavior.

The guides will focus on implementing the use cases, not documenting the features, so we don't necessarily need to cover all of the different effects.

> Additionally, there should be mention of the mobile interaction for interest invokers as well in this one (I don't get into that in the popover guidance much)

Is that a separate use case, or is long press support just a recommendation we'd make for properly implementing a hovercard? Again, the goal of these guides is not to fully document a feature, it's to cover how to properly use them for their primary use cases.

philipwalton:
I'm going to close this in favor of #388, which include the tooltip use cases (but not other popover use cases).

PR #201, opened 2026-03-03
Remove deep-link-to-hidden-content use case

taraojo:
Updated to remove this use case


philipwalton (commented):
Great demo! Though now that I see the demo, I'm thinking maybe we should slightly change the wording of the use case.

With the current phrasing of the use case, I'd expect the guide to explain how to add text fragments to URLs. But I think what you're actually wanting to show is how to structure your page content so that hidden content *can* be deep-linked to via URL fragments, correct?

If so, then how about rewording the use case to this:

"Enable visually hidden content (for example in collapsed accordions, closed tabs, or below "Read more" sections) to be deep linked to via URL fragments and "Scroll to Text Fragment" links."

philipwalton:
Thanks for making this update. Sorry to keep going back-and-forth on this, but after reviewing this and #137, I'm wondering if we actually need two separate use cases. WDYT about just mentioning deep linking via URL fragments as one of the "accessibility features" of this approach.

My concerns is that if we don't combine the use cases, the guide for this will be 99% the same as the guide in #137.

philipwalton (approved):

Issue #60, opened 2026-02-19
interpolate-size


rviscomi:
The interpolate-size CSS property sets whether animations and transitions interpolate between a numeric value and a keyword value, such as from a fixed length to auto or fit-content.

### MDN Documentation
- [interpolate-size](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/interpolate-size)

### Chrome Documentation

<details>
<summary>View 8 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/missing-from-css
- https://developer.chrome.com/en/blog/rfc-customizable-select
- https://developer.chrome.com/en/blog/styling-details
- https://developer.chrome.com/en/docs/css-ui/access-colors-spaces
- https://developer.chrome.com/en/docs/css-ui/animate-to-height-auto
- https://developer.chrome.com/en/docs/devtools/css/reference
- https://web.dev/en/articles/baseline-in-action-container-queries
- https://web.dev/en/learn/css/transitions
</details>

---
Feature ID: interpolate-size

**Use cases:**
- [x] #226
- [x] #242
- [ ] #1232

PR #213, opened 2026-03-04
Use cases: interpolate-size

marcianoskate:
#60 

## Enable animations based on intrinsic size:

Enable modern animations based on keywords like sizing keywords like auto, min-content, max-content, and fit-content.

image


philipwalton:
There's some overlap between this use case and the one mentioned in #187. @devnook do you think it makes sense to combine these?

Based on [this note in the MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/calc-size#:~:text=Note%20however%20that,also%20require%20calculations.), it seems like `interpolate-size` should be our primary recommendation and only recommend `calc-size()` in situations were calculations are required. (That alone makes me think these APIs need to be covered together in the same use case guide, or guides.)

marcianoskate:
I think that the animation of elements based on their intrinsic size is the same use case (`interpolate-size` enables automatic `calc-size()`). 

`calc-size()` is different by the fact that it allows us to compute sizes that are different from the intrinsic one. I'm happy combining them.

Do you have an initial feeling on how to combine them?

philipwalton:
> Do you have an initial feeling on how to combine them?

I added `calc-size` to the list of web features, and also added it to the demo. I'm going to merge this now because I think it's the primary use case for both of these features.

@devnook can you update your PR to remove this use case and only include those specific to calc-size?

PR #250, opened 2026-03-09
Run suite in parallel

paulirish:
image

Using pnpm (a bit of a hack) to both manage parallelism and to group the output so its still nice to read. 

The results are pretty delightful.

This PR parallelizes the actual generation of the tasks.. and also running the graders against the generation results.

- Also a retry w/ exponential backoff on "ApiError: got status: INTERNAL" errors which happen from time to time.


paulirish:
@micahjo7 oops. got mixed up between autoland and skip-that-and-land-right-now. Didn't mean to land already. Will gladly take any comments!

micahjo7:
lgtm, thanks!

PR #259, opened 2026-03-09
Refactor dashboard to support GH pages

micahjo7:
Added instructions to deploy to GH pages in `eval-view/README`, deployed to https://googlechrome.github.io/guidance/.

Updates to include OAuth and client-side fetching to retrieve from GCS.


eval-view/package.json
```diff
@@ -4,13 +4,12 @@
   "type": "module",
   "scripts": {
     "start": "node server.js",
-    "test:e2e": "playwright test"
+    "test:e2e": "playwright test",
+    "deploy-pages": "gh-pages -d ."
```
paulirish:
```suggestion
    "deploy-pages": "gh-pages --dist . --nojekyll"
```

i like longform flags  if they're at rest.

and nojekyll will save ~3s in the deployment. :p

bin/gd.ts
```diff
@@ -137,6 +137,8 @@ ${cBold('Evaluation:')}
   ${cCyan('run')} <tmpl> <prompt>    Run an ad-hoc agent test against a template
 
 ${cBold('Other:')}
+  ${cCyan('deploy')}                 Deploy the dashboard to GitHub Pages
+  ${cCyan('upload')}                 Upload generated evaluation results to GCS
```
paulirish: put that up in the evaluation section

eval-view/dashboard.html
paulirish, line 26:
```suggestion
      <div style="margin-top: 15px;">
        <button id="auth-btn" class="secondary-btn" style="display: none;">
          Sign in with Google
        </button>
      </div>
      <div id="summary-stats">
```

paulirish: in case someone lands directly here and they weren't already authed

eval-view/dashboard.js
```diff
@@ -563,7 +591,7 @@ async function showDetails(testName, runs, stats, testID) {
     });
 
     const runDetails = await Promise.all(runDetailsPromises);
-    body.innerHTML = promptHtml;
+    body.innerHTML = promptHtml; // Insert promptHtml first
```
paulirish, line 612: oh so now we are adding the prompt text? hmm.  (we didnt see it, did we?)

harness/config.ts
```diff
@@ -45,7 +45,7 @@ export const environmentConfig: EnvironmentConfig = {
 // *** Run with: `pnpm suite`         ***
 // **************************************
 export const suiteConfig: SuiteConfig = {
-  name: `full-${new Intl.DateTimeFormat('en-CA', { timeZone: 'America/Los_Angeles' }).format(new Date())}`,
+  name: `full-${new Date().toISOString().replace(/[:.]/g, '-').slice(0, 19)}`,
```
paulirish: isostring is like.. good but the dates get wrong in the afternoon evening. :(  
so thats why i did PT

I can compromise with ET tho :)

eval-view/README.md
```diff
@@ -0,0 +1,38 @@
+# Eval-View Dashboard
+
+The Eval-View Dashboard lets you visualize test results from the Guidance eval harness.
```
paulirish, line 3: needs something about how it can be run two ways.. 1) via the server.js locally.. and 2) static on gh-pages.

and that we have some complexity in the codebase to handle both situations simultaneously.

eval-view/dashboard.js
micahjo7: the logic for this was actually already present, and it gets the prompt by retrieving the associated task. it wasn't working when we were looking at it earlier, because it was broken on the the GH page (only worked locally).

harness/config.ts
micahjo7: ah I did this so we can capture the time, so we don't accidentally override previous runs when uploading

eval-view/dashboard.js
micahjo7: will eventually update this with b/491137229

eval-view/dashboard.html
paulirish: oh looks like we hit an error

if i load https://googlechrome.github.io/guidance/dashboard.html?testId=full-2026-03-08&source=remote right now (with an expired token, i'm guessing)..

i see "sign in with google"  flash by and quickly replaced with 

> Error loading dashboard data: Failed to load data from https://storage.googleapis.com/storage/v1/b/guidance-evals/o/full-2026-03-08%2Fevals.json?alt=media

in red

theres some onerror handler that we gotta soften

paulirish (approved):

PR #262, opened 2026-03-10
Add a `gd gen-negative-suite` to generate tasks & base apps with negative-demo.html files

micahjo7:
And adds `negative` to SuiteConfig to run the suite with all the negative tasks (and base apps).


micahjo7:
@paulirish @rviscomi I know we talked about a lot of changes we could make to improve our tests/apps- lmk if you think this PR is useful to add!

guides/negative-suite-gen.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    // 1. Create base app
+    const baseAppDir = path.join(BASE_APPS_DIR, `${inv.name}-app`);
+    if (!fs.existsSync(baseAppDir)) {
+      fs.mkdirSync(baseAppDir, { recursive: true });
+    }
+
+    const destIndexHtml = path.join(baseAppDir, 'index.html');
+    fs.copyFileSync(negativeDemoPath, destIndexHtml);
+    console.log(`  ✅ Created base app: harness/base_apps/${inv.name}-app/index.html`);
```
paulirish: can we do base_apps/negative/inv.name  ?  

and we can just reference by that in the base_app id:   negative/batch-analytics-events-app

paulirish:
cool results. https://googlechrome.github.io/guidance/dashboard.html?testId=full-negative-2026-03-10T10-39-35-gcli&source=remote

paulirish (changes requested):
love it except for the nit on path.

guides/negative-suite-gen.ts
paulirish: also..  we could get rid of `-app`  now

bin/gd.ts
```diff
@@ -244,6 +245,12 @@ ${cBold('Options:')}
       process.exit(code);
     }
 
+    case 'gen-negative-suite': {
+      // Run the negative-suite-gen.ts script
+      const code = await spawnChild('node', ['--experimental-strip-types', 'guides/negative-suite-gen.ts']);
```
paulirish, line 250: oh.. also this..  i'd prefer we use this kinda pattern

```ts
        const { generateGrader } = await import('../guides/grader-gen.ts');
        await generateGrader(dir);
```

.. remains in the same process and dont need to worry about passing args over CLI etc.  
mostttt of the commands above now use this but clearly not all.

guides/negative-suite-gen.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    // 1. Create base app
+    const baseAppDir = path.join(BASE_APPS_DIR, 'negative', `${inv.name}`);
+    if (!fs.existsSync(baseAppDir)) {
+      fs.mkdirSync(baseAppDir, { recursive: true });
+    }
+
+    const destIndexHtml = path.join(baseAppDir, 'index.html');
+    fs.copyFileSync(negativeDemoPath, destIndexHtml);
```
paulirish, line 57: okay crazy idea.. since it's literally a copy can we do a symlink instead?  feels nice cuz then changes to a negative-demo dont need the double diff

paulirish (approved):
nice! def appreciate you doing this to move us forward in clarifying/simplifying the eval workflow

micahjo7:
addressed latest comments in https://github.com/GoogleChrome/guidance/pull/281

PR #269, opened 2026-03-10
Add Guides for Scroll-Driven Animations Use-Cases

### LINKED ISSUE #263: Create guide and evals for the carousel-slide-effects use case
Create a carousel of slides with images or other visual elements, where each slide animates as they enter/center/exit their scroller. For example, the slides may fade-in/fade-out, rotate, get bigger or smaller, etc.

Affected web-feature IDs: [scroll-driven-animations](https://webstatus.dev/features/scroll-driven-animations), [scroll-snap](https://webstatus.dev/features/scroll-snap)

Use case subdir: [guides/ui-atoms/carousel-slide-effects](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-atoms/carousel-slide-effects)

Related features: #33 #719

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

### LINKED ISSUE #264: Create guide and evals for the parallax-scroll-effects use case
Create scroll-based effects (such as parallax) where foreground and background layers move at different rates, creating a sense of depth as the user scrolls.

Affected web-feature IDs: [scroll-driven-animations](https://webstatus.dev/features/scroll-driven-animations)

Use case subdir: [guides/ui-behaviors/parallax-scroll-effects](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-behaviors/parallax-scroll-effects)

Related features: #33

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

### LINKED ISSUE #265: Create guide and evals for the scroll-entry-exit-effects use case
Create fade-in, scale-up, or other complex reveal-type effects on elements as they enter and exit the scrollport (or viewport) while the user is scrolling.

Affected web-feature IDs: [scroll-driven-animations](https://webstatus.dev/features/scroll-driven-animations)

Use case subdir: [guides/ui-behaviors/scroll-entry-exit-effects](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-behaviors/scroll-entry-exit-effects)

Related features: #33

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

### LINKED ISSUE #266: Create guide and evals for the scroll-progress-indicator use case
Create a scroll progress bar, stepped progress tracker, or any visual affordance that communicates how far through a page or section the user has scrolled.

Affected web-feature IDs: [scroll-driven-animations](https://webstatus.dev/features/scroll-driven-animations)

Use case subdir: [guides/ui-atoms/scroll-progress-indicator](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-atoms/scroll-progress-indicator)

Related features: #33

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

### LINKED ISSUE #267: Create guide and evals for the scrollytelling use case
Animate visual properties on a target element — such as fading a backdrop, shifting a background color, or to create scrollytelling experiences — driven entirely by the scrollport position of a completely different element.

Affected web-feature IDs: [scroll-driven-animations](https://webstatus.dev/features/scroll-driven-animations)

Use case subdir: [guides/ui-behaviors/scrollytelling](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-behaviors/scrollytelling)

Related features: #33

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

### LINKED ISSUE #268: Create guide and evals for the shrinking-header-on-scroll use case
Smoothly animate a fixed header or full-page cover on scroll to dynamically shrink, gain shadows, and transform its layout over a predefined scroll distance.

Affected web-feature IDs: [scroll-driven-animations](https://webstatus.dev/features/scroll-driven-animations)

Use case subdir: [guides/user-experience/shrinking-header-on-scroll](https://github.com/GoogleChrome/guidance/tree/main/guides/user-experience/shrinking-header-on-scroll)

Related features: #33

bramus:
Fixes #263, fixes #264, fixes #265, fixes #266, fixes #267, fixes #268.


bramus:
This is ready for review.

Some things to note:

- In each guide I had to repeat some general guidance for Scroll-Driven Animations. Is there a better way to do this, or is it fine like that?
- I am not convinced the `scrollytelling` use-case should be solved by this. What people want are scroll-triggered animations doing this, not scroll-driven ones (@philipwalton). _([demo](https://codepen.io/bramus/pen/QwKGgzz); Needs Chrome 147)_

guides/user-experience/carousel-item-effects/guide.md
```diff
@@ -7,4 +7,173 @@ web-feature-ids:
...
+    ```html
+    <ul class="scroller">
+      <li class="entry">1</li>
+      <li class="entry">2</li>
+      <li class="entry">3</li>
+      …
+    </ul>
+    ```
+
+2.  **Define the animation:** Create a CSS animation that defines the different states of your slides as they enter, center, and exit the scrollport. For example, you can use the `scale` property to make the slides grow as they approach the center and shrink as they move away.
```
philipwalton: This reads like defining a "center" value is required. I'd clarify that you can define keyframes for any part of the animations, and then give "center" as one such example.

guides/user-experience/carousel-item-effects/guide.md
```diff
@@ -7,4 +7,173 @@ web-feature-ids:
...
+
+# Build Carousel Slide Effects
+
+Carousel slide effects are a great way to add visual interest to a carousel. As the user scrolls through the slides, each slide can animate as it enters, centers, and exits the scrollport. For example, the slides can fade in and out, rotate, or scale in size. This creates a dynamic and engaging user experience. Unlike simple entry/exit animations, this effect uses a single, continuous animation to control the slide's appearance across the entire scrollport.
+
+## How to implement
+
+Here’s how to create carousel slide effects:
+
+1.  **Create a scroller:** This element will act as the container for your carousel slides.
```
philipwalton: This might be obvious for the agent, but I'd include the CSS code needed to make the container a scroller.

guides/user-experience/carousel-item-effects/guide.md
```diff
@@ -7,4 +7,173 @@ web-feature-ids:
...
+    scale: 1;
+  }
+
+  100% {
+    scale: 0.5;
+  }
+}
+
+.scroller > * {
+  view-timeline: --item inline;
```
philipwalton: If you want to include this example, then I'd add some comments to explain this code a bit more. I have no idea how to properly define a named view timeline based on just this example, and I'm not sure if I AI could figure it out either.

guides/user-experience/carousel-item-effects/guide.md
```diff
@@ -7,4 +7,173 @@ web-feature-ids:
...
+  animation: animate auto linear both;
+  animation-timeline: --item;
+}
+```
+
+Prefer a named `view-timeline` when multiple elements or children of the tracked subject need to animate.
```
philipwalton: In this case, multiple elements are animating aren't they (e.g. multiple carousel items)? But I suspect you didn't mean "multiple" in that sense, so if I'm right then I'd clarify this a bit more.

I'd also clarify exactly what "tracked subject" means.

guides/user-experience/carousel-item-effects/guide.md
```diff
@@ -7,4 +7,173 @@ web-feature-ids:
...
+  - If the animation is only considered to be decorative, opt for Progressive Enhancement and **DO NOT** provide a fallback.
+- **DO** respect user preferences: Some users prefer to have less motion on the web. Use the `prefers-reduced-motion` media query to disable or reduce your animations for these users.
+- **DO** try to animate only performant CSS properties: For the smoothest animations, stick to animating properties that can be handled by the browser's compositor thread, such as `transform` and `opacity`. Animating other properties like `width` or `height` can lead to performance issues.
+- **DO** use the correct declaration order: When using the `animation` shorthand property, declare `animation-timeline` and `animation-range` *after* it to prevent the shorthand from resetting the timeline.
+
+When using the `view()` function to create a scroll-driven animation:
+
+- **OPTIONAL** be explicit about the axis to track: When not targeting the default `block` axis (such as in a horizontal scroller), be explicit about which axis to track with `view(block)` or `view(inline)`.
+- When the animation is not applied to the tracked subject itself, use a named view timeline.
```
philipwalton: I *think* I understand what this means, but it's a bit fuzzy so I'd just be a bit more explicit.

guides/user-experience/carousel-item-effects/guide.md
```diff
@@ -7,4 +7,173 @@ web-feature-ids:
...
+
+When using the `view-timeline` property to create a scroll-driven animation:
+
+- **DO** use a CSS `<dashed-ident>` for the name.
```
philipwalton:
```suggestion
- **DO** use a CSS `<dashed-ident>` for the name (e.g. `view-timeline: --my-custom-name`)
```

Just to be explicit about what a dashed-ident is.

guides/user-experience/carousel-item-effects/guide.md
```diff
@@ -7,4 +7,173 @@ web-feature-ids:
...
+- **OPTIONAL** be explicit about the axis to track: When not targeting the default `block` axis (such as in a horizontal scroller), be explicit about which axis to track with `view-timeline-axis`.
+- **DO** make sure the scope of the lookup works: When the element that is declaring the `view-timeline` is not a flat tree ancestor of the animated element, hoist up the visibility of the `view-timeline`’s name by using `timeline-scope` on a shared ancestor.
```
philipwalton, line 141: This is also a bit fuzzy, and at this point I'm wondering if the use of named view timelines just doesn't belong in this use case (and it would be clearer in a use case where using it is the recommendation). I'll revisit this comment after reviewing the rest of the guides.

guides/user-experience/parallax-scroll-effects/guide.md
```diff
@@ -3,7 +3,182 @@ name: parallax-scroll-effects
...
 sources:
   - https://developer.chrome.com/docs/css-ui/scroll-driven-animations
   - https://scroll-driven-animations.style/demos/parallax-carousel/css/
----
\ No newline at end of file
+---
+
+# Build a Parallax Effect on Scroll
+
+A parallax effect on scroll is a visual technique where different layers of content move at varying speeds as the user scrolls down a page. This creates an illusion of depth, with foreground elements appearing to move faster than the background elements, resulting in an engaging and immersive browsing experience. This effect can be achieved using CSS Scroll-Driven Animations, which allow you to link animations to the scroll position of a container.
```
philipwalton:
```suggestion
A parallax effect on scroll is a visual technique where different layers of content move at varying speeds as the user scrolls down a page. This creates an illusion of depth, with foreground elements appearing to move faster than the background elements, resulting in an engaging and immersive browsing experience. This effect is best achieved using CSS Scroll-Driven Animations, which allow you to link animations to the scroll position of a container.
```

Should we make this recommendation stronger?

guides/user-experience/parallax-scroll-effects/guide.md
```diff
@@ -3,7 +3,182 @@ name: parallax-scroll-effects
...
+
+## How to implement
+
+Here’s how to create a basic parallax effect:
+
+1.  **Create a wrapper element:** This element will contain all the layers of your parallax animation.
```
philipwalton: Can the wrapper be the `<html>` or `<body>` elements? If so then maybe say "define the wrapper"?

Also same comment as on the other guide about showing how to make this a scrollable container.

guides/user-experience/parallax-scroll-effects/guide.md
```diff
@@ -3,7 +3,182 @@ name: parallax-scroll-effects
...
+
+    .layer {
+      animation: parallax linear both;
+      animation-timeline: --wrapper;
+    }
+    ```
+
+5.  **Stagger the animations:** To make the layers move at different speeds, you can use one of two approaches:
+
+    *   **Using `sibling-index()` in the keyframes:** This is the simplest approach. The `sibling-index()` function returns the index of a child element amongst its siblings, which you can then use in your keyframes to create a staggered effect.
```
philipwalton: I think using `sibling-index()` makes sense in your demos since it has lots of layers, but would this really make sense in a typical parallax example, which would likely just have one foreground (text) and one background (image) layer—and only the background layer is moving at a different speed?

guides/user-experience/scroll-entry-exit-effects/guide.md
```diff
@@ -7,4 +7,176 @@ sources:
...
+
+While browser support for scroll-driven animations is constantly improving, it's still a good idea to provide a fallback for browsers that don't yet support them. Here's a JavaScript-based fallback that uses an `IntersectionObserver` to create a similar effect.
+
+```html
+<script>
+  if (!CSS.supports('(animation-timeline: view()) and (animation-range: entry)')) {
+    const observer = new IntersectionObserver(
+      (entries) => {
+        for (const entry of entries) {
+          entry.target.style.scale = 0.5 + entry.intersectionRatio * 0.5;
```
philipwalton, line 169: Might want to add a comment here to the effect of: "Customize this to match the logic defined in your CSS."

guides/user-experience/scroll-entry-exit-effects/guide.md
```diff
@@ -7,4 +7,176 @@ sources:
...
+        }
+      },
+      {
+        threshold: Array.from({ length: 101 }, (_, i) => i / 100),
```
philipwalton, line 173: Gross, but I like it :)

guides/user-experience/scroll-entry-exit-effects/guide.md
```diff
@@ -7,4 +7,176 @@ sources:
...
+    );
+
+    document.querySelectorAll('.scroller > *').forEach((el) => {
+      observer.observe(el);
+    });
+  }
+</script>
+```
+
+This script first checks if the browser supports `animation-timeline: view()` and `animation-range: entry`. If not, it creates an `IntersectionObserver` that will fire a callback whenever one of the observed elements intersects with the viewport.
```
philipwalton: I think including this is fine, but at the same time I suspect the AI will be able to figure out what's going on in your fallback example, so maybe it's not needed? (The evals would be able to answer this question.)

guides/user-experience/scroll-progress-indicator/guide.md
```diff
@@ -8,3 +8,131 @@ sources:
...
+      const scrolled = window.scrollY;
+      const progressPercentage = (scrolled / scrollable);
+
+      progress.style.transform = `scaleX(${progressPercentage})`;
+    });
+  }
+</script>
+```
+
+This script first checks if `animation-timeline: scroll()` is supported. If not, it adds a scroll event listener to the `window`. In the event listener, it calculates the scroll progress as a percentage and updates the `scaleX` of the `#progress` element. This provides a similar visual effect for browsers that don't yet support the CSS feature.
```
philipwalton: Similar to above, I think the AI can figure this out and this paragraph is likely not needed.

guides/user-experience/scroll-progress-indicator/guide.md
```diff
@@ -8,3 +8,131 @@ sources:
...
+
+- **DO** use a CSS `<dashed-ident>` for the name.
+- **OPTIONAL** be explicit about the axis to track: When not targeting the default `block` axis (such as in a horizontal scroller), be explicit about which axis to track with `scroll-timeline-axis`.
+- **DO** make sure the scope of the lookup works: When the element that is declaring the `scroll-timeline` is not a flat tree ancestor of the animated element, hoist up the visibility of the `scroll-timeline`’s name by using `timeline-scope` on a shared ancestor.
+
+## Browser support and fallback strategies
+
+{{ BASELINE_STATUS("scroll-driven-animations") }}. Therefore, a fallback strategy is typically required.
+
+While browser support for scroll-driven animations is improving, a fallback is still necessary. You can use JavaScript to create a similar effect.
```
philipwalton: I wonder if a stronger recommendation should be made here (and in the fallback section of the other use case guides as well).

Something like: "...even though Scroll Driven Animations are not yet Baseline Widely available, given their performance benefits, they should still be preferred on browsers that support them..."

guides/user-experience/parallax-scroll-effects/guide.md
```diff
@@ -3,7 +3,182 @@ name: parallax-scroll-effects
 description: Create scroll-based effects (such as parallax) where foreground and background layers move at different rates, creating a sense of depth as the user scrolls.
 web-feature-ids:
   - scroll-driven-animations
+  - sibling-count
```
philipwalton: I'd remove this unless we're saying the use case should not be implemented without it (see other comments for more context).

guides/user-experience/shrinking-header-on-scroll/guide.md
```diff
@@ -0,0 +1,130 @@
...
+      animation-range: 0px 200px;
+    }
+    ```
+
+**Tip:** To prevent the content following the header from being obscured by it, add a `padding-top` to the `body` (or the main content container) that is equal to the initial height of the header.
+
+**Tip:** When your header shrinks from a large height (e.g., `100vh`) to a smaller one (e.g., `10vh`), the `animation-range-end` should be the difference between the start and end sizes (e.g., `90vh`). This ensures the animation completes precisely when the header reaches its final size.
+
+
+## Example code
```
philipwalton, line 67: I'm confused by this example because it comes right after a tip but then it doesn't show what's mentioned in the tip. Should this example be moved up above the tip?

philipwalton (changes requested):

philipwalton:
> * In each guide I had to repeat some general guidance for Scroll-Driven Animations. Is there a better way to do this, or is it fine like that?

I agree there's a lot of repetition here, but I think it's mostly fine as is. The guides should be as independant as possible, which will inevitably mean some duplication.

That said, think you could potentially tailor these best practices a bit more to each use case. You could try asking AI to remove the bullets that aren't explicitly relevent to the guide.

> * I am not convinced the `scrollytelling` use-case should be solved by this. What people want are scroll-triggered animations doing this, not scroll-driven ones (@philipwalton). _([demo](https://codepen.io/bramus/pen/QwKGgzz); Needs Chrome 147)_

I think you could include scroll triggered animations in this guide—especially if it will ship before I/O. You'd just have to be explicit about it in the browser support and fallback section.

In my experience, scrollytelling tends to be a mix of both scroll-driven and scroll-trigger animations, but I agree that "triggered" is a key part of the effect. For example, this new page for the MacBook Neo appears to be using a mix of both (though using JS): https://www.apple.com/macbook-neo/

guides/user-experience/scroll-entry-exit-effects/guide.md
bramus: It was Gemini who came up with it 😀 … and I kept it because I liked it too :)

bramus:
> > * I am not convinced the `scrollytelling` use-case should be solved by this. What people want are scroll-triggered animations doing this, not scroll-driven ones (@philipwalton). _([demo](https://codepen.io/bramus/pen/QwKGgzz); Needs Chrome 147)_
> 
> I think you could include scroll triggered animations in this guide—especially if it will ship before I/O. You'd just have to be explicit about it in the browser support and fallback section.

I think that having both pieces of info in one and the same guide would only be confusing as it has two distinct pieces of information, each with their own fallbacks and browser compat tables.
👉  scroll-triggered use-cases should have their own guides.

philipwalton:
> I think that having both pieces of info in one and the same guide would only be confusing as it has two distinct pieces of information, each with their own fallbacks and browser compat tables.
> 👉 scroll-triggered use-cases should have their own guides.

We have lots of guides that cover multiple APIs, each with their own fallbacks and browser compat tables. For example, Una has a "Build a contextual tooltip" guide that includes popover, popover-hint, anchor positioning, container-anchor-position-queries, interest-invokers, etc.

If a developer is trying to use these APIs to implement this use case, they'll have to figure this stuff out themselves, so the point of the guide is to tell the AI what the best practices are.

It's OK if the best practices are:
- Here's how you implement this use case if you only have to support modern Chrome
- If you have to support older browsers, then we actually don't recommend using any of these APIs yet because the fallback story is more complex than it's worth. Just use scroll events and intersection observer.

That said, given that scroll-triggered animations have not shipped yet, I'd be 100% fine to punt on this use case until a later date.

bramus:
> > I think that having both pieces of info in one and the same guide would only be confusing as it has two distinct pieces of information, each with their own fallbacks and browser compat tables.
> > 👉 scroll-triggered use-cases should have their own guides.
> 
> We have lots of guides that cover multiple APIs, each with their own fallbacks and browser compat tables. For example, Una has a "Build a contextual tooltip" guide that includes popover, popover-hint, anchor positioning, container-anchor-position-queries, interest-invokers, etc.

Yeah, but those are technologies you need to combine to get to the result. With scrollytelling you use either the one (SDA) or the other (SDA), without combining them to get to the result (you can _mix_ them, but you don’t _combine_ them).

The code for each approach ends up being very much different. Yes, there is some overlap in the higher level features — it’s animations, and scrolling is involved — but if I were to write out the guidance for the STA approach there wouldn’t be that much overlap with SDA in how it is explained.

> That said, given that scroll-triggered animations have not shipped yet, I'd be 100% fine to punt on this use case until a later date.

It shipped in M146

guides/user-experience/carousel-item-effects/guide.md
bramus: `timeline-scope` isn’t really needed here (as you animate the carousel item itself, or one of its children) so I could drop it.

guides/user-experience/parallax-scroll-effects/guide.md
bramus: It’s just a wrapper to group the several layers together, so it’s not the thing that should be scrollable. I’ve got a commit that clarifies this.

guides/user-experience/scroll-entry-exit-effects/guide.md
bramus: You mean drop the whole “outro” or only this paragraph?

guides/user-experience/scroll-progress-indicator/guide.md
bramus: You mean drop the whole “outro” or only this paragraph?

guides/user-experience/scroll-progress-indicator/guide.md
bramus: Suggested wording (for all guides):

> For browsers that do not support scroll-driven animations, implement a fallback using the `IntersectionObserver` API to create a similar effect. This fallback is typically less performant than the native CSS approach, so it is recommended to only use this fallback for non-critical animations.

guides/user-experience/shrinking-header-on-scroll/guide.md
bramus: This is the key aspect of https://scroll-driven-animations.style/demos/cover-to-fixed-header/css/.

I originally proposed this as a separate guide, but was told to drop it … so looked for a way to still include that piece of crucial information here. Should I move it to the Best Practices section?

guides/user-experience/carousel-item-effects/guide.md
bramus: Reworded in 2af7083484bfb11b6dca07d39c983e0760a8aa32

guides/user-experience/carousel-item-effects/guide.md
bramus: Added in cd59a04aa744156878a8ec1e312d0025b1b646b6

guides/user-experience/carousel-item-effects/guide.md
bramus: Added in e43c0fb7a7af0259c1a4da112ff575c0cd249991

guides/user-experience/carousel-item-effects/guide.md
bramus: Clarified in f454f40f833c9a0f93a4cb445e8235b4ff3183fe

guides/user-experience/parallax-scroll-effects/guide.md
bramus: Clarified in ec568e32e495bacc443d413996f26390a2f68d3a

guides/user-experience/carousel-item-effects/guide.md
bramus: Squashed into f454f40f833c9a0f93a4cb445e8235b4ff3183fe

guides/user-experience/parallax-scroll-effects/guide.md
bramus: Made `sibling-index()` optional in 2613773109ce2f8f366b7e6dc603e8c2c2a18540

guides/user-experience/scroll-entry-exit-effects/guide.md
bramus: Added in f4491c5d570badb2783259683787241f255175a9

guides/user-experience/parallax-scroll-effects/guide.md
bramus: Dropped in a2557500565c47940a5b4edab5232ac339a6893a

bramus:
Addressed the comments. Some need some more input from your end, @philipwalton.

guides/user-experience/carousel-item-effects/guide.md
philipwalton: TBH, I still don't fully understand. Granted, I don't know this API well, but when I look at this code example it's not clear to me whether the ident is being defined on `view-timeline` and referenced from `animation-timeline`, OR if if the ident is being defined on `animation-timeline` and referenced from `view-timeline`. (I also don't know if `inline` is significant here, or whether/how to use it.)

If you haven't, I'd recommend testing whether AI tools can properly implement this use case following this example.

guides/user-experience/carousel-item-effects/guide.md
philipwalton: Maybe this makes more sense in a skill file that covers general SDA best practices instead of in this guide? WDYT?

guides/user-experience/parallax-scroll-effects/guide.md
philipwalton: Got it. That's more clear ... but now I'm worried it's over-explaining for my misunderstanding.

What about just saying it's a wrapper that group the parallax layers, and it sits inside the scrollable element? WDYT?

guides/user-experience/scroll-entry-exit-effects/guide.md
philipwalton: I'd drop all the parts that explain what the code does. Maybe keep the parts that emphasize that it's not perfect and not every effect can ge recreated using this approach.

guides/user-experience/scroll-progress-indicator/guide.md
philipwalton: Hmmm, I think your language states something different from what I'm suggesting.

My point was to say something to convince the AI that, even though SDA doesn't have full browser support, it's still worth using it today (even if you have to implement both the native version and the fallback) *because* it results in a better use experience for the majority of visitors.

guides/user-experience/scroll-progress-indicator/guide.md
philipwalton: Similar to above. Drop anything that is just explaining the code.

guides/user-experience/shrinking-header-on-scroll/guide.md
philipwalton: Can you clarify why this tip applies in the 100vh > 10vh case, but not in the example below?

guides/user-experience/carousel-item-effects/guide.md
bramus: I don’t know how else to explain it other than _“This creates a named view-timeline on each carousel item.”_ because that is what `view-timeline: --item inline` effectively does.

The **creates** there makes is pretty clear that this creates (and thus defines) a view timeline with that name, no?

The only clarification I can add is that it tracks the inline axis, as `view-timeline` is a shorthand for `view-timeline-name` and `view-timeline-axis`.

guides/user-experience/carousel-item-effects/guide.md
bramus: This is inherent to the skills based approach that is used to organize the guidance, so this wings back to a previous question from during our 1:1: how and where do I define these general best practices for anything that is scroll-driven animations related?

guides/user-experience/parallax-scroll-effects/guide.md
bramus: These additions seem useful, though:

- It is not the scrollable element, so its overflow should be clipped.
- Also give it a `height` that matches the height of one of the layers of the parallax effect.

Or would the LLM derive that automatically from the code snippet?

guides/user-experience/scroll-entry-exit-effects/guide.md
bramus: Done in 23f1e1d27771a6f32d7e15583434f1a8f0d31517

guides/user-experience/scroll-progress-indicator/guide.md
bramus: Addressed in ea089eeecdf7d918dd85b272426ced89bb66eaad

guides/user-experience/scroll-progress-indicator/guide.md
bramus: New attempt (to be included in all use-case guides):

> {{ BASELINE_STATUS("scroll-driven-animations") }}. Therefore, a fallback strategy is typically required.
> 
> For browsers that do not support scroll-driven animations, you can use a fallback to recreate the visual effects. The fallbacks are typically built with either a `scroll` listener (for `ScrollTimeline` effects) or the `IntersectionObserver` API (for `ViewTimeline` effects).
> 
> In browsers with built-in support for scroll-driven animations, ALWAYS use the native CSS implementation as those are more performant.
>
> Note that not every effect can be recreated using the fallbacks approach.

WDYT?

guides/user-experience/shrinking-header-on-scroll/guide.md
bramus: Oh, I see now … the demo used the wrong `animation-range-end`!? The value should be `150px` instead of `200px`. (This was not immediately visible because there was no other content present in the demo).

Fixed in 69a6a9f2a5dcb2b67b8db65e5e31ad8cea6dc697

bramus:
@philipwalton Addressed more comments.

As for the “scrollytelling” use-case I am not sure what to. Like I explained during our 1:1 they are different approaches with different outcomes (and totally different code). Do I drop it for now? Or leave as is?

As for the STA variant: that seems like a different issue to track then, as STA is a separate feature.

guides/user-experience/carousel-item-effects/guide.md
philipwalton: Ok, let's keep it here for now and we can refactor it to a skill later if needed.

guides/user-experience/carousel-item-effects/guide.md
philipwalton: > I don’t know how else to explain it other than _“This creates a named view-timeline on each carousel item.”_ because that is what `view-timeline: --item inline` effectively does.

Ok, this comment actually answers my question. The confusion was that I didn't know *which* line the comment applied to. It seems more obvious now that you've clarified, but I think it would make it more clear to add comments to the individual property declarations. For example:

```css
/*
  Creates a named view-timeline on each carousel item.
  The timeline is used to drive the animation that is
  applied on the same element.
*/
.scroller > * {
  /* Defines a custom view progress timeline name on the inline axis */
  view-timeline: --item inline;
  animation: animate auto linear both;
  /* Links the animation timeline to the view progress timeline named above */
  animation-timeline: --item;
}
```

guides/user-experience/carousel-item-effects/guide.md
```diff
@@ -7,4 +7,179 @@ web-feature-ids:
...
+
+When using the `view-timeline` property to create a scroll-driven animation:
+
+- **DO** use a CSS `<dashed-ident>` for the name (e.g. `view-timeline: --my-custom-name`)
+- **OPTIONAL** be explicit about the axis to track: When not targeting the default `block` axis (such as in a horizontal scroller), be explicit about which axis to track with `view-timeline-axis`.
+- **DO** make sure the scope of the lookup works: When the element that is declaring the `view-timeline` is not a flat tree ancestor of the animated element, hoist up the visibility of the `view-timeline`’s name by using `timeline-scope` on a shared ancestor.
+
+## Browser support and fallback strategies
+
+{{ BASELINE_STATUS("scroll-driven-animations") }}. Therefore, a fallback strategy is typically required.
```
philipwalton, line 145: I think this approach is fine (and I'm not suggestion you change it).

What I've done in the past when creating fallbacks for SDA is to define all of my styles as a function of a single custom property that essentially represents "scroll percentage complete". And then I add a scroll event listener that updates the value of that custom property.

Your solution is more robust (and probably supports more cases), but it also requires duplicating the animation declarations in the JS code.

philipwalton: _(Sorry, looks like this was an unsubmitted draft comments from my previous review)_

guides/user-experience/scroll-progress-indicator/guide.md
philipwalton: Works for me, thanks!

guides/user-experience/shrinking-header-on-scroll/guide.md
philipwalton: Makes sense now, thanks!

philipwalton:
> @philipwalton Addressed more comments.
> 
> As for the “scrollytelling” use-case I am not sure what to. Like I explained during our 1:1 they are different approaches with different outcomes (and totally different code). Do I drop it for now? Or leave as is?

I think it's fine to leave it as is.

philipwalton (approved):
Overall, this LGTM. Feel free to merge once you've addressed my last few comments.

guides/user-experience/parallax-scroll-effects/guide.md
philipwalton: Not sure if you changed it after making this comment, but the current framing LGTM.

guides/user-experience/carousel-item-effects/guide.md
bramus: Addressed in fbc0fd9709fc18b4f8a2b275ac7706f46f1c5f55.

guides/user-experience/scroll-progress-indicator/guide.md
bramus: Updated in b8f698c0a1f1ae77b85dbbf14b7aad7d466ac9c6

PR #284, opened 2026-03-11
Guide validation high level view

micahjo7:
1. Clean up display (move the "Authenticated" button on test view, add a "Back to Results" button at top)
2. Support auto-reload of data on test view when authenticating (so you don't have to manually refresh)
3. Add metrics for guide validation (# of runs that used expected guides, across each test and across the suite)
4. Update dashboard to see those results at suite top level:

Screenshot 2026-03-11 at 6 25 51 PM
image
image


paulirish (approved):
this is great!

PR #292, opened 2026-03-12
Adds skills at the discipline level

micahjo7:
Adds "performance" and "user-experience" skills which contain an "eval-ready" table of the guides in those folders. Adds this structure to the agent harness when skills are enabled in the config, and updates guide validation and dashboard too.

Results from these runs are at:

Full suite (12 eval-ready use-cases): https://googlechrome.github.io/guidance/dashboard.html?testId=negative-2026-03-11T19-16-17&source=remote
Negative suite: https://googlechrome.github.io/guidance/dashboard.html?testId=full-2026-03-11T18-47-25&source=remote

Synonymous runs with MCP:

Full: https://googlechrome.github.io/guidance/dashboard.html?testId=full-2026-03-11T15-46-34&source=remote
Negative: https://googlechrome.github.io/guidance/dashboard.html?testId=negative-2026-03-11T16-14-19&source=remote

image


paulirish:
thanks for making this happen!

harness/lib/utils.ts
```diff
@@ -0,0 +1,143 @@
...
+        taskName: file.replace(/\.md$/, ''),
+        baseApp: data.base_app || 'daily-grind',
+        prompt: content.trim()
+      });
+    }
+  }
+  return taskMap;
+}
+
+export function inventoryGuide(dir: string, taskMap: Map<string, TaskInfo>): GuideInventory {
```
paulirish, line 76: i think this has overlap with both sync-use-cases  and dev-guide's auditGuides()  (which already overlap)

if possible it'd be great to DRY that up into a canonical inventory method.

harness/lib/utils.ts
```diff
@@ -0,0 +1,143 @@
...
+    hasNegativeDemo: fs.existsSync(path.join(dir, NEGATIVE_DEMO_FILE)),
+    hasGrader: fs.existsSync(path.join(dir, GRADER_FILE)),
+    hasPrompts: fs.existsSync(path.join(dir, PROMPTS_FILE)),
+    hasTask: taskMap.has(name),
+  };
+}
+
+export type GuideStatus = 'eval-ready' | 'needs-test' | 'needs-calibration' | 'needs-expectations' | 'stub' | 'incomplete';
+
+export function classifyGuide(inv: GuideInventory): GuideStatus {
```
paulirish, line 123: same deal on overlap

paulirish (changes requested):

harness/lib/utils.ts
paulirish: fyi @rviscomi ... 

at the very least the run() method in sync-use-cases needs to get decomposed.

micahjo7: Okay, I've made an attempt at resolving this.. updated `sync-use-cases` to use `scanAllGuides` and GuideInventory` to address overlap. Then moved the "Steps" from the run() method into separate helper methods. Also moved all constants/helper methods to top of file.

Thought this would produce a cleaner diff but it doesn't really look like it 😭
@rviscomi let me know if you'd like to take care of this separately

harness/lib/utils.ts
micahjo7: Did not use this method yet because `sync-use-cases` currently has slightly different requirements from the `GuideStatus` here (checks for grader, prompt, demo, guide.. but not negative, expectations, etc.)

harness/lib/utils.ts
micahjo7: removed the changes to sync-use-cases. Once https://github.com/GoogleChrome/guidance/pull/312 is merged I can create the changes which optimizes it to use the`utils` methods in the PR (I can do that in the PR or the next).

paulirish (approved):
so good. thank you!

Issue #55, opened 2026-02-19
@function


rviscomi:
The @function CSS at-rule defines a custom function that takes CSS values or custom properties as arguments, and returns a CSS value. It can be based on conditional logic such as by using the @media at-rule.

### Chrome Documentation

<details>
<summary>View 3 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/delaying-shipping-of-css-functions
- https://developer.chrome.com/en/docs/devtools/css/reference
- https://web.dev/en/learn/css/custom-properties
</details>

---
Feature ID: function

**Use cases:**
- [x] #305
- [ ] #1253

PR #293, opened 2026-03-12
feat: Add use cases for CSS @function

micahjo7:
Adds fluid sizing and conditional calculation use cases following the project-use-cases skill. \n\nAddresses feature: https://github.com/GoogleChrome/guidance/issues/55

Generated using https://github.com/GoogleChrome/guidance/blob/guide-agent/.agents/skills/project-use-cases/SKILL.md.

@rviscomi I gave the agent the feature details and used the SKILL, asking it to create one use case, and it came up with the `fluid-sizing-calculation`

Then, I gave it more instructions to focus on the performance benefit of the feature, specifically related to offloading logic from JS into CSS directly- then it created `avoid-javascript-style-calculators`.


guides/user-experience/fluid-sizing-calculation/guide.md
```diff
@@ -0,0 +1,10 @@
+---
+name: fluid-sizing-calculation
+description: Encapsulate complex fluid sizing calculations into a reusable CSS custom function with @function.
```
philipwalton:
```suggestion
description: Reduce excessive style repetition by encapsulating complex or dynamic styling logic into reusable functions (such as a function that computes a fluid size value based on a set of input parameters).
```

I would generalize this use case, while still offering your specific one as an example.

Also, the original framing is fairly similar to [this previously-submitted use case](https://github.com/GoogleChrome/guidance/blob/main/guides/user-experience/fluid-scaling/guide.md), so it's good to be distinct:

guides/performance/avoid-javascript-style-calculators/guide.md
```diff
@@ -0,0 +1,10 @@
+---
+name: avoid-javascript-style-calculators
+description: Eliminate main-thread blocking JavaScript event listeners and layout thrashing by encapsulating conditional style computations directly in CSS with @function.
```
philipwalton: TBH, I'm not sure this use case makes sense. I mean, the use case itself is fine, but I don't know if `@function()` is the appropriate recommendation, since there are many other ways to do this with pure CSS that don't involving using `@function()`.

My recommendation would be to drop this one and just stick with your other use case.

guides/user-experience/fluid-sizing-calculation/guide.md
```diff
@@ -0,0 +1,10 @@
+---
+name: fluid-sizing-calculation
```
philipwalton:
```suggestion
name: reduce-style-repetition
```

guides/user-experience/fluid-sizing-calculation/guide.md
philipwalton: Another potential use case (that is arguably a subset of this use case), would be something like:

- **`improve-style-readability`:** Encapsulate complex or hard-to-understand style logic into self-documenting style functions that clearly communicate the intent of the code.

...though arguably this is getting into skills territory. WDYT?

guides/performance/avoid-javascript-style-calculators/guide.md
micahjo7: removed

guides/user-experience/fluid-sizing-calculation/guide.md
micahjo7: consolidated to the `reduce-style-repetition` use case

philipwalton (approved):

PR #295, opened 2026-03-12
Project-level agent skills to assist contributors

rviscomi:
Skills for:

- creating use cases
- writing guidance
- creating evals

Moves most of the guidance out of AGENTS.md


rviscomi:
@paulirish FYI for the failed CLA check. Are you able to fix that?

paulirish:
> @paulirish FYI for the failed CLA check. Are you able to fix that?

yeah. fixed up.

.agents/skills/project-guides/SKILL.md
```diff
@@ -0,0 +1,50 @@
...
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/API/Feature
+---
+```
+* **web-features**: Must be a list of accurate IDs found via webstatus.dev. If the ID is missing, inform the USER.
+* **sources**: Must be a list of primary source URLs used to synthesize the document.
+
+### 2. Tone and Formatting
+* **MANDATORY:** Use strict imperative directives. Start instructions with `MANDATORY:`, `DO`, and `DO NOT`. Coding agents respond best to rigid constraints.
+* **Focus:** Keep it abstract but short. No fluff. No conversational text. Include a brief overview of the feature and why it is useful for the use-case.
```
philipwalton:
```suggestion
* **Focus:** Keep it abstract but short. No fluff. No conversational text. Include a brief overview of the use case and explanation of why the solution outlined in the guide is the recommended approach.
```

.agents/skills/project-use-cases/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+
+The primary goal of this stage is to translate a technical web platform feature into a carefully selected set of its most common and important use cases. This is the first of three stages in creating guidance:
+
+1. Stage 1: Identifying use cases for a feature (you are here)
+2. Stage 2: Authoring guidance for a use case
+3. Stage 3: Evaluating guidance for a use case
+
+## Manual research and discovery
+
+Use Deep Research in NotebookLM to discover sources and real-world implementations you might not be aware of. Focus on finding complex edge cases, performance implications, and emerging best practices. Deep Research is particularly effective at surfacing GitHub discussions, W3C specifications, and developer blog posts that highlight non-obvious constraints.
```
philipwalton: FWIW, I used Deep Research in the Gemini web app and not NotebookLM. Do we need to be this prescriptive?

Also, this is written for the agent, right? Should we phrase these as "Suggest that the user use Deep Research..."?

.agents/skills/project-use-cases/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+> Thoroughly review the following resources for the `fetchLater()` API, as well as any additional resources you can find on the web, and identify the top 2-5 developer use cases for this API.
+>
+> - https://developer.mozilla.org/en-US/docs/Web/API/Window/fetchLater
+> - https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Deferred_Fetch
+> - https://developer.chrome.com/blog/fetch-later-api-origin-trial
+> - https://github.com/WICG/pending-beacon/blob/main/docs/fetch-later-api.md
+
+## Identifying action-oriented tasks
+
+A "use case" in this project is not a description of a feature; it's how the feature would be used in the wild.
```
philipwalton:
```suggestion
A "use case" in this project is not a description of a feature; it's a task that the user is trying to implement, or a problem they're trying to solve. The feature is only relevant in the sense that it's part of the recommended solution for the use case.
```

.agents/skills/project-use-cases/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+
+* **Action-oriented thinking**: Frame every use case as a task. Instead of "Scroll-driven animations support horizontal scrolling," use something like "Synchronize an animation's progress with the horizontal scroll distance of a container."
```
philipwalton:
```suggestion
* **Action-oriented thinking**: Frame every use case as a task, and make sure it starts with a verb. Instead of "Scroll-driven animations support horizontal scrolling," use something like "Synchronize an animation's progress with the horizontal scroll distance of a container."
```

.agents/skills/project-use-cases/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+* **Bridge the knowledge gap**: Assume the developer knows *what* they want to build (e.g., "I need a sticky header that shrinks on scroll") but might not know *which* modern web feature is the best solution (e.g., scroll-driven animations). Your use cases should facilitate this discovery by focusing on the desired outcome.
```
philipwalton, line 56:
```suggestion
* **Bridge the knowledge gap**: Assume the developer knows *what* they want to build (e.g., "I need a sticky header that shrinks on scroll") but might not know *which* modern web feature is the best solution (e.g., scroll-driven animations). Your use cases should facilitate this discovery by focusing on the desired outcome.
* **Don't get too specific**: The use case must be general enough to match a wide range of relevant user prompts. Try to be as general as possible, while still faithfully representing the use case. For example, instead of saying "Fade an image in/out..." say "Smoothly hide/show a component...".
* **Focus on the WHAT not the HOW**: Do not mention the solution in the use case description. For example, avoid phrases like "...by doing..." or "...through the use of...". Ideally, the use case description should remain constant, even if the recommended features or best practices for implementing it change over time.
```

.agents/skills/project-use-cases/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+* **Targeted scope**: Aim for 2-5 distinct use cases per feature. Each should represent a unique implementation pattern or a significant variation in how the feature is applied.
```
philipwalton:
```suggestion
* **Scope**: Aim for 2-5 distinct use cases per feature. Each use case should represent a distinct implementation pattern or a significant variation in how the feature is applied. IMPORTANT: Not every sub-feature or feature variation needs a use cases.
```

.agents/skills/project-use-cases/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Minimizing overlap
+
+This guidance is ultimately served through a RAG (Retrieval-Augmented Generation) search system. If multiple guides have significant overlap, coding agents may struggle to select the most relevant one, leading to confusing or contradictory advice.
+
+* **Check existing guides**: Before creating a new use case, review existing guides in the same discipline.
+* **Search by web-feature-id**: Each guide lists the web features it relies on in the `web-feature-ids` metadata field. Search for the ID of the feature you're writing about in existing guides and open PRs to see how it's being used.
```
philipwalton, line 68: Ideally we'd be able to run the `search_use_cases` tool to identify any similar, existing use cases.

.agents/skills/project-use-cases/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+* **Merge or differentiate**: If your proposed use case is substantially similar to an existing one, do not create a duplicate. Instead, consider how the existing guide should be updated to include your new scenario as a variation or a specific directive.
+* **Distinct value proposition**: Every new guide must offer a distinct solution to a distinct problem.
+
+## Implementation and scaffolding
+
+The following steps are REQUIRED for creating a new use case:
```
philipwalton, line 74: I'm now realizing that some of my suggestions above are covered here. Not sure where it's better to include them, so I'll make suggestions in both places for now.

.agents/skills/project-use-cases/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+
+* **Step 1: Describe the use case**
+
+  You MUST choose a short, action-oriented description of the problem the feature solves. The description must be a single sentence that captures the essence of the use case.
```
philipwalton:
```suggestion
  You MUST choose a short (max 1024 characters), action-oriented description of the problem the feature solves. The description must be a single sentence, start with a verb, and answer the question: "What is the user trying to DO?"
```

.agents/skills/project-use-cases/SKILL.md
```diff
@@ -0,0 +1,101 @@
...
+    - fetch-priority
+    - fetch
+  sources:
+    - https://web.dev/articles/fetch-priority
+  ---
+  ```
+
+* **Step 5: Create the `demo.html` file**
+
+  Create a `demo.html` file in the new subdirectory. This file should be an ultra-minimal example of a correct implementation of the use case. It should be self-contained with inline scripts and styles for any necessary functionality. Use placeholder URLs for any subresources like images or videos.
```
philipwalton:
```suggestion
  Create a `demo.html` file in the new subdirectory. This file should be minimal example of a correct implementation of the use case, without unnecessary fluff or polish. The demo file should be self-contained with inline scripts and styles for any necessary functionality. Use placeholder URLs for any subresources like images or videos.
```

.agents/skills/project-guides/SKILL.md
```diff
@@ -0,0 +1,50 @@
...
+This is the second of three stages in creating guidance:
+
+1. Stage 1: Identifying use cases for a feature
+2. Stage 2: Authoring guidance for a use case (you are here)
+3. Stage 3: Evaluating guidance for a use case
+
+**MANDATORY RULES FOR WRITING `guide.md`:**
+
+### 1. YAML Frontmatter Schema
+`guide.md` must start with this YAML frontmatter structure:
```
philipwalton:
```suggestion
`guide.md` must start with this YAML frontmatter structure (added in **Stage 1**):
```

.agents/skills/project-use-cases/SKILL.md
paulirish: also this is meant to be "Use Deep Research **and** NotebookLM"  ? 

agree with asking the user to do it.  unless we want to wire this up: https://ai.google.dev/gemini-api/docs/deep-research  or just leave a pointer to it and have the coding agents do it

paulirish:
Amazing review Phil. Thanks for taking such a close look! important stuff here.

.agents/skills/project-use-cases/SKILL.md
paulirish: also do yall have a shared notebooklm  ? can i get in on that?

rviscomi: Yes, I'll update this to clarify that the user should do the research, not the agent. It's here mainly for the agent to guide the user on how to do the research.

I'll also split up this example prompt into two steps to clarify that the process is to deep research sources first, and analyze the use cases second.

As for whether to use deep research in the Gemini app, NotebookLM, or both, my suggestion is to use the deep research option of NotebookLM:

Image

I'm being more prescriptive about using NotebookLM because of the finer-grained control it gives us over adding, removing, or disabling individual sources. OTOH deep research in the Gemini app defaults to producing a huge report, and it's not possible to manage specific sources. Maybe most importantly, NotebookLM tries to base its responses only on the sources, so if you ask it about something unrelated, it'll respond like "The provided sources do not contain direct recommendations on XYZ". Meanwhile the Gemini app will try to answer the question using additional (unrelated, unvetted) sources, muddying the quality and attribution.

.agents/skills/project-use-cases/SKILL.md
rviscomi: Yeah, when we have the CLI set up we can instruct the agent to use that to look for similar use cases. Right now having the MCP server installed in development isn't a prerequisite.

.agents/skills/project-use-cases/SKILL.md
rviscomi: Probably doesn't hurt to have it in both places, so let's roll with it

.agents/skills/project-use-cases/SKILL.md
philipwalton: That makes sense if you want to just stick to the sources, but I think there are cases where you actually want to agent to find sources that you weren't aware existed.

For example, when I did Deep Research in the Gemini App for the `fetchLater()` API, it found an obscure discussion about cross-origin fetch quota that is relevant for Ad Tech use cases, which I was not aware of.

rviscomi: Yeah, NotebookLM will also do the deep research to find additional sources on top of the ones you provide in the prompt.

PR #297, opened 2026-03-13
Updates optimize tasks to use empty app

micahjo7:
In previous runs, the optimize tasks told the agent to create new pages, which the grader did not pick up.

This PR updates the tasks so that they use the original prompts from the guides, but also uses an "empty" app to be compatible with those prompts.

Results: https://googlechrome.github.io/guidance/dashboard.html?testId=optimize-suite&source=remote


paulirish (approved):
nice

Issue #68, opened 2026-02-19
overflow: clip


rviscomi:
The overflow: clip CSS declaration clips an element's overflowing content to the box that's defined by the overflow-clip-margin property. Unlike overflow: hidden, it allows you to hide overflow on one axis while keeping overflow on the other axis visible.

### MDN Documentation
- [overflow](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/overflow)

### Chrome Documentation

<details>
<summary>View 2 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/overflow-replaced-elements
- https://web.dev/en/learn/css/overflow
</details>

---
Feature ID: overflow-clip

**Use cases:**
- [x] #341

PR #298, opened 2026-03-13
use cases: overflow: clip

marcianoskate:
Address the use cases for `overflow: clip` as described in #68 

## clip-content-on-single-axis

image

## disable-programmatic-scrolling 

pending screenshot

## preserve-visual-effects-with-clip-margins 

pending screenshot

## prevent-scroll-animation-conflicts

pending screenshot


guides/user-experience/preserve-visual-effects-with-clip-margins/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: preserve-visual-effects-with-clip-margins
```
philipwalton:
```suggestion
name: preserve-overflow-effects
```

guides/user-experience/preserve-visual-effects-with-clip-margins/guide.md
```diff
@@ -0,0 +1,6 @@
...
+description: Use `overflow-clip-margin` to expand the clipping boundary and preserve visual effects like box-shadow or focus outlines.
```
philipwalton:
```suggestion
description: Allow limited decorative overflow (e.g. box shadows, outlines, badges, etc.) on an HTML element, while otherwise clipping its contents to within a defined box size.
```

Avoid the use of the API in the use case description.

guides/user-experience/preserve-visual-effects-with-clip-margins/guide.md
```diff
@@ -0,0 +1,6 @@
...
+web-feature-ids: 
+  - overflow-clip
```
philipwalton:
```suggestion
  - overflow-clip
  - overflow-clip-margin
```

philipwalton (changes requested):
Thanks @marcianoskate, FYI I had previously discussed with @rviscomi that perhaps we should remove this feature and only covered it a skill.md file (e.g. use `overflow: clip` instead of `overflow: hidden`), but I think your use case that incorporates `overflow-clip-margins` is actually a good one to have a guide for.

So, to restate, I'd remove all but the 3rd use case (which I left some comments on). The other use cases are probably better handled via a skill.

philipwalton:
> Thanks @marcianoskate, FYI I had previously discussed with @rviscomi that perhaps we should remove this feature and only covered it a skill.md file (e.g. use `overflow: clip` instead of `overflow: hidden`), but I think your use case that incorporates `overflow-clip-margins` is actually a good one to have a guide for.

Actually, after playing around with the use case I suggested for a bit, I changed my mind. I don't think it's particularly compelling. Instead I updated this to the more technical and straight-forward use case of having fine-grained control over which box is chosen to contain the overflow.

philipwalton (approved):

Issue #283, opened 2026-03-11
Create guide and evals for the conditional-async-dependencies use case


philipwalton:
Conditionally load or initialize async dependencies (such as importing polyfills for missing web features) without requiring complex orchestration across all of a page's script dependencies.

Affected web-feature IDs: [top-level-await](https://webstatus.dev/features/top-level-await)

Use case subdir: [guides/performance/conditional-async-dependencies](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/performance/conditional-async-dependencies)

Related features: #84

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

PR #304, opened 2026-03-16
Adds the guide for top-level await for fetching async dependencies.

malchata:
Addresses #283 to add a guide for fetching async dependencies using top-level await. First shot at this, so expecting quite a bit of feedback.


guides/performance/conditional-async-dependencies/guide.md
```diff
@@ -4,3 +4,51 @@ description: Conditionally load or initialize async dependencies (such as import
 web-feature-ids:
     - top-level-await
 ---
+
+# Conditional polyfill dependency fetching and initialization with top-level await
```
philipwalton:
```suggestion
# Conditionally load or initialize async dependencies
```

I think it's good to be consistent with the description.

guides/performance/conditional-async-dependencies/guide.md
```diff
@@ -4,3 +4,51 @@ description: Conditionally load or initialize async dependencies (such as import
...
+3. Wrap the dynamic import logic within a `try/catch` block to prevent unhandled rejections from failing the entire module graph.
+4. Execute a dynamic `import()` to fetch and evaluate the required polyfill.
+5. Provide fallback logic in the `catch` block to ensure graceful degradation if the network request fails.
+6. Export the module's public API only after the environment has been successfully patched and verified.
+
+## Example code
+
+```html
+<!-- 1. Ensure the execution environment allows ESM. -->
+<script type="module">
```
philipwalton: Nit: I'd update this entire code block to be a JS rather than an HTML, since that's how most people author JS. (Also, an inline module script can't really export anything, because there no way to import it.)

guides/performance/conditional-async-dependencies/guide.md
```diff
@@ -4,3 +4,51 @@ description: Conditionally load or initialize async dependencies (such as import
...
+  // 2. Check for browser capabilities to avoid running code
+  //    in contexts where a feature doesn't require a polyfill
+  if (typeof globalThis.IntersectionObserver === 'undefined') {
```
philipwalton: I'd avoid using this an example. `IntersectionObserver` has been Baseline widely available for many, many years, and this could confuse AI agents into thinking `IntersectionObserver` code needs to use a polyfill.

guides/performance/conditional-async-dependencies/guide.md
```diff
@@ -4,3 +4,51 @@ description: Conditionally load or initialize async dependencies (such as import
...
+
+# Conditional polyfill dependency fetching and initialization with top-level await
+
+Top-level await enables ECMAScript Modules (ESM) to act as asynchronous functions, pausing their own evaluation and blocking downstream imports until specific Promises resolve. This makes it an ideal mechanism for conditional dependency loading, such as fetching environment-specific polyfills before application code executes.
+
+## How to implement
+
+1. Verify that your module is being parsed as an ECMAScript module.
+2. Programmatically detect the absence of the required API or feature in the current runtime environment.
+3. Wrap the dynamic import logic within a `try/catch` block to prevent unhandled rejections from failing the entire module graph.
```
philipwalton: I think this is only true when importing from a third-party domain. If you're using a bundler that will create a chunk hosted along with all your other chunks, then I don't think it's necessary to use a try/catch. (And since most people use bundlers, I expect that will be the majority case.)

guides/performance/conditional-async-dependencies/guide.md
```diff
@@ -4,3 +4,51 @@ description: Conditionally load or initialize async dependencies (such as import
...
     - top-level-await
 ---
+
+# Conditional polyfill dependency fetching and initialization with top-level await
+
+Top-level await enables ECMAScript Modules (ESM) to act as asynchronous functions, pausing their own evaluation and blocking downstream imports until specific Promises resolve. This makes it an ideal mechanism for conditional dependency loading, such as fetching environment-specific polyfills before application code executes.
+
+## How to implement
+
+1. Verify that your module is being parsed as an ECMAScript module.
```
philipwalton: Yes, but most people are going to be using a bundler, which should take care of that for you.

guides/performance/conditional-async-dependencies/guide.md
```diff
@@ -4,3 +4,51 @@ description: Conditionally load or initialize async dependencies (such as import
...
+2. Programmatically detect the absence of the required API or feature in the current runtime environment.
+3. Wrap the dynamic import logic within a `try/catch` block to prevent unhandled rejections from failing the entire module graph.
+4. Execute a dynamic `import()` to fetch and evaluate the required polyfill.
+5. Provide fallback logic in the `catch` block to ensure graceful degradation if the network request fails.
```
philipwalton: Again, only if the request is to a 3P domain.

guides/performance/conditional-async-dependencies/guide.md
```diff
@@ -4,3 +4,51 @@ description: Conditionally load or initialize async dependencies (such as import
...
+6. Export the module's public API only after the environment has been successfully patched and verified.
```
philipwalton: I think it would be more helpful to move this step higher and frame it as: "Ensure your conditional and top-level await run before the features being loaded or initialized is used.

guides/performance/conditional-async-dependencies/guide.md
```diff
@@ -4,3 +4,51 @@ description: Conditionally load or initialize async dependencies (such as import
...
+
+```html
+<!-- 1. Ensure the execution environment allows ESM. -->
+<script type="module">
+  // 2. Check for browser capabilities to avoid running code
+  //    in contexts where a feature doesn't require a polyfill
+  if (typeof globalThis.IntersectionObserver === 'undefined') {
+    try {
+      // 3 & 4: Dynamically await the polyfill import
+      await import('https://cdn.polyfill.io/v3/polyfill.min.js?features=IntersectionObserver');
```
philipwalton:
```suggestion
      await import('path/to/polyfill.js');
```

I'd just use a generic reference, and definitely don't use a reference to polyfill.io, since that service has been compromised.

guides/performance/conditional-async-dependencies/guide.md
```diff
@@ -4,3 +4,51 @@ description: Conditionally load or initialize async dependencies (such as import
...
+    } catch (error) {
+      // 5: Handle network or loading failures gracefully
+      console.error('Failed to load IntersectionObserver polyfill:', error);
+    }
+  }
+
+  // 6: Export bindings synchronously; downstream consumers are guaranteed the polyfill is evaluated (or handled)
```
philipwalton: I would expect the code here to use `IntersectionObserver` (or whatever you switch this to) freely at this point, rather than creating another wrapper.

guides/performance/conditional-async-dependencies/guide.md
```diff
@@ -4,3 +4,51 @@ description: Conditionally load or initialize async dependencies (such as import
...
+    observer.observe(element);
+    return observer;
+  };
+</script>
+```
+
+## Best practices
+
+* **DO** ensure an ESM execution context. In the browser, top-level await is syntactically invalid in `<script>` blocks that don't use the `type="module"` attribute.
+* **DO** implement strict error boundaries. Unhandled Promise rejections using top-level await will abort module evaluation, resulting in missing dependencies or non-execution of application logic.
```
philipwalton: Again, only if importing from a 3P site (but that's true of all imports).

guides/performance/conditional-async-dependencies/guide.md
```diff
@@ -4,3 +4,51 @@ description: Conditionally load or initialize async dependencies (such as import
...
+  export const createObserver = (element, callback) => {
+    if (!globalThis.IntersectionObserver) return null;
+    const observer = new IntersectionObserver(callback);
+    observer.observe(element);
+    return observer;
+  };
+</script>
+```
+
+## Best practices
```
philipwalton: This section should include some recommendations to avoid this bug (which is what is preventing this feature from being Baseline): https://bugs.webkit.org/show_bug.cgi?id=242740

guides/performance/conditional-async-dependencies/guide.md
philipwalton, line 1: The use case in this file is generic and mentions both loading or initializing async dependencies, but this guide only mentions loading polyfills. I think you should add an example that includes a non-polyfill async dependency.

philipwalton, line 1: This file is also missing a fallback section, which in this case I think should discuss how to use the feature in a way that doesn't trigger the Safari bug.

philipwalton (changes requested):

malchata:
Took another pass at this by having Gemini CLI generate a new guide using the appropriate agent skills. I resolved all prior open comments and considering this a bit of a fresh start.

malchata:
@rviscomi, this is ready for a review.

rviscomi:
Oh this is a tricky one.

BCD considers Safari as having "partial support" because of the bug. That means top-level await isn't actually Baseline.

The `### Fallback strategies` section says:

> If you must support older browsers that lack top-level `await` support entirely, you cannot use it to block module execution. Instead, you must use standard asynchronous functions or dynamic `import()`

But the thing is that coding agents don't look at when the feature landed in the browser (Safari 15), they're instructed to look only at the Baseline target (Widely Available by default) and the feature's Baseline status (Limited Availability in this case).

**So coding agents' default behavior should be to fall back to regular (not top-level) await within an async function.** Only when a coding agent is set to target Limited Availability should it actually use top-level await.

That means for the purposes of eval'ing this use case, I think we actually have to write the expectations "the Widely Available way" without top-level await at all 🫤 

@paulirish @micahjo7 let's think about a way to eval multiple Baseline targets. We may also want to stub features to have specific Baseline statuses to avoid sudden eval failures when they mature to Widely Available.

malchata:
(Accidentally closed this in trying to reply 😅)

So, my understanding is that the example in the guide already does this by calling `await import(/* ... */)` in an exported `async` function (`initializeDependencies` in this example), then calls that exported function as a promise using `.then`. I think I'm missing something here—isn't this example already being done in the Widely available way without using top-level `await`?

rviscomi:
What I'm suggesting is that there's a conflict between the fallback strategy in the guide and the modern approach shown in the demo.

Because top-level await [is not Baseline](https://webstatus.dev/features/top-level-await), agents will follow the instructions in the fallback section _to not use top-level await at all_. That would mean that the correct thing to show in the demo (and later, the thing we want to eval against) would be the approach without top-level await.

Instead, if what we actually want agents to do is use top-level await in a way that avoids the Safari bug, we should drop the fallback recommendation and clarify that because all browsers have supported this for years, no additional action is needed.

My understanding is that @philipwalton does think top-level await it safe to use if care is taken to avoid the Safari bug (which the guide explains well), so we should go the latter route.

guides/performance/conditional-async-dependencies/guide.md
```diff
@@ -2,5 +2,96 @@
...
+
+// You must manually await the initialization before running your app logic
+initializeDependencies().then(() => {
+  // Safe to use the polyfilled feature here
+  const myPopover = document.getElementById('my-popover');
+  if (myPopover) {
+    myPopover.showPopover();
+  }
+});
+```
```
rviscomi:
```suggestion
Top-level `await` has been supported in all major browsers since 2021 (Chrome 89, Firefox 89, Safari 15). Because of this broad support, **you do not need to implement a fallback strategy for modern web applications.**

As long as you follow the guidance in the previous section to **avoid the Safari execution order bug**, you can safely rely on top-level `await` directly to manage your async dependencies.

You only need to avoid top-level `await` and fall back to standard asynchronous functions or dynamic `import()` orchestration if your application is explicitly required to support legacy browsers released before 2021.
```

rviscomi (changes requested):

rviscomi (approved):
LGTM

PR #306, opened 2026-03-16
Add Chrome Extensions & Chrome WebStore skill draft

sebastianbenz:
Both are still WIP and I'll update them over time.


skills-drafts/chrome-webstore/references/chrome-webstore.skill
rviscomi, line 1: Is this file needed?

sebastianbenz: nope, removed

paulirish (approved):

PR #96, opened 2026-02-20
fix: Resolves outdated antigravity selectors 

pattishin:
**Environment:** Antigravity IDE from v1.27.3 / Node v25.6.1
**Issue:**
* While running `pnpm run suite`, the agent seemed to be hanging and no longer filling out the agent conversation box.
Screenshot 2026-02-18 at 10 25 00 AM

It looks like the selectors were updated recently, so the following: 
* updates selectors to re-target the conversation box
* removes the iframe locator
* selects the model before the latest default before prompt is submitted to ensure test stability

PR #308, opened 2026-03-16
Fix MCP logging for Jetski, enable trajectories in dashboard

micahjo7:
Jetski needed some updates after the structural changes we've made recently.

The MCP logging was updated to be configured properly, and we rename the trajectory file to prepend it with "session-" so that the dashboard knows which HTML file to use as the trajectory.

To be merge after/with https://github.com/GoogleChrome/guidance/pull/96.

### DIFF
--- a/harness/agents/jetski-agent.ts
+++ b/harness/agents/jetski-agent.ts
@@ -101,7 +101,7 @@ function setupIsolatedWorkDir(): string {
  */
 async function bypassInitialDialogs(page: Page): Promise<void> {
   console.log('Bypassing initial dialogs...');
-  
+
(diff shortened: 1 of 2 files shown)


harness/agents/jetski-agent.ts
```diff
@@ -151,7 +152,7 @@ async function bypassInitialDialogs(page: Page): Promise<void> {
   } catch (e: any) {
     console.log('Error bypassing Welcome Wizard:', e.message);
   }
-
```
paulirish, line 154: aw the poor vibecoded whitespace. we will miss you!  :)

paulirish (approved):

PR #309, opened 2026-03-16
Added form-best-practice SKILL.md

samdutton:


paulirish (approved):

paulirish:
Thanks Sam! Feels good to see the evolution of that chrome extension and how the same kinda thing ends up here :)

PR #312, opened 2026-03-17
Refactor sync-use-cases and harden the requirements

rviscomi:
**Summary**

  - Refactored sync-use-cases.ts into focused, testable functions making run() a thin 10-line orchestrator
  - Added comprehensive test suite (sync-use-cases.test.ts) with 62 tests
  - Dry run is now the default — only `DRY_RUN=false` disables it; updated package.json scripts and workflow accordingly
  - Project board status is now fetched before syncing — the script paginates through all project items to retrieve current status values and skips updates when the status is already correct, eliminating noisy no-op updates
  - Dry run output now only logs what actually changed — title, body, labels, and state are each suppressed when already up to date; "up to date" is only logged  when nothing at all changed
  - Validation errors now block issue sync for the affected use case (previously only a missing name would block it)
  - grader/prompts mismatch is now an error (was a warning), but only when the guide has content — stub guides are exempt
  - Added placeholder demo.html files for 17 use cases missing them
  - Fixed prompts.md bullet formatting for pull-to-reveal, scroll-target-on-load, and autofill-highlight-inputs so task prompts are generated correctly
  - Renamed `prompts.md` to `_prompts.md` for any use cases that would have otherwise triggered validation errors for having partial eval files (no `grader.ts`)

**Test plan**

  - `npm test` in guides/ passes all 62 unit tests and dry run completes cleanly
  - Dry run output shows only genuine changes (no spurious status updates)
  - Merging to main triggers the sync workflow with DRY_RUN=false and applies
  changes to GitHub

**Affected issues**

- This will create a new issue for the `flicker-free-client-side-ab-testing` use case, which was accidentally ignored by the script because it contained the word `test`


guides/user-experience/pull-to-reveal/grader.ts
rviscomi, line 1: @pattishin I generated this with `gd dev`. Could you give it a look and make sure that it's implementing the expectations correctly?

I needed to do it in this PR because I'm changing the severity of the warnings about missing eval-related files, and not having them would start to cause CI errors.

guides/user-experience/scroll-target-on-load/grader.ts
rviscomi, line 1: @pattishin same comment, could you take a look?

guides/user-experience/pull-to-reveal/grader.ts
rviscomi: As discussed offline, @pattishin will review the grader changes and make any changes after merging this PR

rviscomi:
I ended up reverting all of my `gd dev` generated changes, which would have resulted in those use cases getting closed without SME review. Instead, to work around the new CI failures, I renamed `prompts.md` files to `_prompts.md` to signify that they're still a WIP and would be ignored by the script. Contributors would need to rename it back before generating the eval files.

I'll merge this PR now, which will more strictly enforce having all eval files instead of some, and will reopen closed feature issues when their use cases are still open.

cc @philipwalton

PR #323, opened 2026-03-17
Add trajectory artifacts view for Claude Code.

formgit:
- Implement a simple trajectory viewer just for Claude Code as the exiting trajectory viewer doesn't work for Claude Code.
- Hook the new Caude Code trajectory viewer with the dashboard


paulirish (approved):
thanks! this'll be nice :) 
We can upstream this into go/trajectory too, at some point

PR #333, opened 2026-03-17
Run typecheck in grader-gen

rviscomi:
A few times I've run grader-gen or `gd dev`, it's created typechecking issues that I needed to manually resolve. This should make it so that the auto-generated grader file already passes typechecking.


guides/grader-gen.ts
```diff
@@ -139,6 +139,12 @@ async function runGraderGeneration(targetDir: string, prompt: string): Promise<v
       console.error(`Error: grader.ts was not generated by Gemini CLI in ${workDir}`);
     }
 
+    try {
+      execSync('npm run typecheck', { cwd: __dirname, stdio: 'inherit' });
+    } catch {
+      throw new Error(`Typecheck failed on generated grader.ts. Review the file for type errors.`);
```
micahjo7, line 145: if this throws, will `gd dev` pick up where it left off after fixing the errors?

maybe it should include the errors, or tell the user they need to run typecheck to figure out what to fix?

paulirish: stdio inherit should flush the errors to terminal. so.. hypothetically this is fine.

paulirish (approved):

PR #347, opened 2026-03-18
Set the default agent to Gemini CLI

rviscomi:
I think it might still be useful to have a way to configure agents via CLI args, but for now I pretty much always prefer Gemini CLI as my default agent for quick tests, and I assume most other people do too


micahjo7:
Leiming will work on b/493704095 to address this problem.

micahjo7 (approved):

PR #348, opened 2026-03-18
Adds Skills + CLI approach

micahjo7:
Adds the modern-web-use-cases SKILL and infrastructure to call the guide search and retrieval with CLI.

Results: https://googlechrome.github.io/guidance/dashboard.html?testId=skills-cli-2026-03-23T08-48-17&source=remote


guides/modern-web-use-cases/SKILL.md
```diff
@@ -0,0 +1,58 @@
...
+## Usage Instructions
+
+You can execute these tools using standard `node` from the command line from **any** directory. 
+
+### 1. Search Use Cases
+
+Search with an action-oriented query summarizing what you want to achieve.
+
+```bash
+node /Users/micahjo/guidance/serving/scripts/search.ts "<query>"
```
micahjo7: TODO: will need to fix this, and below

Because we run in an isolated home dir we need to tell the agent where the script is...

serving/scripts/search.ts
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  const store = new Store();
+  const embedder = Embedder.getInstance();
+
+  try {
+    const vector = await embedder.embed(query);
+    const results = await store.search(vector);
+    logToolResult("search_use_cases", results.map(r => ({ id: r.id, distance: r.distance })));
+    console.log(JSON.stringify(results, null, 2));
```
paulirish: i'd suggest this inner stuff moves into serving/lib/search.ts
and similar for serving/lib/retrieve

then i'd go with a singular CLI endpoint

`serving/bin/modern-web`

with a node hashbang. and use node's `util.parseArgs()`

in my head i'm thinking

```sh
./bin/modernweb --search <query>
./bin/modernweb --retrieve < use_case_id | use_case_ids >
```

https://clig.dev/ has thoughts on CLI design which don't seem bad.

open to alternatives tho. wdyt

harness/lib/guide_validation.ts
```diff
@@ -66,3 +45,99 @@ export async function collectGuidesUsed(dirPath: string, enableSkills: boolean):
     return [...new Set(guidesFromLog)];
   }
 }
+
+async function collectGeminiCliGuides(dirPath: string): Promise<string[]> {
```
paulirish: nit: i'd move these guide collection methods into the individual agent harness files.

harness/lib/guide_validation.ts
```diff
@@ -66,3 +45,99 @@ export async function collectGuidesUsed(dirPath: string, enableSkills: boolean):
...
+      }
+    }
+  } catch (e) {
+    console.error(`Error reading session files in ${dirPath}:`, e);
+  }
+  return [...new Set(guidesFromSkills)];
+}
+
+async function collectJetskiGuides(dirPath: string): Promise<string[]> {
+  // TODO: Implement skills guide collection for Jetski
```
paulirish: i know how but it's not something we can land in github. we'd need to rely on a glinux-only binary

guides/modern-web-use-cases/SKILL.md
paulirish: we can do this relative to where the skill md is. 

in a simple world it's actually WITHIN the modern-web-use-cases folder.. that would be the standard skill convention.

paulirish (commented):
this is great stuff. thanks for getting this working

harness/lib/guide_validation.ts
micahjo7: we should still have the MCP server logs at least (renamed to `modern-web.log`)

serving/scripts/search.ts
micahjo7: sgtm, updated

guides/modern-web-use-cases/SKILL.md
micahjo7: updated to remove the hardcoded path in prep for the changes we'll be making next (single packaged dist)

paulirish (approved):

PR #350, opened 2026-03-19
Include feature names in seach results

rviscomi:
Feature info could be useful for a coding agent when trying to understand if a use case is appropriate for a given task. Use case descriptions are more focused on the problem statement, so this adds the names of the web features directly to the search results.

Example result (what an agent would see):

```json
{
  "id": "optimize-preload-priority",
  "description": "Optimize the relative priority of preloaded content to reduce critical resource load delays.",
  "category": "performance",
  "featuresUsed": [
    "Fetch priority",
    "<link rel=\"preload\">"
  ],
  "distance": "1.08"
}
```

### DIFF
--- a/serving/mcp-server/data/baseline.ts
+++ b/serving/mcp-server/data/baseline.ts
@@ -176,6 +176,16 @@ export function resolveFeatureId(featureId: string): string[] {
 }
 
 /**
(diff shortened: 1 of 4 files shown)


philipwalton (approved):
LGTM

PR #354, opened 2026-03-19
Refactor suite configs to support runtime overrides

formgit:
- Refactored suiteConfig to be mutable and handle runtime overrides
- Added default local config file and `--config` flag to override suite configuration at runtime
- Saved runtime configuration snapshots to ensure accurate evaluation and result collection


CONTEXT.md
```diff
@@ -124,6 +124,7 @@ pnpm link --global && gd setup-completion
 |---|---|
 | `gd eval` | Run the full evaluation suite (discovers all tasks in `harness/tasks/`) |
 | `gd eval [task1] [task2]` | Run specific tasks only |
+| `gd eval [options]` | Run with config overrides (`--agent`, `--numRuns`, `--negative`, `--enableSkills`, `--mcpServers`) |
```
paulirish: @formgit do you have any thoughts on the usability of this as a CLI? 

also.. we have defaults for all these things right? (what are they?)

formgit: ahh, I see that the description is not detailed enough for the CLI. The default values (from config.ts) are also not listed.

Although @micahjo7 suggested we should support a user custom runtime config file instead of CLI args which is a easier way to set configs without needing to change/upload config.ts accidentally. So I'm thinking about using a json (an alternative is to use .ts format to enforce enum selection but need additional imports) config file as opposed to CLI flags. WDYT @paulirish

harness/config.ts
```diff
@@ -40,16 +40,24 @@ export const environmentConfig: EnvironmentConfig = {
   mcpApiKey: process.env.MCP_API_KEY || '', // For google-developer-knowledge MCP server
 };
 
-export const suiteConfig: SuiteConfig = {
+export const defaultSuiteConfig: SuiteConfig = {
   name: `full-${new Date().toLocaleString('sv-SE', { timeZone: 'America/Los_Angeles' }).replace(' ', 'T').replace(/:/g, '-')}`,
   numRuns: 2,
   tasks: [], // Empty = discover all tasks in harness/tasks/. Set explicitly to run a subset.
   mcpServersToEnable: ['modern-web'], // Available servers: 'modern-web', 'google-developer-knowledge'
```
micahjo7: can you remove this entire comment ("// Available servers: ...")

also, in the runtime config, you don't need to include `mcpServersToEnable`. This can be purely configured by the new `serving` config, and the MCP config is not updated if the `serving` is not set to `MCP`: https://github.com/GoogleChrome/guidance/pull/400/changes#diff-6aa3f038a7fd47602798336fd4ba6f51104e8f2169bcdeda5ab01e12437897e0R33

CONTEXT.md
formgit: End up choosing the `.ts` format as I feel like I need to provide an example config anyway.

harness/config.ts
formgit: Done

harness/lib/collection.ts
micahjo7: nit: `config` not used

bin/gd.ts
```diff
@@ -151,6 +152,7 @@ ${cBold('Options:')}
   ${cDim('-h, --help')}             Show this help
   ${cDim('--verbose')}              Show additional output
   ${cDim('--usecases')}             (Audit) Group by categories/usecases (default is features)
+  ${cDim('--config <path>')}         (Eval) Path to a custom TS suite config file (defaults to local_config_override.ts, or falls back to defaults in harness/config.ts)
```
micahjo7: nit: spacing

harness/agents/claude-code-agent.ts
```diff
@@ -2,8 +2,10 @@ import fs from 'fs';
 import path from 'path';
 import { spawn } from 'child_process';
 import { fileURLToPath } from 'url';
-import { createIsolatedHome, cleanupIsolatedHome, parseAgentArgs, copyFileIfExists, updateMcpConfig, copyResultsToTarget, createWorkDir, copySkills, watchLogFile } from '../lib/agent-shared.ts';
+import { initSuiteConfig, createIsolatedHome, cleanupIsolatedHome, parseAgentArgs, copyFileIfExists, updateMcpConfig, copyResultsToTarget, createWorkDir, copySkills, watchLogFile } from '../lib/agent-shared.ts';
 import config, { Agents, Serving } from '../config.ts';
+
+initSuiteConfig();
```
micahjo7: does this need to be here or should we put it in the run() method?

harness/lib/collection.ts
```diff
@@ -69,7 +69,7 @@ export function extractModelFromResults(resultsDir: string): string {
   return 'unknown';
 }
 
-export async function collectResults(resultsDir: string) {
+export async function collectResults(resultsDir: string, suiteConfig: any) {
```
micahjo7: nit: set explicit type?

harness/config.ts
```diff
@@ -48,16 +48,24 @@ export const environmentConfig: EnvironmentConfig = {
   mcpApiKey: process.env.MCP_API_KEY || '', // For google-developer-knowledge MCP server
 };
 
-export const suiteConfig: SuiteConfig = {
+export const defaultSuiteConfig: SuiteConfig = {
   name: `full-${new Date().toLocaleString('sv-SE', { timeZone: 'America/Los_Angeles' }).replace(' ', 'T').replace(/:/g, '-')}`,
   numRuns: 1,
   tasks: [], // Empty = discover all tasks in harness/tasks/. Set explicitly to run a subset.
-  mcpServersToEnable: [], // Available servers: 'modern-web', 'google-developer-knowledge'
+  mcpServersToEnable: [],
```
micahjo7: initialize with `'modern-web'` as default

harness/evaluate.ts
```diff
@@ -21,15 +21,27 @@ export async function evaluateSuite(resultsDir: string, suiteName: string) {
     return;
   }
 
+  let suiteConfig = config.suite;
```
micahjo7: it might be better to remove `config` from this file and solely rely on the `configPath`; it should always be present, right?

this way we can prevent accidentally using `config` in future edits to this file

harness/agents/claude-code-agent.ts
micahjo7: in case we need it in global scope (might be safer), can we put it after the imports?

bin/gd.ts
micahjo7, line 252: this block I think will still use the default config

harness/lib/agent-shared.ts
```diff
@@ -2,11 +2,24 @@ import fs from 'fs';
...
-import { Agents } from '../config.ts';
+import { Agents, mergeSuiteConfig } from '../config.ts';
 import { classifyGuide, scanAllGuides } from './utils.ts';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
+/**
+ * Initializes the suite configuration from environment variables.
+ */
+export function initSuiteConfig(): void {
```
micahjo7: in `run_suite` we merge the configs and save to `process.env.GD_SUITE_CONFIG`. So in here, we don't have to merge again, right? We can just extract the JSON into ServingConfig object?

harness/config.ts
```diff
@@ -48,16 +48,24 @@ export const environmentConfig: EnvironmentConfig = {
...
   agent: Agents.GEMINI_CLI,
   negative: false, // When `true`, runs the suite on all tasks in `tasks/negative/`
 };
 
+export let suiteConfig: SuiteConfig = { ...defaultSuiteConfig };
+
+export function mergeSuiteConfig(overrides: Partial<SuiteConfig>): SuiteConfig {
+  suiteConfig = { ...suiteConfig, ...overrides };
+  config.suite = suiteConfig;
+  return suiteConfig;
```
micahjo7: it might be better to just return the `suiteConfig` here, and not reset the `config.suite`?

and then in `run_suite` we use the returned value

and everywhere else, we should not use the config value directly, but extract from the `suite_config.json`?

harness/run_suite.ts
micahjo7: this method needs to be updated too, right?

needs to merge config, and pass the new suite config file in runCommand

harness/run_suite.ts
```diff
@@ -190,7 +199,7 @@ const args = [
     templateDir
   ])}
 ];
-const result = spawnSync('node', args, { stdio: 'inherit', cwd: ${JSON.stringify(process.cwd())} });
+const result = spawnSync('node', args, { stdio: 'inherit', cwd: ${JSON.stringify(process.cwd())}, env: { ...process.env, GD_SUITE_CONFIG: ${JSON.stringify(JSON.stringify(config.suite))} } });
```
micahjo7: I think we only need to set the GD_SUITE_CONFIG env variable and not rest of env

harness/run_suite.ts
```diff
@@ -314,23 +323,24 @@ function restoreLogging(originals: any) {
   }
 }
 
-async function runCommand(command: string, args: string[] = [], cwd?: string) {
+async function runCommand(command: string, args: string[] = [], cwd?: string, envOverrides?: Record<string, string>) {
```
micahjo7: we can update to take the suite config file and set GD_SUITE_CONFIG env variable only in the impl

micahjo7 (changes requested):

README.md
```diff
@@ -85,7 +86,21 @@ All configuration is centralized in [`harness/config.ts`](./harness/config.ts).
...
 -   **Suite**: Agent selection, number of runs, tasks to run, enabled MCP servers, and skills.
 
-All settings must be adjusted in `harness/config.ts` or via environment variables in `.env` at the `guidance/` root.
+### Runtime Configuration Overrides
+
+You can override suite configurations without modifying `harness/config.ts` directly. The `gd eval` command automatically looks for a `local_config_override.ts` file in the project root. If this file doesn't exist and no `--config` flag is provided, it safely falls back to the defaults in `harness/config.ts`.
+
+To get started, copy the template:
+```bash
+cp example_config_override.ts local_config_override.ts
```
paulirish: filename bikeshedding

how about

    cp config.ts.example config.ts

would love to avoid extra terms like 'override' and 'local' and 'custom'  if we dont need them. :)

harness/lib/collection.ts
formgit: Removed

bin/gd.ts
formgit: Fixed

harness/agents/claude-code-agent.ts
formgit: initSuiteConfig() is now removed in favor of the getSuiteConfig() function, which parse the config on demand from the saved json config file.

harness/lib/collection.ts
formgit: Done

harness/config.ts
formgit: Done

harness/evaluate.ts
formgit: Done

bin/gd.ts
formgit: fixed

harness/lib/agent-shared.ts
formgit: Done. Now the the config is read from the saved json file specified by GD_SUITE_CONFIG

harness/config.ts
formgit: Done. Now the merged result is directly returned and config.suite won't be overwritten.

harness/run_suite.ts
formgit: sure, added providedSuiteConfig explicitly here. The configs are written to suite_config.json and its path is passed on.

harness/run_suite.ts
formgit: Here the new envOverrides env (contains the saved json file path) is the only new addition to the process.env. IIUC, before this change, anything under process.env is passed on inexplicit. So I think this approach is not breaking existing behavior and not introducing unnecessary risk.

harness/run_suite.ts
formgit: done

README.md
formgit: done

harness/config.ts
```diff
@@ -87,7 +91,7 @@ export interface SuiteConfig {
 
 export const config = {
   environment: environmentConfig,
-  suite: suiteConfig,
+  suite: defaultSuiteConfig,
```
micahjo7: we can remove `suite` altogether to prevent using this on accident somewhere?

README.md
```diff
@@ -71,6 +71,7 @@ gd dev [dir] [options]        # auto-generate/calibrate
 # Evaluation
 gd eval                       # run the full evaluation suite
 gd eval [task1] [task2]       # run specific tasks
+gd eval [config]              # run with config overrides (e.g. --config my_custom_config.ts, defaults to config.ts, or falls back to defaults in harness/config.ts)
```
micahjo7: `gd eval --config <custom_config>`

harness/run_suite.ts
```diff
@@ -2,7 +2,8 @@ import fs from 'fs';
 import path from 'path';
 import { fileURLToPath } from 'url';
 import { spawn } from 'child_process';
-import { config, Agents } from './config.ts';
+import { Agents, defaultSuiteConfig } from './config.ts';
+import type { SuiteConfig } from './config.ts';
```
micahjo7: combine imports

bin/gd.ts
```diff
@@ -205,8 +208,26 @@ ${cBold('Options:')}
...
+      if (fs.existsSync(resolvedConfigPath)) {
+        const fileUrl = pathToFileURL(resolvedConfigPath).href;
+        const customConfig = await import(fileUrl);
+        overrides = customConfig.default || customConfig;
+      } else if (configPath) {
+        console.error(cRed('⚠️ Specified config file not found: ' + resolvedConfigPath));
+        process.exit(1);
+      }
+
+      const mergedSuiteConfig = mergeSuiteConfig(overrides);
```
micahjo7: can we move this to a helper (212-227)

harness/run_suite.ts
micahjo7: I think we can remove this

the `if (positionalArgs.length === 2 && args.includes('--with-template'))` block

harness/run_suite.ts
```diff
@@ -190,9 +201,9 @@ const args = [
     templateDir
   ])}
 ];
-const result = spawnSync('node', args, { stdio: 'inherit', cwd: ${JSON.stringify(process.cwd())} });
```
micahjo7: why remove this?

harness/run_suite.ts
```diff
@@ -52,14 +57,15 @@ export async function runAgent(templateDirRaw: string, promptContentRaw: string)
...
+    const suiteConfigPath = path.resolve(targetDir, 'suite_config.json');
     await runCommand('node', [
       '--experimental-strip-types', 
       agentScript, 
       JSON.stringify(promptContent), 
       'guided', // Default to guided for ad-hoc tool execution
       targetDir,
       templateDir
-    ]);
+    ], undefined, { GD_SUITE_CONFIG: suiteConfigPath });
```
micahjo7: maybe we can switch order of the last two args so we don't need to put `undefined` here

guides/README.md
```diff
@@ -140,10 +140,11 @@ To verify that guides improve agent performance starting from a "bad" implementa
    - A task in `harness/tasks/negative/<guideName>-task-negative.md` pointing to that base app.
 
 2. **Run the negative suite**:
-   Update `suiteConfig.negative = true` in [`harness/config.ts`](../harness/config.ts) and run:
+   Ensure your local configuration (`config.ts`) has `negative: true`, then run:
```
micahjo7: maybe specify `guidance/config.ts`

harness/run_suite.ts
micahjo7: so we don't need to pass the env overrides here?

formgit: No, it's not required at this point as it's already running with variables in its environment.

harness/config.ts
formgit: Removed, it was used by the codex cli agent after rebase. Now the codex cli agent is using the `getSuiteConfig()` to read config from the config file just like other agents.

README.md
formgit: Done

harness/run_suite.ts
formgit: done

bin/gd.ts
formgit: done

harness/run_suite.ts
formgit: done

harness/run_suite.ts
formgit: It was accidentally removed when addressing a previous comment. Now it's restored.

harness/run_suite.ts
formgit: done

guides/README.md
formgit: done

micahjo7 (approved):
Thank you Leiming!! The changes look good, sorry for all the intermediate changes. If https://github.com/GoogleChrome/guidance/pull/445 ends up being checked in before this, there will be one more place to update the config obj (`guides/dev-guide.ts`).

formgit:
> Thank you Leiming!! The changes look good, sorry for all the intermediate changes. If #445 ends up being checked in before this, there will be one more place to update the config obj (`guides/dev-guide.ts`).

Thanks for your review! I have just resolved merge conflicts and updated guides/dev-guide.ts`

Issue #148, opened 2026-02-25
Create guide and evals for the search-hidden-content use case


paulirish:
Hide content from view using patterns such as accordions, tabs, and "Read more" sections, while ensuring the hidden text reveals itself during native "Find in page" searches, allows search engine indexing, supports URL fragment deep links, and maintains ARIA accessibility.

Affected web-feature IDs: [details](https://webstatus.dev/features/details), [details-name](https://webstatus.dev/features/details-name), [hidden-until-found](https://webstatus.dev/features/hidden-until-found)

Use case subdir: [guides/ui-behaviors/search-hidden-content](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-behaviors/search-hidden-content)

Related features: #31

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

PR #355, opened 2026-03-19
Fix evals for search-hidden-content use case

taraojo:
Fixing issues in #148. Currently there is no difference between guided and unguided results. Clarify `hidden="until-found"` expectations and consolidate search-hidden-content prompts and tasks to ensure each prompt generates examples for both `<details>` and `hidden=until-found`. This means that any test that relies on one state or the other will always have an example to check.


rviscomi:
This is obsoleted by https://github.com/GoogleChrome/guidance/pull/387

Issue #53, opened 2026-02-19
Event timing


rviscomi:
The event and first-input performance entries and the PerformanceEventTiming API measures the latency of user input events, such as mouse clicks or keypresses. They're used to calculate Interaction to Next Paint (INP), a common metric for perceived responsiveness.

### MDN Documentation
- [PerformanceEventTiming](https://developer.mozilla.org/docs/Web/API/PerformanceEventTiming)

### Chrome Documentation

<details>
<summary>View 95 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/aligning-input-events
- https://developer.chrome.com/en/blog/aurora-update-2023
- https://developer.chrome.com/en/blog/chrome-2024-recap
- https://developer.chrome.com/en/blog/chrome-at-io23
- https://developer.chrome.com/en/blog/devtools-grounded-real-world
- https://developer.chrome.com/en/blog/devtools-insights-sidebar
- https://developer.chrome.com/en/blog/devtools-realtime-cwv
- https://developer.chrome.com/en/blog/first-input-delay-in-crux
- https://developer.chrome.com/en/blog/inp-tools-2022
- https://developer.chrome.com/en/blog/inside-browser-part4
- https://developer.chrome.com/en/blog/introducing-scheduler-yield-origin-trial
- https://developer.chrome.com/en/blog/lighthouse-11-0
- https://developer.chrome.com/en/blog/lighthouse-6-0
- https://developer.chrome.com/en/blog/lighthouse3
- https://developer.chrome.com/en/blog/new-soft-navigations-origin-trial
- https://developer.chrome.com/en/blog/next-third-parties
- https://developer.chrome.com/en/docs/aurora/inp-in-frameworks
- https://developer.chrome.com/en/docs/crux/dashboard
- https://developer.chrome.com/en/docs/crux/guides/bigquery
- https://developer.chrome.com/en/docs/crux/guides/looker-studio-dashboard
- https://developer.chrome.com/en/docs/crux/guides/pagespeed-insights
- https://developer.chrome.com/en/docs/crux/methodology/metrics
- https://developer.chrome.com/en/docs/crux/vis
- https://developer.chrome.com/en/docs/devtools/performance/overview
- https://developer.chrome.com/en/docs/devtools/performance/reference
- https://developer.chrome.com/en/docs/devtools/performance/timeline-reference
[...]

PR #356, opened 2026-03-20
event-timing: add use cases

paulirish:
Progress on https://github.com/GoogleChrome/guidance/issues/53


guides/sync-use-cases.ts
```diff
@@ -633,10 +633,10 @@ async function processUseCases(
     let guideBody = '';
 
     if (hasGuide || inv.isStub) {
-      const { errors, data, body } = validateGuide(path.join(subdir, 'guide.md'));
-      guideErrors = errors;
-      guideData = data;
-      guideBody = body;
+      const validation = validateGuide(path.join(subdir, 'guide.md'));
```
paulirish, line 677: without this i fell into some infinite error loop when running  `pnpm --filter guides test`  when i had invalid web-feature-ids

guides/performance/first-input-delay/guide.md
```diff
@@ -0,0 +1,10 @@
+---
+name: first-input-delay
+description: Measure the First Input Delay (FID) using the Event Timing API to assess the delay between the first user interaction and when the browser begins processing its event handlers.
```
philipwalton: Do we want to include a use case for a metric we don't recommend people track anymore? My sense is we can omit this one.

guides/performance/interaction-latency-attribution/guide.md
```diff
@@ -0,0 +1,11 @@
+---
+name: interaction-latency-attribution
+description: Decompose interaction latency into input, processing, and presentation phases to diagnose responsiveness bottlenecks.
```
philipwalton:
```suggestion
description: Identify slow user interactions, and attribute them for diagnostic purposes (e.g. determining the target element of an interaction or decomposing its phases into input delay, processing duration, and presentation delay).
```

I think it makes sense to generalize this to cover more attribution use cases (in addition to the phase breakdown).

guides/performance/interaction-to-next-paint/guide.md
```diff
@@ -0,0 +1,10 @@
+---
+name: interaction-to-next-paint
+description: Monitor session-wide interaction latency to identify and optimize the Interaction to Next Paint (INP) score.
```
philipwalton: For this one, is our recommendation for the AI to use event timing, or is it to use the `web-vitals` library?

I'm concerned that if we include use cases where our actual recommendation is to use `web-vitals` then we may end up with a bunch of sites that poorly track Core Web Vitals metrics. WDYT?

philipwalton (changes requested):
These are good use cases, but as I mentioned in one of the inline comments, I wonder if we want to stick to use cases that are not already well-covered by the `web-vitals` JS library.

IMO custom attribution is the main reasons someone would use event timing in addition to (or instead of) the `web-vitals` library.

guides/sync-use-cases.ts
rviscomi: Ah good catch. I included this change in #358 and added tests to make sure it doesn't happen again.

guides/performance/first-input-delay/guide.md
paulirish: okay will drop.

guides/performance/interaction-latency-attribution/guide.md
paulirish: sounds good. so i'll flesh out this usecase.. and feels like i'll delete the others.

guides/performance/interaction-to-next-paint/guide.md
paulirish: sure. turning event-timing into INP is nontrivial, thats for sure. 

so IMO we should just delete this usecase, no? if it mostly recommends the web-vitals lib then it doesnt map to a feature-id anymore.  we leave that to a more general improve-cwv usecase or performance skill.

philipwalton: Yeah, I think we can drop it, and I agree that this subject can easily be covered in a performance skill.

paulirish:
@philipwalton updated if you're still around

Issue #142, opened 2026-02-25
Create guide and evals for the consistent-cross-document-transitions use case


paulirish:
Ensure critical page state is loaded and stable before initiating a cross-document view transition. This means critical CSS styles are loaded and applied, critical JavaScript is loaded and run, and the HTML visible for the user's initial view of the page has been parsed before the transition runs.

Affected web-feature IDs: [blocking-render](https://webstatus.dev/features/blocking-render), [link-rel-expect](https://webstatus.dev/features/link-rel-expect), [cross-document-view-transitions](https://webstatus.dev/features/cross-document-view-transitions)

Use case subdir: [guides/ui-behaviors/consistent-cross-document-transitions](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-behaviors/consistent-cross-document-transitions)

Related features: #40 #50

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

PR #357, opened 2026-03-20
Create guidance for the blocker=render use cases

philipwalton:
Addresses #142 #343


guides/user-experience/consistent-cross-document-transitions/guide.md
```diff
@@ -3,4 +3,230 @@ name: consistent-cross-document-transitions
...
+
+## The Solution
+
+Use `blocking="render"` on critical `<link>` and `<script>` elements in the new page's `<head>`, and use `<link rel="expect">` to block rendering until specific DOM elements have been parsed. This ensures the browser does not begin the view transition animation until the new page's visual state is stable. The browser continues parsing the HTML in the background — only painting is deferred.
+
+### Implementation Strategy
+
+1. **MANDATORY:** Opt in to cross-document view transitions with the `@view-transition` CSS at-rule on both pages.
+2. **MANDATORY:** Ensure critical stylesheets are in the `<head>`. Stylesheets in the `<head>` are render-blocking by default. Dynamically injected stylesheets require explicit `blocking="render"`.
+3. **DO** use `blocking="render"` on `<script>` elements that must execute before the transition animates (e.g., scripts that apply a theme or effect the layout).
```
rviscomi:
```suggestion
3. **DO** use `blocking="render"` on `<script>` elements that must execute before the transition animates (e.g., scripts that apply a theme or affect the layout).
```

guides/user-experience/consistent-cross-document-transitions/guide.md
```diff
@@ -3,4 +3,230 @@ name: consistent-cross-document-transitions
...
+  (link clicks, form submissions, back/forward).
+*/
+@view-transition {
+  navigation: auto;
+}
+```
+
+### Step 2: Block Rendering Until Critical Scripts Execute
+
+If a non-blocking script in the `<head>` must run before the transition animates (e.g., to apply a theme class or effect the layout), mark it with `blocking="render"`. Without this, `async`, `defer`, or `type="module"` scripts may execute after the transition has already started.
```
rviscomi:
```suggestion
If a non-blocking script in the `<head>` must run before the transition animates (e.g., to apply a theme class or affect the layout), mark it with `blocking="render"`. Without this, `async`, `defer`, or `type="module"` scripts may execute after the transition has already started.
```

guides/user-experience/consistent-cross-document-transitions/expectations.md
```diff
@@ -0,0 +1,9 @@
+- Both the source and destination pages include the `@view-transition { navigation: auto; }` CSS at-rule.
+- `<link rel="expect" href="#element-id" blocking="render">` is used in the `<head>` to block rendering until above-the-fold DOM elements have been parsed, ensuring the transition snapshot captures a visually complete page.
+- Scripts that must run before the transition (e.g., theme application) are marked with `blocking="render"` in the `<head>`.
+- The `media` attribute is used on `<link rel="expect">` when different viewport sizes require blocking on different DOM elements.
+- Render blocking is limited to resources and elements visible in the initial viewport. Non-critical or below-the-fold content is NOT render-blocked.
+- If `view-transition-name` values are assigned dynamically via `pagereveal`, the listener is registered in a parser-blocking script or a `blocking="render"` script in the `<head>`.
```
rviscomi: To confirm, is it ok to block parsing? The guide only recommends setting up the listener in a non-parser-blocking script (`<script async blocking=render>`)

guides/user-experience/consistent-cross-document-transitions/expectations.md
```diff
@@ -0,0 +1,9 @@
+- Both the source and destination pages include the `@view-transition { navigation: auto; }` CSS at-rule.
```
rviscomi, line 1: @micahjo7 this use case requires eval'ing multiple files, which isn't currently supported. Do you have any ideas how we might do that?

guides/user-experience/flicker-free-client-side-ab-testing/guide.md
```diff
@@ -3,4 +3,175 @@ name: flicker-free-client-side-ab-testing
...
+
+## The Solution
+
+The `blocking=render` attribute allows a `<script>` or `<link>` element placed in the `<head>` to block rendering—but not parsing—until the resource has been fetched and executed. This gives experimentation scripts the same render-blocking behavior that stylesheets have by default, ensuring the browser never paints the page until the experiment variant has been applied. No opacity hacks, no arbitrary timeouts, and no flicker.
+
+### Implementation Strategy
+
+1. **MANDATORY:** Place the experimentation script in the document `<head>`.
+2. **MANDATORY:** Add the `blocking="render"` attribute alongside `async` on the script tag.
```
rviscomi: This might confuse agents when implementing the inline script approach, where the solution is to use `type=module` rather than `async`

rviscomi:
Also note there are a couple of linter issues:

```
  ❌ Error in guides/user-experience/consistent-cross-document-transitions: Missing grader.ts. Must have BOTH grader.ts and prompts.md.
  ❌ Error in guides/user-experience/flicker-free-client-side-ab-testing: Missing grader.ts. Must have BOTH grader.ts and prompts.md.
```

It tries to avoid only having some eval files but not all

guides/user-experience/consistent-cross-document-transitions/expectations.md
philipwalton: Good point. It's technically OK for this use case (and AFAICT it was taken from [this bit](https://developer.chrome.com/docs/web-platform/view-transitions/cross-document#:~:text=Important%3A%20The,%3Cscript%3E%20tag.) in Bramus's doc), but it's probably not something we want to recommend given that `blocking=render` works.

guides/user-experience/flicker-free-client-side-ab-testing/guide.md
philipwalton: Potentially. The thing is we can't really recommend using <script type=module> here because not all experiment scripts can run as modules. Also with these types of scripts you typically want the `async` behavior rather than the `defer` behavior because they're independent of all other scripts and you want them to load ASAP.

rviscomi: Maybe relax the "MANDATORY" bit, or add a condition like "... if the script isn't inline"

guides/user-experience/consistent-cross-document-transitions/expectations.md
micahjo7: hmm I haven't given it too much thought (this is kinda hard to support)

we could have the grader file take multiple target files with additional env variables. grader-gen/grader template would have to updated, along with the infrastructure in `dev-guide.ts` and `collection.ts` (which are the two places where the graders run)

we would also need to enforce sync between the prompts and the graders, to make sure the agent creates specific files that the grader expects

philipwalton: For the MVP ahead of I/O, I think there will be relatively few of these cases. It might be easiest to manually create the tests if they have complicated requirements like this.

But long term I think we will need a solution. For example, there are folks wanting to write WASM use cases that will almost certainly require advanced test setup to properly validate.

guides/user-experience/flicker-free-client-side-ab-testing/guide.md
philipwalton: Addressed in https://github.com/GoogleChrome/guidance/pull/357/commits/c29fd5f44e0e53c4e25494169a299092cd97ca0c

guides/user-experience/consistent-cross-document-transitions/expectations.md
philipwalton: Addressed in https://github.com/GoogleChrome/guidance/pull/357/commits/c29fd5f44e0e53c4e25494169a299092cd97ca0c

philipwalton:
> Also note there are a couple of linter issues:
> 
> ```
>   ❌ Error in guides/user-experience/consistent-cross-document-transitions: Missing grader.ts. Must have BOTH grader.ts and prompts.md.
>   ❌ Error in guides/user-experience/flicker-free-client-side-ab-testing: Missing grader.ts. Must have BOTH grader.ts and prompts.md.
> ```
> 
> It tries to avoid only having some eval files but not all

Added in https://github.com/GoogleChrome/guidance/pull/357/commits/be0fe2614b2a4bd786a51d1f737065a1999ebec2

rviscomi:
I ran both use cases through `gd dev` locally to evaluate the guidance pass rates:

| Use Case | Unguided Pass Rate | Guided Pass Rate | Guide Impact |
| :--- | :--- | :--- | :--- |
| `consistent-cross-document-transitions` | 63% (5/8) | 63% (5/8) | +0% |
| `flicker-free-client-side-ab-testing` | 10% (1/10) | 20% (2/10) | +10% |

Are you seeing something similar?

guides/user-experience/consistent-cross-document-transitions/expectations.md
micahjo7: filed b/496651090

rviscomi:
Guided pass rates are now 90% and 75% for `flicker-free-client-side-ab-testing` and `consistent-cross-document-transitions` respectively.

The second one isn't performing as well because of a limitation with the eval infrastructure (b/496651090) but the guidance is sound.

PR #360, opened 2026-03-20
Automate use case research

rviscomi:
image


bin/gd.ts
```diff
@@ -263,6 +269,18 @@ ${cBold('Options:')}
       break;
     }
 
+    case 'research': {
+      // Pass remaining args through to the research script
+      const researchArgs = process.argv.slice(3);
+      await spawnChild('node', [
```
paulirish: would prefer the 

```
     const { generateGrader } = await import('../guides/grader-gen.ts');
        await generateGrader(dir);
``` 

style instead of spawning a new node process.  if that's compatible with your stuff.

guides/research-use-cases.ts
```diff
@@ -0,0 +1,615 @@
...
+
+// ---------------------------------------------------------------------------
+// Resolve inputs (from --issue or explicit flags)
+// ---------------------------------------------------------------------------
+
+let featureId: string;
+let seedSources: string[];
+
+if (values.issue) {
+  const issueData = fetchIssue(values.issue);
```
paulirish: this'll happen via the spawn-> import fix but lets put this stuff in a function

guides/research-use-cases.ts
```diff
@@ -0,0 +1,615 @@
...
+  });
+
+  return { category: obj.category, useCases };
+}
+
+// ---------------------------------------------------------------------------
+// Guide stub creation
+// ---------------------------------------------------------------------------
+
+function collectExistingDescriptions(): string[] {
```
paulirish, line 466: feels like we might have overlap between this and some of our util stuff in sync-use-cases etc. (but didnt verify) would love to DRY up that

guides/research-use-cases.ts
paulirish: oh. it can go into main()

guides/research-use-cases.ts
```diff
@@ -0,0 +1,615 @@
...
+  }
+
+  return { featureId, sources };
+}
+
+function fetchIssue(issueNumber: string): IssueData {
+  console.log(`Fetching issue #${issueNumber} from GitHub…`);
+  let json: string;
+  try {
+    json = execSync(`gh issue view ${issueNumber} --json title,body`, { encoding: 'utf8' });
```
paulirish, line 54: just fyi my event-timing issue has 95 articles. and many are not relevant  https://github.com/GoogleChrome/guidance/issues/53

but.. i expect that to be pretty exclusive to event-timing which has a very general name.

bin/gd.ts
rviscomi: Done

guides/research-use-cases.ts
rviscomi: Done

guides/research-use-cases.ts
rviscomi: Pulled in scanAllGuides, LMK if you had something else in mind

guides/research-use-cases.ts
rviscomi: Yeah, what I hoped was for people to manually prune irrelevant articles from their GH issue descriptions, but that's ok. I have a script that'll scoop them up and saves them for other features-to-sources uses in the future.

.agents/skills/project-use-cases/SKILL.md
```diff
@@ -11,42 +11,60 @@ The primary goal of this stage is to translate a technical web platform feature
...
+gd research \
+  --feature-id bfcache-blocking-reasons \
+  --sources https://web.dev/articles/bfcache \
+            https://github.com/rakina/bfcache-explainer
 ```
 
-This seeds the model with known good resources and asks it to perform deep research to surface additional technical details, constraints, and discussions.
+The command will:
+1. Search the web using Google Search grounding to find additional authoritative sources (W3C specs, MDN, developer blogs, WICG proposals, GitHub discussions)
+2. Propose 2–5 use cases in the correct format
```
paulirish, line 37: i think the CLI should check in with the human to confirm the URL sources list. (both me and phil had problems with the source list in our features. :) 

i also like the idea of the human getting a chance to add more urls to the list

so like... 

"we found these obvious ones (mdn, explainer, webdev canonical).  we also found these (our blog posts). do you have any you'd like to add?  any to remove?"

guides/research-use-cases.ts
```diff
@@ -0,0 +1,724 @@
...
+    console.log(`  Interaction ID: ${id} — polling for completion…`);
+  }
+
+  // Poll until the interaction completes (max 60 minutes).
+  const POLL_INTERVAL_MS = 15_000;
+  const MAX_WAIT_MS = 60 * 60 * 1000;
+  const start = Date.now();
+  while (Date.now() - start < MAX_WAIT_MS) {
+    await new Promise((r) => setTimeout(r, POLL_INTERVAL_MS));
+
```
paulirish, line 199: i'd like a log here to clarify we're making a request.  

you could do process.stderr.write('poll ⏳. ')  to not do a newline and keep it tiny
or something

guides/research-use-cases.ts
```diff
@@ -0,0 +1,724 @@
...
+  featureId: string,
+  info: ReturnType<typeof getFeatureInfo>,
+  seedUrls: string[],
+  issueContext?: { title: string; body: string }
+): string {
+  const lines: string[] = [];
+
+  if (issueContext) {
+    lines.push(
+      'Thoroughly research the web platform feature described in the following GitHub issue.',
```
paulirish, line 252: is 'github issue' relevant?

guides/research-use-cases.ts
```diff
@@ -0,0 +1,724 @@
...
+      '',
+      '## Issue title',
```
paulirish, line 254: same for these headers.

guides/research-use-cases.ts
```diff
@@ -0,0 +1,724 @@
...
+      lines.push('- **Spec URL(s)**:');
+      for (const s of info!.spec) lines.push(`  - ${s}`);
+    }
+  }
+
+  lines.push(
+    '',
+    '## Task',
+    'Use the above as a starting point and search for additional high-quality information',
+    'from MDN, developer blogs, Chrome developer documentation, WICG proposals, and GitHub',
```
paulirish, line 280: Maybe something like 

expert discussion
novel techniques


like.. i want web-vitals-discuss to come up for some webperf issues. or httparchive discuss. 
and personally i really want all the crazy clever techniques from folks like https://kizu.dev/fit-to-width/ or https://codepen.io/thebabydino, etc.

not sure exactly how to generalize that but.. ideally i'd have the corpus stretched to wherever they are.  :)

paulirish: okay well tbh the latter part is perhaps less relevant to usecases. but... i don't know that i'd want different sources that are used for usecase generation vs guide generation

paulirish (approved):
love this work

rviscomi:
@paulirish I'm exploring a different approach in https://github.com/GoogleChrome/guidance/pull/429 for researching and generating discipline-level skills. Instead of a CLI, I put all of the logic in [.agents/skills/project-skills-research/SKILL.md](https://github.com/GoogleChrome/guidance/blob/skill-research/.agents/skills/project-skills-research/SKILL.md) with a couple of tiny helper scripts, so the contributor's own coding agent does most of the heavy lifting. One of the scripts calls out to Gemini's Deep Research model because I still think that's an irreplaceable part of the pipeline, but otherwise it's all done by the local agent.

I generated three skill drafts using this process and I'm pretty happy with the results. So I'm tempted to revert most of this PR and convert it from CLI to Skill format, but I wanted to see if you had any strong preferences for the CLI approach before I went through with it.

PR #369, opened 2026-03-21
Avoid static assertions in grader.ts

rviscomi:
I've found static assertions to be an antipattern. They're typically implemented as regexes, which are much more brittle than querying the DOM, getComputedStyle, or similar.

Open to suggestions if there's a better way around it.


rviscomi:
Ping @paulirish @micahjo7 (bit by this again today)

.agents/skills/project-evals/SKILL.md
```diff
@@ -17,7 +17,10 @@ This is the third of three stages in creating guidance:
   * Avoid over-constraining implementation details that don't matter for correctness (e.g., don't assert that an element must be a direct child if a descendant also works).
 
 ## Grading Note
-* Graders (`grader.ts`) live within their respective guide folders. These are Playwright test files, but they are permitted to perform non-browser tests (like `str.includes()` on file contents) as well as actual browser automation checks. A huamn may manually edit them if the generator struggles to get it perfectly tailored.
+* Graders (`grader.ts`) live within their respective guide folders. These are Playwright test files.
+* **AVOID** using static assertions (like regex or `str.includes()` on `fs.readFileSync`) to test CSS or HTML syntax whenever possible. These are extremely brittle and will fail if the agent uses a different class name, semantic element, or formatting.
```
micahjo7, line 51: should this type of info go in grader-gen [prompt](https://github.com/GoogleChrome/guidance/blob/main/guides/grader-gen.ts#L17) instead? this SKILL is just instructing the agent to call `gd dev` which runs the grader-gen, right?

rviscomi: Ah yes I think it should go in the prompt too (I missed that) but I think we should also keep it here in the skill. Reason being that we want contributors' coding agents to be able to keep generated graders in check as we go, or to retroactively apply grader best practices to previously generated files.

I don't mind the duplication so much, but if needed we could find a way to centralize the instructions. (I don't love it but we could scrape the skill md and inject it into the prompt)

paulirish: definitely agree that its more important to have in the grader prompt. 

but if you want it here as well, sure.  i'm fine with the duplication. (less fine with scrape+inject :)

paulirish:
> They're typically implemented as regexes, which are much more brittle than querying the DOM, getComputedStyle, or similar.

agree the regexes are bad. i want some proper parsers (like html5lib, a JS parser->ast ish)  that would make this a little bit easier. in the meantime i'm fine to add your guidance but i may revert it when i introduce proper parsers .

.agents/skills/project-evals/SKILL.md
rviscomi: Updated

paulirish:
> > They're typically implemented as regexes, which are much more brittle than querying the DOM, getComputedStyle, or similar.

> agree the regexes are bad. i want some proper parsers (like html5lib, a JS parser->ast ish) that would make this a little bit easier. in the meantime i'm fine to add your guidance but i may revert it when i introduce proper parsers .

well... of course the distance between thought and code is very short now! 🤣 

https://github.com/GoogleChrome/guidance/pull/422  <= wdyt about this.?

rviscomi:
> #422 <= wdyt about this.?

I like it! I'll push this through for now and we can continue exploring the parser option in your PR.

Issue #64, opened 2026-02-19
Long animation frames


rviscomi:
The long-animation-frame performance event type and the PerformanceLongAnimationFrameTiming API provide information about rendering updates that take longer than 50 milliseconds. Also known as LoAFs.

### MDN Documentation
- [PerformanceLongAnimationFrameTiming](https://developer.mozilla.org/docs/Web/API/PerformanceLongAnimationFrameTiming)

### Chrome Documentation

<details>
<summary>View 33 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/devtools-insights-sidebar
- https://developer.chrome.com/en/blog/how-to-measure-browser-graphics-performance
- https://developer.chrome.com/en/blog/loaf-has-shipped
- https://developer.chrome.com/en/blog/profiling-long-paint-times-with-devtools-continuous-painting-mode
- https://developer.chrome.com/en/docs/chromium/renderingng-architecture
- https://developer.chrome.com/en/docs/devtools/layers
- https://developer.chrome.com/en/docs/devtools/performance/reference
- https://developer.chrome.com/en/docs/devtools/performance/timeline-reference
- https://developer.chrome.com/en/docs/devtools/rendering/performance
- https://developer.chrome.com/en/docs/performance/insights/dom-size
- https://developer.chrome.com/en/docs/performance/insights/forced-reflow
- https://developer.chrome.com/en/docs/web-platform/long-animation-frames
- https://web.dev/en/articles/animations-overview
- https://web.dev/en/articles/avoid-large-complex-layouts-and-layout-thrashing
- https://web.dev/en/articles/css-paint-times
- https://web.dev/en/articles/css-web-vitals
- https://web.dev/en/articles/custom-metrics
- https://web.dev/en/articles/debug-performance-in-the-field
- https://web.dev/en/articles/find-slow-interactions-in-the-field
- https://web.dev/en/articles/mobile-optimization-and-performance
- https://web.dev/en/articles/optimize-inp
- https://web.dev/en/articles/optimize-vitals-lighthouse
- https://web.dev/en/articles/rail
- https://web.dev/en/articles/reduce-the-scope-and-complexity-of-style-calculations
- https://web.dev/en/articles/rendering-performance
- https://web.dev/en/articles/smoothness
- https://web.dev/en/articles/speed-rendering
[...]

PR #373, opened 2026-03-21
Add loaf use cases and demos

philipwalton:
Addresses #64.

@tunetheweb these use cases and demo were generated based on the `project-use-cases` agent skill. Can you review and make any changes you think need to be made?

The seem mostly fine to me, my only concern is whether we want to drop the `identify-scripts-delaying-rendering` use case since it's about debugging INP, and I assume our actual recommendation there is to just use `web-vitals`.


guides/performance/correlate-interaction-with-long-frame/demo.html
```diff
@@ -0,0 +1,44 @@
...
+          document.getElementById('output').innerText = 
+            `Interaction was delayed by a frame lasting ${entry.duration.toFixed(0)}ms!`;
+        }
+      }
+    });
+
+    try {
+      observer.observe({ type: 'long-animation-frame', buffered: true });
+    } catch (e) {
+      console.warn('Long Animation Frames API not supported.');
```
tunetheweb: I don't think this works does it? An observer of an unsupported type is just ignored AFAIK?

So instead you wanna use something like this:

```js
if (PerformanceObserver.supportedEntryTypes.includes('long-animation-frame')) {
  // Monitor LoAFs
}
```

guides/performance/identify-scripts-delaying-rendering/demo.html
```diff
@@ -0,0 +1,46 @@
...
+          console.log(`URL: ${script.sourceURL}:${script.sourceCharPosition}`);
+        }
+      }
+    });
+    
+    // Start observing long-animation-frame entries
+    try {
+      observer.observe({ type: 'long-animation-frame', buffered: true });
+    } catch (e) {
+      console.warn('Long Animation Frames API not supported in this browser.');
```
tunetheweb: Ditto

guides/performance/measure-total-frame-delay/demo.html
```diff
@@ -0,0 +1,45 @@
...
+        // The Long Tasks API would not catch this, but LoAF does.
+        const totalScriptTime = entry.scripts.reduce((total, script) => total + script.duration, 0);
+        console.log(`Combined script time: ${totalScriptTime}ms`);
+      }
+    });
+    
+    try {
+      observer.observe({ type: 'long-animation-frame', buffered: true });
+    } catch (e) {
+      console.warn('Long Animation Frames API not supported.');
```
tunetheweb: And again

tunetheweb (changes requested):
> The seem mostly fine to me, my only concern is whether we want to drop the `identify-scripts-delaying-rendering` use case since it's about debugging INP, and I assume our actual recommendation there is to just use web-vitals.

I think that one is quite good, and is basically one of the ones that I have in my draft branch. It identifies all long running scripts whether it's the INP one or not. So it's basically what I advise here: https://developer.chrome.com/docs/web-platform/long-animation-frames#report-to-analytics

However I think the `correlate-interaction-with-long-frame` has the issue you're talking about and for that I'd rather advise we use web-vitals since we should be concentrating on the INP interaction. Or perhaps we can leave that and add a 4th use case specifically for INP?

Shall I merge my two use cases into this branch/PR?

guides/performance/correlate-interaction-with-long-frame/demo.html
tunetheweb: Removed in https://github.com/GoogleChrome/guidance/pull/373/commits/f9a7d8137dc400395651f066a9c2ce8a264912a0

philipwalton:
> > The seem mostly fine to me, my only concern is whether we want to drop the `identify-scripts-delaying-rendering` use case since it's about debugging INP, and I assume our actual recommendation there is to just use web-vitals.
> 
> I think that one is quite good, and is basically one of the ones that I have in my draft branch. It identifies all long running scripts whether it's the INP one or not. So it's basically what I advise here: https://developer.chrome.com/docs/web-platform/long-animation-frames#report-to-analytics

The use case description specifically mentions "to debug Interaction to Next Paint (INP)", which is why I said that. If you wanted to instead focus this use case on scripts that delay the initial page load, I think that could work because it's distinct from how web-vitals exposes LoAF.

> However I think the correlate-interaction-with-long-frame has the issue you're talking about and for that I'd rather advise we use web-vitals since we should be concentrating on the INP interaction. Or perhaps we can leave that and add a 4th use case specifically for INP?

I lean toward only having a single use case for anything related to identifying/attributing the causes of poor interaction latency (including INP). Otherwise you end up with a lot of overlapping content, and we're already seeing issues where the agent is matching with the wrong use cases because of use-case similarity.

One option would be to have a single guide that focuses on INP and shows how to do it with `web-vitals`. Then it could have a separate example that shows how to use the raw LoAF API to debug other slow interactions, and offer some guidance as to when it makes sense to go beyond INP.

philipwalton (commented):
FYI, I noticed some spelling/grammar issues, so I asked Gemini to review and pushed those changes.

guides/performance/correlate-interaction-with-long-frame/guide.md
```diff
@@ -3,6 +3,7 @@ name: correlate-interaction-with-long-frame
 description: Correlate a specific user interaction with the long animation frame that delayed its visual feedback.
 web-feature-ids:
   - long-animation-frames
+  - event-timing
```
tunetheweb: This use case doesn't use event-timing. Going to remove this line.

tunetheweb:
> The use case description specifically mentions "to debug Interaction to Next Paint (INP)", which is why I said that. If you wanted to instead focus this use case on scripts that delay the initial page load, I think that could work because it's distinct from how web-vitals exposes LoAF.

Ah I missed that in the description.

OK I'm veering towards this now:

Merging these two into one "identify-causes-of-poor-inp"
- correlate-interaction-with-long-frame
- identify-causes-of-poor-inp
As you say above, they are kind of similar and confusing which to serve.

Merging these two into one "identifying-heaviest-javascipt-on-your-site"
- identify-scripts-delaying-rendering
- measure-total-frame-delay
This goes beyond just "interactions" and also beyond just "loading" (which is usually easily measured in the lab anyway so LoAF isn't as useful) to ALL JavaScript on your site causing potential problems (e.g. loading delays, animations, potential INP issues...etc).

So down to 2 use cases.

WDYT?

Ironically this is what I originally had in my use cases but never finished off to submit a PR!

guides/performance/correlate-interaction-with-long-frame/guide.md
philipwalton: Gah! Embarrassing that I just trusted Gemini on that one!

philipwalton:
So just to confirm, you're proposing these two, correct?

- `identify-causes-of-poor-inp`
- `identifying-heaviest-javascipt-on-your-site`

If so, that SGTM, and I'd just rename them to this instead:

- `identify-inp-causes`
- `identify-heaviest-script`

Arguably, that last one could be `identify-heavy-scripts` and the guide could discuss how many scripts it makes sense to track, but I'm fine with either (and personally I'd just track the heaviest script anyway).

tunetheweb (approved):
OK I'm happy for this to be reviewed and merged if you're all happy!

PR #387, opened 2026-03-22
Fine tune evals for various use cases

rviscomi:
Looking at the [latest results](https://googlechrome.github.io/guidance/dashboard.html?testId=full-2026-03-21T16-53-56&source=remote), I thought I'd try improving the pass rates of the lowest performing use cases:

- animate-scrollbar-color-on-scroll
- customize-scrollbar-color-and-thickness
- improve-next-page-load-performance
- search-hidden-content
- full-session-analytics
- select-menu-interaction

This PR updates guides and evals for those specific use cases as well as fine-tuning the trigger conditions in the tool descriptions, which should have a global effect on all other use cases.

Before:

image

After:

image


paulirish:
this is great!

guides/dev-guide.ts
```diff
@@ -250,16 +250,18 @@ async function generatePrompts(targetDir: string, baseApp: string): Promise<void
...
-- Don't reference the name of the base app.. a real developer wouldn't do that.
-- Each be on its own line, prefixed with "- "
+Generate 1–4 realistic test prompts that a web developer would send to an AI coding assistant to accomplish the goal described in this guide. Write these to a file called ${PROMPTS_FILE}.
+
+Rules:
+- Write prompts as a developer talking to an AI coding assistant — casual, lowercase, sometimes vague.
+- Phrase prompts as ACTION REQUESTS or directives (e.g. "add X", "can you build Y", "implement Z"). NEVER phrase them as advisory questions (e.g. "how can I?", "what's the best way to?", "can you explain?") — the agent must implement, not just explain.
+- The first prompt is the most important: it must be specific enough that an agent implementing it would produce a grader-testable result.
+- Vary specificity: include at least one vague/intent-based prompt and one specific/technical ask.
+- Assume the developer is working on the existing app seen in base-app.html. Reference its real assets and content where relevant.
```
micahjo7, line 260: not originally in this PR since it is also listed above, but this is potentially confusing to the agent, right? the file is called `index.html` and is within the base_app

serving/mcp-server/tools/modern-web.ts
```diff
@@ -55,7 +59,9 @@ export function registerModernWebTools(server: McpServer) {
   server.registerTool(
     "get_best_practices",
     {
-      description: "MANDATORY: After finding a relevant 'use_case_id' from 'search_use_cases', call this tool to retrieve the complete, actionable implementation guide. This guidance is framework-agnostic; adapt it to whatever library or framework is being used. Do not guess or hallucinate APIs; you must use the patterns in this guide.",
+      description: `This is a retrieval tool that returns the complete, actionable implementation guide for a given web development use case.
```
micahjo7, line 47: this tool didn't seem to have any issues- is an update here necessary?

just worried in case it causes regression for any reason.

but if you tested and verified it looks the same as before, this is probably fine

guides/dev-guide.ts
rviscomi: Looks like it gets copied into `base-app.html`, so I think this would be correct

https://github.com/GoogleChrome/guidance/blob/9c952688ca9dbaa729d174766b197e47296ca367/guides/dev-guide.ts#L234

serving/mcp-server/tools/modern-web.ts
rviscomi: No you're right, this isn't strictly necessary. It was just for consistency, to actually describe what the tool does. My only tests were with both tool description changes, and guide usage is up a lot, so it's unclear if this change alone is causing regressions, but it's ok AFAICT.

guides/performance/full-session-analytics/guide.md
```diff
@@ -75,15 +75,17 @@ setInterval(queueBeacon, 10000);
...
+
+The recommended fallback strategy is to use the polyfill below, which internally handles unsupporting browsers using `fetch()` with `keepalive` or `navigator.sendBeacon()`. Your own code MUST call `fetchLater()` directly — never call `fetch()`, `sendBeacon()`, or other beacon APIs yourself.
 
 ### `fetchLater()` polyfill
 
 Use the following minimal `fetchLater()` polyfill, which implements the API as closely as possible in unsupporting browsers.
 
-The only notable behavior difference with this polyfill is instead of sending the payload when the user leaves the page, it sends it whenever the page's `visibilityState` changes to "hidden", since this is the most reliable end-of-session signal that's widely available today.
+The only notable behavior difference with this polyfill is that it uses `visibilitychange` to detect when the user leaves, rather than relying on the browser's native unload handling. This is an internal implementation detail — your code does not need to listen for `visibilitychange` or any other page lifecycle events. Just call `fetchLater()` and the polyfill handles delivery.
```
philipwalton, line 88: This is fine if it's working. But I wonder if a simpler fix could be to just not mention the `visibilitychange` event at all, so there's nothing to confuse the agent.

The only thing the agent needs to be aware of is that—since there's no reliable end-of-session events on the web—the polyfill, just to be safe, could send the event before the user's session actually ends. So a robust analytics package should always check to see if the fetchLater result  has been activated, and respond accordingly.

And looking at the code now, I see the example doesn't do that, so I should fix that...

micahjo7 (approved):

guides/performance/full-session-analytics/guide.md
rviscomi: Thanks, it does seem to help so I'll merge it with this PR, but feel free to remove it if you'll be changing the example later.

PR #399, opened 2026-03-23
skills-alpha packaging and release process

paulirish:
publishing to https://github.com/googleChrome/skills-alpha 

image

image

PR #400, opened 2026-03-23
Update config to use new serving enum instead of enableSkills

micahjo7:
To be merged only after: https://github.com/GoogleChrome/guidance/pull/399


bin/gd.ts
```diff
@@ -227,7 +228,13 @@ ${cBold('Options:')}
         process.exit(1);
       }
       
-      const buildCode = await runNpm(['build:mcp']);
+      let buildCode = 0;
+      if (config.suite.serving === Serving.MCP) {
+        buildCode = await runNpm(['build:mcp']);
+      } else if (config.suite.serving === Serving.SKILLS_CLI) {
+        buildCode = await runNpm(['--filter', 'modern-web-mcp', 'build-dist']);
```
paulirish, line 228: lol funny that the skill script is within the modern-web-mcp scope. :p

but.. thats fine we can address later.

bin/gd.ts
paulirish: do  me a favor and remove this 4 lines around the moved command?

harness/config.ts
```diff
@@ -68,7 +76,7 @@ export interface SuiteConfig {
   numRuns: number;
   tasks: string[];
   mcpServersToEnable: string[];
-  enableSkills: boolean;
+  serving: Serving;
```
paulirish, line 79: this obviously has some implications for https://github.com/GoogleChrome/guidance/pull/354  

sorry @formgit for asking you to do something that is in the middle of high-churn!

paulirish (approved):

bin/gd.ts
micahjo7: b/495783350

PR #401, opened 2026-03-23
Add typechecking to express server. 

paulirish:
I wanted to drop the lib/ folder but in doing so, I uncovered we didnt have typechecking turned on for the dashboard.

both fixed now. nothing that interesting going on here.


micahjo7 (approved):

micahjo7:
curious, why drop lib? i think we'll have it again btw, after Philip's PR https://github.com/GoogleChrome/guidance/pull/420

paulirish:
haha well yes. i wanted to remove a toplevel directory. i added it and.. didnt see a good reason for it. 

but... yes when i saw phil's PR i was like "OH NO" . i can't really argue with that one though. 

and it's probably reasonable that we take some of that code we DRY'd up in sync-usecases and throw that into lib too. 

I don't love it because.. the presence of a lib/util folder just begets all sorts of ambiguity about where code belongs. so.. thats why i'm hesitant to open the door.  :/

paulirish:
> curious, why drop lib? i think we'll have it again btw, after Philip's PR #420

btw i retained lib/colors when this merged (despite the commit message. oops).

PR #404, opened 2026-03-24
Add guidance tool validation to dashboard; Add model column & filter to dashboard landing page

micahjo7:
MCP: guidance tool activation is determined by whether `modern-web.log` is present in the result dir.
Skills (CLI): determined by session trajectory logs to see if `modern-web-use-cases` Skill is used; collects the list of all Skills activated
Skills (megaskill): uses session trajectory logs to collect Skills activated, validates that the specific "category" Skill is used (e.g. `performance`); also updates the megaskill files to test

The session logs checks are only supported for GCli and Claude Code, since we cannot parse the trajectory files in Jetski.

https://screenshot.googleplex.com/BrKaztJCaVeSwNb
https://screenshot.googleplex.com/5bJ7vAixicSiKvS

Model: https://screenshot.googleplex.com/TeTMcQfG9iV4EiF


skills-drafts/megaskill-accessibility/SKILL.md
```diff
@@ -0,0 +1,20 @@
+---
+name: accessibility
```
paulirish, line 2: deliberate add, yes?

paulirish: should this be in drafts/megaskill-*  ? not sure..

harness/lib/collection.ts
```diff
@@ -1,14 +1,59 @@
...
+        const obj = JSON.parse(line);
+        if (obj.message && obj.message.model) {
+          return obj.message.model;
+        }
+      }
+    } else {
+      const session = JSON.parse(content);
+      if (session.messages) {
+        for (const m of session.messages) {
+          if (m.type === 'gemini' && m.model) return m.model;
```
paulirish: fwiw the model can change on any turn so.. its not a slamdunk that first model found  is the most-used model (and it might use a lite model for some readline completion early on)..

but.....  i'm fine with this. :)    

alternative would be collect them all and see whats used the most.  i mean.. i guess that's easy too.

all good either way

paulirish (approved):

skills-drafts/megaskill-accessibility/SKILL.md
micahjo7: moved back, was setting this up to do a fresh set of runs across all the dimensions

harness/lib/collection.ts
micahjo7: interesting, thanks! most used seems like a better check- updated.

PR #361, opened 2026-03-20
Autofill x 5: add content to guide.me, and add grader.ts, negative-demo.html

samdutton:
Fixes #223, fixes #319, fixes #320, fixes #321, fixes #322.

PR #408, opened 2026-03-25
Autofill review

rviscomi:
This PR refines the 5 autofill use case guides and evals introduced in https://github.com/GoogleChrome/guidance/pull/361, aligning them with project conventions, and fully calibrating their evaluation suites:

- `autofill-address-form`
- `autofill-highlight-inputs`
- `autofill-payment-form`
- `autofill-sign-in-form`
- `autofill-sign-up-form`

**Key Changes:**

*   **Refined Task Prompts:** Updated `prompts.md` and their corresponding harness tasks (`harness/tasks/*.md`) to explicitly instruct the agent to modify `index.html` and not create new pages. This better simulates the environment of the simulated developer app being used in the tests.
*   **Cleaned up Demo Files:** Radically simplified the `demo.html` implementations for all five autofill use cases. Removed unnecessary CSS boilerplate, extraneous JavaScript, and visual styling. The demos now serve strictly as minimal, reference implementations focused exclusively on demonstrating proper HTML attribute usage for autofill, meeting the expectations outlined in the `project-guides` skill.
*   **Calibrated Graders (`grader.ts`):** 
    *   Updated Playwright assertions to correctly target the simplified DOM structures.
    *   Fixed edge cases involving regular expression evaluations in Playwright, ensuring that regex strings passed from `pattern` attributes correctly perform full-string matches (by wrapping them in `^(?:...)$`) to mimic HTML5 validation behaviors (e.g., verifying that fields correctly reject non-matching string patterns instead of passing implicitly).
    *   Corrected the pseudo-class matching regex in the `autofill-highlight-inputs` grader to reliably catch variations like `:auto-fill` and `:-webkit-autofill`.
*   **Refined Negative Demos:** Regenerated all `negative-demo.html` files. The new files explicitly violate every tested expectation without relying on comments, hints, or generic fallback content, resulting in 100% failure rates and perfectly calibrated evaluation test suites.
*   **Preflight Passed:** Confirmed that all use cases successfully pass the internal `gd dev <guide> --test-grader` and `npm run preflight` checks.

| Use case | Pass rate |
|---|---|
| autofill-address-form | 100% (9/9) |
| autofill-highlight-inputs | 100% (4/4) |
| autofill-payment-form | 100% (9/9) |
| autofill-sign-in-form | 100% (14/14) |
| autofill-sign-up-form | 79% (11/14) |


samdutton:
Thank you! Thanks especially for sorting out the demos. 

Looks good, though I'm not clear what happens about expectations that have been removed. For example, for [autofill-address-form](https://github.com/GoogleChrome/guidance/pull/408/changes#diff-61e668810b6b5a38e3a179c6accd3fff9df17cd03499e37c99a4674b7626d319), stuff like this:

```
- The attribute `inputmode="numeric"` **MUST** be included in an `<input>`
element that is used for input of a number, such as a payment card number, 
date in numeric format, numeric-only postal code, one-time code (OTP), 
telephone number, membership number, or quantity.
```

Are these expectations redundant, or implied/stated elsewhere, or otherwise out of scope?

rviscomi:
Thanks for pointing that out, I think my first pass might have been too aggressive.

Here's an audit of the requirements mentioned in the guide that should probably be added to the expectations:

**autofill-address-form**
- Submit button must be `<button>` or `<input type="submit">`
- Single name input preferred (don't split given/family unless required)

**autofill-highlight-inputs**
- `:-webkit-autofill` vendor prefix should be used alongside `:autofill`
- Use `box-shadow` to style background (can't override `background-color` directly)

**autofill-payment-form**
- Submit button must be `<button>` or `<input type="submit">`
- Allow spaces in card number input (don't block them)
- Security code must allow 3 or 4 digits
- Card number pattern must handle 14–20+ digit ranges

**autofill-sign-in-form**
- `type="email"` on the email input
- Submit button must be `<button>` or `<input type="submit">`
- Allow password pasting

**autofill-sign-up-form**
- `type="email"` on the email input
- Submit button must be `<button>` or `<input type="submit">`
- Allow password pasting
- `autocomplete="username"` if a username field is present

WDYT? Also, do you think the `:autofill` CSS should be part of every autofill guide and expectation?

Specifically for `inputmode=numeric`, the address form might have a numeric-only postal code, but it might not. So I'm not sure it should be a strict requirement. That said, we do have it as an expectation on the payment use case:

https://github.com/GoogleChrome/guidance/blob/ff4fff0b31323a2b0fa0912eee9720b55b9d1528/guides/user-experience/autofill-payment-form/expectations.md?plain=1#L11-L12

samdutton (approved):
LGTM (overall).

You're right about `inputmode="numeric"`: this should only be for fields that are purely numeric (quantity, numeric phone number, size, membership number, etc., as well as CVC).

There are some other removals that I'm not sure about. 

For example...

```
An `<input>` element used for entry of a user's personal name must
have an `autocomplete` attribute, and the value of that `autocomplete`
attribute must include the value `name`.
```

(and for other autocomplete values)

```
- An `<input>` element used for entry of a search query **MUST**
include the attribute `type="search"`.
```

(and other autocomplete values)

Also a few random things like `type="number"` (which is a common problem with forms):

```
- The `type="number"` attribute **MUST NOT** be included in an `<input>`
element used for entry of a number that is not meant to be incremented,
such as a payment card number, date, postal code, one-time code (OTP),
ID number, membership number, or telephone number.
```

Like you say though, we can always add things later.

rviscomi:
Ok I'll merge this as-is and approve/merge your PR. We can always expand the expectations later if we find examples of coding agents _not_ following the guides. Thanks Sam!

PR #409, opened 2026-03-25
Adds Codex CLI agent & infrastructure

micahjo7:
Results:

https://screenshot.googleplex.com/NiWyzaV7oCfwNdb


harness/agents/codex-cli-agent.ts
```diff
@@ -0,0 +1,209 @@
...
+      stdio: ['ignore', 'pipe', 'pipe']
+    });
+
+    let stdoutData = '';
+    let stderrData = '';
+
+    child.stdout.on('data', (data) => {
+      const chunk = data.toString();
+      stdoutData += chunk;
+      process.stdout.write(chunk);
```
paulirish: dunno if our others are like this but it looks weird that we both have 'pipe' set above and we force write to out/err here. 

feels like that would mean we double print all output.

harness/agents/codex-cli-agent.ts
```diff
@@ -0,0 +1,209 @@
...
+      throw new Error(`Codex exited with code ${exitCode}`);
+    }
+
+    copyResultsToTarget(workDir, targetDir);
+
+    const chatLogPath = path.join(targetDir, 'chat_log.txt');
+    fs.writeFileSync(chatLogPath, stdoutData, 'utf8');
+    console.log(`Saved output to: ${chatLogPath}`);
+
+    // Export Codex trajectory as an inline HTML viewer
```
paulirish: i'd suggest extracting this bit to its own function.

harness/lib/collection.ts
```diff
@@ -21,9 +21,9 @@ export function getGuideCategory(guideName: string): string | null {
   return null;
 }
 
-export function extractModelFromResults(resultsDir: string): string {
+export function extractModelFromResults(resultsDir: string, agent: string): string {
```
paulirish, line 28: maybe this should be a fn that lives in each agent file?

harness/lib/guidance_validation.ts
paulirish: also feels cleaner if each agent file retained its own implementation for this. 

then its nice cuz all trajectory parsing lives in the same spot and that'll make for some happy unit tests.

paulirish (approved):
nice work! great to have this coverage added.

harness/agents/codex-cli-agent.ts
micahjo7: the others had the same, and I don't think there is duplicate output. upon querying the agent:

Screenshot 2026-03-26 at 11 44 58 AM

moved this to a shared helper so it isn't duplicated across the agent files, to make it easier to update if needed in the future

harness/lib/collection.ts
micahjo7: done, added agent-specific impl

harness/lib/guidance_validation.ts
micahjo7: extracted to agent files

harness/agents/codex-cli-agent.ts
micahjo7: moved to separate function (and did the same for claude)

PR #416, opened 2026-03-26
rag: add isolated benchmarking suite

paulirish:
Read the official  results breakdown here: [go/rxolz](http://go/rxolz).

- **MiniLM reigns supreme:** We statistically tested MiniLM against heavier ~300M parameter models (Nomic, Gemma) across an 850-query pool. MiniLM consistently hit ~72%+ Top-1 accuracy and won out against everything else, so we're aggressively sticking to the lightweight 23MB choice for the MCP server.
- **Chunking rules:** We originally thought throwing full, unchunked markdown files directly at the embedder worked better based on a tiny 85-query test. However, scaling up to 850 complex queries rigorously proved that chunking strictly outperforms no-chunking on harder edge cases. We've officially reverted the "no-chunking" hypothesis; parsing by headers is explicitly remaining enabled!
- **Data Cleanup:** All the heavy evaluation logic (`test-variance.ts`, the Gemini query generators, and Plotly visualization scripts) have been extracted cleanly out of the build flow into `benchmarks/rag/`.
- **Lighter payload:** We formally booted the `gpt4all` C++ engine bindings out of `dependencies` down to `devDependencies`, massively protecting the core CLI distribution size from unnecessary native payload bloat.
- **Better precision parsing:** The `Embedder` object now correctly parses quantization instructions natively via the `@dtype` string suffix in the pipeline (e.g., `model@q8` or `model@q4`).

To run the full suite and dynamically generate interactive accuracy plots locally on your machine at any time, just run `npm run benchmark:rag`. 

---

@micahjo7  @rviscomi i tried to keep this isolated (so we have it to experiment with but it wont get in the way of the impl).   The noise in embedder and build-guides is as good as i could get it.


micahjo7 (approved):
the changes to embedder will eventually make it into the new `skills-dist` when we build and publish, i think. i guess that's fine though?

paulirish:
> the changes to embedder will eventually make it into the new `skills-dist` when we build and publish, i think. i guess that's fine though?

agree

PR #417, opened 2026-03-26
Handle new project status "Needs investigation"

rviscomi:
1. **Don't Close:** If a use case would normally be closed by the script (because the files are all there and valid), the script will skip closing it if it sees the `Needs investigation` project status.
2. **Reopen:** If a use case issue was previously closed, but someone has set the project status to `Needs investigation`, the script will automatically **reopen** the issue on GitHub.

Same logic applies to feature issues. If a feature has a use case that is being investigated, it'll be considered active and the feature issue will stay open accordingly.

This clears the way for the eval pipeline to reopen use case issues if they're not performing well

cc @paulirish


paulirish (approved):
love it thank you

PR #418, opened 2026-03-26
Fix grader-gen errors (typecheck failures, bash syntax)

micahjo7:
Adds typecheck to grader creation instructions for agent.

Adds rule to prevent invalid Bash syntax.

https://chat.google.com/room/AAQAPIjpZaI/4WjdJuPaUz8/eXpdRvG_e0M?cls=10


rviscomi (approved):

paulirish:
love it thank you

PR #420, opened 2026-03-26
Update gd to work with git worktrees

philipwalton:
This PR adds `lib/root.ts`, a shared utility that uses `git rev-parse --show-toplevel` to resolve the repo root (with `__dirname`-based fallback for non-git contexts).

This fixes an issue when `gd` is a `pnpm`-linked command, Node resolves the symlink to the original repo's `bin/gd.ts`, and `__dirname` always points to the original repo. Now `git rev-parse --show-toplevel` correctly returns the worktree root when run from a worktree.


paulirish (approved):

paulirish:
lg! resolve those conflicts and land away. im gonna tweak some things afterwards.. 

great upgrade!

PR #421, opened 2026-03-27
export common paths from paths.ts

paulirish:
a bit of a revision on @philipwalton's #420  (which was a nice inspiration :)


guides/grader-gen.ts
micahjo7: not related but this got added back in https://github.com/GoogleChrome/guidance/pull/420/changes#diff-fdd94fe8ce146fbd3248c9d22d834a4651103e14fc966038c185ad1af4e03cebR152, could you remove it if possible?

harness/upload_suite.ts
```diff
@@ -42,7 +39,7 @@ async function main() {
   // Strip trailing slashes and normalize to just the suite ID
   suiteName = path.basename(suiteName);
 
-  const resultsDir = path.join(__dirname, 'results', suiteName);
+  const resultsDir = path.join(import.meta.dirname, 'results', suiteName);
```
micahjo7: use `resultsDir` from paths.ts?

micahjo7 (approved):

PR #423, opened 2026-03-27
#363 dynamic-sibling-animations

### LINKED ISSUE #363: Create guide and evals for the dynamic-sibling-animations use case
Stagger animation or transition timing across sibling elements so each one starts after a computed delay based on its position in the sibling list.

Affected web-feature IDs: [sibling-count](https://webstatus.dev/features/sibling-count)

Use case subdir: [guides/ui-behaviors/dynamic-sibling-animations](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-behaviors/dynamic-sibling-animations)

Related features: #78

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

jamesnw:
#363 
This seems like a fairly small use case, I'm wondering if there are things I should expand on?

Also, I'm fairly uncertain on what to do for expectations here, or what is reasonable to expect playwright to test here. Also, the negative tests end up using the fallback, meaning that the only way to test is by asserting implementation details rather than the effect. Any thoughts on how to approach this?


guides/user-experience/dynamic-sibling-animations/guide.md
```diff
@@ -9,3 +9,57 @@ sources:
   - https://css-tricks.com/how-to-wait-for-the-sibling-count-and-sibling-index-functions/
   - https://utilitybend.com/blog/styling-siblings-with-CSS-has-never-been-easier.-Experimenting-with-sibling-count-and-sibling-index/
 ---
+
+# Creating a stagger animation
+
+Stagger animations provide an interesting effect where multiple ordered elements animate
+sequentially with a slight delay between each, rather than all animating
+at once. This technique is often used in lists, galleries, or navigation
+menus to guide the user's eye and add a polished, rhythmic feel to interactions.
```
rviscomi:
```suggestion
Stagger animations provide an interesting effect where multiple ordered elements animate sequentially with a slight delay between each, rather than all animating at once. This technique is often used in lists, galleries, or navigation menus to guide the user's eye and add a polished, rhythmic feel to interactions.
```

guides/user-experience/dynamic-sibling-animations/guide.md
```diff
@@ -9,3 +9,57 @@ sources:
...
+#stagger-list > .item {
+  --stagger-time: 0.1s;
+  /* Define the animation first */
+  animation: fade-in 0.4s;
+  /* Set the `animation-delay` to a time multipled by the `sibling-index()` */
+  animation-delay: calc(sibling-index() * var(--stagger-time))
+}
+```
+
+Do respect user preferences by disabling the animation for users who prefer reduced motion.
```
rviscomi:
```suggestion
**MANDATORY:** Respect user preferences by disabling the animation for users who prefer reduced motion. 
```

guides/user-experience/dynamic-sibling-animations/guide.md
```diff
@@ -9,3 +9,57 @@ sources:
...
+}
+```
+
+Do respect user preferences by disabling the animation for users who prefer reduced motion. 
+
+```css
+@media (prefers-reduced-motion: reduce){
+  /* Disable animation for users who prefer reduced motion. */
+  animation: none;
+}
```
rviscomi, line 39: IIUC omitting the selector implies CSS nesting, as it's done in demo.html. That's fine for the demo, but the guide should probably expand this out so that it doesn't rely on nesting, which isn't quite widely available yet:

```suggestion
@media (prefers-reduced-motion: reduce){
  /* Disable animation for users who prefer reduced motion. */
  #stagger-list > .item {
    animation: none;
  }
}
```

guides/user-experience/dynamic-sibling-animations/guide.md
```diff
@@ -9,3 +9,57 @@ sources:
...
+  [...staggerList.children].forEach((el, index)=>el.style.setProperty('--sibling-index', index + 1));
+}
+```
+
+Add an `animation-delay` declaration that uses the `--sibling-index` custom property. It must be before the `animation-delay` declaration that uses the `sibling-index()` function.
+
+```css
+#stagger-list > .item {
+  animation-delay: calc(var(--sibling-index) * var(--stagger-time));
+  animation-delay: calc(var(sibling-index()) * var(--stagger-time));
```
rviscomi:
```suggestion
  animation-delay: calc(sibling-index() * var(--stagger-time));
```

guides/user-experience/dynamic-sibling-animations/expectations.md
```diff
@@ -0,0 +1,4 @@
+* The animation on the first element starts before the animation on the second element.
+* `sibling-index()` is multiplied by a time and used as the `animation-delay`.
+* If a legacy fallback is required, JavaScript checks if `sibling-index()` is supported, and if not, adds a custom property on each sibling element being animated, starting with 1. The `animation-delay` uses this property in its calculation, and a second `animation-delay` uses `sibling-index()`.
```
rviscomi: Each expectation should correspond to a single, concrete assertion
```suggestion
* The implementation provides a fallback for older browsers using CSS custom properties.
* The custom property values start at 1 and increment for each sibling.
* The fallback is applied conditionally only when `sibling-index()` is not supported.
```

rviscomi (changes requested):
Ready for `gd dev` to generate prompts and grader

guides/user-experience/dynamic-sibling-animations/prompts.md
```diff
@@ -0,0 +1,4 @@
+- add a 0.4s fade-in animation to the cards in the .grid so they stagger in sequentially. use the css sibling-index() function for the animation-delay multiplied by a 0.1s stagger time. also include a js fallback that sets a --sibling-index variable for older browsers, but make sure to wrap it in a CSS.supports check so it doesn't run if it's not needed. disable the animation entirely if the user prefers reduced motion.
```
rviscomi: The ideal prompt wouldn't be this prescriptive, to better emulate a developer who isn't necessarily aware of the latest features and best practices. 

Was this generated or handwritten? (Just asking so we can adjust the generation prompt if needed)

guides/user-experience/dynamic-sibling-styling/guide.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: dynamic-sibling-styling
-description: Stagger animation or transition timing across sibling elements so each one starts after a computed delay based on its position in the sibling list.
+description: Compute visual properties like position, color, or size dynamically for each child element based on how many siblings are present and its position among them.
```
rviscomi, line 3: I think the old description actually does a better job of describing the use case that's solved by this guide, which should make it more discoverable and likely to be used by agents. Was there something about it that you didn't like?

rviscomi (commented):
Ran this locally:

```
Agent test results:
  Base app (pre):    0/6 checks passed (0%)
  Unguided:          1/6 checks passed (17%)
  Guided:            1/6 checks passed (17%)
  Guide impact:      +0% (vs unguided)
```

Worth a closer look into why the guided checks are so low. Seems like maybe the agent never thought to query the CLI, so the prompt might need to be adjusted.

guides/user-experience/dynamic-sibling-animations/prompts.md
jamesnw: This was generated. I agree it seems overly prescriptive .

jamesnw: @rviscomi Should I be manually editing these prompts to improve them? Or is the correct course of action waiting until the generation prompt is improved, and regenerating?

rviscomi: Yeah, manually edit for now.

If useful, I've found this prompt helpful for fine-tuning the task.md:

>look at task.md. if a coding agent with access to guide.md implements the first prompt, would you expect it follow everything in expectations.md?

rviscomi (approved):
Since the guide is in good shape, I'm actually inclined to merge this as-is and follow up on fine-tuning the evals later as needed. Thanks @jamesnw!

Issue #275, opened 2026-03-11
Create guide and evals for the content-based-styling use case


paulirish:
Build a component that changes its layout based on whether it contains specific child elements (or not). For example, if the component contains an image, use a multi-column layout, otherwise default to a single-column layout.

Affected web-feature IDs: [has](https://webstatus.dev/features/has), [not](https://webstatus.dev/features/not)

Use case subdir: [guides/css/content-based-styling](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/css/content-based-styling)

Related features: #56 #67

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

Issue #274, opened 2026-03-11
Create guide and evals for the child-state-based-styling use case


paulirish:
Build a component that changes its styling based on the state of one of its child elements. For example, a component that renders in light or dark mode based on whether a theme toggle is checked (or not).

Affected web-feature IDs: [has](https://webstatus.dev/features/has), [not](https://webstatus.dev/features/not)

Use case subdir: [guides/css/child-state-based-styling](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/css/child-state-based-styling)

Related features: #56 #67

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

PR #426, opened 2026-03-27
Create guides and evals for :has use cases

philipwalton:
Addresses #56 #274 #275


rviscomi (approved):
The guidance looks good, but the evals are the same for both unguided and guided runs. It doesn't seem like the agent used the CLI at all.

This won't be as much of an issue when we add support for the high-level CSS skills, which include `:has()` so merging this now LGTM and we can fine-tune later as needed.

Issue #146, opened 2026-02-25
Create guide and evals for the form-fields-automatically-fit-contents use case


paulirish:
Allow form fields to grow and shrink to fit the user input, e.g. as the user types or selects a different option. Apply maximum and minimum size limits to create dynamic and responsive form fields that conform with the page design.

Affected web-feature IDs: [field-sizing](https://webstatus.dev/features/field-sizing)

Use case subdir: [guides/forms/form-fields-automatically-fit-contents](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/forms/form-fields-automatically-fit-contents)

Related features: #54

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

Issue #54, opened 2026-02-19
field-sizing


rviscomi:
The field-sizing CSS property allows form controls such as <textarea> to be sized based on their content.

### MDN Documentation
- [field-sizing](https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/field-sizing)

### Chrome Documentation

<details>
<summary>View 4 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/docs/css-ui/css-field-sizing
- https://web.dev/en/blog/contenteditable-plaintext-only-baseline
- https://web.dev/en/learn/forms/fields
- https://web.dev/en/learn/forms/form-fields
</details>

---
Feature ID: field-sizing

**Use cases:**
- [x] #146
- [ ] #1242

rowan-m:
Use cases complete, ready for guidance.

rviscomi:
Tracking the use case in https://github.com/GoogleChrome/guidance/issues/146

PR #428, opened 2026-03-27
Add guide and evals for field-sizing use case

philipwalton:
Addresses #54 #146


rviscomi (approved):
LGTM with a couple of changes

PR #429, opened 2026-03-28
Use case research skill

rviscomi:
Alternative skills-based implementation to https://github.com/GoogleChrome/guidance/pull/360


rviscomi:
Closing in favor of https://github.com/GoogleChrome/guidance/pull/486 which is branched directly off of main

PR #431, opened 2026-03-28
Forms skill and subdir

rviscomi:
- Creates a new `guides/forms` subdir
- Marks the form skill draft "prod-ready" by moving it out of `skill-drafts`
- Moves existing form-related use cases into the new subdir

@samdutton the [form skill](https://github.com/GoogleChrome/guidance/pull/431/changes#diff-7883c88054584f40b3134b8dbf296ce3b5177cde18dad0db78ebd4d39e39a653) was generated based on the content at http://web.dev/learn/forms, so hopefully nothing new or surprising, but I'd really love your feedback on it in case anything is off.

Tip: use "expand all lines" to see and comment on the full skill file

image


guides/forms/SKILL.md
```diff
@@ -261,18 +263,18 @@ form.addEventListener('submit', (e) => {
...
-> [!TIP]
-> **Password Visibility**: Do not hide passwords unconditionally; provide a visibility toggle to reduce entry errors rather than forcing double-entry verification.
 
 ## 9. Address Collection
 
 ### Guidelines
 
 - **DO** use a single field for names.
 - **DO** use `autocomplete="street-address"`.
+- **DO** use free-form textareas for addresses to accommodate global diversity.
```
samdutton: Maybe something like "If the site has users in different countries, **DO** use the `<textarea>` element for addresses, to accommodate different address formats in different geographical regions."

I think "to accommodate global diversity" sounds a bit weird.

Also, I suspect a lot of sites (for better or worse) will insist on using a "multiline" address form, with address data entered one part/input at a time (e.g. "Street name", "Town or city", etc.) In that case, an address form should use inputs with autocomplete values address-line1, address-line2, etc.

guides/forms/SKILL.md
```diff
@@ -254,47 +256,45 @@ form.addEventListener('submit', (e) => {
 <form action="/checkout" method="POST">
```
samdutton: Not sure what this form is for.

Maybe just the one from the [web.dev sign-up form best practices](https://web.dev/articles/sign-up-form-best-practices) [example](https://codepen.io/web-dot-dev/pen/VYvvJpj?editors=1010).

samdutton (changes requested):
Did a commit with some tweaks, and suggested a couple of other changes.

guides/forms/SKILL.md
rviscomi: Done

guides/forms/SKILL.md
rviscomi: Done

rviscomi:
Thank you!

samdutton (approved):
LGTM.

@rviscomi What's the workflow for testing SKILL.md against the demos?

rviscomi:
> @rviscomi What's the workflow for testing SKILL.md against the demos?

We're still working out the details but hopefully we'll have a process for testing it this week.

Issue #163, opened 2026-02-27
Create guide and evals for the design-token-reactivity use case


paulirish:
Define higher-order design tokens, like density modes (compact, comfortable, spacious) or themes and have descendant components react to changes directly and in component-appropriate ways.

Affected web-feature IDs: [container-style-queries](https://webstatus.dev/features/container-style-queries)

Use case subdir: [guides/css/design-token-reactivity](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/css/design-token-reactivity)

Related features: #48

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

philipwalton:
@Snugug can you submit a demo for this use case so that the particular implementation technique you had in mind is captured here?

PR #433, opened 2026-03-28
Design Token Demo

Snugug:
All style container queries are _roughly_ the same pattern, what distinguishes this from the "position aware" example is that is typically powered by some config a user has chosen (like site density). It's similar to light mode/dark mode, but that can usually _mostly_ be done with changing values for custom properties, whereas something like density can change actual layout and changes may be more localized than what would be appropriate to do through flipping a custom property.


guides/user-experience/design-token-reactivity/demo.html
rviscomi, line 1: FYI these demo files are only used for calibrating the grader, never shown to developers or their coding agents. So while this is a beautiful demo, it's a bit overkill. See the project-use-cases/SKILL.md for more info:

https://github.com/GoogleChrome/guidance/blob/15f0bdb87b3d18f52bb6dc2fc7fcbc141afbe8ef/.agents/skills/project-use-cases/SKILL.md#L117-L123

Gemini suggested this as a minimal implementation of the use case. Could this work?

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <style>
    .container {
      --density: compact;
      /* No container-type required for style queries! */
    }

    .item {
      padding: 1rem;
      border: 1px solid #ccc;
    }

    @container style(--density: compact) {
      .item {
        padding: 0.5rem;
      }
    }

    @container style(--density: spacious) {
      .item {
        padding: 2.5rem;
      }
    }
  </style>
</head>
<body>
  <div class="container" id="target-container">
    <div class="item" id="target-item">I react to density</div>
  </div>

  <button id="toggle-spacious">Toggle Spacious</button>

  <script>
    document.getElementById('toggle-spacious').addEventListener('click', () => {
      const c = document.getElementById('target-container');
      const current = getComputedStyle(c).getPropertyValue('--density').trim();
      c.style.setProperty('--density', current === 'spacious' ? 'compact' : 'spacious');
    });
  </script>
</body>
</html>
```

Snugug: It sounded like from @philipwalton that this was because people didn't quite understand what I had in mind here, which is why I made it a larger demo. Happy for it to be more compact, though, if people can get what they need from it.

rviscomi: Reading https://github.com/GoogleChrome/guidance/issues/163#issuecomment-4143874400 I think he was mostly just asking for _a_ demo, since it was missing entirely

rviscomi (approved):
Converted to a minimal demo

PR #434, opened 2026-03-28
Accessibility skill

### LINKED ISSUE #531: Create a high-level accessibility skill
https://github.com/GoogleChrome/guidance/blob/main/skills-drafts/accessibility/SKILL.md

rviscomi:
Progress on https://github.com/GoogleChrome/guidance/issues/531


guides/accessibility/SKILL.md
knowler:
```suggestion
.visually-hidden:where(:not(:focus-within, :active)) {
  position: absolute !important;
  clip-path: inset(50%) !important;
  overflow: hidden !important;
  width: 1px !important;
  height: 1px !important;
  margin: -1px !important;
  padding: 0 !important;
  border: 0 !important;
  white-space: nowrap !important;
}
```

- Uses a single class that just has some `:not()` conditions for `:focus-within` and `:active`.
  - This is a bit different than the direction you were going with this. I think the not focused/active styles could vary situation-to-situation. As much as avoiding the layout shift is important, I don’t think it’s necessary for every case. Could this be something we call out separately? Or show an example where the styles for the thing that `.visually-hidden` is applied to now take over when it’s focused/active.
- `:focus-within` is more encompassing than `:focus` which makes the utility more versatile (it can wrap a set of focusables — see Eric Meyer’s website for an example of this in the wild). Unfortunately, it also can tend to be a pattern in React et al. to create a component that wraps other things (e.g. `<VisuallyHidden><a href="#content">Skip to main content</a></VisuallyHidden>`).
- Use `:where()` for the `:not()` conditions to avoid increasing specificity which will make it work well with other utility classes.
- I considered removing `!important`, but decided to leave it as I think exceptions should be included in the `:not()` condition.

knowler: I think we should find a different section for “Provide a Standard Visually Hidden Utility Class” (and the corresponding example) as this is not related to “Modals and Native Dialogs.” Or does it even matter if this isn’t meant for humans? I think it’s related to accessible names (which might make for a section on its own), but also maybe being able to navigate the page since it’s so common for skip links (e.g. use of headings, landmarks, etc.) and even live regions.

knowler: For this section, I think it could be better to focus on content navigability/discovery:

- Placing all content within landmarks.
- Structuring the main content with headings.
- Using lists for contiguous, repeated content so they’re skippable + pre-counted for assistive tech.
- Using skip links to skip past repeated content across views (e.g. site header/banner) or long/infinite lists within a view, so they’re skippable by keyboard users.
- Avoiding overusing landmarks.
  - Avoid labeling `<section>` elements as this makes them `region` landmarks. `region` landmarks should be used as a last resort, when you need a landmark and the other options aren’t suitable.

A separate section could cover semantic HTML and ARIA:

- Prefer HTML elements and attributes to ARIA roles and properties.
  - Example: `<button>` implies `role=button`, so setting the latter explicitly is unnecessary.
  - Example: `required` implies `aria-required`, so setting the latter explicitly is unnecessary when using the former.
  - Caveat: sometimes `aria-disabled` is often preferable over `disabled` (the latter is a bit of a nuclear option as it prevents focusability — `tabindex=0` doesn’t work).
- Avoid redundant ARIA roles.
  - Caveat: Safari removes list semantics from list elements (used outside of a `<nav>`) that use flex/grid or set `list-style: none`. There are CSS techniques to get around this, but in many cases it’s unavoidable and you must use `role=list`.
    - Using flex/grid with tables can have a similar expect which I believe impacts more browsers (so you end up needing to set the roles anyway).
- Be aware that custom elements can have implicit ARIA roles/properties set with `ElementInternals`. Some automated testing is unable to detect these values which can cause false negatives.
- Ensure that ARIA implementations don’t just add the attributes, but also match their affordances by providing the expected functionality.
  - This means that some ARIA patterns don’t work without JavaScript (CSS-only is not an option) and need suitable alternatives.

knowler:
```suggestion
:where(a:any-link, button):focus-visible {
```

- I tend to only style `:focus-visible` for focus outline (which I believe is what the UA styles do).
- Minor nit: `a` is inclusive of `<a>` without `href` which is not `role=link`. I prefer to use `:any-link` to only style hyperlinks, and more specifically `a:any-link` to exclude `area`.

knowler:
```suggestion
<main id="main" tabindex="-1">...</main>
```

There’s a lot of nuance that could be said here, but I think it’s safest to set `tabindex=-1` for skip link targets that aren’t otherwise focusable as that seems to work better across the board.

knowler:
```suggestion
  <input id="pwd" type="password" aria-describedby="pwd-hint" autocomplete="current-password" required>
```

Redundant `aria-required` when `required` is used.

knowler:
```suggestion
<span id="pwd-hint">Must contain at least 8 characters.</span>
<input id="pwd" type="password" aria-describedby="pwd-hint" autocomplete="current-password" required aria-required="true">
```

Generally, it’s better to place descriptions above the input so that autocomplete popovers don’t block them while editing the field.

knowler:
```suggestion
- **Provide Pause mechanism**: Allow users to stop auto-running carousels banners or other persistent animations.
```

Even minor animations can be distracting for some folks (but shouldn’t necessarily fall into the `prefers-reduced-motion` category).

jamesnw: I'm not sure what is meant by "allow standard zoom levels".

jamesnw:
```suggestion
- **Empty Alt properties for decorative visuals**: Use `alt=""` or `role="presentation"` to hide decorative images from screen readers.
```
We need to differentiate from visually hiding it. Maybe "removing it from the accessibility tree"? Although I'm no sure the accessibility tree is introduced. (Should it be?)

jamesnw, line 267: Is this example overly prescriptive? The advice is "have a transcript", not "use details/summary". 

Also, I'm thinking the answer is no, but is there any way we can associate the audio file with yhe transcript? `aria-describedby` isn't quite right.

guides/accessibility/SKILL.md
```diff
@@ -307,20 +318,12 @@ article {
...
-
 ## 8. Testing Validations
 
 ### Actionable Guidelines
 
 #### DOs
-- **Run Automated checks via axe-core audits**: Catch missing alt texts or low contrasts.
+- **Run Automated checks via axe-core or Lighthouse audits**: Catch missing alt texts or low contrasts (e.g., via Lighthouse in Chrome DevTools MCP).
 - **Validate Sequential Navigations using keyboards alone**: Disconnect mouse traps to verify logic.
 - **Use Screen Readers with calibrated browsers**: Rely on standard bindings (e.g., VoiceOver with Safari).
```
jamesnw:
```suggestion
- **Test on Screen Readers with calibrated browsers**: Rely on standard bindings (e.g., VoiceOver with Safari).
```

guides/accessibility/SKILL.md
```diff
@@ -307,20 +318,12 @@ article {
...
-```
-
 ## 8. Testing Validations
 
 ### Actionable Guidelines
 
 #### DOs
-- **Run Automated checks via axe-core audits**: Catch missing alt texts or low contrasts.
+- **Run Automated checks via axe-core or Lighthouse audits**: Catch missing alt texts or low contrasts (e.g., via Lighthouse in Chrome DevTools MCP).
 - **Validate Sequential Navigations using keyboards alone**: Disconnect mouse traps to verify logic.
```
jamesnw: What is meant by "Disconnect mouse traps"?

guides/accessibility/SKILL.md
jamesnw: When would the `dialog` element not be available? It's not browser support limited, so it would be some use cases?

guides/accessibility/SKILL.md
```diff
@@ -239,7 +250,7 @@ element.addEventListener('keydown', (e) => {
 - **Minimum contrast standards**: Maintain 4.5:1 for normal text and 3:1 for large text or icons.
 - **Use multiple state indicators**: Do not denote success/errors ONLY with color. Use icons or text.
 - **Relative font size units**: Use `rem` or `em` for font sizes instead of `px`.
-- **Left text alignment**: Cap paragraph blocks to a maximum of 80 characters width.
+- **Consistent or Start alignment (avoid justify)**: Cap paragraph blocks to a maximum of 80 characters width.
```
jamesnw:
```suggestion
- **Consistent or Start alignment**: Avoid `justify` alignment as it can be more difficult to read.
-  ** Avoid long lines of text**: Cap paragraph blocks to a maximum of 80 characters width.
```
This seems to be 2 separate items.

guides/accessibility/SKILL.md
```diff
@@ -239,7 +250,7 @@ element.addEventListener('keydown', (e) => {
...
 - **Support user preference media queries**: Implement `@media (prefers-color-scheme: dark)` and `@media (prefers-contrast: high)`.
```
jamesnw, line 358:
```suggestion
- **Support user zoom preferences**: Allow users to resize text up to 200%.

```

guides/accessibility/SKILL.md
```diff
@@ -239,7 +250,7 @@ element.addEventListener('keydown', (e) => {
...
```
jamesnw: I think this needs more nuance. `prefers-color-scheme` should be about supporting light and dark (and mention the `color-scheme` property).

`prefers-contrast` should describe the cases where you would want to use it, as it shouldn't be needed in most websites, unless they are doing something abnormal.

guides/accessibility/SKILL.md
jamesnw: Several things going on here-

1. Do we really want to discourage italics as strongly as all-caps?
2. Do we want to mention semantic elements (em/strong) here, or is this section solely for visual styling.
3. The second sentence is confusing - maybe remove the word "sparingly", but then add a separate caution to limit the use of emphasis in general. 
4. I don't think "dense text" is useful as part of the sentence.

guides/accessibility/SKILL.md
```diff
@@ -254,7 +265,7 @@ element.addEventListener('keydown', (e) => {
 body {
   font-size: 1rem;
   line-height: 1.5;
-  text-align: left;
+  text-align: start; /* Supports LTR and RTL */
 }
 article {
   max-width: 65ch; /* Caps character lengths */
```
jamesnw: The prose says max of 80 characters- do we want to use 80ch here, or add a comment about why it's less?

jamesnw (commented):
I agree with @knowler's suggestions for organization.

I'm also wondering if we could add a few more points to the intro-

1. Accessibility is the minimum, and users should strive for a higher standard of usability
2. Accessibility patterns are very use-case specific, and need to be tested (including by users) in order to ensure accessible content.

guides/accessibility/SKILL.md
knowler: - [`aria-details`](https://developer.mozilla.org/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-details) ([support table from 2022](https://a11ysupport.io/tech/aria/aria-details_attribute)) might work better for that as it’s meant for descriptions that might include interactive content (in this case the `<summary>`). I’d be curious about the effect it might have on the role as [`<audio>` doesn’t have an ARIA role](https://www.w3.org/TR/html-aam-1.0/#el-audio) (i.e. maybe it’ll announce as “group”—I might do some testing).
- I think the transcript being colocated in the DOM seems ok.

- - -

Some quick + limited screen reader testing of the snippet with and without `aria-details`:

- VoiceOver + Safari announces “Audio Controls group” whether `aria-details` is present or not, so that seems safe.
- NVDA + Firefox on Windows 11 announces “has details Play button.”

So I think it could be a harmless addition.

knowler, line 8: Some DOs and DON’Ts for an additional “Accessible names and descriptions” section (I’ll update this comment with the description guidance):

#### DOs

- **DO** prefer native HTML naming mechanisms such as `<label>` for form controls, `<caption>` for `<table>`, `<legend>` for `<fieldset>`, `<figcaption>` for `<figure>`.
- **DO** explicitly associate a `<label>` element with a form control using the `for` attribute even when associating by nesting: this helps improve assistive technology support.
- **DO** prefer `aria-labelledby` over `aria-label` when a visual label exists: this avoids duplication, improves maintainability, and has better support for translation.
- **DO** prefer to use the same name for hyperlinks with the same `href`.
- **DO** use visually hidden text to ensure links/controls with different effects have different accessible names even though they might have the same name visually.

#### DON’Ts

- **DON'T** use `aria-label` or `aria-labelledby` for elements with a role that should not be named (e.g. `<div>` or `<span>` elements or custom elements without a `role`).
  - Be aware that custom elements can have ARIA applied with `ElementInternals`, so the `role` attribute alone is not sufficient for determining if a custom element has an ARIA role.
- **DON'T** use the same accessible name for controls that have different effects and exist in the same context (i.e. this would exclude close buttons for different modal dialogs as they wouldn’t be open at the same time).
- **DON'T** use the same accessible name for hyperlinks that have different `href`s.
- **DON'T** include extended descriptions, error message within labels.
- **DON'T** include state that has already been reflected with ARIA (e.g. `aria-expanded`, `aria-checked`, `aria-selected`, `aria-pressed`): this avoid redundancy and ambiguity.
- **DON'T** include the name of an element’s role or anything synonymous with it in the label (e.g. `<nav aria-label="Primary navigation">`): this is redundant and extra verbose for screen readers.
- **DON'T** use the `title` or `placeholder` attributes for naming.
- **DON'T** include interactive elements whose text content does not otherwise make sense within a description referred to by `aria-describedby`.

knowler, line 310: We should create a dedicated section to cover live regions as they aren’t form-specific. Some other considerations for live regions:

- For live regions that are not also visible, try to use centralized live regions to help coordinate messages, create consistency, and improve maintainability. It’ll likely be necessary to include one for `assertive` and one for `polite`, but perhaps others depending on the `aria-atomic` setting. Often web frameworks have their own method for managing announcements.
- If a live region can update rather quickly, it should be debounced to avoid spamming users (e.g. combobox search result counts).
- For situations where other announcements might be made (e.g. user is typing; context/managed focus change), it can be advantageous to add a slight delay when adding new messages to a live region.
- Avoid using live regions for interstitial states (e.g. “Loading” or “Updating”). This can create unnecessary noise.
- Making live region updates as modal dialogs open/close or content becomes inert, may require extra coordination to ensure messages are not added to DOM that is inert. This is especially important when debouncing or delaying messages.

knowler: This might be a bit too permissively worded.

- Anything that is focusable whether by keyboard or programmatically _should_ have an implicit or explicit ARIA role (so not everything should be made focusable).
- Recommending `tabindex=0` also means it becomes a part of the tab order. That needs to be an intentional decision. `tabindex=-1` would make it focusable, but not a part of the tab order which could also be an option.

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,451 @@
...
+<header>
+  <h1>Platform Dashboard</h1>
+  <nav aria-label="Primary">
+    <ul>
+      <li><a href="/">Home</a></li>
+    </ul>
+  </nav>
+</header>
+<main>
+  <section>
```
knowler, line 47:
```suggestion
```html
<!-- Good: Semantic landmarks, heading hierarchy, skip link -->
<header>
  <a href="#content" class="skip-link visually-hidden">Skip to content</a>
  <nav aria-label="Primary">
    <ul>
      <li><a href="/">Home</a></li>
    </ul>
  </nav>
</header>
<main id="content" tabindex="-1">
  <h1>Platform Dashboard</h1>
  <section>
```

- Moves the `<h1>` to the main landmark as the banner landmark is meant for _site-wide_ information that’s repeated on every page (and therefore something that’s worth skipping).
- Demonstrates skip link with visually hidden class (presumably the `skip-link` class here would be providing styles for when the link is focused — something that the visually hidden styles we’ve shared do not do).

rviscomi:
@jamesnw @knowler thanks for all of your feedback! It should be addressed now, LMK if you have any other suggestions or if you think this is in good shape to merge.

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,451 @@
...
+customWidget.addEventListener('keydown', (e) => {
+  if (e.key === 'Enter' || e.key === ' ') {
+    e.preventDefault(); // Prevent page scrolling on Spacebar
+    // Toggle state for screen readers
+    const pressed = customWidget.getAttribute('aria-pressed') === 'true';
+    customWidget.setAttribute('aria-pressed', !pressed);
+    performCustomAction();
+  }
+});
+```
```
knowler, line 211: When implementing a custom keyboard handler for button-like elements, <kbd>Enter</kbd> should be a `keydown` handler and <kbd>Space</kbd> should be a `keyup` handler. This matches how `<button>` work where <kbd>Enter</kbd> will repeat and <kbd>Space</kbd> will be activated just once.

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,451 @@
...
+<!-- Good: Skip to main content -->
+<a href="#main" class="skip-link">Skip to main content</a>
+<main id="main" tabindex="-1">...</main>
+```
+
+```javascript
+// Good: Keyboard handlers for complex custom widgets (e.g., Tree items).
+// NOTE: This pattern applies ONLY to non-standard UI where no native HTML tag exists.
+// Always prioritize native <button> or <input> elements for standard interactions.
+// Elements MUST have the appropriate ARIA role (e.g., role="treeitem" or role="button").
```
knowler:
```suggestion
// Elements MUST have the appropriate ARIA role (e.g., role="treeitem" or role="tab").
```
Let’s use a non-`role=button` example to avoid seeming like we’re encouraging custom button implementations.

knowler (commented):
I have a few more changes that I’m researching, but here are a few I’ve noticed while I do so.

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,451 @@
...
+
+#### DOs
+- **Connect Labels Programmatically**: Use `<label for="id">` linked to `<input id="id">`.
+- **Use Autocomplete**: Set valid standard `autocomplete` options (e.g., `"email"` or `"given-name"`) for user profiles.
+- **Link hints to inputs via `aria-describedby`**: Associate help text with inputs, and place the hint above the input so autocomplete popovers don't cover it during editing.
+- **Announce dynamic errors via live regions**: Use `aria-live` or shift focus to error lists.
+- **Provide form validation constraints**: Use `required` (or `aria-required="true"` only when `required` isn't applicable) to signal mandatory inputs.
+
+#### DON'Ts
+- **Don't rely on placeholders alone**: Placeholders are not persistent labels.
```
knowler: I think we can put this more strongly:

```suggestion
- **Don't use placeholders as labels**: Placeholders are not persistent labels.
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,451 @@
...
+
+### Actionable Guidelines
+
+#### DOs
+- **Support Reduced Motion media queries**: Support `@media (prefers-reduced-motion: reduce)` media queries.
+- **Provide Pause mechanism**: Allow users to stop auto-running carousels banners or other persistent animations.
+- **Default to static views**: Consider defaulting to static states and allowing users to opt-in to motion.
+
+#### DON'Ts
+- **Don't exceed flash limits (three per second)**: Avoid rapid light-to-dark flashing to prevent seizures.
```
knowler: Strengthening this.

```suggestion
- **Don't exceed flash limits (three per second)**: Never include rapid light-to-dark flashing. Such effects can cause seizures.
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,451 @@
...
+- **Don't reuse an accessible name across controls with different effects in the same context** (close buttons for two different open dialogs are fine because only one is reachable at a time; two "Submit" buttons on the same form are not).
+- **Don't reuse an accessible name across hyperlinks pointing to different `href`s.**
+- **Don't pack descriptions, error messages, or instructions into the label.**
+- **Don't repeat state already exposed via ARIA** (`aria-expanded`, `aria-checked`, `aria-selected`, `aria-pressed`) inside the accessible name — it creates redundancy and ambiguity.
+- **Don't include the role name in the label**: `<nav aria-label="Primary navigation">` reads as "Primary navigation navigation."
+- **Don't use `title` or `placeholder` as a naming mechanism.**
+- **Don't include interactive elements in an `aria-describedby` target** unless their text content reads sensibly as a description on its own.
```
knowler: Expanding this prior addition to include an example.

```suggestion
- **Don't include interactive elements in an `aria-describedby` target** unless their text content reads sensibly as a description on its own (e.g. if a link’s text is the same as how it’s labelled elsewhere, it can be included within a description).
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,451 @@
...
+
+#### DON'Ts
+- **Don't use ARIA when native HTML exists**: Avoid `<div role="button">` or `<a role="button">` if `<button>` works.
+- **Don't add redundant ARIA roles or properties**: Avoid `<ul role="list">`, `<nav role="navigation">`, or `<input required aria-required="true">`.
+  - **Caveat**: Safari removes list semantics from `<ul>`/`<ol>` outside `<nav>` when `list-style: none` or `display: flex`/`grid` is applied. In that case `role="list"` is required to restore them. A similar effect impacts `<table>` with flex/grid across more browsers.
+- **Don't assume custom elements have no ARIA**: Custom elements can attach ARIA via `ElementInternals`, which some automated test tools can't see — so the absence of `role`/`aria-*` attributes in markup doesn't prove the element has no semantics. Verify with a screen reader or the browser's accessibility-tree inspector.
+
+## 3. Accessible Names and Descriptions
+
+Every interactive element needs an accessible name, and many benefit from an accessible description. Names are short and identify the control; descriptions add context.
```
knowler: This section has content that applies to naming/describing in general, so we should make this preamble inclusive of landmarks as well.

```suggestion
Every interactive element and some landmarks need an accessible name, and many benefit from an accessible description. Names are short and identify the element; descriptions add context.
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,451 @@
...
+- **Explicitly associate `<label>` with its control via `for`/`id`**, even when nesting the input inside the label — explicit association improves assistive-tech support.
+- **Prefer `aria-labelledby` over `aria-label` when a visible label exists**: avoids duplication, improves maintainability, and translates better.
+- **Reuse the same accessible name for hyperlinks that share an `href`.**
+- **Use visually hidden text to disambiguate controls** that look identical visually but do different things (e.g. multiple "Edit" buttons in a list).
+
+#### DON'Ts
+- **Don't put `aria-label`/`aria-labelledby` on elements that shouldn't be named** — e.g. plain `<div>`, `<span>`, or custom elements without a role. Custom elements may have an implicit role set via `ElementInternals`, so the absence of a `role` attribute isn't conclusive.
+- **Don't reuse an accessible name across controls with different effects in the same context** (close buttons for two different open dialogs are fine because only one is reachable at a time; two "Submit" buttons on the same form are not).
```
knowler: Adding a better example here.

```suggestion
- **Don't reuse an accessible name across controls with different effects in the same view** (close buttons for two different open dialogs are fine because only one is reachable at a time; multiple “Edit” buttons for different content is not).
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,451 @@
...
+
+Every interactive element needs an accessible name, and many benefit from an accessible description. Names are short and identify the control; descriptions add context.
+
+### Actionable Guidelines
+
+#### DOs
+- **Prefer native naming mechanisms**: `<label>` for form controls, `<caption>` for `<table>`, `<legend>` for `<fieldset>`, `<figcaption>` for `<figure>`.
+- **Explicitly associate `<label>` with its control via `for`/`id`**, even when nesting the input inside the label — explicit association improves assistive-tech support.
+- **Prefer `aria-labelledby` over `aria-label` when a visible label exists**: avoids duplication, improves maintainability, and translates better.
+- **Reuse the same accessible name for hyperlinks that share an `href`.**
```
knowler: This is a more of a recommendation than a requirement.

```suggestion
- **Prefer to reuse the same accessible name for hyperlinks that share an `href`.**
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,451 @@
...
+- **Be deliberate about `disabled` vs `aria-disabled`**: `disabled` removes the element from the focus order entirely (and `tabindex="0"` won't bring it back), which is often wrong for toolbar buttons or links. `aria-disabled="true"` keeps the element focusable so users can land on it and learn it's disabled.
+
+#### DON'Ts
+- **Don't use ARIA when native HTML exists**: Avoid `<div role="button">` or `<a role="button">` if `<button>` works.
+- **Don't add redundant ARIA roles or properties**: Avoid `<ul role="list">`, `<nav role="navigation">`, or `<input required aria-required="true">`.
+  - **Caveat**: Safari removes list semantics from `<ul>`/`<ol>` outside `<nav>` when `list-style: none` or `display: flex`/`grid` is applied. In that case `role="list"` is required to restore them. A similar effect impacts `<table>` with flex/grid across more browsers.
+- **Don't assume custom elements have no ARIA**: Custom elements can attach ARIA via `ElementInternals`, which some automated test tools can't see — so the absence of `role`/`aria-*` attributes in markup doesn't prove the element has no semantics. Verify with a screen reader or the browser's accessibility-tree inspector.
```
knowler: We should prefer the accessibility-tree inspector for this (testing with a screen reader is still important, but it could cause some false negatives).

```suggestion
- **Don't assume custom elements have no ARIA**: Custom elements can attach ARIA via `ElementInternals`, which some automated test tools can't see — so the absence of `role`/`aria-*` attributes in markup doesn't prove the element has no semantics. Verify with the browser's accessibility-tree inspector.
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,463 @@
...
+- **Accessibility is the minimum, not the ceiling.** Conformance to standards is the floor; aim for genuine usability.
+- **Patterns are use-case specific.** No checklist replaces real testing — including testing with disabled users — to confirm a given implementation is actually accessible in context.
+
+## 1. Content Navigability and Structure
+
+### Actionable Guidelines
+
+#### DOs
+- **Place all content within landmarks**: Wrap the page in `<header>`, `<nav>`, `<main>`, `<aside>`, and `<footer>` so assistive-tech users can jump between regions.
+- **Structure main content with headings**: Use `<h1>`–`<h6>` sequentially (no jumping `<h1>` → `<h4>`) so screen-reader users get a navigable outline.
```
knowler, line 21:
```suggestion
- **Structure main content with headings**: Use `<h1>`–`<h6>` sequentially (no jumping `<h1>` → `<h4>`) so screen-reader users get a navigable outline.
- **Avoid using headings with `<details>` and `<summary>`**: If a heading is important for understanding or navigating a content structure, avoid placing the heading within `<summary>` or as a part of the `<details>` contents, as they may not be accessible using common navigation methods (i.e. navigating by heading).
  - If a heading must be used as a part of a disclosure trigger, use a more robust disclosure approach that allows wrapping the disclosure trigger with the heading (e.g. `<h2><button type="button" aria-expanded="false" aria-controls="significant-section-content">Significant section</button></h2>`). This ensures the heading semantics aren’t lost, and the button and its state are announced.
```

For example, in JAWS a heading within `<summary>` cannot be navigated to using the very popular methods of navigating by heading. They aren’t shown in its heading list either. For headings within the details contents, these cannot be navigated to unless the details element is open.

See my comment on the HTML skill for more info about how the details open/closed state is communicated when navigating by heading: https://github.com/GoogleChrome/guidance/pull/609#discussion_r3164222226

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,463 @@
...
+- **Use lists for repeated, contiguous content**: `<ul>`/`<ol>` give assistive tech a count up front and let users skip the entire group.
+- **Provide skip links**: Add a "Skip to content" link at the top of the page so keyboard users can bypass repeated content like site headers or long/infinite lists. Make sure the target is focusable (e.g. `<main id="main" tabindex="-1">`).
```
knowler: Some rewording so the advice isn’t so focused on one use case (i.e. the “Skip to content” link at the start of the page — skip links can be within content as well).

```suggestion
- **Provide skip links** prior to repeated content like site headers with navigation or long/infinite lists, so that keyboard users can easily bypass them. Make sure the target is focusable (e.g. `<main id="content" tabindex="-1">`).
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,463 @@
...
+```
+
+## 12. Testing Validations
+
+### Actionable Guidelines
+
+#### DOs
+- **Run Automated checks via axe-core or Lighthouse audits**: Catch missing alt texts or low contrasts (e.g., via Lighthouse in Chrome DevTools MCP).
+- **Validate Sequential Navigations using keyboards alone**: Using only keyboard shortcuts, such as Tab/Shift+Tab, arrow keys, Enter, Space, and Esc, confirm every interactive element is reachable and operable, and that focus never gets stuck.
+- **Test on Screen Readers with calibrated browsers**: Rely on standard bindings (e.g., VoiceOver with Safari).
```
knowler: JAWS, NVDA, and iOS VoiceOver are the most popular screen readers.<sup><a href="https://webaim.org/projects/screenreadersurvey10/#primary">1</a></sup> <sup><a href="https://webaim.org/projects/screenreadersurvey10/#mobilescreenreaders">2</a></sup>

```suggestion
- **Test on Screen Readers with calibrated browsers**: Rely on standard bindings (e.g., JAWS with Chrome, NVDA with Firefox, Narrator with Edge, VoiceOver with Safari on macOS and iOS, TalkBack with Chrome for Android).
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,463 @@
...
+  <button onclick="extendSession()">Extend Session</button>
+</div>
+```
+
+## 9. Color, Contrast, and Typography
+
+### Actionable Guidelines
+
+#### DOs
+- **Minimum contrast standards**: Maintain 4.5:1 for normal text and 3:1 for large text or icons.
```
knowler, line 346: This guidance focuses on text and icons, but there are minimum contrast requirements for distinguishing user interface components as well. The relevant WCAG 2.2 success criterion is [1.4.11 Non-text Contrast (Level AA)](https://www.w3.org/TR/WCAG22/#non-text-contrast).<sup>1</sup>

For the sake of this guidance, we should advise that:

- The visual elements (e.g. borders, backgrounds, box-shadows, underlines) used to form the visual boundary between a user interface component and its context or that are used to indicate the presence of a user interface component should have a minimum contrast ratio of 3:1 (e.g. the border around a text field and the background of the view it’s included within; the underline for a link and the background).
- The visual elements used within a user interface component to indicate state should also have a minimum contrast ratio of 3:1 (e.g. a checkbox’s checkmark on inner fill of the box; the thumb of a switch against its track).

A challenge with this guidance is that it doesn’t easily produce good design, especially for more minimalistic design paradigms (it can come across as being brutalistic). In many cases the only solution is to introduce more complicated design paradigms (e.g. using gradients, inset/outset shadows) to help soften the increased contrast.

Many websites and, even some fairly robust and otherwise accessible, design systems tend to fail this SC.

- - -

<sup>1</sup> [Understanding SC 1.4.11 Non-text Contrast (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html)

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+
+#### DON'Ts
+- **Don't use ARIA when native HTML exists**: Avoid `<div role="button">` or `<a role="button">` if `<button>` works.
+- **Don't add redundant ARIA roles or properties**: Avoid `<ul role="list">`, `<nav role="navigation">`, or `<input required aria-required="true">`.
+  - **Caveat**: Safari removes list semantics from `<ul>`/`<ol>` outside `<nav>` when `list-style: none` or `display: flex`/`grid` is applied. In that case `role="list"` is required to restore them. A similar effect impacts `<table>` with flex/grid across more browsers.
+- **Don't assume custom elements have no ARIA**: Custom elements can attach ARIA via `ElementInternals`, which some automated test tools can't see — so the absence of `role`/`aria-*` attributes in markup doesn't prove the element has no semantics. Verify with the browser's accessibility-tree inspector.
+- **Don't place headings inside `<summary>`, and avoid relying on headings inside `<details>` content**: Headings inside `<summary>` are hidden from screen-reader heading lists and heading-navigation shortcuts entirely; headings inside `<details>` content are only reachable via heading navigation when the disclosure is open.
```
knowler: The heading discoverability issue isn’t universal for screen readers (e.g. with NVDA it seems to be fine—but I wouldn’t mention specific support for screen readers in the guide), it’s just an area that tends to have problems. The attitude should be more that **because headings and heading structure play such a vital navigation role for screen reader users, developers should be careful with how they use them** (e.g. placing headings for significant topics/sections of the page within `<summary>` is to be avoided because they’re very important for understanding the shape of the content; including headings for FAQ questions in `<summary>` or sub-question headings within the details contents is likely fine, because the contents of FAQ pages tends to be less cohesive).

```suggestion
- **Don't place headings inside `<summary>`, and avoid relying on headings inside `<details>` content**: Headings inside `<summary>` may be hidden from screen-reader heading lists and heading-navigation shortcuts entirely; headings inside `<details>` content are only reachable via heading navigation when the disclosure is open.
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+  - **Caveat**: If a heading must act as a disclosure trigger, wrap a `<button>` (with `aria-expanded` and `aria-controls`) in the heading instead, e.g. `<h2><button type="button" aria-expanded="false" aria-controls="…">…</button></h2>`.
```
knowler: Reword to avoid mentioning implementation details and focus on the prescribed patterns instead.

```suggestion
  - **Caveat**: If a heading must act as a disclosure trigger, use a more robust alternative to `<details>`/`<summary>` instead, e.g. an accordion or a disclosure implemented with ARIA where the heading wraps the button.
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+- **Match ARIA implementations to actual behavior**: If you set `role="tab"`, the element must behave like a tab — including keyboard interactions. Many ARIA patterns can't be implemented in CSS alone and need JavaScript.
+- **Be deliberate about `disabled` vs `aria-disabled`**: `disabled` removes the element from the focus order entirely (and `tabindex="0"` won't bring it back), which is often wrong for toolbar buttons or links. `aria-disabled="true"` keeps the element focusable so users can land on it and learn it's disabled.
+
+#### DON'Ts
+- **Don't use ARIA when native HTML exists**: Avoid `<div role="button">` or `<a role="button">` if `<button>` works.
+- **Don't add redundant ARIA roles or properties**: Avoid `<ul role="list">`, `<nav role="navigation">`, or `<input required aria-required="true">`.
+  - **Caveat**: Safari removes list semantics from `<ul>`/`<ol>` outside `<nav>` when `list-style: none` or `display: flex`/`grid` is applied. In that case `role="list"` is required to restore them. A similar effect impacts `<table>` with flex/grid across more browsers.
```
knowler: Let’s drop this bit. I believe the issue has to do with `<tr>`, but I’d want to verify it before including it in the guide.

```suggestion
  - **Caveat**: Safari removes list semantics from `<ul>`/`<ol>` outside `<nav>` when `list-style: none` or `display: flex`/`grid` is applied. In that case `role="list"` is required to restore them.
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+- **Don't assume custom elements have no ARIA**: Custom elements can attach ARIA via `ElementInternals`, which some automated test tools can't see — so the absence of `role`/`aria-*` attributes in markup doesn't prove the element has no semantics. Verify with the browser's accessibility-tree inspector.
+- **Don't place headings inside `<summary>`, and avoid relying on headings inside `<details>` content**: Headings inside `<summary>` may be hidden from screen-reader heading lists and heading-navigation shortcuts entirely; headings inside `<details>` content are only reachable via heading navigation when the disclosure is open.
+  - **Caveat**: If a heading must act as a disclosure trigger, use a more robust alternative to `<details>`/`<summary>` instead, e.g. an accordion or a disclosure implemented with ARIA where the heading wraps the button.
```
knowler: I do think this would be better in the Content Navigability & Structure section as that’s the reason for the guidance.

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+  width: 1px !important;
+  height: 1px !important;
+  margin: -1px !important;
+  padding: 0 !important;
+  border: 0 !important;
+  white-space: nowrap !important;
+}
+```
+
+When the hidden content is focusable (skip links, focus-receiving wrappers), the `:focus-within`/`:active` exception lets it become visible. Style the visible state per situation — a skip link typically wants fixed positioning at the top-left of the viewport so the rest of the page doesn't shift.
```
knowler:
```suggestion
When the hidden content is focusable (skip links, focus-receiving wrappers), the `:focus-within`/`:active` exception lets it become visible. Style the visible state per situation, e.g. a skip link to the main content typically wants fixed positioning at the top-left of the viewport so the rest of the page doesn't shift.
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+</html>
+```
+
+## 5. Keyboard and Focus Management
+
+### Actionable Guidelines
+
+#### DOs
+- **Logical Tab Order**: Ensure tab order matches visual layouts (top-to-bottom).
+- **Visible Focus Indicators**: Always style `:focus-visible` states explicitly. If disabling defaults, provide high-contrast overrides.
```
knowler: High-contrast has a particular meaning, so let’s avoid it here.

```suggestion
- **Visible Focus Indicators**: Always style `:focus-visible` states explicitly. If disabling defaults, provide overrides with sufficient contrast.
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+- **Skip Navigation Links**: Provide a "skip to content" link at the top of the page, prior to repeated content or long lists so keyboard users can easily bypass them.
```
knowler: Duplicate. We have this in the first section.

```suggestion
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+### Actionable Guidelines
+
+#### DOs
+- **Informative Visual Descriptions**: Describe the purpose of the image (e.g., "Search", not "Magnifying glass").
+- **Empty Alt properties for decorative visuals**: Use `alt=""` or `role="presentation"` to remove decorative images from the accessibility tree so they aren't announced.
+- **Synchronous Captions for videos**: Supply WebVTT captions for video tracks.
+- **Transcripts for audio**: Provide text transcripts for purely audio podcasts.
+- **Informative View Descriptions for inline SVGs**: Apply `role="img"` and a nested `<title>` tag for informative visuals.
+- **Decorative SVGs removal**: Apply `aria-hidden="true"` to remove decorative SVGs from reading flows.
+- **Long descriptions for complex images**: Use `<figure>`/`<figcaption>` or `aria-describedby` for charts and infographics.
```
knowler, line 224: We could also suggest providing data tables as alternatives for charts and other data visualizations.

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+
+#### DON'Ts
+- **Don't use clichéd prefixes**: Avoid "Image of..." or "Picture of...".
+- **Don't use underscores in filenames**: Use dashes if the filename might be announced as fallback.
+
+### Code Examples
+
+```html
+<!-- Decorative -->
+<img src="divider.png" alt="" role="presentation">
```
knowler:
```suggestion
<img src="divider.png" alt="">
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+}
+```
+
+## 6. Alternate Text and Media
+
+### Actionable Guidelines
+
+#### DOs
+- **Informative Visual Descriptions**: Describe the purpose of the image (e.g., "Search", not "Magnifying glass").
+- **Empty Alt properties for decorative visuals**: Use `alt=""` or `role="presentation"` to remove decorative images from the accessibility tree so they aren't announced.
```
knowler: For `` I’d almost always use `alt=""`.

```suggestion
- **Empty Alt properties for decorative visuals**: Use `alt=""` to remove decorative images from the accessibility tree so they aren't announced.
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+  </div>
+</details>
+```
+
+### Content Visibility Decision Matrix
+
+| Intent | Visual | Screen Reader | Focusable | Structural Pattern |
+| :--- | :--- | :--- | :--- | :--- |
+| **Visible to all** | Yes | Yes | Yes | Standard rendering |
+| **Screen Reader only** | No | Yes | Yes (if interactive) | `.visually-hidden` (`clip-path`) |
```
knowler:
```suggestion
| **Screen Reader only** | No | Yes | Yes (if interactive) | Visually hidden utility (e.g. `.visually-hidden`) |
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+Modern browsers provide native mechanisms for focus trapping and modal overlays that bypass the need for heavy manual JavaScript event tracking.
+
+### Actionable Guidelines
+
+#### DOs
+- **Use the Native `<dialog>` Element**: Invoke the dialog using the `.showModal()` method to automatically lock focus into the popup and dim the background.
+- **Use the `inert` Attribute for Custom Overlays**: When `<dialog>` doesn't fit (e.g., non-modal overlays, framework constraints, or layouts where `<dialog>`'s top-layer/positioning behavior conflicts with the design), apply `inert` to background app-shells to isolate focus.
+
+#### DON'Ts
+- **Don't rebuild focus traps manually**: Avoid using manual `keydown` listeners for focus loops.
```
knowler:
```suggestion
- **Don't implement focus traps for native modal dialogs**: When a `<dialog>` element is opened in a modal state, browsers set outside content as inert which is sufficient for ensuring only the dialog’s content can be focused.
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+```
+
+## 11. Modals and Native Dialogs
+
+Modern browsers provide native mechanisms for focus trapping and modal overlays that bypass the need for heavy manual JavaScript event tracking.
+
+### Actionable Guidelines
+
+#### DOs
+- **Use the Native `<dialog>` Element**: Invoke the dialog using the `.showModal()` method to automatically lock focus into the popup and dim the background.
```
knowler:
```suggestion
- **Use the Native `<dialog>` Element**: Invoke the dialog using the `.showModal()` method to open it in a modal state. When in a modal state, the browser sets outside content as inert (i.e. the outside content is hidden from the accessibility tree and cannot be interacted with nor be focused).
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+- **Use the `inert` Attribute for Custom Overlays**: When `<dialog>` doesn't fit (e.g., non-modal overlays, framework constraints, or layouts where `<dialog>`'s top-layer/positioning behavior conflicts with the design), apply `inert` to background app-shells to isolate focus.
```
knowler:
```suggestion
- **Use the `inert` Attribute for Custom Overlays**: When `<dialog>` cannot be used (e.g., some non-modal overlays, framework constraints, or layouts where `<dialog>`'s top-layer/positioning behavior conflicts with the design), apply `inert` to outside content to ensure it cannot be interacted with by keyboard, pointer, or assistive technology. This requires structuring elements in such a way that the custom overlay is not a descendant of the element with `inert` set on it.
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+  .spinner {
+    animation: none;
+    opacity: 0.5;
+  }
+}
+```
+
+## 11. Modals and Native Dialogs
+
+Modern browsers provide native mechanisms for focus trapping and modal overlays that bypass the need for heavy manual JavaScript event tracking.
```
knowler:
```suggestion
Modern browsers provide native solutions for creating modal dialogs which avoid the need for focus traps, managing the accessibility of outside content, ensuring the content is on top, and dimming the background content — all of which can be error prone and require heavy JavaScript event tracking to maintain.
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+```
+
+## 5. Keyboard and Focus Management
+
+### Actionable Guidelines
+
+#### DOs
+- **Logical Tab Order**: Ensure tab order matches visual layouts (top-to-bottom).
+- **Visible Focus Indicators**: Always style `:focus-visible` states explicitly. If disabling defaults, provide overrides with sufficient contrast.
+- **Lock Modal Focus**: Ensure focus cannot leave open modal dialogs.
```
knowler:
```suggestion
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+  - **Caveat**: Meeting 3:1 non-text contrast can challenge minimalistic designs. Soft gradients or subtle inset/outset shadows can soften visual boundaries while satisfying accessibility requirements.
+- **Use multiple state indicators**: Do not denote success/errors ONLY with color. Use icons or text.
+- **Relative font size units**: Use `rem` or `em` for font sizes instead of `px`.
+- **Consistent or Start alignment**: Avoid `justify` alignment as it can be more difficult to read.
+- **Avoid long lines of text**: Cap paragraph blocks to a maximum of 80 characters width.
+- **Support user zoom preferences**: Allow users to resize text up to 200% without loss of content or functionality.
+- **Support light and dark color schemes**: Honor `@media (prefers-color-scheme: dark)` and pair it with the `color-scheme` CSS property so form controls, scrollbars, and other UA-rendered surfaces match.
+- **Use `prefers-contrast` only when warranted**: Reach for `@media (prefers-contrast: more)` when the design uses low-contrast accents (e.g., subtle borders, muted secondary text) that need to be reinforced; most sites that already meet baseline contrast won't need it.
+
+#### DON'Ts
```
knowler, line 359:
```suggestion
#### DON'Ts
- **Don't use color alone to indicate the presence of a user interface component or its state**: Use iconography and/or shape to help differentiate.
```

guides/accessibility/SKILL.md
```diff
@@ -0,0 +1,469 @@
...
+- **Don't use Ornate fonts**: Omit cursive typefaces for main reading content.
+- **Don't rely on all-caps for emphasis**: Prefer bolding for visual emphasis, and use `<em>`/`<strong>` when the emphasis is semantic.
+- **Limit emphasis overall**: Emphasis loses meaning when it's everywhere — apply it only where it changes how the content should be read.
+
+### Code Examples
+
+```css
+/* Good: Relative sizing and line caps */
+body {
+  font-size: 1rem;
```
knowler:
```suggestion
```

knowler (approved):
This is my final batch of changes.

rviscomi:
Thank you @knowler and @jamesnw for the extremely thorough review!

PR #442, opened 2026-03-29
deploy dashboard results to gh-pages

paulirish:
no more gcs login required. 

Okay, this is ready.

repro:

```
STATIC=true gd dashboard
http://localhost:8081/?source=static
```

And we're actually ready to basically delete pretty much all of server.js. I left a plan in this branch on doing that, but I figure we'll do that cleanup in a follow-up just to provide a little bit of transition.

the `gd backfill`  script remains pretty key and i fixed some more bugs with our historical data. it's now very resilient and makes all our historical stuff fit the shape of the model day `evals.json` etc

And of course, the other thing is that we run ` eval-view/generate-manifests.js`  right at server startup (or deploy). It generates a bunch of static files that essentially give us the lookups we needed from the dynamic web server.  we were doing a lightweight version of this already but now it's expanded a bit


eval-view/generate-manifests.js
```diff
@@ -0,0 +1,130 @@
...
+    const outputFilePath = path.join(outputDir, 'features_mapping.gen.js');
+    const jsContent = `// Automatically generated by generate-manifests.js. Do not edit.\nwindow.__featuresMapping = ${JSON.stringify(mapping, null, 2)};\n`;
+    fs.writeFileSync(outputFilePath, jsContent);
+    return mapping;
+}
+
+/** 
+ * Generates suites.gen.json based on folders in harness/results that contain evals.json.
+ */
+export function generateSuitesManifest(outputDir = '.', resultsSourceDir = resultsDir) {
```
micahjo7: should this generation merge the local results with the one on `guidance-dash`?

if not, when we deploy, it will overwrite the file with only the local results, right?

eval-view/server.js
```diff
@@ -4,8 +4,79 @@ import fs from 'fs';
...
+    try {
+      generateSuitesManifest('.');
+      console.log('✅ Static manifests generated.');
+    } catch (e) {
+      console.error('Failed to trigger manifest generation:', e);
+    }
+  }
+
+  // Create dist/dashboard directory to mimic deployment structure
+  const distDir = path.resolve('dist/dashboard');
```
micahjo7, line 27: nit: I noticed this folder is present in `guidance-dash`, we only need it locally, right?

i guess `deploy.ts` can be updated to omit this folder in the upload, or we can add dist to gitignore in eval-view?

eval-view/generate-manifests.js
paulirish: great catch!  fixed with a merge operation.

eval-view/server.js
paulirish: true!  done

micahjo7 (approved):

PR #453, opened 2026-03-31
delivery: ship the CLI via clone&install rather than vendored

paulirish:
update: the plan was: publish to npm  and the skill will use cli via `npx -y @scope/pkg ...`  but we're gonna skip publish given the size of our testing population.

now the skill tells the agent to do a global install. This is also how the chrome-devtools-mcp CLI skill does things.


guides/modern-web-use-cases/SKILL.md
```diff
@@ -23,12 +23,18 @@ Must use this skill:
 
 You can execute these tools using standard `node` from the command line from **any** directory. 
 
+### Prerequisite check: Ensure the binary is installed!
+If you find `modern-web` is not available out of the box in your path, you can use standard `pnpm link --global` inside `dist/skills-cli` to set it up:
```
micahjo7: maybe I'm not understanding how this works fully- does the agent have access to `dist/skills-cli`? isn't that only in the guidance repo?

paulirish: yup good call. sorry i was updating some text in RELEASING.md and this one was stale.

paulirish:
\o/

serving/skills-cli/build-dist.ts
micahjo7: nit: improve formatting of the feature names so they show up cleaner in the README?

```
const featureNamesCsv = allFeaturesSorted.map(f => `\`${f.name.replace(/</g, '&lt;')}\``).join(', ');
```

serving/lib/practices.ts
```diff
@@ -27,8 +28,12 @@ export function getUseCasesByCategory(category?: string): UseCase[] {
 export async function getGuide(useCaseId: string): Promise<string | null> {
   const useCase = USE_CASES.find((u) => u.id === useCaseId);
   if (!useCase) return null;
-
-  const guidesDir = path.resolve(import.meta.dirname, "../../build/guides");
+  let guidesDir = path.resolve(import.meta.dirname, "../build/guides");
```
micahjo7: nit (suggestion): name variables like `devGuidesDir` and `prodGuidesDir`?

serving/lib/practices.ts
```diff
@@ -27,8 +28,12 @@ export function getUseCasesByCategory(category?: string): UseCase[] {
...
+  if (!existsSync(guidesDir)) {
+    // We are in production: either Option A (../guides) or self-contained Option B (../../guides)
+    const selfContainedPath = path.resolve(import.meta.dirname, "../../guides");
+    guidesDir = existsSync(selfContainedPath) ? selfContainedPath : path.resolve(import.meta.dirname, "../guides");
```
micahjo7: why do we need both `../../guides` and `../guides`?

serving/lib/store.ts
```diff
@@ -2,8 +2,10 @@ import lancedb from "@lancedb/lancedb";
 import path from "path";
 import fs from "fs";
 
-// Data directory for LanceDB
-const DATA_DIR = path.resolve(import.meta.dirname, "../../.modern-web-data");
+let DATA_DIR = path.resolve(import.meta.dirname, "../vector_store");
+if (!fs.existsSync(DATA_DIR) && fs.existsSync(path.resolve(import.meta.dirname, "../../vector_store"))) {
```
micahjo7, line 8: nit: add comment for dev vs. prod

serving/skills-cli/template/README.md
```diff
@@ -25,14 +25,16 @@ gemini extensions install https://github.com/GoogleChrome/skills-alpha --auto-up
...
 * In VSCode, open the Command Palette (`Cmd+Shift+P` or `Ctrl+Shift+P`)
 * Select "Extensions: Install from Location..."
 * Navigate to the `skills-alpha` directory and select it.
 
-*Note: This is a local extension and will not be automatically updated.*
+Compatibility with VSCode forks: unknown. 
 
+---
+
+🎓
```
micahjo7: were you going to add something else here?

micahjo7:
we also need to update the agent guide validation to support the new CLI calls that the agent makes (`modern-web.cjs` --> `modern-web`)

https://github.com/GoogleChrome/guidance/blob/main/harness/agents/claude-code-agent.ts#L162
https://github.com/GoogleChrome/guidance/blob/main/harness/agents/codex-cli-agent.ts#L165
https://github.com/GoogleChrome/guidance/blob/main/harness/agents/gemini-cli-agent.ts#L139

I can do this in the next PR if not done in this one

paulirish:
super appreciate the review! good call outs. 

- agent guide validation now handles `modern-web`
- split `practices.ts` paths into clear `devGuidesDir` / `prodGuidesDir`
- added `store.ts` comments for local vs prod
- dropped the stray emoji in the readme template
- backtick-wrapped feature names in the publishing script for a cleaner generated readme

all tested and green locally ✌️

serving/lib/practices.ts
paulirish: because dumb reasons but.. probably not in this case. good catch

serving/lib/practices.ts
paulirish: thxx

serving/skills-cli/build-dist.ts
paulirish: i like

micahjo7 (approved):

PR #455, opened 2026-03-31
Baseline skill

rviscomi:
Skill file to teach coding agents

- about Baseline
- where to find project-level Baseline target configs
- how to interpret target configs
- using Widely Available by default
- how to compare targets and statuses

Note: This relies on the agent to do the target-status comparison. I've reverted the changes that introduced a new CLI command for the sake of simplicity. I've also reverted the "Baseline TODO" instructions for now so we can discuss whether agents should also handle TODOs in the code, which would require them to know about features' statuses.

Progress on https://github.com/GoogleChrome/guidance/issues/543


paulirish:
image

viscomi out here challenging the github PR page to a fight. 🤣

rviscomi:
Haha yeah I really don't love bundling web-features into the skill's script. Should we expose it through the `modern-web` CLI instead?

paulirish:
> Should we expose it through the modern-web CLI instead?

that makes a lot of sense to me. seems good.

I'd give some more thoughts but.. i literally can't view the SKILL.md in the GH ui.. without the tab freezing. 🤣 haha.  So.. Yeah i think that's a good sign that keeping this data in our deps is ideal.  :) 

aside... we need to get a dependabot rule to keep our web-features module up to date like.. frequently. (or some other similar solution without the PR noise)

rviscomi:
Ok lol I removed the huge file, try now

paulirish:
> Ok lol I removed the huge file, try now

I CAN SEE! \o/

serving/scripts/sync-baseline-features.ts
```diff
@@ -0,0 +1,85 @@
...
+ * for the automated agent skill folder.
+ * 
+ * Run via: `pnpm run sync:baseline` from workspace root.
+ */
+
+import fs from 'node:fs';
+import path from 'node:path';
+import { features } from 'web-features';
+
+interface FilteredFeature {
```
paulirish: dont love that we have this interface and the one in baseline-src and they're super similar and probably baseline has its own typedef that we could maybe use here?

serving/scripts/check-baseline-src.ts
```diff
@@ -0,0 +1,111 @@
...
+    return false; // Feature is not baseline yet
+  }
+
+  if (normalizedTarget.includes('widely available')) {
+    return status.baseline === 'high';
+  }
+
+  if (normalizedTarget.includes('newly available') || normalizedTarget === 'baseline') {
+    return status.baseline === 'low' || status.baseline === 'high';
+  }
```
paulirish: whats going on here? feels like something the getFeatureStatus method does.

serving/scripts/check-baseline-src.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (status.baseline !== 'high' || !status.baseline_high_date) return false;
+    return status.baseline_high_date <= dateMatch[1];
+  }
+
+  return false;
+}
+
+const args = process.argv.slice(2);
+if (args.length < 2) {
+  console.log('Usage: node check-baseline.js <command> <args...>');
```
paulirish: lol its my turn to ask about overlap ... i feel like serving/cli/baseline-status.ts has a bunch of this.

tbh i wrote it with the intention of it being dev-exposed.. not just for us.

so... wdyt about it as the CLI?  (and/or lets tweak it to cover your usecases)

or convince me this is better and we should delete that one. :)

skills-drafts/baseline/SKILL.md
```diff
@@ -0,0 +1,151 @@
...
+- Use the canonical ID for feature verification.
+- **Resolving Moved vs Split IDs:**
+  - **Moved:** Replace the stale ID with the single modern canonical ID directly in the source code.
+  - **Split:** Analyze the context of the code block. Determine which sub-feature(s) apply. Use the specific sub-feature ID(s).
+  - **Single ID per Comment (Recommendation):** If multiple split sub-features apply to the same code block, add a separate `// TODO(baseline/feature-id)` comment for each (stack them). Do **NOT** use comma-separated lists inside `TODO` comments (e.g., `// TODO(baseline/id1,id2)`) as it breaks 1:1 compatibility with CLI tools.
+
+### Data Synchronization & Maintenance
+
+The skill relies on a static `features.json` snapshot to avoid runtime dependencies. To keep this data fresh as the web platform evolves:
+- Run the workspace sync command: `pnpm run sync:baseline`
```
paulirish: probably npm run. not pnpm.  

if only others were as hip as us.

skills-drafts/baseline/SKILL.md
```diff
@@ -0,0 +1,151 @@
...
+- This script pulls the latest `web-features` dataset (installed in `guides/node_modules`) and regenerates the paired-down snapshot inside the skill folder.
```
paulirish: buttttttttttttt also yeah i guess this is our docs and not developer-agent facing docs. righhhtttt?

guides/modern-web/SKILL.md
```diff
@@ -21,7 +21,19 @@ Must use this skill:
...
+
+Before you can check if a feature satisfies the target, you need to find the project's Baseline target:
+
+- **`baseline.config.json`**: Check for a `"baseline_target"` property in this file at the project root. This takes precedence over all other sources.
+- **Browserslist**: Check for `baseline` queries in `.browserslistrc` or `package.json` (e.g., `baseline widely available`).
+- **Linters**: Check linter configurations (e.g., `eslint.config.js`, `stylelint.config.js`) to see if they explicitly specify a Baseline target.
+
+Do NOT attempt to interpret or map general browser support matrices (like `> 1%` or `last 2 versions`) to Baseline equivalents.
+
+If no explicit configuration is found, default to `'widely available'`.
```
malchata: Does `'widely available'` here relate to a Browserslist config, or is this the working syntax for `baseline.config.json`?

guides/modern-web/SKILL.md
```diff
@@ -41,20 +53,48 @@ node <modern-web-directory>/modern-web.mjs search "<query>"
...
 
+### Step 4. Verify Baseline Satisfaction
+
+Compare the project's Baseline target (from Step 1) with the Baseline status of **each feature** recommended in the guide (typically found in the **Fallback strategies** section).
+
+#### Project Target Types
+Project targets can take several forms:
+- **Relative**: `widely available`, `newly available`, `limited availability`
+- **Yearly**: `baseline 2024` (Available by end of 2024)
+- **Date-specific**: `baseline widely available on 2027-01-01` (Newly available 30 months prior)
```
malchata: These bullet points feel in conflict with one another in terms of syntax—some are fully qualified (e.g., contain `baseline` in the config value like `baseline 2024`, while the first bullet point doesn't contain `baseline` in them. I'm sure this is by design, but poking at it in case it's an incosistency.

malchata (approved):
A couple questions, nothing to get in a twist over. Otherwise it seems LGTM.

guides/modern-web/SKILL.md
rviscomi: It's the agent's conceptual interpretation of the project's Baseline target. I updated it to `Widely Available` to align with the table in Step 4 and added `(default)` there to make it clearer.

guides/modern-web/SKILL.md
rviscomi: Good call, updated

guides/modern-web/SKILL.md
```diff
@@ -21,7 +21,19 @@ Must use this skill:
 
 ## Usage Instructions
 
-### Step 1. Search Use Cases
+### Step 1. Determine Project Target
```
paulirish, line 24: I’m hesitant to land the current Baseline Target integration as-is. I appreciate everyone hearing me out on these topics and apologize for the churn and drama. Anyway...


Our current conflict is that static tools (like ESLint) treat Baseline as a **ceiling**—blocking features newer than the target—while our guidance treats it as a **floor** for determining when to drop fallbacks. Trying to force one identifier to do both doesn't really serve developers well.

Here’s the alternative approach I’m pitching:

### Default to Widely Available for production safety
The skill should explicitly state that its default is **Widely Available**. 
* It should explain this optimizes for the optimal  browser support out of the gate. (etc etc)
* It should explain it follows expert guidance fallback scenarios, including progressive enhancement and polyfills (when appropriate).
* We’re providing a high-confidence floor by default, so developers can use modern features without faffing about with manual compatibility checks.

### Reactive discovery of non-default browser-support configuration
Instead, the agent should act reactively when it identifies specific project constraints.
* If the agent sees indicators of a browser monoculture (like **Electron** or **Tauri**)
* If the developer mentions clear support exclusions (e.g., "we don't need to support X Browser")…
* If the developer shares hesitations about polyfill invasiveness…

All of these are triggers for the agent to ask the user to clarify their support story. Being reactive to this  keeps the interaction focused on the task at hand rather than an upfront configuration wizard.

### Persistence via agents.md
When the agent reactively identifies a non-default support policy, it should suggest persisting that constraint directly into **`AGENTS.md`**.
* This avoids the overhead of a rigid JSON configuration.  The "policy" is free-form text.
* Storing these constraints as project context solves the floor/ceiling conflict by letting the agent interpret the support matrix within the specific use case. 
* (Yes it means the same "config" cannot be shared by other tools.  A bummer. ... I'm honest I think that's preferable -- but this is a longer discussion)
* Once it's in `agents.md`, all subsequent interactions benefit from that project memory without repeated questioning. \o/

paulirish (changes requested):

rviscomi:
Closing this and handing over to @paulirish to move it forward

LeaVerou:
Agreed 100% with everything @paulirish wrote above. One question though:

> Instead, the agent should act reactively when it identifies specific project constraints.
> - If the agent sees indicators of a browser monoculture (like Electron or Tauri)
> - If the developer mentions clear support exclusions (e.g., "we don't need to support X Browser")…

Given that `BASELINE_STATUS` currently prints out summary info like e.g. "Temporal has limited availability.", how are agents supposed to get browser support info for limited availability features? Web search? Or do we provide something for that?

PR #464, opened 2026-04-01
Add standardized launch commands for base apps

formgit:


harness/base_apps/cards-app/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "cards-app",
+  "version": "1.0.0",
+  "scripts": {
+    "start": "pnpm dlx serve -p 8080"
```
paulirish: can you use `statikk` instead of `serve` ? 

also.. lets pick different ports. 8080 is everywhere and i get conflicts personally.

how about like

11221
11222
11223

for these 3

formgit: makes sense, done

paulirish (approved):

PR #465, opened 2026-04-01
Gemini CLI custom policy

rviscomi:
Adding a policy file for Gemini CLI so that it allows agents to execute commands needed by our skill while in Plan Mode, which is restricted to read-only tools by default.

https://geminicli.com/docs/cli/plan-mode/#custom-policies
https://github.com/gemini-cli-extensions/conductor/blob/main/policies/conductor.toml


paulirish (approved):
cool! danke

Issue #469, opened 2026-04-02
Fix path resolution failure in eval


paulirish:
https://screenshot.googleplex.com/5fX5neV8izUz9b5
https://googlechrome.github.io/guidance/dashboard.html?testId=debug1&source=remote
https://googlechrome.github.io/guidance/dashboard.html?testId=debug2&source=remote

paulirish:
(obviously i didnt write this, but did guide it with some context)

# Path Resolution Analysis

Investigation of the "path resolution" problem revealed that the `guides/` and `vector_store/` directories are not correctly positioned relative to the `modern-web.mjs` CLI in all execution modes.

## Summary of Execution Modes

| Mode | Entry Point Execution Path | Guides Location (relative) | Vector Store Location (relative) | Status |
| :--- | :--- | :--- | :--- | :--- |
| **Dev Mode** (Direct) | `serving/bin/modern-web.ts` | `../build/guides/` | `../vector_store/` | ✅ Works |
| **Dist Mode** (Real World) | `dist/skills-cli/skills/modern-web-use-cases/modern-web.mjs` | `../../guides/` | `../../vector_store/` | ✅ Works |
| **Suite Mode** (`gd eval`) | `.../.gemini/jetski/skills/modern-web-use-cases/modern-web.mjs` | `../../guides/` (MIA) | `../../vector_store/` (MIA) | ❌ Fails |

---

## Detailed Folder Structures

### 1. Dev Mode (Direct Execution)
Used by SMEs and during unit tests.

```
guidance/
├── guides/ (original source)
├── serving/
│   ├── bin/
│   │   └── modern-web.ts
│   ├── build/
│   │   └── guides/ (built markdown files)
│   ├── lib/
│   │   ├── practices.ts (resolves guides to ../build/guides)
│   │   └── store.ts (resolves vector_store to ../vector_store)
│   └── vector_store/ (lancedb database)
```

### 2. Dist Mode (Real World / Installed `skills-cli`)
The structure of the package generated for distribution.

```
dist/skills-cli/
├── package.json
├── README.md
├── guides/ (copied from serving/build/guides)
├── vector_store/ (copied from serving/vector_store)
└── skills/
    └── modern-web-use-cases/
        ├── modern-web.mjs (bundled CLI)
        └── SKILL.md
```
> [!NOTE]
> The bundled `modern-web.mjs` resolves `guides/` at `../../guides/` and `vector_store/` at `../../vector_store/` successfully.

[...]

PR #475, opened 2026-04-03
Introduce installation tests for Claude, Gemini, and Skills CLI

paulirish:
Introduces a suite of integration tests to verify that the skills package can be installed and used locally across Claude Code, Gemini CLI, and the skills CLI.

- **New Tests**: Added `install-claude.test.ts`, `install-gemini.test.ts`, and `install-skills.test.ts` in `serving/skills-cli/`.  Verifies that the agents can actually find and use the skill (they run real prompts and answer using the guide).
- **Harness Utility Verification**: Added `copy-skills.test.ts` to verify that the internal `copySkills` utility correctly isolates skills and data, and to confirm that the evaluation runner for Gemini (`gemini-cli-agent.ts`) works end-to-end.  This is a regression test for #469 
- **Publishing Tweaks**: Updated the `publish-skills.ts` tip to suggest running the full pre-flight suite.

These are really heavy tests. So I put them up behind an environment variable: 
`env FULL=1 node --test serving/skills-cli/install-* harness/tests/copy-skills.test.ts`


micahjo7 (approved):
thank you!!

PR #477, opened 2026-04-03
Support for multiple tasks per guide

micahjo7:
The eval harness is updated to handle running more than one task per guide.
Ways to run evals:

### Single Guide Eval (to run `guides/performance/batch-analytics-events/tasks/task.md`)
`gd eval guides/performance/batch-analytics-events/tasks/task.md`
`gd eval batch-analytics-events`
`gd eval batch-analytics-events/task`

Any of these will work.

Once we have more tasks per guide (e.g. `guides/performance/batch-analytics-events/tasks/task2.md`), they can be run with:

`gd eval guides/performance/batch-analytics-events/tasks/task2.md`
`gd eval batch-analytics-events/task2`

### Negative Task (runs against guide's `negative-demo.html`)
`gd eval batch-analytics-events/negative`

*Putting any of these paths/identifiers in `tasks` within `config.ts` will also work.*

### Eval Runner UI

This PR also introduces a UI for setting config, tasks, and kicking off the evals. This is currently behind a flag and can be run with:

`gd eval --ui`

image

This will make it easier to run suite evals and select the tasks that we want to run.


harness/run_suite.ts
```diff
@@ -130,34 +129,50 @@ export async function runSuite(options: RunSuiteOptions = {}) {
...
       const tasksToRun = options.tasks && options.tasks.length > 0
         ? options.tasks
         : (suiteConfig.tasks.length > 0
           ? suiteConfig.tasks
-          : Array.from(taskMap.keys()));
+          : Array.from(taskMap.keys()).filter(key => key.endsWith('/task')));
 
       for (const task of tasksToRun) {
-        const taskInfo = taskMap.get(task);
+        let resolvedTask = task;
```
paulirish: Can you do me a favor and extract some of this function into some smaller methods?  It's getting pretty big.  
Not really your fault, but While we're making changes I just like to tidy it up a bit.

paulirish (approved):

harness/run_suite.ts
micahjo7: thanks! good call, updated

eval-view/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "start": "node generate-mapping.js && node server.js",
     "test:e2e": "playwright test",
-    "typecheck": "tsc server.js --allowJs --checkJs --noEmit --module nodenext --target esnext --skipLibCheck",
+    "typecheck": "pnpm --filter serving build && tsc server.js --allowJs --checkJs --noEmit --module nodenext --target esnext --skipLibCheck --allowImportingTsExtensions",
```
paulirish, line 8: we need to build serving before this?  cuz of build-guides i guess?

paulirish (approved):
thanks. the extracted methods look great :)

eval-view/package.json
micahjo7: yeah, was getting a typecheck error in CI, i think from adding `--allowImportingTsExtensions`

PR #488, opened 2026-04-06
only include logging to modern-web.log in local dist

micahjo7:


paulirish:
we  only want this in the harness right? 
if so i think we can skip the forPublish stuff and just set ENABLE_FILE_LOGGING in the env where we do the childprocess for our harness agents.

micahjo7:
right, thanks! updated :)

paulirish (approved):

PR #492, opened 2026-04-06
Support evals for discipline SKILLs

micahjo7:
## 1. Added `demo.html`, `expectations.md` to accompany the existing `forms` and `performance` SKILLs
*Note: generated with AI for the purposes of building/testing the `gd dev` and `gd eval` infrastructure, not reviewed for quality*

## 2. Updated `gd dev` workflow to support discipline SKILLs
It should be run with:
```
gd dev guides/<discipline>
```
This generated the `grader.ts`, `negative-demo.html`, and `tasks/task.md` files for both `forms` and `performance`.

## 3. Update harness to support running the discipline tasks
Run with:
```
gd eval guides/<discipline>/tasks/<task-name>.md
gd eval <discipline>/<task-name>
gd eval <discipline> (runs `<discipline>/tasks/task.md`)
```

Can also be run with `gd eval --ui`:

Screenshot 2026-04-06 at 4 07 59 PM

## 4. Update dashboard to show discipline eval results

In the overall metrics calculation, the discipline evals contribute to the `Tool Activation` percentage, but since we don't expect any guides, nothing is contributed to the `Guide Usage` metric.

image

Also adds a yellow highlight in guide evals when the corresponding discipline SKILL is used:

Screenshot 2026-04-06 at 5 37 37 PM


guides/forms/expectations.md
rviscomi, line 1: From a quick look, it seems like this is missing a ton of nuance in the skill file. Are we ok not eval'ing most of the guidance?

rviscomi (commented):
I have some doubts about the feasibility of creating a single demo and expectations file per high-level skill that faithfully captures most of the best practices. These skill files tend to cover A LOT of ground.

guides/forms/expectations.md
micahjo7: I've updated this PR to remove the intermediary files (demo, negative, expectations) for the Skills and updated the grader to be a stub impl with one simple test and a TODO to populate it.

This PR can focus on adding the infrastructure to test the evals, and we can address the content of the grader and the generation process in another PR.

rviscomi: Ok SG. If this PR will be more infra-focused I'll remove myself as a reviewer.

micahjo7:
Updated after changes made in https://github.com/GoogleChrome/guidance/pull/507. Some extra properties & logic was needed to calculate the correct guide usage #s. @paulirish

guides/performance/grader.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  test('Styles do not use @import', async ({ page }) => {
+    const hasImport = await page.evaluate(() => {
+      const styles = Array.from(document.querySelectorAll('style'));
+      return styles.some(s => s.innerText.includes('@import'));
+    });
+    expect(hasImport).toBe(false);
+  });
+
+  // TODO: Populate the grader with comprehensive tests based on SKILL.md
```
paulirish, line 42: you flagged it and i'm flagging it too.

paulirish (approved):

paulirish:
> I have some doubts about the feasibility of creating a single demo and expectations file per high-level skill that faithfully captures most of the best practices. These skill files tend to cover A LOT of ground.

i dont disagree. the value of evals for these is pretty low, but that doesnt mean the value of the skills is low. 🤷

LeaVerou:
> I have some doubts about the feasibility of creating a single demo and expectations file per high-level skill that faithfully captures most of the best practices. These skill files tend to cover A LOT of ground.

Ditto. It would be useful *in general* for the infra to be able to support multiple evals for the same guide, but it's particularly useful for skills.

That said, even without it, data from it can still be useful, if used with awareness of its limitations. E.g. if we see that a certain guideline is used in the unguided condition across all three agents, it can be safely removed, even if the evals don't cover all guidelines.

PR #496, opened 2026-04-07
Adds guide, demo, and negative demo for #365.

malchata:
Adds the guide, demo, and negative demo for the `VisibilityStateEntry` API use case on calculating total user foreground time for a page.


rviscomi (changes requested):
Could you also fill out the `expectations.md` file?

rviscomi (approved):
Dropped the eval files again otherwise LGTM

PR #501, opened 2026-04-07
docs: add guide and evals for `animate-element-entry-exit`

### LINKED ISSUE #229: Create guide and evals for the animate-element-entry-exit use case
Smoothly hide/show elements as they are added/removed from the DOM or as their display values are toggled.

Affected web-feature IDs: [starting-style](https://webstatus.dev/features/starting-style), [transition-behavior](https://webstatus.dev/features/transition-behavior)

Use case subdir: [guides/ui-behaviors/animate-element-entry-exit](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-behaviors/animate-element-entry-exit)

Related features: #35 #85

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

castastrophe:
## Description
This PR introduces a new guide for animating element entry and exit using modern web platform features.

The guide focuses on using `@starting-style` and `transition-behavior: allow-discrete` to create smooth entry and exit animations for elements that transition to/from `display: none`, without JavaScript animation libraries.

## Validation Results

Successfully ran the `gd dev` pipeline:
   - Calibration: Passed on Attempt 1.
     - demo.html: 100% pass (8/8 tests).
     - negative-demo.html: 0% pass (8/8 tests failed correctly).
   - Agent Test: Verified that the grader correctly identifies implementations following the guidance vs. unguided baselines.

## Checklist
   - [x] guide.md with citations and examples
   - [x] demo.html demonstrates the "Do" patterns in the guide
   - [x] grader generated and calibrated
   - [x] task.md generated

### DIFF
--- a/guides/user-experience/animate-element-entry-exit/demo.html
+++ b/guides/user-experience/animate-element-entry-exit/demo.html
@@ -124,6 +124,24 @@
         transform: translateY(-20px) scale(0.95);
       }
     }
+
+    /* Respect user preference for reduced motion */
+    @media (prefers-reduced-motion: reduce) {
+      .card {
+        transform: none;
+        transition-duration: 0.1s;
+      }
+
+      @starting-style {
+        .card {
+          transform: none;
+        }
+      }
+
+      .card:where([hidden], .hidden) {
+        transform: none;
+      }
+    }
   </style>
 </head>
 <body>
--- a/guides/user-experience/animate-element-entry-exit/guide.md
+++ b/guides/user-experience/animate-element-entry-exit/guide.md
@@ -4,4 +4,104 @@ description: Smoothly hide/show elements as they are added/removed from the DOM 
 web-feature-ids:
   - starting-style
   - transition-behavior
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/@starting-style
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/transition-behavior
+  - https://web.dev/blog/baseline-entry-animations
+  - https://www.smashingmagazine.com/2025/01/transitioning-top-layer-entries-display-property-css/
 ---
+
+In the past, CSS transitions could not animate elements when they were first added to the DOM or when their `display` property changed from `none`. The `@starting-style` at-rule and `transition-behavior: allow-discrete` provide a declarative way to create smooth entry and exit animations.
+
+## Implementation
+
+### 1. Animating `display: none` Toggles
+
+To animate an element when toggling its visibility via an attribute (e.g., `hidden` with `display: none`):
+
+1. **Define the visible state**: Set the final property values (e.g., `opacity: 1`) on the base class.
+2. **Define the entry starting state**: Use `@starting-style` to specify the values to transition *from* when the element becomes visible.
+3. **Enable discrete transitions**: Include `display` in the `transition` property and use the `allow-discrete` keyword.
+4. **Define the exit state**: Set the target values in the `hidden` attribute.
+
+```css
+.card {
+  display: block;
+  opacity: 1;
+  transform: translateY(0);
+  /* MANDATORY: Use allow-discrete for display transition */
+  transition:
+    display 0.4s allow-discrete,
+    opacity 0.4s ease-out,
+    transform 0.4s ease-out;
+}
+
+/* Entry animation: transition FROM these values when first rendered or display changes from none */
+@starting-style {
+  .card {
+    opacity: 0;
+    transform: translateY(-20px);
+  }
+}
+
+/* Exit animation: transition TO these values when the hidden class or attribute is present */
+.card:where(.hidden, [hidden]) {
+  display: none; /* display none is only needed for the class-based implementation as the hidden attribute applies it by default */
+  opacity: 0;
+  transform: translateY(-20px);
+}
+```
+
+### 2. Animating DOM Insertion and Removal
+
+For elements added via `appendChild()` or removed via `remove()`:
+
+- **Entry**: Use `@starting-style` as shown above. The browser will automatically detect the style change from "nothing" to the element's initial styles and trigger the transition from the `@starting-style` values.
+- **Removal**: Since `element.remove()` is instantaneous and doesn't trigger a CSS transition on its own, you must trigger the exit transition first (e.g., by adding a class) and wait for it to finish before removing the node from the DOM.
+
+```javascript
+// Trigger exit transition
+element.setAttribute('hidden', true);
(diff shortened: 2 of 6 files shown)


guides/user-experience/animate-element-entry-exit/negative-demo.html
```diff
@@ -0,0 +1,168 @@
...
+    <p>Toggling the <code>hidden</code> attribute which sets <code>display: none</code>.</p>
+    <div class="controls">
+      <button id="toggleBtn">Toggle Display</button>
+    </div>
+    <div class="container">
+      <div id="toggleCard" class="card">Item 1</div>
+    </div>
+  </div>
+
+  <div class="demo-section">
```
malchata, line 125: More a question for @rviscomi and @philipwalton—I see a lot of the generated demos use non-semantic HTML elements for containers. For example, I've seen `<div>` used as a main page container where `<main>` would be better. Is this something that the accessibility agent skill would be looking for?

malchata: Also noting that I posted this comment on the negative demo, where it probably wouldn't really matter (or even be used as future eval criteria for a "bad" demo), but the question still stands :)

guides/user-experience/animate-element-entry-exit/guide.md
```diff
@@ -4,4 +4,104 @@ description: Smoothly hide/show elements as they are added/removed from the DOM
...
+
+## Constraints & Accessibility
+
+- **MANDATORY**: Use `allow-discrete` (either via `transition-behavior: allow-discrete` or the `allow-discrete` keyword in the `transition` shorthand) when transitioning `display`. Without it, the element will instantly disappear during exit.
+- **MANDATORY**: Use `@starting-style` for entry animations. Browsers skip transitions on an element's first style update (initial render or `display: none` change) unless this is provided.
+- **DO**: Include `overlay` in the `transition` list if animating top-layer elements like `<dialog>` or `popover` to ensure they stay in the top layer during the exit animation.
+- **DO**: Respect user preferences for reduced motion using the `prefers-reduced-motion` media query.
+- **DO NOT**: Rely on `@starting-style` for exit animations; it only defines the *starting* point for an entry transition. Exit animations are defined by the transition to the hidden state.
+
+## Fallback strategies
```
malchata, line 104: @rviscomi ,the agent skill insists on this being a level 3 heading, but I don't think that would make sense here, because why would we want to nest it under `## Constraints & Accessibility` if fallback strategies were an entirely separate concern?

guides/user-experience/animate-element-entry-exit/guide.md
```diff
@@ -4,4 +4,104 @@ description: Smoothly hide/show elements as they are added/removed from the DOM
...
+// 3. Finally remove the node from the DOM
+element.remove();
+```
+
+## Constraints & Accessibility
+
+- **MANDATORY**: Use `allow-discrete` (either via `transition-behavior: allow-discrete` or the `allow-discrete` keyword in the `transition` shorthand) when transitioning `display`. Without it, the element will instantly disappear during exit.
+- **MANDATORY**: Use `@starting-style` for entry animations. Browsers skip transitions on an element's first style update (initial render or `display: none` change) unless this is provided.
+- **DO**: Include `overlay` in the `transition` list if animating top-layer elements like `<dialog>` or `popover` to ensure they stay in the top layer during the exit animation.
+- **DO**: Respect user preferences for reduced motion using the `prefers-reduced-motion` media query.
```
malchata, line 101: Using `prefers-reduced-motion` is mentioned here, but not shown in action in the code snippets in this guide.

guides/user-experience/animate-element-entry-exit/demo.html
malchata, line 1: `prefers-reduced-motion` is also not used in the demo—if this was generated, maybe modify your generation criteria to explicitly mention this, particularly if it's an accessibility constraint.

guides/user-experience/animate-element-entry-exit/guide.md
```diff
@@ -4,4 +4,104 @@ description: Smoothly hide/show elements as they are added/removed from the DOM
...
+
+## Fallback strategies
+
+{{ BASELINE_STATUS("starting-style") }}
+
+For browsers that do not support these features, elements will toggle `display: none` instantly. For smooth animations in older browsers, use JavaScript-based animation libraries or manually coordinate `display` toggles using `requestAnimationFrame` and `transitionend` events.
+
+### Manual Entry Animation (JS Fallback)
+
+```javascript
```
malchata, line 123: I think using `CSS.supports()` is worth considering.

malchata (changes requested):
Some comments to consider, but I do think some of these need to be changed. Gemini did suggest a number of these, and they seem sensible from my perspective.

castastrophe:
@malchata Thanks for the review, I'll make those updates today and push changes up.

guides/user-experience/animate-element-entry-exit/negative-demo.html
rviscomi: No it doesn't matter. The demo files are only used for calibrating the grader. See https://github.com/GoogleChrome/guidance/blob/main/.agents/skills/project-guides/SKILL.md for more info

guides/user-experience/animate-element-entry-exit/guide.md
rviscomi: Yeah, the skill hardcodes the H3 level, so it should be relaxed

> You **MUST** include a `### Fallback strategies` section regardless of Baseline status, as developers may have older baseline targets.

rviscomi (approved):
LGTM after some fine tuning

- I updated the prompt to specify the same IDs and class names the grader expected from the demo.
- Rewrote a chunk of the grader to use a more reliable animation spy rather than timeouts to remove flakiness

malchata (approved):

guides/user-experience/animate-element-entry-exit/guide.md
castastrophe: just pushed an update incorporating it

guides/user-experience/animate-element-entry-exit/demo.html
castastrophe: just pushed an update incorporating it

guides/user-experience/animate-element-entry-exit/guide.md
castastrophe: updated now

PR #504, opened 2026-04-07
CI for Publish, update dashboard deployment CI

micahjo7:
Combines the existing `publish-npm` workflow impl into a general `publish` workflow.

It currently calls the `publish-skills` which will build the `dist/skills-cli` with incremented version, and publish it to `skills-alpha`. Then, it will commit the version upgrade change directly into `main` so this repo stays up-to-date (skipping CI).

This is configured to run whenever any changes to:
```
- 'serving/skills-cli/**'
- 'guides/**'
- '.github/workflows/publish.yml'
```

are merged to `main`. The publish to NPM (when ready) would also trigger with the same `dist/skills-cli` version so `skills-alpha` and the NPM package are always in sync (is this something we want/need?).

In order for this workflow to work, the `GH_PAT` [key](https://github.com/GoogleChrome/guidance/settings/secrets/actions) or owner of that key needs write access to `guidance` and `skills-alpha`. The owner of the key also needs to be added to `Allow specified actors to bypass required pull requests` list within `Settings > Branches > main` Settings (so that the version bump can be committed to this repo.

Also updates the dashboard deployment so that it only runs when changes are made to:

```
- 'eval-view/**'
- 'guides/**'
- '.github/workflows/deploy-dashboard.yml'
```

instead of on every push to `main`.


.github/workflows/publish.yml
```diff
@@ -0,0 +1,76 @@
...
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write # Required for npm OIDC provenance when enabled
+
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          token: ${{ secrets.GH_PAT }}
```
paulirish: @rviscomi suggests we may not need a GH_PAT and can use our GH APP's auth to do this.

he said

> PAT would be someone's individually provisioned token. The GH app would be its own auth mechanism.

> Here's how we use the GH app in the sync use cases workflow: https://github.com/GoogleChrome/guidance/blob/main/.github/workflows/guide-sync.yml#L47-L53

and we might need to change something here: https://github.com/organizations/GoogleChrome/settings/apps/chrome-devrel-infra/permissions  but we can do that. :)

micahjo7: thanks! updated authenticate similarly to sync use cases.

and I've updated https://github.com/organizations/GoogleChrome/settings/apps/chrome-devrel-infra/permissions to support reading/writing to the repositories

rviscomi: I saw the permission request email and hit approve

paulirish (approved):
this is a very good looking PR.

PR #507, opened 2026-04-08
separate guide usage metrics

paulirish:
###  main focus: split out "guide usage" metrics" & improved reporting 
*   Differentiated between guides retrieved via tools and those **read directly from files**. Updated dashboard  to match.
*   Added tracking for 'generation errors' when a task straight up failed.

#### evals.json types and backfill. 

*   **Formal Types**: Added formal TypeScript interfaces for `evals.json` to keep things consistent across the pipeline (with loose/strict variants to handle legacy data).
*   Added backfill.ts to update evals.json to modern format.  Handling legacy stuff and upgrading it.   
*    Introduced a workflow to update our GCS results' evals.json safely.  Added `--summary-only` flag to `upload_suite.ts` to only push `evals.json`/`evals.md`. Details covered in `harness/eval-results.md` 

image

#### Drivebys
*   Fixed some lockfile stuff in `build-dist.ts` --  check if the lock-owning PID is still alive. it was buggy before
*   Better logging around build-dist/test-dist
*   nice gemini-cli-core types in gcli agent for trajectory parsing :)

### DIFF
--- a/eval-view/evals.d.ts
+++ b/eval-view/evals.d.ts
@@ -1,9 +1,6 @@
 import type { EvalsReport } from '../harness/lib/metrics.ts';
 
-export interface LooseEvalsReport extends Partial<Omit<EvalsReport, 'results'>> {
-  results?: Record<string, any>;
-  [key: string]: any;
-}
+
 
 declare global {
   interface Window {
--- a/eval-view/evals.d.ts
+++ b/eval-view/evals.d.ts
@@ -1,0 +1,29 @@
+import type { Metrics, RunResult as HarnessRunResult } from '../harness/lib/metrics.ts';
+
+export interface FullRunResult extends HarnessRunResult {
+  taskName?: string;
+  baseApp?: string;
+  prompt?: string;
+}
+
+export interface EvalsReport {
+  summary: Metrics['summary'];
+  results: Record<string, FullRunResult[]>;
+  stats: Metrics['testStats'];
+  timestamp: string;
+  runCount: number;
+  agent: string;
+  serving: string;
+  model: string;
+}
+
+export interface LooseEvalsReport extends Partial<Omit<EvalsReport, 'results'>> {
+  results?: Record<string, any>;
+  [key: string]: any;
+}
+
+declare global {
+  interface Window {
+    google: any;
+  }
+}
--- a/eval-view/evals.d.ts
+++ b/eval-view/evals.d.ts
@@ -1,21 +1,4 @@
-import type { Metrics, RunResult as HarnessRunResult } from '../harness/lib/metrics.ts';
-
-export interface FullRunResult extends HarnessRunResult {
-  taskName?: string;
-  baseApp?: string;
-  prompt?: string;
-}
-
-export interface EvalsReport {
-  summary: Metrics['summary'];
-  results: Record<string, FullRunResult[]>;
-  stats: Metrics['testStats'];
-  timestamp: string;
-  runCount: number;
-  agent: string;
-  serving: string;
-  model: string;
-}
+import type { EvalsReport } from '../harness/lib/metrics.ts';
 
 export interface LooseEvalsReport extends Partial<Omit<EvalsReport, 'results'>> {
   results?: Record<string, any>;
--- a/eval-view/tsconfig.json
+++ b/eval-view/tsconfig.json
@@ -13,6 +13,7 @@
     "lib": ["esnext", "dom", "dom.iterable"]
   },
   "include": [
+    // TODO: remove this 'include' block so we typecheck ALL of eval-view!
     "server.js",
     "utils.js",
     "evals.d.ts"
--- a/eval-view/tsconfig.json
+++ b/eval-view/tsconfig.json
@@ -13,7 +13,9 @@
     "lib": ["esnext", "dom", "dom.iterable"]
   },
   "include": [
-    "server.js"
+    "server.js",
+    "utils.js",
+    "evals.d.ts"
   ],
   "exclude": [
     "node_modules",
--- a/harness/agents/gemini-cli-agent.ts
+++ b/harness/agents/gemini-cli-agent.ts
@@ -117,6 +117,11 @@ async function run() {
   }
 }
 
+function readTrajectory(filePath: string): ConversationRecord {
+  const content = fs.readFileSync(filePath, 'utf8');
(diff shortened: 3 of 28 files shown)


eval-view/tsconfig.json
```diff
@@ -13,7 +13,10 @@
     "lib": ["esnext", "dom", "dom.iterable"]
   },
   "include": [
-    "server.js"
+    // TODO: remove this 'include' block so we typecheck ALL of eval-view!
```
micahjo7, line 16: nit: can we not do this now?

harness/agents/gemini-cli-agent.ts
```diff
@@ -111,40 +117,57 @@ async function run() {
...
+              } else if (tc.name === 'read_file' && tc.args && tc.args.file_path) {
+                const filePath = tc.args.file_path as string;
+                  if (filePath.includes('/skills/')) {
+                    if (filePath.endsWith('/guide.md')) {
+                      const match = filePath.match(/\/skills\/[^/]+\/([^/]+)\/guide\.md$/);
+                      if (match) {
+                        fileReadGuides.push(match[1]);
+                      }
+                    } else if (filePath.endsWith('.md')) {
+                      const match = filePath.match(/\/skills\/[^/]+\/(?:references\/)?(?:[^/]+\/)*([^/]+)\.md$/);
```
micahjo7: what is this path for?

harness/agents/gemini-cli-agent.ts
```diff
@@ -111,40 +117,57 @@ async function run() {
...
+          if (msg.type === 'gemini' && msg.toolCalls) {
             for (const tc of msg.toolCalls) {
-              if (serving === Serving.SKILLS && tc.name === 'read_file' && tc.args && tc.args.file_path) {
-                const filePath = tc.args.file_path;
-                if (filePath.includes('/skills/') && filePath.endsWith('/guide.md')) {
-                  const match = filePath.match(/\/skills\/[^/]+\/([^/]+)\/guide\.md$/);
-                  if (match) {
-                    guidesFromSkills.push(match[1]);
-                  }
+              if (tc.name.includes('get_best_practices')) {
```
micahjo7: so, there are two different ways that a retrieved guide could be captured in the session file (here and below in 163), and two different ways a file-read guide can be captured?

micahjo7 (approved):

eval-view/evals.d.ts
```diff
@@ -0,0 +1,29 @@
...
+  runCount: number;
+  agent: string;
+  serving: string;
+  model: string;
+}
+
+export interface LooseEvalsReport extends Partial<Omit<EvalsReport, 'results'>> {
+  results?: Record<string, any>;
+  [key: string]: any;
+}
```
micahjo7: would it be possible to do the following:

1. update `RunResult` in `metrics.ts` with:
```
export interface RunResult {
  runNumber: number;
  results: ScenarioCheck[];
  guidesUsed?: string[];
  guidanceToolsUsed?: string[];
  expectedToolPrefixes?: string[];
  guideName?: string;
  taskName?: string;
  baseApp?: string;
  prompt?: string;
}
```

2. remove `EvalsReport` from `reporting` and move it to `metrics.ts`

3. update `evals.d.ts` to be:
```
import type { EvalsReport } from '../harness/lib/metrics.ts';

declare global {
  interface Window {
    google: any;
  }
}
```

eval-view/tsconfig.json
paulirish: we could but it'd be a lot of diff noise. i'm happy to do in a a standalone followup. :)

harness/agents/gemini-cli-agent.ts
paulirish: this is the path used in my (unlanded) megaskill branch :)

eval-view/evals.d.ts
paulirish: love it. done https://github.com/GoogleChrome/guidance/pull/507/commits/5d4e7ba18e4bbc7a8fe2f2ba2bb197caf1cb06a4

harness/agents/gemini-cli-agent.ts
paulirish: yeah.. we need both ways but... the code was weird so i redid that. [`9935ad4` (this PR)](https://github.com/GoogleChrome/guidance/pull/507/commits/9935ad4cb9e24d77a30ccf78903a54e3e7c64186)

paulirish:
@micahjo7 fyi i just did my backfill on gcs things and did a dashboard deploy.   eval-results.md has details on how that works.

rviscomi:
_I hit an issue while trying to eval a use case and my coding agent traced it back to this PR and wrote up this nice, long bug report:_


Running the eval harness (`gd eval`) or the developer workflow (`gd dev`) fails immediately with a `ReferenceError`. This occurs because the project is configured as an ES Module (`"type": "module"` in `package.json`), where the CommonJS global `__dirname` is not available.

#### **Steps to Reproduce**
1. Run an evaluation or test a grader locally:
   ```bash
   ./bin/gd.ts dev guides/user-experience/improve-body-text-layout-and-legibility --test-grader
   ```
2. Observe the execution failure as Playwright attempts to load the shared configuration file.

---

#### **Error / Stack Trace**
```
ReferenceError: __dirname is not defined in ES module scope
    at file:///Users/rviscomi/git/guidance/guides/playwright.config.ts:10:12
    at ModuleJob.run (node:internal/modules/esm/module_job:413:25)
    at onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:660:26)
    at requireOrImport (/Users/rviscomi/git/guidance/node_modules/.pnpm/playwright@1.58.2/node_modules/playwright/lib/transform/transform.js:223:12)
```

---

#### **Root Cause**
The file `guides/playwright.config.ts` uses `__dirname` directly on line 10:
```typescript
// guides/playwright.config.ts
export default defineConfig({
  // ...
  testDir: __dirname,
  // ...
});
```
Since `package.json` defines `"type": "module"`, Node treats TypeScript execution as an ES module where `__dirname` does not exist.

---

#### **Recommended Solution**
Update `guides/playwright.config.ts` to derive the directory path using ES module-compatible utilities (`import.meta.url`):

```typescript
import { fileURLToPath } from 'node:url';
import * as path from 'node:path';

const __filename = fileURLToPath(import.meta.url);
const __dirname = path.dirname(__filename);
```

PR #517, opened 2026-04-09
Add passkey skills

agektmr:
Add passkey skills


rviscomi (approved):
Thanks @agektmr! LGTM

As we discussed the other day, the next step will be to flatten the separate markdown guides into a single SKILL.md file. We should also use the some of the project-level best practice skills to recommend opportunities to dedupe or reorganize the file to fit with existing conventions. After merging this, I can take a pass at it and send you a PR for review.

PR #519, opened 2026-04-09
fix playwright config ESM crash

paulirish:
ref: https://github.com/GoogleChrome/guidance/pull/507#issuecomment-4210986603 ..  (rick flagging that eval running failed because `__dirname` isn't defined in ESM scope.)

Switched it to `import.meta.dirname` in `playwright.config.ts` to keep the absolute path resolution but make it ESM-friendly. 

improved test coverage.. added regression test


paulirish:
@micahjo7 plz feel free to give me a review and i can do followups

micahjo7:
lgtm!

Issue #176, opened 2026-02-27
Create guide and evals for the persistent-top-layer-ui use case


paulirish:
Keep a modal dialog, fullscreen element, or native popover visibly open and functionally active when its underlying DOM node is moved or reparented in the DOM.

Affected web-feature IDs: [move-before](https://webstatus.dev/features/move-before)

Use case subdir: [guides/ui-behaviors/persistent-top-layer-ui](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-behaviors/persistent-top-layer-ui)

Related features: #66

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

Issue #175, opened 2026-02-27
Create guide and evals for the move-dom-element-without-losing-state use case


paulirish:
Move or reparent a DOM element without losing important element state, such as interactivity states (:focus/:active), `<iframe>` loading state, animation/transition state, etc

Affected web-feature IDs: [move-before](https://webstatus.dev/features/move-before)

Use case subdir: [guides/ui-behaviors/move-dom-element-without-losing-state](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-behaviors/move-dom-element-without-losing-state)

Related features: #66

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

PR #524, opened 2026-04-09
Add guides for persistent top layer

taraojo:
#176 Add guides for persistent top layer guide, including implementation demo, and verification tests.

There is an unusually long grader test for fallback behaviour that simulates two different browsers. The first part creates an environment where moveBefore exists, and second creates a browser environment where moveBefore is deleted to pretend it is an older browser. I've kept this as it seems useful to test the different scenarios.

Note: there are a lot of similarities in the guidance with this use case and https://github.com/GoogleChrome/guidance/issues/175 as they are the same feature, but it seems worth keeping both as the demos are quite different and additional info for handling top layer elements has been added.


rviscomi (approved):
LGTM

WAI in Chrome and Safari

PR #525, opened 2026-04-09
Passkey skill

rviscomi:
Flatten the passkey draft skill guides into a single SKILL.md, hopefully without losing any important nuance

I kept skills-drafts/passkeys/expectations.md under skills-drafts in case we want to repurpose it for eval'ing the skill at some point, intentionally leaving it out of guides/passkeys for now.

Progress on https://github.com/GoogleChrome/guidance/issues/542


rviscomi:
@agektmr are you able to take a look?

agektmr:
Hmm, I thought by flatten you meant not having multiple levels of folders. I don't want to merge markdown files into a single SKILL.md to be honest. I believe splitting files helps optimize the context consumption.

rviscomi:
The flattened skill is about 200 LOC, which has a reasonably small token footprint. But I take your point that branching out into sub-skills lets agents take only what they need. My biggest concern with that is the dropoff in discoverability, and unfortunately we don't really have a way to test that until we get skill evals set up.

In order to run that test though, we'd still need to validate that the contents of the skill file are technically correct and we didn't lose any important nuance from your drafts. Are you able to review the draft as-is and LMK if there are any factual inaccuracies or omissions?

agektmr:
I've tested this skill with the following prompts on Jetski with Gemini Next:
> Implement passkey registration

> Implement passkey authentication

> Implement passkey management

Here's the score I get with my evaluater:

Status | Count
-- | --
PASS | 36
FAIL | 32
N/A | 16

`N/A` means they are not implemented (reauthentication and conditional create).

For reference, here's the score without the skill (but I applied server side skill for convenience. Still required manual fixes on frontend though)

Status | Count
-- | --
PASS | 30
FAIL | 27
N/A | 17

Not a direct comparison, but here's the score when I applied the original skills to a more complex personal project (Note that the evaluater skills are not identical as I'm incrementally adding new stuffs):

Status | Count
-- | --
PASS | 55
FAIL | 2
N/A | 21

Other concern is, passkey skills include server side bits, but the suggested version only shows SimpleWebAuthn sample code. I'm not sure if this will work on non-JavaScript/TypeScript environment with completely different library. See my approach in the original skills.

rviscomi:
@agektmr after thinking more about this, a good next step would probably be to pull out any parts of the skill draft that would be a better fit as a use case-level guide. For example, after a quick audit, we could have guides for individual use cases like:

- Register a new passkey for a user
- Authenticate a user with a passkey
- Manage registered passkeys
- Re-authenticate a user for sensitive actions
- Conditionally create a passkey during sign-up
- Authenticate users using passkey autofill
- Signal credential state changes to the browser
- Identify the passkey provider from the credential

That will give each use case the breathing room to go into more detail, and we already have the infrastructure in place to run evals against frontend code. We'll have to figure out a way to eval the server-side code, but that doesn't have to block getting the guidance in.

Does that sound like a good plan?

agektmr:
The goal for me is to provide the skill in the same or better quality than it is now. I don't know if it continues to be as good if I split this into multiple chunks. Give me some time to test.
I don't know how you plan to run evals, but any chance it can be flexible enough to evaluate multiple skill files like it is in the passkey skills?

rviscomi:
Ok I will close this PR and keep the skill as a draft while we regroup.

> I don't know how you plan to run evals, but any chance it can be flexible enough to evaluate multiple skill files like it is in the passkey skills?

@micahjo7 could confirm. I think it's technically possible but the bigger question is if we want or need to have skill subresources at all.

micahjo7:
Currently, we don't have support for multiple skill files. If we see increasing need for it, it could be added.

PR #526, opened 2026-04-09
#340 individual-transform-properties

### LINKED ISSUE #340: Create guide and evals for the individual-transform-properties use case
Animate or override individual CSS transform properties (e.g. translate, rotate, scale) independently of other transform properties on a single element.

Affected web-feature IDs: [individual-transforms](https://webstatus.dev/features/individual-transforms)

Use case subdir: [guides/css/individual-transform-properties](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/css/individual-transform-properties)

Related features: #58

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

jamesnw:
Addresses #340


guides/user-experience/individual-transform-properties/guide.md
```diff
@@ -4,5 +4,102 @@ description: Animate or override individual CSS transform properties (e.g. trans
...
+The `transform` property and individual transform properties impact the layout and rendering of the page and may cause unexpected behavior with the z-index or anchor positioning. MANDATORY: If an element may have a transform applied as part of a state change like `:hover`, or a transition or animation, apply an identity transformation to the base element. This ensures that the element's stacking context and containment do not change when a transform is applied.
+
+```css
+.element{
+  /* Identity transformation for `translate`. */
+  translate: 0px;
+  /* Identity transformation for `rotate`. */
+  rotate: 0deg;
+  /* Identity transformation for `scale`. */
+  scale: 1;
```
malchata: Consider condensing these three code comments and using a single comment that goes a bit more into the "why" of this approach. Suggested by Gemini (which seems correct):

```suggestion
  /* MANDATORY: Apply identity transformations for properties that will
     change on state changes (like :hover). This prevents unexpected layout
     or z-index shifts caused by creating a new stacking context only on hover. */
  translate: 0px;
  rotate: 0deg;
  scale: 1;
```

guides/user-experience/individual-transform-properties/guide.md
```diff
@@ -4,5 +4,102 @@ description: Animate or override individual CSS transform properties (e.g. trans
...
+
+```css
+.card {
+  /* Define independent animations that don't overwrite each other */
+  animation: float 3s infinite ease-in-out;
+  
+  /* Transition only the scale property for hover states */
+  transition: scale 0.3s ease;
+
+  /* Specify the identity for the scale property. */
```
malchata: Reinforce the "why" a bit more in these comments, they're pretty focused on the "what". For example, in this case:

```suggestion
  /* Establish the base scale to prevent a sudden stacking context shift when transitioning on hover. */
```

I think going through and changing these to reinforce this pattern would help things a bit :)

malchata (approved):
Overall, I think this works quite well. There were some suggestions on focusing on the "why" of code comments over the "what". I think that could be a good improvement, but not worth blocking merge over IMO.

rviscomi (approved):
LGTM

PR #528, opened 2026-04-09
zero-dependency pure-JS vector search

paulirish:
Having node dependencies for our CLI is tricky because it means step 0 of our skill is ["install  us please"](https://github.com/GoogleChrome/guidance/blob/97a39ebdffba60f3c89c9fbaadbefcfe36207575/guides/modern-web-use-cases/SKILL.md#step-0-install-the-modern-web-cli).

Not only that but our dependencies have native components that differ per platform. And they weigh a lot.

Here, we swap out lancedb and onnxruntime (native) to a **pure JS pipeline.** 

| | before | after |
|-|-|-|
| rag top1 | 69% | 69% | 
| rag MRR |  0.812 |  0.812 | 
| cold latency | 33ms | 328ms | 
| warm latency | 34ms | 227ms | 
| size | 507 MB  (after install) | 30 MB |
| file count | 4,757 | 12| 
| setup | `npm install` | _none :)_ | 

### The Journey

A bit of iteration to land here..

*   **baseline: Transformers.js + ONNX Native**:  Performs great but demands compiled Node bindings. Complicates our distribution and packaging.. and requires a npm install step (which is very atypical for skills).
*   **Experiment 1: The WASM Fallback**: First attempt, to transformersjs + onnx wasm.
*   **Experiments 2-??**: Trying  a bunch of things that didnt work or regressed the benchmark.  Also different models but MiniLM is such a champ.
*   **The winner: Pure JS with TFJS**: We dropped ONNX entirely and wrote a targeted loader using the pure JS CPU backend of **TensorFlow.js** (also hard to keep track that transformers.js and tensorflow.js are two separate things but.. very different!). We needed to convert the  model weights to tensorflow format _manually_ because a prebuilt version wasnt available. 
    - I did some extra work to golf down the bundled JS size from 6MB to 832KB.
    - all said and done: no network calls, no system compiling, no WASM bloat, no native dependencies. Just javascript.

image

***

### other things
* **no more lancedb**: we precompute the embeddings and just keep them on disk. cosine similarity right on the main thread. Super fast and easy.   I decided to gzip the embeddings too (5.2MB to 1.8MB)
*   **Lazy imports**: The CLI is now instant for simple operations like `--help` or `--retrieve`. It only pulls in the heavy TFJS weights if you actually execute a search.


serving/lib/macros.ts
```diff
@@ -1,4 +1,4 @@
-import { validateFeature, getStatusMessage } from "../data/baseline.ts";
+import { validateFeature, getStatusMessage } from "../mcp-server/data/baseline.ts";
```
micahjo7: nit: should we move `baseline.ts` out of `mcp_server` dir?

serving/lib/transformers-embedder.ts
```diff
@@ -33,7 +37,7 @@ export class Embedder {
     this.pipe = (await pipeline("feature-extraction", repo, { dtype: dtype as any })) as any as FeatureExtractionPipeline;
   }
 
-  public async embed(text: string): Promise<number[]> {
+  public async embed(text: string, _isQuery = false): Promise<number[]> {
```
micahjo7, line 40: nit: remove?

serving/scripts/build-guides.ts
```diff
@@ -69,15 +79,16 @@ async function processGuides() {
   }
   
   let embedder: any;
-  if (modelName && (modelName.includes(".gguf") || modelName.includes("nomic"))) {
-    embedder = Gpt4AllEmbedder.getInstance(modelName);
+  if (!modelName || modelName === "transformers" || modelName === "tfjs") {
+    const { Embedder } = await import("../lib/transformers-embedder.ts");
+    embedder = Embedder.getInstance();
   } else {
+    const { Embedder } = await import("../lib/transformers-embedder.ts");
```
micahjo7: move outside of if/else?

serving/skills-cli/template/policies/modern-web-use-cases.toml
```diff
@@ -1,10 +1,3 @@
-[[rule]]
-toolName = "run_shell_command"
-commandRegex = "setup.sh"
-decision = "ask_user"
-priority = 100
-modes = ["plan"]
-
 [[rule]]
```
micahjo7: we only need the last `rule` now, right?

serving/skills-cli/test-dist.test.ts
```diff
@@ -38,22 +38,7 @@ try {
...
-
-  // template/package.json (placed at root of dist, for global publishing)
-  assert.strictEqual(templatePkg.dependencies?.["@huggingface/transformers"], transformersVersion, "template package.json should match serving package.json @huggingface/transformers version");
-  assert.strictEqual(templatePkg.dependencies?.["@lancedb/lancedb"], lancedbVersion, "template package.json should match serving package.json @lancedb/lancedb version");
-
-  // template/skills/modern-web-use-cases/package.json (placed adjacent to SKILL.md, for skill users)
-  assert.strictEqual(skillPkg.dependencies?.["@huggingface/transformers"], transformersVersion, "skill package.json should match serving package.json @huggingface/transformers version");
-  assert.strictEqual(skillPkg.dependencies?.["@lancedb/lancedb"], lancedbVersion, "skill package.json should match serving package.json @lancedb/lancedb version");
-});
+// Dependency parity test dropped as requested.
```
micahjo7: nit: remove

serving/skills-cli/template/skills/modern-web-use-cases/package.json
```diff
@@ -6,7 +6,6 @@
     "modern-web": "./modern-web.mjs"
   },
   "dependencies": {
-    "@huggingface/transformers": "^3.8.1",
-    "@lancedb/lancedb": "^0.26.2"
+    "onnxruntime-node": "^1.24.3"
```
micahjo7: do we need this in the dist? it's not being used, right?

serving/skills-cli/template/package.json
```diff
@@ -24,7 +24,6 @@
     "modern-web": "./skills/modern-web-use-cases/modern-web.mjs"
   },
   "dependencies": {
-    "@huggingface/transformers": "^3.8.1",
-    "@lancedb/lancedb": "^0.26.2"
+    "onnxruntime-node": "^1.24.3"
```
micahjo7: same comment, do we need this?

micahjo7 (commented):
this is awesome, thank you!

besides the comments, I was wondering if we should test this against our evals to check for any regressions in the search?

serving/skills-cli/template/policies/modern-web-use-cases.toml
paulirish: yup

serving/scripts/build-guides.ts
paulirish: thx!

serving/skills-cli/template/skills/modern-web-use-cases/package.json
paulirish: yah ... transformers.js is kind of a bear and makes it really hard to only use the parts you need and this is a artifact of that, but I managed to find a better way to handle this. That also makes our CI happy and resolves the failures you might have seen.

micahjo7 (approved):

paulirish:
> I was wondering if we should test this against our evals to check for any regressions in the search?

i ended up doing this at the embeddings level. The embeddings that come out now are the same as the previous ones. So all search queries from before end up with the same results now. :)

Issue #159, opened 2026-02-27
Create guide and evals for the complex-computations-and-loops use case


paulirish:
Break up heavy synchronous processing or DOM updates, to let the browser handle user input and repaint the screen.

Affected web-feature IDs: [scheduler](https://webstatus.dev/features/scheduler)

Use case subdir: [guides/performance/complex-computations-and-loops](https://github.com/GoogleChrome/guidance/tree/main/guides/performance/complex-computations-and-loops)

Related features: #71

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [ ] Use case guidance (guide.md)
- [ ] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

PR #539, opened 2026-04-10
Add guidance for complex-computations-and-loops

taraojo:
https://github.com/GoogleChrome/guidance/issues/159 Add guidance for complex-computations-and-loops and add scheduler.yield() and postTask() demo files and simplify demo implementation.


rviscomi:
@tunetheweb are you able to give this a look?

guides/performance/complex-computations-and-loops/guide.md
```diff
@@ -2,5 +2,109 @@
...
-    - scheduler
----
\ No newline at end of file
+  - scheduler
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/API/Scheduler
+  - https://github.com/WICG/scheduling-apis
+  - https://developer.mozilla.org/en-US/docs/Web/API/Scheduling
+  - https://developer.mozilla.org/en-US/docs/Web/API/Prioritized_Task_Scheduling_API
+  - https://www.w3.org/TR/task-scheduler/
```
malchata: Not in the interest of self-promotion 😅  but wondering if it might sense to include the long tasks guide from web.dev, as it covers `scheduler.yield`.

```suggestion
  - https://www.w3.org/TR/task-scheduler/
  - https://web.dev/articles/optimize-long-tasks
```

guides/performance/complex-computations-and-loops/guide.md
```diff
@@ -2,5 +2,109 @@
...
+    // Process the item
+    processItem(items[i]);
+    
+    // MANDATORY: Yield to the main thread periodically to keep the UI responsive.
+    // The chunk size (e.g., 50) should be tuned based on how long each item takes to process.
+    // Aim for chunks that complete within 50ms.
+    if (i > 0 && i % 50 === 0) {
+      await scheduler.yield();
+    }
+  }
```
malchata: I think a deadline-based approach here would be better:

```suggestion
async function processLargeArray(items) {
  // DO: Set a time-based deadline 50 milliseconds into the future. 50
  // milliseconds is the boundary for when a task becomes a long task.
  let deadline = performance.now() + 50; // 50ms budget

  for (let i = 0; i < items.length; i++) {
    // Process the item
    processItem(items[i]);
    
    // MANDATORY: Yield to the main thread periodically to keep the UI
    // responsive. This can be done by checking if the deadline set earlier
    // has been exceeded. When it has been, yield, then reset the deadline
    // another 50 milliseconds into the future.
    if (performance.now() >= deadline) {
      await scheduler.yield();
      deadline = performance.now() + 50;
    }
  }
```

guides/performance/complex-computations-and-loops/guide.md
```diff
@@ -2,5 +2,109 @@
...
+    if (options.priority === 'user-blocking') {
+      // MessageChannel is a faster macrotask than setTimeout, yielding to the 
+      // event loop but executing before standard timeouts.
+      const channel = new MessageChannel();
+      channel.port1.onmessage = () => resolve(callback());
+      channel.port2.postMessage(null);
+    } else if (options.priority === 'background') {
+      // Fallback for background tasks (since requestIdleCallback lacks Safari support).
+      // Using a short timeout helps defer execution until the main thread is less busy.
+      setTimeout(() => resolve(callback()), 200);
```
malchata: Wouldn't it be a potentially better solution to feature detect `requestIdleCallback()` here and then use that if possible? I know Safari compatibility is an issue, but is there a reason why we wouldn't want to instruct agents to suggest it in environments where it is available?

Note, I am aware that `requestIdleCallback` in busy environments could result in "background" tasks (as described in this case) never completing. If that's the rationale for avoiding it, that's fine, but it's worth bringing up!

```suggestion
    } else if (options.priority === 'background') {
      // DO: Detect if requestIdleCallback is available.
      if ('requestIdleCallback' in window) {
        requestIdleCallback(() => resolve(callback()));
      } else {
        // DO: Since requestIdleCallback has a lack of
        // support in earlier versions of Safari, using
        // setTimeout with a longer timeout value can work:
        setTimeout(() => resolve(callback()), 200);
      }
```

malchata (changes requested):

guides/performance/complex-computations-and-loops/guide.md
```diff
@@ -2,5 +2,109 @@
...
+  - scheduler
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/API/Scheduler
+  - https://github.com/WICG/scheduling-apis
+  - https://developer.mozilla.org/en-US/docs/Web/API/Scheduling
+  - https://developer.mozilla.org/en-US/docs/Web/API/Prioritized_Task_Scheduling_API
+  - https://www.w3.org/TR/task-scheduler/
+---
+
+Long-running JavaScript tasks block the main thread, preventing the browser from rendering updates or responding to user input (like clicks or scrolling). By intentionally breaking up these heavy synchronous processes and yielding control back to the main thread, you ensure your application remains responsive.
```
tunetheweb: Scrolling _can_ be affected. But mostly this happens outside of the main thread, so confusing to include that. Let's drop that.

```suggestion
Long-running JavaScript tasks block the main thread, preventing the browser from rendering updates or responding to user input (like clicks). By intentionally breaking up these heavy synchronous processes and yielding control back to the main thread, you ensure your application remains responsive.
```

guides/performance/complex-computations-and-loops/guide.md
```diff
@@ -2,5 +2,109 @@
...
+
+// Schedule analytics or background sync
+scheduler.postTask(() => sendAnalyticsData(), { priority: 'background' });
+```
+
+### Fallback strategies
+
+{{ BASELINE_STATUS("scheduler") }}
+
+Older browsers may not support the `scheduler` API. You MUST implement a fallback using `setTimeout` (or `requestIdleCallback`) to ensure code executes without breaking.
```
tunetheweb: The only browser that doesn't support `scheduler` is Safari - which also doesn't support `rIC`. Also maybe drop "older" older that reason.


```suggestion
Some browsers may not support the `scheduler` API. You MUST implement a fallback using `setTimeout` to ensure code executes without breaking.
```

guides/performance/complex-computations-and-loops/guide.md
tunetheweb: +1 definitely include this.

guides/performance/complex-computations-and-loops/guide.md
tunetheweb: To be honest, I could go either way here.

Deadline works better if the tasks are of short or of variable length. Just yielding every so often is a bit simpler (simplistic?) but not a bad strategy as long as the tasks aren't too short (where the overhead of yielding is a bit unnecessary).

So you're probably technically right here @malchata but I also wouldn't be totally against leaving it as is.

malchata: Generally agree. I think either approach is better than no yielding at all, but it seems a bit more precise to check for the 50 millisecond boundary. I could go either way on it, though, too!

guides/performance/complex-computations-and-loops/guide.md
tunetheweb: As per below, Safari is the only browser that doesn't support rIC. So feature detecting would maybe add some support for really old Chrome or Firefox but otherwise wouldn't be that much use. So not sure it's worth the extra code to be honest.

If you want the best support you should use the polyfill. But I also argue that's overkill for most.

Talking of which, we should probably mention the polyfill: https://github.com/GoogleChromeLabs/scheduler-polyfill

guides/performance/complex-computations-and-loops/guide.md
malchata: Yeah, I forgot that `requestIdleCallback` is in Safari TP (behind a flag, I think), and conflated that as recent support. I favor dropping "Older" if this suggestion gets applied.

guides/performance/complex-computations-and-loops/guide.md
taraojo: Added, strange but I didn't come across this in my Googles!

guides/performance/complex-computations-and-loops/guide.md
taraojo: Thanks both, I'll commit the change. It seems easier to read for me too.

guides/performance/complex-computations-and-loops/guide.md
taraojo: My general approach to fallbacks is that the feature should (in most cases) be Baseline widely available which `requestIdleCallback` wasn't. I wasn't aware of the polyfill, does it make sense to add that as the fallback? Although if Safari isn't supporting scheduler, the polyfill will be used in almost all cases. Would that be a problem? In what cases is using the polyfill not overkill?

tunetheweb: Yeah that’s kind of my point too. rIC isn’t available to the one browser we’d want to fallback for (ignoring the small number of older Chrome’s or Firefox’s) so I think it’s mostly pointless to fallback to that.

If you DO need the full fallback then you’re better going all in and using the polyfill to get even better fallback support for Safari.

Whether using the polyfill is worth it (over a simple setTimeout fallback like you currently have) is quite subjective and something we never really landed with definitive guidance on for scheduler. I’m with you of simplicity ruling as someone who’s spent far too long managing dependency upgrades (not to mention the downside of too many polyfills!). So I’m quite negative on polyfills when they aren’t needed. On the other hand, setTimeout is not as good as yield, so we are given a (slightly) lesser experience to Safari users and in some cases that may not be acceptable. The counter argument is Safari users tend to be on higher end devices and if there was enough of a differentiator in this then Safari would implement yield. @philipwalton what’s your opinion here?

taraojo: Chatted with @rviscomi about this and have now updated the PR so this is use case is focused on using `scheduler.yield()` with `setTimeout` as the fallback in agreement that the polyfill is overkill here. I've updated the guide and demo with this in mind and have done the same for #550 PR. Please take another look!

guides/performance/break-up-long-tasks/guide.md
```diff
@@ -0,0 +1,74 @@
...
+  - scheduler
+sources:
+  - https://web.dev/articles/optimize-long-tasks
+  - https://developer.mozilla.org/en-US/docs/Web/API/Scheduler/yield
+  - https://developer.mozilla.org/en-US/docs/Web/API/Prioritized_Task_Scheduling_API
+  - https://developer.mozilla.org/en-US/docs/Web/API/Scheduler
+  - https://github.com/WICG/scheduling-apis
+  - https://developer.mozilla.org/en-US/docs/Web/API/Scheduling
+  - https://www.w3.org/TR/task-scheduler/
+  - https://github.com/GoogleChromeLabs/scheduler-polyfill
```
tunetheweb: Shall we remove this since not using it?


```suggestion
```

tunetheweb (approved):
LGTM with one suggestion.

guides/performance/break-up-long-tasks/guide.md
taraojo: Ah yes, that should not be in this PR

rviscomi (approved):

Issue #160, opened 2026-02-27
Create guide and evals for the prioritized-continuation-after-yielding use case


paulirish:
Yield execution of a long-running task to the main thread while guaranteeing its continuation is immediately prioritized and resumed, preventing queue jumping.

Affected web-feature IDs: [scheduler](https://webstatus.dev/features/scheduler)

Use case subdir: [guides/performance/prioritized-continuation-after-yielding](https://github.com/GoogleChrome/guidance/tree/main/guides/performance/prioritized-continuation-after-yielding)

Related features: #71

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [ ] Use case guidance (guide.md)
- [ ] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

tunetheweb:
I think this is covered by #159 (with PR #539 ). Do you agree @paulirish ?

rviscomi:
Yeah these use cases have significant overlap so I agree we could close this one.

tunetheweb:
@rviscomi should we leave thus issue open until we clean up the directory that was already created for this (https://github.com/GoogleChrome/guidance/tree/main/guides/performance/prioritized-continuation-after-yielding)?

rviscomi:
Done! https://github.com/GoogleChrome/guidance/pull/578

Issue #161, opened 2026-02-27
Create guide and evals for the prioritized-preloading use case


paulirish:
Orchestrate low priority work, such as background network requests, without competing with critical rendering.

Affected web-feature IDs: [scheduler](https://webstatus.dev/features/scheduler)

Use case subdir: [guides/performance/prioritized-preloading](https://github.com/GoogleChrome/guidance/tree/main/guides/performance/prioritized-preloading)

Related features: #71

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [ ] Use case guidance (guide.md)
- [ ] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

PR #550, opened 2026-04-10
Add guides for prioritized-preloading

taraojo:
#161 Add testing framework, guide documentation, and demo implementations for prioritized task preloading using scheduler.postTask(). This PR follows a similar approach (specifically fallback strategy) to #539


rviscomi:
@tunetheweb similar request 🙏

guides/performance/prioritized-preloading/guide.md
```diff
@@ -2,5 +2,65 @@
...
+You MUST implement a fallback to simulate background priority. While `requestIdleCallback` is often used for background tasks, it is also not yet a widely available Baseline feature. Therefore, a delayed `setTimeout` is the most reliable cross-browser fallback.
+
+```javascript
+// A helper function for scheduling background tasks with a fallback
+function fallbackScheduleTask(callback) {
+  return new Promise((resolve) => {
+    // DO: Fallback for background tasks.
+    // Using a short timeout (e.g., 200ms) helps defer execution until 
+    // the main thread is less busy, giving priority to immediate user interactions.
+    setTimeout(() => resolve(callback()), 200);
```
malchata: Similar to what I said in #539 about `requestIdleCallback`—if we're avoiding it for a good reason (i.e., `requestIdleCallback` tasks may never complete on busy pages), that's fine, but genuinely asking if it should be mentioned here as a fallback strategy.
```suggestion
    if ('requestIdleCallback' in window) {
      requestIdleCallback(() => resolve(callback()));
    else {
      setTimeout(() => resolve(callback()), 200); 
    }
```

guides/performance/prioritized-preloading/guide.md
```diff
@@ -2,5 +2,65 @@
 name: prioritized-preloading
 description: Orchestrate low priority work, such as background network requests, without competing with critical rendering.
```
tunetheweb: Hmmm I'm not sure about this use case for the scheduler API at all and think it should 1) be removed or 2) refactored to look at different preloading APIs, or 3) be renamed to look at "low priority work (such as sending analytics)" rather than "prioritized preloading".

The schedular API is mostly for reducing CPU work, rather than for reducing network work. Of course, it _can_ be used for that (in that it can be used for anything!) but it's not a good use case IMHO and there are better APIs (primarily `<link rel=prefetch>`, `<link rel=preload>`, and Speculation Rules).

I noticed @philipwalton [made a comment in this direction in his review](https://github.com/GoogleChrome/guidance/pull/112#discussion_r2851229638) (you'll need to expand the last resolved comment).

@NourNabil , since you added this use case in the first place can you give a but more background of what you were thinking of here? Or examples from docs?

malchata: Seconding this. I had wondered if I missed something, because this was a thought I had. Seems like a job for Speculation Rules.

The `fetch` part of it also seems absent—it's not necessarily a bad idea to deprioritize a `fetch` call, but the guide doesn't invoke `fetch`—I could be misunderstanding, though, but even if that were the case, the `priority` argument on the `fetch` call would be a more appropriate, I would think.

NourNabil: The original idea was based on [this blog post ](https://medium.com/airbnb-engineering/building-a-faster-web-experience-with-the-posttask-scheduler-276b83454e91). The gist is Airbnb uses the scheduler API to preload images in a carousel for example because it provides the ability to stagger requests using a delay, cancel them outright if the user navigates away from the carousel. The API also can be used to offload background CPU work.

That is the reason the description reads 

> "Orchestrate low priority work, such as background network requests, without competing with critical rendering."

I wanted to specify the background network requests can be part of the low priority work as well. I understand why the naming might be confusing in this case as I wanted to say "You can choose the priority for low priority work such as preloading" but reading it out of context would make it sound like "You can make preloading have a higher priority". 
 
I am happy to refactor the use case to something like the following and steer away from the preloading angle completely since it is covered by <link rel=prefetch> and <link rel=preload>.
 
>name: orchestrate-background-tasks
>description: Orchestrate background low priority work (such as sending and processing analytics), allowing the browser to yield instantly to critical user input.
>web-feature-ids:
>    - scheduler

This would also be supported by the demo. @malchata I was using the heavy maths functions as a stand in for a fetch in this case.

tunetheweb: > The original idea was based on [this blog post ](https://medium.com/airbnb-engineering/building-a-faster-web-experience-with-the-posttask-scheduler-276b83454e91). The gist is Airbnb uses the scheduler API to preload images in a carousel for example because it provides the ability to stagger requests using a delay, cancel them outright if the user navigates away from the carousel. The API also can be used to offload background CPU work.

That's interesting! I hadn't seen that one.

However, since that case study came out, `loading=lazy` has become Baseline Widely available for images, and `fetchpriority` further allows tweaking the priority of network resources (though admittedly not delaying them completely).

The fact this use of `scheduler` for reducing network load, still isn't widespread 5 years later shows it was a bit of a niche use case IMHO.

> I am happy to refactor the use case to something like the following and steer away from the preloading angle completely since it is covered by and .

Yes, I think that would be a a better angle. And couples quite nicely with [the other use case of splitting up higher priority tasks](https://github.com/GoogleChrome/guidance/pull/539).

taraojo: To make sure I understand the proposed changes, instead of focusing on the preloading aspect of this current guide we should turn this into the orchestration angle that Nour has detailed above. Which if I understand correctly is the same code (demo and guidance examples) we have in this PR? 

Then for the scheduler API we have:

1. split high priority tasks in #539 
2. orchestrate low priority tasks here
3. prioritise tasks after yielding (todo in #160) - does this still make sense to keep?

tunetheweb: > To make sure I understand the proposed changes, instead of focusing on the preloading aspect of this current guide we should turn this into the orchestration angle that Nour has detailed above. Which if I understand correctly is the same code (demo and guidance examples) we have in this PR?

Yes mostly. Though change mentions of preloading (in the use case, the demo) to "orchestrate lower priority tasks" or something like that. There is also some overlap with https://github.com/GoogleChrome/guidance/tree/main/guides/performance/full-session-analytics but think that's more about batching the analytics, whereas here it's more about sending them in down time. Plus analytics isn't the only low-priority request.

> Then for the scheduler API we have:
>
> 1. split high priority tasks in https://github.com/GoogleChrome/guidance/pull/539
> 2. orchestrate low priority tasks here
> 3. prioritise tasks after yielding (todo in https://github.com/GoogleChrome/guidance/issues/160) - does this still make sense to keep?

I'm not sure it does. I think 1 and 3 are basically the same. Will comment on that issue.

taraojo: Great thanks! I'll make those changes. On the third one, I have WIP code/guide and the code examples are identical to number one, I could combine the descriptions to ensure the guide is selected if a user's prompt contains references to either option.

rviscomi: Catching up, I've just closed #160 which I agree had a lot of overlap with #159.

Looking at this one, it's still unclear to me how distinct this is from the `postTask` guidance covered by #539. It includes an example very similar to the suggested direction for this use case:

```js
// Schedule analytics or background sync
scheduler.postTask(() => sendAnalyticsData(), { priority: 'background' });
```

To help make each use case more distinct, what if we effectively make this one the `postTask` use case, and make #539 the `yield` use case?

tunetheweb: Yes. It's

1. Splitting high-priority tasks - yield
2. Defering low-priority work - postTask/rIC

Annoyingly setTimeout is the easist fallback for both so the lines blurs for Safari :-(

taraojo: I've now renamed this guide to `schedule-tasks-by-priority` to focus on scheduling of different types and included the polyfill as a fallback in this case. Please take another look!

rviscomi (approved):
LGTM

Tested in Chrome and Safari

guides/performance/prioritized-preloading/guide.md
rviscomi: Thanks @taraojo!

@tunetheweb could you take another look?

guides/performance/schedule-tasks-by-priority/guide.md
```diff
@@ -0,0 +1,71 @@
...
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/API/Scheduler
+  - https://developer.mozilla.org/en-US/docs/Web/API/Prioritized_Task_Scheduling_API
+  - https://github.com/WICG/scheduling-apis
+  - https://developer.mozilla.org/en-US/docs/Web/API/Scheduling
+  - https://www.w3.org/TR/task-scheduler/
+  - https://github.com/GoogleChromeLabs/scheduler-polyfill
+---
+
+When building complex web applications, tasks have different levels of urgency. Fetching data for the current view is more important than sending analytics or prefetching assets. The Prioritized Task Scheduling API allows you to schedule work with specific priorities, ensuring the browser remains responsive to user input.
```
tunetheweb: Let's tweak this slightly away from fetch:

```suggestion
When building complex web applications, tasks have different levels of urgency. Completing tasks for the current view is more important than sending analytics or prefetching assets. The Prioritized Task Scheduling API allows you to schedule work with specific priorities, ensuring the browser remains responsive to user input.
```

guides/performance/schedule-tasks-by-priority/guide.md
```diff
@@ -0,0 +1,71 @@
...
+  // DO: Perform heavy background work that is not time-critical
+  sendAnalytics();
+}, { priority: 'background' });
+```
+
+### Fallback strategies
+
+{{ BASELINE_STATUS("scheduler") }}
+
+To support browsers that do not have the Prioritized Task Scheduling API, you must use a polyfill to maintain task prioritization. Do not rely solely on `requestIdleCallback` or `setTimeout` as they do not support multiple priorities.
```
tunetheweb: I think this is a little strong.

```suggestion
For browsers that do not have the Prioritized Task Scheduling API, you have two options: 1) complete critical work in the current task, rather than breaking it up, or 2) use the scheduler polyfill.

The first option involves deferring non-critical tasks using `setTimeout`, but executing critical work in the current task. This can impact INP since you are not breaking up the current task. It also does not offer more than two priorities (now, and deferred with setTimeout) so offers only basic scheduling.

A better option when explicitly looking to schedule task by priority, is to use the scheduler polyfill to support multiple priorities.
```

Or is it better to be opinionated here @rviscomi rather than nuanced?

rviscomi: Nuance is generally good, I'd just say that if we present multiple options we should also help agents decide how to know which option to take. If the recommendation at the end is to simply use the polyfill, then having the other option might be confusing.

> Do not rely solely on `requestIdleCallback` or `setTimeout` as they do not support multiple priorities.

What if we drop this part instead, to avoid needing to clarify when they might be useful? I think the intention behind it was to help explain why the polyfill is useful, not to imply that these methods aren't useful on their own.

tunetheweb: SGTM

taraojo: Removed!

tunetheweb (approved):

PR #554, opened 2026-04-10
Guide and evals for `stabilize-reactive-state` use case for temporal

### LINKED ISSUE #381: Create guide and evals for the stabilize-reactive-state use case
Manage task deadlines or schedules in data-driven views without unexpected side effects from shared mutable state.

Affected web-feature IDs: [temporal](https://webstatus.dev/features/temporal)

Use case subdir: [guides/js/stabilize-reactive-state](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/js/stabilize-reactive-state)

Related features: #79

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

pattishin:
**Relates to:** #381 

**Use cases:**
`stabilize-reactive-state`

**Test by:**
```
gd eval <use-case-name>
gd dashboard 
```
Screenshot 2026-04-10 at 6 54 48 PM

Screenshot 2026-04-10 at 6 54 41 PM


guides/user-experience/stabilize-reactive-state/guide.md
```diff
@@ -3,4 +3,88 @@ name: stabilize-reactive-state
...
\ No newline at end of file
+sources:
+  - https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Temporal
+  - https://tc39.es/proposal-temporal/docs/
+  - https://www.w3schools.com/js/js_temporal.asp
+---
+
+# Stabilize Reactive State with Temporal
+
+In modern web development, reactive systems (such as React, Vue, or Svelte) rely on reference equality to detect state changes. When an object's reference remains the same, the system assumes the state has not changed and skips re-rendering the UI.
```
rviscomi: Is this strictly true? Seems like it might be true for React, but Vue and Svelte may use `Proxy` objects and not necessarily rely on reference equality.

This doesn't affect the guidance, but just want to make sure the statement is factually correct.

rviscomi: FWIW I just ran a couple of tests in Vue and Svelte 5 and they _are_ able to watch for changes to a simple object like `{count: 1}` but NOT for `Date` objects.

So the statement seems to hold when talking about some objects, but not all.

WDYT about something like this (combining the first and second paragraphs):

> While some reactive systems (like React) rely strictly on reference equality to detect state changes, others (like Vue and Svelte) can track mutations to plain objects. However, for built-in objects like the legacy `Date` object, internal mutations (like `setHours()`) do not change the object's reference and are generally not tracked by *any* framework's default reactivity system. This leads to missed UI updates and hard-to-debug side effects.

pattishin: Yeah! sgtm 👍🏼 will update.

pattishin: Updated!

guides/user-experience/stabilize-reactive-state/expectations.md
```diff
@@ -0,0 +1,7 @@
+- The implementation MUST feature-detect the `Temporal` API using `typeof Temporal === 'undefined'` or `!globalThis.Temporal` before usage.
+- The implementation MUST conditionally load a Temporal polyfill (e.g., `@js-temporal/polyfill`) only if native support is absent.
```
rviscomi: The demo file wasn't changed in this PR so I can't comment on it directly, but I noticed that it unconditionally loads the polyfill, which isn't aligned with this expectation and the guidance in the doc.

```html
<script src="https://cdn.jsdelivr.net/npm/@js-temporal/polyfill@0.4.4/dist/index.umd.js"></script>
```

It should be updated to conditionally load the polyfill using feature detection.

guides/user-experience/stabilize-reactive-state/guide.md
```diff
@@ -3,4 +3,86 @@ name: stabilize-reactive-state
...
+- **DO NOT** mutate `Date` objects in place when they are part of a component's state.
+- **DO** ensure you handle environments without native support by conditionally loading a polyfill.
+
+### Fallback strategies
+{{ BASELINE_STATUS("temporal") }}
+
+Since the `Temporal` API is a newer feature and may not be supported in all browsers, you should feature-detect it and conditionally load a polyfill if needed.
+
+```html
+<!-- Conditionally load the Temporal polyfill only if not natively supported -->
```
rviscomi, line 70: I was testing in Safari and got this error:

> ReferenceError: Can't find variable: Temporal

IIUC we also need to manually assign the global `Temporal` object, since the polyfill doesn't do it automatically.

Not sure if other Temporal guides do it that way and this is the outlier, or I just missed it from the other ones. If so we should update those as well.

rviscomi (changes requested):

guides/user-experience/stabilize-reactive-state/guide.md
pattishin: 10-4 - taking a look!

pattishin: 🤦🏼‍♀️ Okay updated and I went back to check on the other temporal use cases against Safari.  https://github.com/GoogleChrome/guidance/pull/587 Switching them all to use the esm.sh temporal polyfill.

rviscomi (approved):
LGTM

Tested in Chrome and Safari

PR #556, opened 2026-04-11
Add warning about animating scrollbar-color, closes #512

### LINKED ISSUE #512: Review for "Animate scrollbar color on scroll" guide
Link: https://github.com/GoogleChrome/guidance/blob/main/guides/user-experience/animate-scrollbar-color-on-scroll/guide.md

## Overview

This is a cool demo pulling in multiple features (scrollbar-color, scrollbar-width, scroll-driven-animations, registered-custom-properties). It would work brilliantly as a feature-focused tutorial aimed at helping human developers understand how these pieces fit together.

However, it seems less well-suited to a Guidance use case, as it is unlikely to arise in real-world scenarios, and much more likely to come up as a false positive. A quick survey of common patterns and developer needs suggests this is not something practitioners are actively trying to solve. 

Real-world scroll-driven animations tend to fall into two broad categories:

1. **Communicating scroll progression:** These rely on quantitative [visual variables](https://towardsdatascience.com/data-visualization-explained-part-2-an-introduction-to-visual-variables/) — visual variables with an inherent ordering, such as position or length (e.g. progress bars, carousel dots). Color — particularly hue — does not encode order well and is generally considered a poor choice for conveying quantitative change on its own.
2. **Using scroll as a timeline for richer effects**: Examples include parallax or scrollytelling. In these cases, animating scrollbar color alone is too subtle and disconnected from the content to carry meaningful value.

By contrast, real-world uses of `scrollbar-color` are typically about **adapting to context:** matching surrounding UI, responding to color schemes (e.g. dark mode), or accommodating user preferences (e.g. high contrast mode). A scroll-driven animation that changes it could theoretically be useful together with another effect (e.g. to have scrollbars adapt to a changing context), but provides little value by itself, and even that is relatively uncommon.

Including it as a standalone “use case” is not without cost. In practice, **we’ve already seen agents select guides based on shallow keyword matching**. Having it increases the likelihood that this example would be surfaced for scenarios like “show scroll progress” where it represents an **antipattern** rather than a recommended approach.

As a result, this content may be better positioned as a feature demonstration rather than presented as a representative real-world use case.

## Content/demo review

Although my recommendation is to remove the guide altogether, I did a pass on the content as well, in case.

### Critical: Severely broken in Safari

Scrollbar flickers on scroll!
**Turns out this is a Safari bug, which I reported, see [WebKit bug #311752](https://bugs.webkit.org/show_bug.cgi?id=311752).**

https://github.com/user-attachments/assets/10206bfc-6223-435b-aad1-b949e2f4240c

No attempt to work around it (e.g. set the color in a different way) succeeded, it seems to be fundamentally triggered via anything that changes `scrollbar-color` via any means, even JS. I was also unsuccessful in programmatically detecting the bug. This adds another reason to remove this guide and potentially even add **guidance against animating or transitioning `scrollbar-color`**.

**The rest of this was written before I tested in Safari :)**

-----

### Moderate: False claim that `scrollbar-color` is discrete

This is false:

> Because the standard `scrollbar-color` property is mathematically defined as "discrete"

`scrollbar-color` is [defined](https://drafts.csswg.org/css-scrollbars/#scrollbar-color) as a `<color>{2}` and interpolates as one. It _may_ not animate between `auto` and specified colors, but that's not what we're doing here.

Indeed, removing the registered property from the demo and just animating `scrollbar-color` works fine. 

It _would_ be useful to have the track color in a variable (e.g. `--scrollbar-track-color`) since that does not change, but that does not need property registration (and guidance about when to use a variable for DRY-ness should probably live in the CSS skill, not individual guides).

And in a real-world scenario we'd probably have both colors in variables, and we'd be animating between variable values too (design tokens) but that is out of scope for this guide, as it is not specific to this use case.

## Fallback 

Given that this is progressive enhancement, it doesn’t necessarily _need_ a fallback.

But also, a lot of this is generic fallback for `scrollbar-color` and `scrollbar-width` rather than specific to the use case discussed, and as currently structured they would need to be repeated across every guide that uses these features which could quickly get unmanageable.

I wonder if it would be easier to manage if we were able to associate a feature’s “standard” fallbacks with the feature itself, leaving only use case specific fallbacks to the guides, rather than repeating it across every guide that uses the feature.

Additionally, the fallback currently in the guide is only useful if `scrollbar-color` is not supported but [scroll-driven-animations](https://coda.io/d/_d4wB-Md4c-M#Features_tuHwPoik/r6&view=modal) are. The browsers that fit this criterion are rather few, making this tradeoff a bit questionable:

Browser / versions | Market share
-- | --
Chrome 115 - 120 | 0.37% (incl Edge & Opera)
Chrome Android (?) | ?
Safari 26.0 - 26.1 | 0.07%
Safari iOS 26.0 - 26.1 | 0.46%
Opera Mobile 80 | 0.9%
Total (minus Chrome Android) | **1.8%**

If a “proper” fallback is desired, the solution is typically a `scroll` event listener that sets a CSS variable that goes from 0 to 1, which would apply much more broadly, but is also more heavyweight. Yet another reason to encode the feature fallbacks and their tradeoffs together with the feature, since this would be useful for scroll-driven animations in general; then which fallback to use could be decided on a case by case basis taking into account how important the effect is for the user's individual scenario as well.

### Nit: No need to repeat CSS variable fallback value in every `var()`

From a CSS comment in `demo.html`:
> You MUST include a static fallback color inside the `var()` function (e.g., `var(--thumb-color, hotpink)`) so the scrollbar remains visible on legacy WebKit browsers that do not support `@property`.

There is **no browser with support for scroll-based animations, but not `@property`** so this is unnecessary — just add a guard for `animation-timeline: scroll(self)` as well in the `@supports` to avoid breakage in older browsers.

But also, you can just set `--thumb-dynamic-color` on `.scroller` — the animation will override it anyway. 

If the motivation is to support inherited `--thumb-dynamic-color` you can use a “pseudo-private" variable:

```css
.scroller.animated-scrollbar {
  --_thumb-dynamic-color: var(--thumb-dynamic-color, #f06292);
}
```

and then use `var(--_thumb-dynamic-color)` throughout.

In fact, the 0% keyframe is also unnecessary: it is automatically generated based on the existing element state, which is exactly the value we want. It would only be needed if we had a different thumb color in the first frame than the one the element has without the animation.

### Nit: No way to specify interpolation color space

In the CSS guide we recommend interpolating colors in oklab/oklch, but here we use a technique that cannot (currently) do that.

Using a nested `color-mix()` and only interpolating a single `--scroll-progress` variable would not only make the animation more reusable across different elements and effects but would also let us specify _how_ colors should interpolate.

### Nit: Fallback for `scrollbar-width: thin` only applies to vertical scrollbar

Note: To try out horizontal scrolling in the demo, add `h2 { width: 100vw; }` under .scroller.animated-scroller .

Fix: add height: 12px as well.

### Nit: Corner is a different color (fallback)

Fix: add `::-webkit-scrollbar-corner` as well.

----

If we end up keeping the guide, I'm happy to send a PR for any of these!

On a more meta point, is this format useful for the qualitative reviews? I'm thinking it could end up adding a lot of clutter if I post one of these per guide, so perhaps it's better to keep it in the report and post more targeted issues here?

LeaVerou:


malchata (approved):
I'm fine with these changes, but wait on @rviscomi before merging in case there's something I didn't account for.

rviscomi (approved):

PR #559, opened 2026-04-13
More eval investigation skill fine tuning

rviscomi:
- use checklists in eval investigations
- require the agent to share artifacts during the investigation
- sync other prompt-related guidance for consistency, including the `gd dev` prompt generation


paulirish (approved):
this is AMAZING

hell yeah

PR #560, opened 2026-04-13
#235 keep-preceding-text-static-when-editing

### LINKED ISSUE #235: Create guide and evals for the keep-preceding-text-static-when-editing use case
Reduce visual noise in editable text by ensuring text before the edit point remain stable and only the new text is re-wrapped.

Affected web-feature IDs: [text-wrap](https://webstatus.dev/features/text-wrap)

Use case subdir: [guides/user-experience/keep-preceding-text-static-when-editing](https://github.com/GoogleChrome/guidance/tree/main/guides/user-experience/keep-preceding-text-static-when-editing)

Related features: #81

jamesnw:
Fixes #235 

Note- I question whether this is a useful use case. UAs can choose their algorithm, and it seems that UAs often pick this, so I was unable to come up with a case where specifying `text-wrap: stable` actually makes a difference. It's also rather niche, and I can't imagine this guide solving a real world problem.


rviscomi:
> Note- I question whether this is a useful use case. UAs can choose their algorithm, and it seems that UAs often pick this, so I was unable to come up with a case where specifying `text-wrap: stable` actually makes a difference. It's also rather niche, and I can't imagine this guide solving a real world problem.

Thanks for calling that out. Your reasoning makes sense and I agree that it'd be best to drop it.

Sorry you already put in the time to get the guide and evals written up!

malchata:
I had a review request outstanding for this one, but it appears we won't be moving forward with it—so I'll close this out. Reopen it if you need to for any reason.

rviscomi:
Only thing left is to delete the use case subdirectory. @jamesnw can you do that in this PR?

jamesnw (commented):
@rviscomi I've deleted the directory.

rviscomi (approved):

rviscomi:
Thanks!

PR #562, opened 2026-04-13
#190 interactive-content-reveal

### LINKED ISSUE #190: Create guide and evals for the interactive-content-reveal use case
Create interactive reveal effects, such as a spotlight that follows the user's pointer to uncover details within an image or UI section.

Affected web-feature IDs: [masks](https://webstatus.dev/features/masks), [registered-custom-properties](https://webstatus.dev/features/registered-custom-properties)

Use case subdir: [guides/ui-behaviors/interactive-content-reveal](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-behaviors/interactive-content-reveal)

Related features: #65

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

jamesnw:
Addresses #190


guides/user-experience/interactive-content-reveal/guide.md
```diff
@@ -1,6 +1,134 @@
...
+    transparent var(--outer-size, 0%)
+  );
+}
+```
+
+{{ BASELINE_STATUS("masks") }}
+
+### Basic Mask Support
+For browsers that do not support CSS masking at all:
+1. **Prefixed property:** Use the `--webkit-mask-image` prefixed property for broader browser support.
```
malchata: It's a vendor prefix, correct? I think one dash then, so this looks like a typo :)

```suggestion
1. **Prefixed property:** Use the `-webkit-mask-image` prefixed property for broader browser support.
```

guides/user-experience/interactive-content-reveal/guide.md
```diff
@@ -1,6 +1,134 @@
...
+}  
+```
+
+### 3. Update Coordinates with JavaScript
+Track the pointer position and update the CSS variables. Because the properties are registered and have a `transition` defined, the spotlight will move smoothly even if the pointer events are infrequent.
+
+```javascript
+const container = document.querySelector('.container');
+
+container.addEventListener('mousemove', (e) => {
```
malchata: Should `'pointermove'` be used here, or are we scoping strictly to fine input devices?

guides/user-experience/interactive-content-reveal/guide.md
```diff
@@ -1,6 +1,134 @@
...
+  // Calculate position as a percentage of the container
+  const rect = container.getBoundingClientRect();
```
malchata: This feels like a performance smell a bit—`onmousemove` (and I assume `onpointermove`) are high frequency events, and calling `getBoundingClientRect()` here could be computationally expensive. Wouldn't mind getting a little input from @rviscomi (or even @tunetheweb if he wants to flex his perf muscles a bit).

guides/user-experience/interactive-content-reveal/guide.md
```diff
@@ -1,6 +1,134 @@
...
+{{ BASELINE_STATUS("masks") }}
+
+### Basic Mask Support
+For browsers that do not support CSS masking at all:
+1. **Prefixed property:** Use the `--webkit-mask-image` prefixed property for broader browser support.
+2. **Progressive Enhancement:** Design the base state of the UI to be fully functional and legible without the reveal effect. This is useful when the effect is only adds visual flair, and not a requirement for reading content.
+3. **Feature Detection:** Use `@supports` to only apply the masking styles if the browser supports them.
+
+```css
+@supports not (mask-image: radial-gradient(circle, black, transparent)) {
```
malchata: It _does_ add a bit more noise, but would it be work checking for both of the properties here?

```suggestion
@supports not ((mask-image: radial-gradient(circle, black, transparent)) or (-webkit-mask-image: radial-gradient(circle, black, transparent))) {
```

malchata (approved):
More questions than actual suggestions, but wanted to kick it back to you for your opinion before moving it along.

guides/user-experience/interactive-content-reveal/guide.md
jamesnw: I did a bit more looking into the support of the prefixed version, and I think it's sufficient that we can just drop the Feature Detection fallback option here.

malchata: Works for me!

guides/user-experience/interactive-content-reveal/guide.md
jamesnw: Good call, I think `pointermove` makes more sense here.

guides/user-experience/interactive-content-reveal/guide.md
jamesnw: You're not wrong, and I think the right method will likely be context specific here. 

I went with the simplest fix here, moving the `getBoundingClientRect()` out of the event listener. This works in this case because the demo container never changes size. Should we make this more robust with a ResizeObserver, or another tactic?

malchata: `ResizeObserver` is a good call here.

malchata (changes requested):
I think add in a ResizeObserver example for the `getBoundingClientRect()` issue and we've got it.

guides/user-experience/interactive-content-reveal/guide.md
jamesnw: Added!

guides/user-experience/interactive-content-reveal/guide.md
```diff
@@ -1,6 +1,130 @@
...
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/mask-image
+  - https://web.dev/articles/css-masking
+  - https://www.gperrucci.com/blog/css/mask-image
+  - https://web.dev/learn/css/paths-shapes-clipping-masking#masking
+  - https://frontendmasters.com/blog/css-spotlight-effect/
+  - https://css-tricks.com/image-fragmentation-effect-with-css-masks-and-custom-properties/
 ---
+
+Add performant, interactive reveal effects to your site with CSS masks and registered custom properties. By using a radial gradient as a mask and registering its stop values, we can smoothly transition the entry and exit, while following a user's pointer with minimal JavaScript.
```
rviscomi: I was just doing a project-wide accessibility review with Gemini and it flagged this use case even before any content was written:

> - **Content gated behind pointer position.** Non-pointer users (keyboard, switch, screen reader) have no access path to the hidden content.
>- **CSS mask-based hiding** may leave content in the DOM but invisible, or may clip it from the AT — behavior varies by implementation.
>- **No keyboard or focus equivalent can replicate "following the cursor."** This pattern needs a fundamentally different alternative for non-pointer users.

Would you say this use case is safe for any application, or should we emphasize when NOT to use it, eg revealing critical content?

jamesnw: Good question- I think it's use-case specific, and we should be clear how to handle other cases. I had some notes in the Accessibility and Interaction section, but expanded that. Do we also need to add cautions in this intro?

rviscomi: Thanks, I think that's a good start. We'll also have an accessibility specialist coming on board soon and I'll make sure they give this another look.

rviscomi (approved):
LGTM

Tested in Chrome and Safari

Made a note of this for the accessibility specialist to take a closer look at later

PR #563, opened 2026-04-13
#244 physics-based-easing

### LINKED ISSUE #244: Create guide and evals for the physics-based-easing use case
Create custom, physics-based animation and transition effects, like bounce and spring, that feel more natural and engaging than traditional easing curves.

Affected web-feature IDs: [linear-easing](https://webstatus.dev/features/linear-easing)

Use case subdir: [guides/ui-behaviors/physics-based-easing](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-behaviors/physics-based-easing)

Related features: #63

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

jamesnw:
Addresses #244

### DIFF
--- a/guides/user-experience/physics-based-easing/guide.md
+++ b/guides/user-experience/physics-based-easing/guide.md
@@ -15,7 +15,7 @@ Traditional CSS easing functions like `ease-in` or `cubic-bezier()` are limited 
 ### Implementation Steps
 
 1.  **Generate the curve stops:**
-    Manually plotting dozens of points for a spring or bounce is impractical. Use a timing function from an external library, or use an external generator to convert an existing JavaScript easing function or an SVG path into the `linear()` syntax. Optional: store these timing functions as CSS custom properties for reuse throughout your site.
+    Manually plotting dozens of points for a spring or bounce is impractical. Use a timing function from an external library, or use a  tool to convert an existing JavaScript easing function or an SVG path into the `linear()` syntax. Optional: store these timing functions as CSS custom properties for reuse throughout your site.
 2.  **Define the timing function:**
     Apply the generated stops to the `transition-timing-function` or `animation-timing-function` property, or through the `transition` or `animation` shorthands.
 3.  **Adjust the duration:**
--- a/guides/user-experience/physics-based-easing/guide.md
+++ b/guides/user-experience/physics-based-easing/guide.md
@@ -93,6 +93,7 @@ For browsers that do not support `linear()`, provide a standard easing function 
 ```
 
 #### JavaScript Library Fallback (Motion/GSAP)
+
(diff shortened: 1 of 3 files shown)


guides/user-experience/physics-based-easing/guide.md
```diff
@@ -3,4 +3,126 @@ name: physics-based-easing
...
+  /* Fallback for older browsers (standard smooth exit) */
+  transition: transform 0.8s ease-out;
+  
+  /* Modern browsers will override with the physics-based easing */
+  transition-timing-function: linear(0, 1.1, 0.95, 1.02, 1);
+}
+```
+
+#### JavaScript Library Fallback (Motion/GSAP)
+Optional: If a high-fidelity physics animation is critical even in older browsers, use a JavaScript library like **Motion** (motion.dev) or **GSAP** (greensock.com) to handle the animation when `linear()` is unsupported.
```
malchata, line 97: Just a nit, and not sure it matters, but I think Markdown syntax often recommends a line break after a heading.

```suggestion
#### JavaScript Library Fallback (Motion/GSAP)

Optional: If a high-fidelity physics animation is critical even in older browsers, use a JavaScript library like **Motion** (motion.dev) or **GSAP** (greensock.com) to handle the animation when `linear()` is unsupported.
```

malchata (approved):
I think this looks good! Before merging, maybe check in with @rviscomi on evals. Both guided and unguided are at 0%.

guides/user-experience/physics-based-easing/guide.md
jamesnw: Adjusted

guides/user-experience/physics-based-easing/guide.md
malchata: Gemini got tripped up on "external generator", complaining that it can't use external resources:

```
this instruction is impossible for them to follow and may cause them to hallucinate `linear()` values.
```

Which, I can't tell if that's a misunderstanding, but I think what you meant by "external generators" are websites that calculate custom easing functions for you. I guess you could tweak this a bit, but Gemini didn't complain about it before, so I'm not sure why it's complaining about it now 😅

malchata (approved):

guides/user-experience/physics-based-easing/guide.md
jamesnw: I changed to `use a tool`, although I'm not sure if that's better.

What I want from this as a user is not an implementation, but rather encouragement to go make the curve myself, but I don't think these tools are set up to be that interactive.

malchata: I agree. I think having the developer fish a little for what they want in this case helps them build a UI that's better-suited to their needs and preferences.

rviscomi (approved):
LGTM

Tested in Chrome and Safari

PR #568, opened 2026-04-14
#338 cross-document-transitions

### LINKED ISSUE #338: Create guide and evals for the cross-document-transitions use case
Create smooth, seamless transitions between full page navigations, such as cross-fades, custom reveal effects, or morphing of content from one page to the next.

Affected web-feature IDs: [view-transitions](https://webstatus.dev/features/view-transitions), [cross-document-view-transitions](https://webstatus.dev/features/cross-document-view-transitions), [navigation](https://webstatus.dev/features/navigation)

Use case subdir: [guides/ui-behaviors/cross-document-transitions](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-behaviors/cross-document-transitions)

Related features: #87 #50 #667

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

jamesnw:
Addresses #338 

View transition types for cross document view transitions are not covered in other use cases, so I added those here. I think other parts (like shared element transitions) will be covered in the other use cases.

### DIFF
--- a/guides/user-experience/cross-document-transitions/guide.md
+++ b/guides/user-experience/cross-document-transitions/guide.md
@@ -4,6 +4,7 @@ description: Create smooth, seamless transitions between full page navigations, 
 web-feature-ids:
   - view-transitions
   - cross-document-view-transitions
+  - navigation
 sources:
   - https://developer.chrome.com/docs/web-platform/view-transitions/cross-document
   - https://blog.robino.dev/posts/cross-document-view-transitions
@@ -20,10 +21,13 @@ Cross-document view transitions allow you to create smooth, app-like transitions
 Both the source and destination pages must opt-in to view transitions for the browser to trigger them on navigation.
 
 ```css
-/* Add to a global stylesheet shared by both pages */
-@view-transition {
-  /* Enables transitions for same-origin navigations */
-  navigation: auto;
+/* Respect user's preference for reduced motion */
+@media (prefers-reduced-motion: no-preference) {
+  /* Add to a global stylesheet shared by both pages */
+  @view-transition {
+    /* Enables transitions for same-origin navigations */
+    navigation: auto;
+  }
 }
 ```
 
@@ -60,10 +64,12 @@ You may want different transition effects depending on the pages you are navigat
 If the page you are navigating to will always have the same transition type, regardless of how you get to the page, you can specify it with `types` in the `@view-transition` rule.
 
 ```css
- @view-transition {
-  navigation: auto;
-  /* Specify the types of view transitions that will always be used on this page. */
-  types: previous;
+@media (prefers-reduced-motion: no-preference) {
+  @view-transition {
+    navigation: auto;
+    /* Specify the types of view transitions that will always be used on this page. */
+    types: previous;
+  }
 }
 ```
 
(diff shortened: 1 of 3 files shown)


guides/user-experience/cross-document-transitions/guide.md
```diff
@@ -4,4 +4,127 @@ description: Create smooth, seamless transitions between full page navigations,
...
+
+#### 4. Respect User Preferences for Reduced Motion
+
+
+```css
+@media (prefers-reduced-motion: no-preference) {
+  @view-transition {
+    navigation: auto;
+  }
+}
```
malchata: Would it make more sense to move this into step one so that this pattern is the default?

guides/user-experience/cross-document-transitions/guide.md
```diff
@@ -4,4 +4,127 @@ description: Create smooth, seamless transitions between full page navigations,
...
+```
+
+You can also conditionally specify transition types inside of an event listener for `pagereveal`.
+
+```js
+window.addEventListener("pagereveal", async (e) => {
+  if (e.viewTransition) {
+    // Use application-specific logic to compute a transition type
+     const transitionType = yourTransitionTypeLogic(navigation.activation.from, navigation.activation.entry);
```
malchata, line 82: This is the Navigation API right? If so, would it make sense to add `navigation` to `web-feature-ids` in the front matter?

guides/user-experience/cross-document-transitions/guide.md
```diff
@@ -4,4 +4,127 @@ description: Create smooth, seamless transitions between full page navigations,
...
+  /* Specify the types of view transitions that will always be used on this page. */
+  types: previous;
+}
+```
+
+You can also conditionally specify transition types inside of an event listener for `pagereveal`.
+
+```js
+window.addEventListener("pagereveal", async (e) => {
+  if (e.viewTransition) {
```
malchata: I think we want to do a Navigation API check as well.

```suggestion
  if (e.viewTransition && window.navigation?.activation) {
```

malchata (changes requested):

guides/user-experience/cross-document-transitions/guide.md
jamesnw: Good point. I've added this, and updated the demo to use it as well. I am also open to removing it from the guide and demo, as the core use case is possible without the Navigation API.

malchata: I think it's fine if we add it, if the pattern is sustainable and interoperable, and I think it is. Though I wouldn't mind hearing what @bramus has in the way of an opinion :)

bramus: I’d keep it in, as that’s the only way to customize things based on which URL you are coming from and which URL you are going to.

bramus: https://view-transitions.chrome.dev/pagination/navigation-api/ has a demo

malchata: Perfect. I think we can roll with it James having added it then :)

malchata (approved):

rviscomi (approved):
LGTM

Tested in Chrome and Firefox

Issue #162, opened 2026-02-27
Create guide and evals for the resolution-optimized-pseudo-elements use case


paulirish:
Use resolution-optimized images in CSS pseudo-elements (such as `::before` and `::after`) to reduce the number of DOM nodes.

Affected web-feature IDs: [image-set](https://webstatus.dev/features/image-set)

Use case subdir: [guides/performance/resolution-optimized-pseudo-elements](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/performance/resolution-optimized-pseudo-elements)

Related features: #57

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

PR #569, opened 2026-04-14
Add guide for resolution-optimized-pseudo-elements

taraojo:
#162 

This guide also includes a rename, as this was missed last time in the use case PR.

Question: the image-set has been added to the `content`, does this make sense or is `content: ""` and `background-image` preferred?


rviscomi (approved):
LGTM

Issue #167, opened 2026-02-27
Create guide and evals for the deliver-optimized-decorative-images use case


paulirish:
Deliver optimized decorative images (such as backgrounds, UI icons, or complex masks) by simultaneously providing next-generation image formats (like AVIF or WebP) alongside multiple pixel densities (like 1x and 2x) so the browser can dynamically negotiate the best combination of file size and visual quality for the user's device capabilities.

Affected web-feature IDs: [image-set](https://webstatus.dev/features/image-set)

Use case subdir: [guides/user-experience/serve-alternative-image-formats](https://github.com/GoogleChrome/guidance/tree/main/guides/user-experience/serve-alternative-image-formats)

Related features: #57

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [ ] Use case guidance (guide.md)
- [ ] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

PR #570, opened 2026-04-14
Add guide for deliver-optimized-decorative-images

taraojo:
#167 

This guide also includes a rename, as this was missed last time in the use case PR.


guides/user-experience/deliver-optimized-decorative-images/guide.md
```diff
@@ -0,0 +1,61 @@
...
+  - https://css-tricks.com/almanac/functions/i/image-set/
+  - https://css-tricks.com/using-performant-next-gen-images-in-css-with-image-set/
+  - https://uploadcare.com/blog/image-set-for-responsive-images
+  - https://web.dev/articles/preload-responsive-images
+  - https://web.dev/learn/design/responsive-images
+  - https://web.dev/articles/responsive-images
+---
+
+Delivering optimized decorative images via CSS improves perceived performance without sacrificing visual quality. By using the `image-set()` CSS function, you can provide the browser with multiple options for a single background or mask image. You can specify modern formats (like AVIF or WebP) alongside different resolutions (like `1x` and `2x`). The browser will dynamically select the smallest compatible image that provides the appropriate pixel density for the user's device.
+
```
rviscomi: It might be worth providing a cautionary note about avoiding CSS background images if it's the LCP image, for consistency with our Performance skill.

>DO declare the LCP image in standard HTML: Ensure the  tag is present in the raw HTML response so the preload scanner discovers it immediately.

cc @tunetheweb 

```suggestion
**CAUTION**: If the image is likely to be the Largest Contentful Paint (LCP) element (e.g., a large hero banner), be aware that images referenced in CSS via image-set() are not discoverable by the browser's preload scanner. This can significantly delay image loading and harm LCP. For LCP candidates, consider using a standard HTML <img> or <picture> tag instead.
```

tunetheweb: Let’s suggest preloading responsive images instead to get around this:

https://web.dev/articles/preload-responsive-images#image-set

Usually that’s easier than switching to `` or `<picture>`.

rviscomi: Are you sure? This guidance is for AI coding agents, so it'd be trivial for them to convert a background image to `` if that's truly the best way to serve it.

rviscomi: Ping @tunetheweb

tunetheweb: I don't think it's always truly the best way. Preload has advantages when the `` is buried way down HTML. But has the disadvantage that it can get out of sync (but by your same rational, if it's for AI coding, then that's less of a concern).

Let's go with your suggestion but with a little tweak:

```suggestion
**CAUTION**: If the image is likely to be the Largest Contentful Paint (LCP) element (e.g., a large hero banner), be aware that images referenced in CSS via image-set() are not discoverable by the browser's preload scanner. This can significantly delay image loading and harm LCP. For LCP candidates, consider using a standard HTML <img> or <picture> tag instead or alternatively, preloading the image as well using `<link rel=preload>` option with a `media` attribute.
```

guides/user-experience/deliver-optimized-decorative-images/guide.md
```diff
@@ -0,0 +1,61 @@
...
+### Implementation
+
+The `image-set()` function is used anywhere CSS expects an `<image>` value, most commonly in `background-image`, `content`, or `mask-image`. Note that while providing both the image format via `type()` and the resolution (like `1x` or `2x`) yields the best results, both of these arguments are optional. 
+
+```css
+.hero-banner {
```
tunetheweb: Given Rick's comment above, maybe using this for a "hero-banner" isn't the best example since hero banners often are LCP elements? Maybe this could just be `.gallery`?

guides/user-experience/deliver-optimized-decorative-images/guide.md
```diff
@@ -0,0 +1,61 @@
...
+  /* Provide multiple resolutions and formats using image-set() */
+  /* MANDATORY: Always order your formats from most optimized (AVIF) to least optimized (JPEG/PNG). 
+     The browser will stop at the first supported format. */
+  background-image: image-set(
+    url("hero.avif") type("image/avif") 1x,
+    url("hero-2x.avif") type("image/avif") 2x,
+    url("hero.webp") type("image/webp") 1x,
+    url("hero-2x.webp") type("image/webp") 2x,
+    url("hero.jpg") type("image/jpeg") 1x,
+    url("hero-2x.jpg") type("image/jpeg") 2x
```
tunetheweb: Same comment regarding the use of the term "hero"

guides/user-experience/deliver-optimized-decorative-images/guide.md
```diff
@@ -0,0 +1,61 @@
...
+.hero-banner {
+  /* MANDATORY: Fallback for browsers that do not support image-set() */
+  background-image: url("hero.jpg");
+  
+  /* Modern browsers will apply this and override the fallback */
+  background-image: image-set(
+    url("hero.avif") type("image/avif") 1x,
+    url("hero-2x.avif") type("image/avif") 2x,
+    url("hero.jpg") type("image/jpeg") 1x,
+    url("hero-2x.jpg") type("image/jpeg") 2x
```
tunetheweb: Ditto

guides/user-experience/deliver-optimized-decorative-images/guide.md
tunetheweb: https://bsky.app/profile/tunetheweb.com/post/3mk3jjwq3qs24

We’re so relevant 😁

malchata: This is making me wonder if we're not stressing the LCP pitfalls enough throughout the guide to ensure coding agents understand. The overlap between what might be considered a "decorative" image and an image that could be the LCP element is probably pretty decent. Worst case in my mind is that the generated artifacts beyond this guide (`expectations.md` and `demo.html`) might need to be regenerated.

Or I might be expanding this too much, but I could see the results impacting LCP across the web in these cases. It probably will anyway, but I would seek to minimize it as we're trying to do, just more so.

guides/user-experience/deliver-optimized-decorative-images/guide.md
rviscomi: Good call

guides/user-experience/deliver-optimized-decorative-images/guide.md
taraojo: Thanks for feedback! I've removed all references to hero images (excluding the caution) from the guide and demo. Do we think the caution is enough to stress the LCP pitfalls? Or maybe we need another test/expectation to check we're not applying this guidance to LCP elements?

tunetheweb (approved):

guides/user-experience/deliver-optimized-decorative-images/guide.md
tunetheweb: I think it's fine as is. Thanks for taking on board the feedback!

rviscomi (approved):
LGTM

PR #571, opened 2026-04-15
Fix Claude yolo flag

rviscomi:
https://code.claude.com/docs/en/permission-modes

> the --allow- variant adds the mode to the cycle without activating it

`--allow-dangerously-skip-permissions` is a half-flag that enables the option but doesn't turn it on. The canonical form is `--permission-mode bypassPermissions` and `--dangerously-skip-permissions` is the documented shortcut.


paulirish (approved):
thx! i fixed that elsewhere but not in this file. 
sorry for that tripping you up for a bit!

PR #576, opened 2026-04-15
Add design-token-reactivity guide

### LINKED ISSUE #163: Create guide and evals for the design-token-reactivity use case

knowler:
Resolves #163 

My recommended guidance is to not use container style queries for core/baseline experiences and to use it for feature that are a progressive enhancement instead (e.g. a UI density use preference that does not need to be a part of every cross-browser experience). I have included how to implement it as a progressive enhancement in the guide as well as how to create an adequate fallback experience. I only updated the demo to hide the toggle when style queries are not supported.

A challenge with this feature is that it’s not super clean to create a fallback for and, as explained in the guide, using a fallback defeats a lot of the purpose of using the feature (i.e. style queries help with maintainability and separation of concerns, but a fallback undermines that).

### DIFF
--- a/guides/user-experience/design-token-reactivity/demo.html
+++ b/guides/user-experience/design-token-reactivity/demo.html
@@ -2,9 +2,12 @@
 <html lang="en">
 <head>
   <style>
+		:root {
+			--style-queries-supported: check;
+		}
+
     .container {
       --density: compact;
-      /* No container-type required for style queries! */
     }
 
     .item {
@@ -23,6 +26,16 @@
         padding: 2.5rem;
       }
     }
+
+		#toggle-spacious {
+			display: none;
+		}
+
+		@container style(--style-queries-supported) {
+			#toggle-spacious {
+				display: revert;
+			}
+		}
   </style>
 </head>
 <body>
--- a/guides/user-experience/design-token-reactivity/demo.html
+++ b/guides/user-experience/design-token-reactivity/demo.html
@@ -2,9 +2,9 @@
 <html lang="en">
 <head>
   <style>
-		:root {
-			--style-queries-supported: check;
-		}
+    :root {
+      --style-queries-supported: check;
+    }
 
     .container {
       --density: compact;
@@ -27,15 +27,15 @@
       }
     }
 
-		#toggle-spacious {
-			display: none;
-		}
+    #toggle-spacious {
+      display: none;
+    }
 
-		@container style(--style-queries-supported) {
-			#toggle-spacious {
-				display: revert;
-			}
-		}
+    @container style(--style-queries-supported) {
+      #toggle-spacious {
+        display: revert;
+      }
+    }
   </style>
 </head>
 <body>
--- a/guides/user-experience/design-token-reactivity/expectations.md
+++ b/guides/user-experience/design-token-reactivity/expectations.md
@@ -1,0 +1,1 @@
+- Only use container style queries with custom properties.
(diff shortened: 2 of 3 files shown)


rviscomi:
@knowler are you still making changes to this PR or is it ok to mark it as ready for review?

knowler:
@rviscomi I could use some help with the expectations. It might be my interpretation, but I don’t think the mechanism to change the token value is a necessary feature, so I don’t have any expectations for that.

Also, I’m realizing now that I had intended on going back to work using nesting into the guide. Perhaps, I’ll follow up with that when I finish `usage-aware-component-variations` which will likely use the same wording.

knowler:
I’ve added more expectations which should suffice, so this is ready for review from my end.

rviscomi (commented):
I've been struggling with this one as a non-expert but I think I get it now.

The fallback strategy section is definitely complex. It tries to cover several contingencies: core vs not-core, `@scope` support, progressive enhancement or not, etc. It seems like the most straightforward fallback strategy is to simply discourage using style queries if it doesn't meet the target and unconditionally fall all the way back to data attribute selectors.

Does that also address the concerns you raised?

knowler:
@rviscomi I’ve made some updates:

- I removed the `@scope` fallback as it doesn’t bring a whole lot of benefit and just complicates things.
- I emphasized the limitation for nesting elements with `data-density` for the fallback approach.
- I separated the basic fallback from the progressively enhanced fallback using style queries (this enhancement would make nested cases work).

- - -

There is a way to get around the limitation, but I think it complicates things significantly.

At its core it’s quite simple: instead of applying `data-density` to a container and then styling with a descendant selector (e.g. `[data-density="compact"] .component-that-consumes-density`), each child element that consumes the density token needs to have `data-density` set for it to avoid the order-of-appearance limitations (i.e. because an element should only have one `data-density` attribute set for it vs. needing to reconcile potentially multiple from ancestors).

Often in a design system implemented with React, this is handled with React Context. For example, each component that consumes the design token would be a context consumer and then at a higher level in the tree, there’d be a component that provides context for that design token. Each individual component could override the design token value from the context.

The rendered HTML looks like this:

```html
<div class="component-that-provides-density">
  <!-- presumably these components would be setting their density from context -->
  <div class="component-that-consumes-density" data-density="compact"></div>
  <div class="component-that-consumes-density" data-density="compact"></div>
  <!-- presumably this component would be overriding the density from context -->
  <div class="component-that-consumes-density" data-density="spacious"></div>
</div>
```

Notice that there’s no information in DOM about the density for the provider or even which component is overriding the context’s density… all of that lives in React. One could encode that information into the HTML, but its not necessary for the CSS to work.

So, because you effectively need a framework to create a fallback that comes close to mimicking style queries, this is why I’ve opted to just include the fallback with limitations and to discourage for application features that can’t be considered a progressive enhancement.

rviscomi (approved):
LGTM

Issue #237, opened 2026-03-06
Create guide and evals for the prevent-incorrect-theme-flash use case


paulirish:
Prevent a flash of the incorrect color theme before page styles are loaded, for example a flash of a white background before the dark theme CSS is applied.

Affected web-feature IDs: [color-scheme](https://webstatus.dev/features/color-scheme)

Use case subdir: [guides/performance/prevent-incorrect-theme-flash](https://github.com/GoogleChrome/guidance/tree/main/guides/performance/prevent-incorrect-theme-flash)

Related features: #42

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [ ] Use case guidance (guide.md)
- [ ] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

PR #580, opened 2026-04-16
feat: create guide and evals for prevent-incorrect-theme-flash

castastrophe:
Progress on #237

## Summary
This PR adds the guidance, demo implementation, and evaluation expectations for the `prevent-incorrect-theme-flash` use case. This use case addresses the common issue of a "flash of white" during initial page load for users with dark mode enabled, by leveraging the `color-scheme` feature.

Key additions:
- **`guide.md`**: Provides clear, imperative guidance on using the `<meta name="color-scheme">` tag for immediate browser hints and the `color-scheme` CSS property for ongoing style consistency. Includes details on inheritance and "themed islands."
- **`demo.html`**: A reference implementation demonstrating best practices, including a nested section with its own color scheme.
- **`expectations.md`**: A set of mandatory assertions for evaluation, focusing on the presence and correctness of both the meta tag and CSS property.

## Test plan
- Verified the `demo.html` locally (simulated with standard browser behavior).
- Ensured all mandatory requirements from the `project-guides` and `project-evals` skills are met.
- Manually checked for clarity in technical terminology (distinguishing between meta tag and CSS property).


rviscomi (commented):
Our fault for not catching this sooner, but it seems like this use case has significant overlap with [browser-ui-color-theme](https://github.com/GoogleChrome/guidance/blob/main/guides/user-experience/browser-ui-color-theme/guide.md?plain=1), with the main difference being the localStorage override here.

From your perspective, are these use cases distinct enough to warrant staying separate? Or could we merge the localStorage guidance into browser-ui-color-theme? (either way, this should count towards your # of guides)

castastrophe:
@rviscomi Ah yes, good catch. I think we could merge them and incorporate the note about localStorage. We'll probably want to make sure the expectations are expanded as well if needed. I can pick that up if you want me to.

rviscomi:
That'd be great, thanks!

rviscomi:
I'll close this PR and delete the use case

PR #584, opened 2026-04-16
Update expectations for scroll-progress-indicator

philipwalton:
While I was creating slides for my talk, I took a screenshot of the `scroll-progress-indicator` evals, and in my script I mentioned that the evals test that the generated output uses SDA instead of scroll event listeners. BUT then I noticed that the evals in the screenshot do not actually check to ensure scroll event listeners aren't used, so I figured we should add that.


rviscomi (approved):
LGTM

Pass Rate - Unguided: 86%, Guided: 100%

PR #590, opened 2026-04-17
A few fixes for `animate-element-entry-exit` guide

### LINKED ISSUE #229: Create guide and evals for the animate-element-entry-exit use case

LeaVerou:
Fixes for some of the more straightforward issues in my review (https://github.com/GoogleChrome/guidance/issues/229#issuecomment-4269057452)

- Replaced `allow-discrete` in shorthand with `transition-behavior` longhand and added guidance recommending against shorthand use
- Replaced `transform: translate()` with more composable `translate` property
- Added 2s failsafe timeout so that elements don't hang around forever if a looping animation is applied
- Reset `display` to stylesheet value instead of hardcoding `block`

### DIFF
--- a/guides/user-experience/animate-element-entry-exit/guide.md
+++ b/guides/user-experience/animate-element-entry-exit/guide.md
@@ -124,7 +124,7 @@ if (!supportsModernTransitions) {
 
 ```javascript
 // To show:
-el.style.display = 'block';
+el.style.display = '';
 requestAnimationFrame(() => {
   requestAnimationFrame(() => {
     el.classList.remove('hidden');
--- a/guides/user-experience/animate-element-entry-exit/guide.md
+++ b/guides/user-experience/animate-element-entry-exit/guide.md
@@ -56,18 +56,18 @@ To animate an element when toggling its visibility via an attribute (e.g., `hidd
 @media (prefers-reduced-motion: reduce) {
   .card {
     /* Disable movement and shorten duration for a simple fade */
-    translate: 0;
+    translate: none;
     transition-duration: 0.1s;
   }
 
   @starting-style {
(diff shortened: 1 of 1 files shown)


castastrophe:
Looks like pretty solid updates but a note on this one:
> Replaced allow-discrete in shorthand with transition-behavior longhand and added guidance recommending against shorthand use

@malchata had suggested demonstrating both shorthand _and_ longhand examples. My coding preference is to only use long-hand for systems work (but shorthand if it's just a personal project or standalone page).

LeaVerou:
Let's wait for @malchata then, there could be a factor I'm missing. 

Just to clarify — *in general* including examples for both is good practice. This is an exception because it affects how unsupporting browsers parse this, and there are many cases where we want them to parse the rest of the `transition` without having to specify it twice (once with and once without the `allow-discrete`).

malchata:
I agree with with @LeaVerou—both are best in practice. However, I think if the shorthand is an isolated case, and doesn't require additional context, I think that can work here.

castastrophe (approved):

PR #592, opened 2026-04-17
#335 group-element-transitions

### LINKED ISSUE #335: Create guide and evals for the group-element-transitions use case
Transition a group of similar elements simultaneously using the same transition logic, such as removing a product from a shopping cart and having all the other products animate into their new positions.

Affected web-feature IDs: [view-transitions](https://webstatus.dev/features/view-transitions), [view-transition-class](https://webstatus.dev/features/view-transition-class)

Use case subdir: [guides/ui-behaviors/group-element-transitions](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-behaviors/group-element-transitions)

Related features: #87 #86

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

jamesnw:
Addresses #335


guides/user-experience/group-element-transitions/demo.html
```diff
@@ -1,117 +1,159 @@
...
+        animation-timing-function: ease-in-out;
+      }
+      @media (prefers-reduced-motion: reduce) {
+        ::view-transition-group(.list-item) {
+          animation: none !important;
+        }
+        ::view-transition-old(.list-item){
+          display: none;
+        }
+      }
```
jamesnw: I would have expected disabling animations or setting an `animation-duration: 0ms` to handle this, but removed items were staying in the snapshot until the view transition finished.

guides/user-experience/group-element-transitions/demo.html
```diff
@@ -1,117 +1,159 @@
...
+        padding: 4px 10px;
+        font-size: 12px;
+        background: #cc3333;
+      }
+    </style>
+  </head>
+
+  <body>
+    <h1>Group Element Transitions</h1>
+    <button onclick="add()">Add Item</button>
```
rviscomi, line 99: I expected to be able to spam the button with lots of clicks and add lots of items, but it doesn't seem to do anything while the items are animating. It's probably a browser limitation, but are there any best practices for styling UI that is temporarily non-interactive while the transitions are in flight?

guides/user-experience/group-element-transitions/demo.html
```diff
@@ -1,117 +1,159 @@
...
+      }
+
+      ::view-transition-group(.list-item) {
+        animation-duration: 0.5s;
+        animation-timing-function: ease-in-out;
+      }
+      @media (prefers-reduced-motion: reduce) {
+        ::view-transition-group(.list-item) {
+          animation: none !important;
+        }
```
rviscomi: I enabled prefers-reduced-motion and tried my "Add Item" button spamming test, but it seems like there's still an invisible view transition delaying the clicks from happening.

Gemini suggested this change, which is also recommended in the [Chrome doc](https://developer.chrome.com/docs/web-platform/view-transitions/same-document#react_to_the_reduced_motion_preference). WDYT?

(We'd also need to update the guide accordingly)
```suggestion
        ::view-transition-group(*), 
        ::view-transition-old(*), 
        ::view-transition-new(*) {
          animation: none !important;
        }
```

guides/user-experience/group-element-transitions/guide.md
```diff
@@ -4,4 +4,111 @@ description: Transition a group of similar elements simultaneously using the sam
...
+### Implementation steps
+
+1. **Assign unique names and a shared class**
+
+Each element that needs to be tracked individually during a transition must have a unique `view-transition-name`.
+
+```html
+<!-- Mandatory: Each element must have a unique view-transition-name -->
+<li style="view-transition-name: item-1" class="item">Item 1</li>
+<li style="view-transition-name: item-2" class="item">Item 2</li>
```
malchata, line 24: Are inline styles a pattern we want to be using here?

guides/user-experience/group-element-transitions/guide.md
```diff
@@ -4,4 +4,111 @@ description: Transition a group of similar elements simultaneously using the sam
...
+@media (prefers-reduced-motion: reduce) {
+  ::view-transition-group(.list-item) {
+    animation: none !important;
+  }
+}
+```
+
+3. **Optional: Define entry and exit animations**
+
+Use the `:only-child` selector to add specific transitions to the elements that are added or removed. `::view-transition-new()` and `::view-transition-old()` pseudo-elements are children of a `::view-transition-image-pair()` psuedo-element, so we can determine it is an added or removed element if it is the only child.
```
malchata:
```suggestion
Use the `:only-child` selector to add specific transitions to the elements that are added or removed. `::view-transition-new()` and `::view-transition-old()` pseudo-elements are children of a `::view-transition-image-pair()` pseudo-element, so we can determine it is an added or removed element if it is the only child.
```

guides/user-experience/group-element-transitions/guide.md
```diff
@@ -4,4 +4,111 @@ description: Transition a group of similar elements simultaneously using the sam
...
+
+```css
+/* A `::view-transition-new()` element is the only child if it wasn't present before the view transition, so it is an added element. */
+::view-transition-new(.list-item):only-child {
```
malchata, line 63: Won't the UA default duration be set if `animation-duration` isn't specified? Or is that a non-issue for this guide?

guides/user-experience/group-element-transitions/guide.md
```diff
@@ -4,4 +4,111 @@ description: Transition a group of similar elements simultaneously using the sam
...
+  animation-name: slide-in;
+}
+/* A `::view-transition-old()` element is the only child if it isn't present after the view transition, so it is a removed element. */
+::view-transition-old(.list-item):only-child {
```
malchata, line 69: Same question here as for `::view-transition-new`

guides/user-experience/group-element-transitions/guide.md
```diff
@@ -4,4 +4,111 @@ description: Transition a group of similar elements simultaneously using the sam
...
+```
+
+3. **Optional: Define entry and exit animations**
+
+Use the `:only-child` selector to add specific transitions to the elements that are added or removed. `::view-transition-new()` and `::view-transition-old()` pseudo-elements are children of a `::view-transition-image-pair()` psuedo-element, so we can determine it is an added or removed element if it is the only child.
+
+```css
+/* A `::view-transition-new()` element is the only child if it wasn't present before the view transition, so it is an added element. */
+::view-transition-new(.list-item):only-child {
+  animation-name: slide-in;
```
malchata, line 64: I see there are `@keyframes` definitions for `animation-name` in `demo.html`, but shouldn't they be here in the guide for agents?

guides/user-experience/group-element-transitions/guide.md
```diff
@@ -4,4 +4,111 @@ description: Transition a group of similar elements simultaneously using the sam
...
+}
+/* A `::view-transition-old()` element is the only child if it isn't present after the view transition, so it is a removed element. */
+::view-transition-old(.list-item):only-child {
+  animation-name: slide-out;
```
malchata, line 70: Same concern as above for missing `@keyframes` definition.

malchata (changes requested):
Small stuff, but the main concern is missing `@keyframes` definitions in `guide.md` that are present in `demo.html`.

guides/user-experience/group-element-transitions/demo.html
jamesnw: This is broader than needed for this use case, but in most cases, I'm thinking we don't need to be more specific, so I'll use this.

guides/user-experience/group-element-transitions/guide.md
jamesnw: Yeah, I think that's the best way, since there are an unknown number of items, and the view-transition-name has to be stable.

guides/user-experience/group-element-transitions/demo.html
jamesnw: Resolved by disabling the root transition.

guides/user-experience/group-element-transitions/guide.md
jamesnw: Added

guides/user-experience/group-element-transitions/guide.md
jamesnw: Yeah, it defaults to the UA duration, which is 1/2 a second. Is there a reason to be explicit, or different?

guides/user-experience/group-element-transitions/demo.html
jamesnw: I was able to make it interactive, and added a section to the guide on that.

guides/user-experience/group-element-transitions/guide.md
malchata: All good 😁  When I see `style=`, I tend to ask questions!

guides/user-experience/group-element-transitions/guide.md
malchata: It might be worth making a note in a comment. Something like:

```
/* Specify an animation duration if you want something different than the UA default of 0.5s */
animation-duration: 1s;
```

jamesnw: Added!

rviscomi (approved):
LGTM

Tested in Chrome and Firefox, with and without `prefers-reduced-motion`

PR #593, opened 2026-04-17
Fix mangled markdown

LeaVerou:
Code block was left open consuming the rest of the page. Fixed it and removed duplicate guideline.


paulirish:
Great catch thx!

added regression tests for this and some more: https://github.com/GoogleChrome/guidance/pull/594

paulirish (approved):

PR #595, opened 2026-04-17
build: generate THIRD_PARTY_NOTICES for skills-cli distribution

paulirish:


micahjo7 (approved):
I reviewed the changes and verified that the build works and tests pass. The THIRD_PARTY_NOTICES file is generated correctly.\n\nA few minor notes (not blocking):\n- The `allowedLicenses` list is hardcoded. If we add a dependency with a license like `Unlicense` or a dual license string like `MIT OR Apache-2.0`, the build will fail and we'll need to update the list.\n- Some dependencies (like TensorFlow) didn't have their license text included because it wasn't found in the package root with the checked filenames. This is likely fine as the license name is still listed.\n\nApproving as it works as intended.

paulirish:
fwiw impl was inspired/pulled by https://github.com/GoogleChrome/lighthouse/blob/44e76c97fc80cfd23376b1bf633ff0a6d462f39d/build/build-bundle-mcp.js#L326-L336  and the dtmcp rollup listed below.



> Some dependencies (like TensorFlow) didn't have their license text included because it wasn't found in the package root with the checked filenames. 

yup. 


>  The `allowedLicenses` list is hardcoded. If we add a dependency with a license like `Unlicense` or a dual license string like `MIT OR Apache-2.0`, the build will fail and we'll need to update the list.\

Yeah, good call. I stole the allowedLicenses bit from https://github.com/ChromeDevTools/chrome-devtools-mcp/blob/0a6aaa52ebacb2db834ffa437863e5844aa3730b/rollup.config.mjs#L34

and I feel okay about it.   But it certainly would be better if we got notified somehow, since now this is an automated process.

PR #596, opened 2026-04-17
publish: transition to tag-based versioning

paulirish:
Transition from file-based versioning  to Git tag-based versioning for skills distribution. 

This eliminates noisy `chore: bump version [skip ci]` commits on the `main` branch.  I also realized we had a race condition between when the commit landed .. it wasn't guaranteed to land after the commit that was published. 

- **publish-skills.ts**: Derives the next version from the latest Git tag (`v*.*.*`) using `git describe`. Falls back to `package.json` version if no tags are found.
- **build-dist.ts**: Accepts the derived version as an argument and injects it into distribution files (`package.json`, manifests) during the build process, ensuring the generated README reflects the correct version.
- **publish.yml**: Removed automated version bump commits. The release process is now fully driven by the script, including creating and pushing the release tag.
- **Tests**: Added unit tests for `getNextVersion` to verify tag resolution and fallback behavior, refactoring the function to use dependency injection for clean and idiomatic mocking.
- **Moved version injection into the build process (`build-dist.ts`)**: The distribution's `README.md` is auto-generated by reading the version from `package.json` during the build. To ensure the README reflects the *new* version (and not the old template version), we must inject the derived version into the `dist/` files *during* the build process (right after copying template files but before generating the README), rather than after the build completes.

Also caught a bug.. build-dist would exit with 0 if there was an error. ruh roh.  fixed that so fatal problems are loud and will fail the build/publish.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -70,10 +70,4 @@ jobs:
       #     npm publish --dry-run --provenance --access public --registry https://registry.npmjs.org/
       # =========================================================================
 
-      - name: Create and Push Tag
-        run: |
-          VERSION=$(node -p "require('./dist/skills-cli/package.json').version")
-          TAG="v$VERSION"
-          echo "Creating tag $TAG"
-          git tag $TAG
-          git push origin $TAG
+
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -70,15 +70,10 @@ jobs:
       #     npm publish --dry-run --provenance --access public --registry https://registry.npmjs.org/
       # =========================================================================
 
-      - name: Commit and Push Version Bump
+      - name: Create and Push Tag
         run: |
-          git add serving/skills-cli/template/
-          if ! git diff --quiet --exit-code --cached; then
-            git commit -m "chore: bump version [skip ci]"
-
-            # Fetch latest changes and rebase our commit on top of them
-            git fetch origin main
-            git rebase origin/main
-
-            git push origin main
-          fi
+          VERSION=$(node -p "require('./dist/skills-cli/package.json').version")
+          TAG="v$VERSION"
+          echo "Creating tag $TAG"
+          git tag $TAG
+          git push origin $TAG
--- a/serving/skills-cli/build-dist.ts
+++ b/serving/skills-cli/build-dist.ts
@@ -41,47 +41,31 @@ async function acquireLock(lockFilePath: string) {
 function updateVersionsInDir(publishCliDir: string, newVersion: string) {
   // Gemini
   const geminiPath = path.join(publishCliDir, "gemini-extension.json");
-  try {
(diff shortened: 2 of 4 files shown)


micahjo7:
thank you! should we update the version in the files with a placeholder? e.g. `0.0.0-dev` in these files https://github.com/GoogleChrome/guidance/commit/9d2f5653e7f02122779044547fc7bf18351aba7b

we could do that in follow-up or do this before merging:

```
git tag v0.0.26
git push origin v0.0.26
```

micahjo7 (approved):

paulirish:
> thank you! should we update the version in the files with a placeholder? e.g. `0.0.0-dev` in these files [9d2f565](https://github.com/GoogleChrome/guidance/commit/9d2f5653e7f02122779044547fc7bf18351aba7b)
> 
> we could do that in follow-up or do this before merging:

i dont think we need to...?  because of the fallback.

```ts
  try {
    // Get the latest tag that looks like v*.*.*
    const latestTag = getLatestTag();
    currentVersion = latestTag.startsWith('v') ? latestTag.slice(1) : latestTag;
    console.log(`Found latest tag: ${latestTag}`);
  } catch (e) {
    console.warn("No version tags found, falling back to package.json version.");
    const vscodePath = path.join(SKILLS_CLI_TEMPLATE_DIR, "package.json");
    const vscodeData = JSON.parse(await fs.readFile(vscodePath, 'utf8'));
    currentVersion = vscodeData.version;
  }
```

--

are you pitching that we basically backfill git tags for our releases so far? or something else.

micahjo7:
> > thank you! should we update the version in the files with a placeholder? e.g. `0.0.0-dev` in these files [9d2f565](https://github.com/GoogleChrome/guidance/commit/9d2f5653e7f02122779044547fc7bf18351aba7b)
> > we could do that in follow-up or do this before merging:
> 
> i dont think we need to...? because of the fallback.
> 
> ```ts
>   try {
>     // Get the latest tag that looks like v*.*.*
>     const latestTag = getLatestTag();
>     currentVersion = latestTag.startsWith('v') ? latestTag.slice(1) : latestTag;
>     console.log(`Found latest tag: ${latestTag}`);
>   } catch (e) {
>     console.warn("No version tags found, falling back to package.json version.");
>     const vscodePath = path.join(SKILLS_CLI_TEMPLATE_DIR, "package.json");
>     const vscodeData = JSON.parse(await fs.readFile(vscodePath, 'utf8'));
>     currentVersion = vscodeData.version;
>   }
> ```
> 
> --
> 
> are you pitching that we basically backfill git tags for our releases so far? or something else.

ah, I meant so that we don't have the `0.0.26` hardcoded into those files, to prevent confusion. 
but, yeah this is a nit, doesn't need to be addressed

paulirish:
success!    https://github.com/GoogleChrome/guidance/releases/tag/v0.0.27

I think!  https://github.com/GoogleChrome/modern-web-guidance doesnt show it though..

PR #598, opened 2026-04-17
Add grader and resources for discipline SKILL evals (forms & performance)

micahjo7:
Started by constructing comprehensive list of expectations that attempt to capture everything in the guides comprehensively. Then created `demo.html` for each.

Ran `gd dev` for each.

This resulted in poor performance (`~50%` pass rate on unguided/guided with almost no delta).

Tweaked the prompts so they explicitly ask the agent to use the SKILLs if available.

Iterated and updated expectations and grader files to remove brittle checks and be flexible towards simpler apps like `daily-grind` (introducing dynamic checks based on what the app contains).

Results:

image
image

https://googlechrome.github.io/guidance/?tests=discipline-skills-2026-04-17T16-02-38%7C%7C%7Cremote%2Cdiscipline-skills-2026-04-17T15-57-23%7C%7C%7Cremote

*Main takeaway: it is hard to get the agent to implement everything within the SKILL because of the large amount of content, especially without giving some direction within the prompt itself toward those specific areas. Asking the agent to use the SKILL explicitly also gave a significant boost to the guided percentage.*


guides/performance/demo.html
```diff
@@ -0,0 +1,239 @@
...
+        taskStatus.textContent = 'Failed to load feature.';
+      }
+      
+      heavyTaskBtn.disabled = false;
+    });
+
+    // Service Worker Registration
+    if ('serviceWorker' in navigator) {
+      window.addEventListener('load', () => {
+        navigator.serviceWorker.register('sw.js')
```
paulirish: oooooooooof. can i suggest that we ignore service workers for now? i'm pretty sure we'll end up debugging weird things about them for a while. 

i see the skill has text about them.. and .. obviously if we add text like "Don't use a service worker" to the task.. we can't evaluate that text but......  that's what I'm suggesting. :)

paulirish (approved):
nice job!

making the un-eval-able actually fairly evalable.

i have hesitations on the SW, but the rest looks nice

guides/performance/demo.html
micahjo7: thanks! removed the service worker evals

Issue #32, opened 2026-02-19
Popover


rviscomi:
The popover HTML attribute creates an overlay to display content on top of other page content. Popovers can be shown declaratively using HTML, or using the showPopover() method.

### MDN Documentation
- [Popover API](https://developer.mozilla.org/docs/Web/API/Popover_API)

### Chrome Documentation

<details>
<summary>View 33 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/anchor-positioning-api
- https://developer.chrome.com/en/blog/chrome-2024-recap
- https://developer.chrome.com/en/blog/command-and-commandfor
- https://developer.chrome.com/en/blog/css-ui-ecommerce-popover
- https://developer.chrome.com/en/blog/entry-exit-animations
- https://developer.chrome.com/en/blog/introducing-popover-api
- https://developer.chrome.com/en/blog/new-in-web-ui-io-2024
- https://developer.chrome.com/en/blog/new-in-web-ui-io-2025-recap
- https://developer.chrome.com/en/blog/pop-ups-theyre-making-a-resurgence
- https://developer.chrome.com/en/blog/popover-hint
- https://developer.chrome.com/en/blog/tether-elements-to-each-other-with-css-anchor-positioning
- https://developer.chrome.com/en/blog/web-at-io25
- https://developer.chrome.com/en/blog/what-is-the-top-layer
- https://developer.chrome.com/en/docs/css-ui/anchor-positioning-api
- https://developer.chrome.com/en/docs/extensions/develop/ui/add-popup
- https://developer.chrome.com/en/docs/extensions/develop/ui
- https://developer.chrome.com/en/docs/extensions/get-started/tutorial/service-worker-events
- https://developer.chrome.com/en/docs/extensions/reference/api/action
- https://web.dev/en/articles/baseline-in-action-dialog-popover
- https://web.dev/en/articles/building/a-split-button-component
- https://web.dev/en/articles/centering-in-css
- https://web.dev/en/articles/components-howto-tooltip
- https://web.dev/en/blog/baseline-digest-feb-2025
- https://web.dev/en/blog/interop-2024-midyear
- https://web.dev/en/blog/interop-2024-wrapup
- https://web.dev/en/blog/new-in-the-web-io2024
- https://web.dev/en/blog/popover-api
- https://web.dev/en/blog/popover-baseline
[...]

PR #601, opened 2026-04-18
Create a navigation-drawer use case and guide

philipwalton:
This PR is an alternate take on #392, one of the use cases for Popover in #32. It updates the name of the use case as well as the implementation to be a bit more robust.


guides/user-experience/navigation-drawer/demo.html
```diff
@@ -0,0 +1,312 @@
...
+
+    initDrawer();
+
+
+  </script>
+</head>
+<body>
+  <div class="Drawer" id="drawer" popover="manual">
+    <div class="Drawer-scroller">
+      <nav class="Drawer-sheet">
```
rviscomi: Should this have a `tabindex` to demonstrate the "Keyboard focus is moved inside the drawer after it opens" expectation?
```suggestion
      <nav class="Drawer-sheet" tabindex="-1">
```

guides/user-experience/navigation-drawer/guide.md
```diff
@@ -0,0 +1,358 @@
...
+  @keyframes fade-drawer-backdrop {
+    /* Scroll position 0 = drawer fully open = backdrop visible. */
+    0% { --drawer-backdrop: 1 }
+    /* Scroll position 100% = drawer fully closed = backdrop hidden. */
+    100% { --drawer-backdrop: 0 }
+  }
+}
+```
+
+### 3. Open and close the drawer
```
rviscomi, line 209: WDYT about recommending some sort of smooth-scroll escape hatch if `prefers-reduced-motion` is set?

guides/user-experience/navigation-drawer/guide.md
```diff
@@ -0,0 +1,358 @@
...
+Detect with `CSS.supports('animation-timeline: scroll()')` and write `--drawer-backdrop` from a `scroll` event listener if not supported. The CSS `@supports` block in step 2 ensures the keyframes never apply in unsupported browsers, so the JavaScript value is the only writer.
+
+```js
+if (!CSS.supports('animation-timeline: scroll()')) {
+  scroller.addEventListener('scroll', () => {
+    // Same mapping as the @keyframes: 0 scroll = 1 (open),
+    // sheet-width scroll = 0 (closed).
+    const ratio = 1 - scroller.scrollLeft / sheet.offsetWidth;
+    drawer.style.setProperty('--drawer-backdrop', ratio);
+  });
```
rviscomi, line 325: Would this be a good candidate for a passive scroll listener?

```suggestion
  scroller.addEventListener('scroll', () => {
    // Same mapping as the @keyframes: 0 scroll = 1 (open),
    // sheet-width scroll = 0 (closed).
    const ratio = 1 - scroller.scrollLeft / sheet.offsetWidth;
    drawer.style.setProperty('--drawer-backdrop', ratio);
  }, { passive: true });
```

rviscomi (commented):
Tested the demo in Chrome, Firefox, and Chrome for Android. A few suggestions courtesy of the [project-guide-validation](https://github.com/GoogleChrome/guidance/blob/main/.agents/skills/project-guide-validation/SKILL.md) skill.

Haven't looked at the evals yet since these suggestions might change some of the expectations / tests.

guides/user-experience/navigation-drawer/demo.html
philipwalton: Ahh, yeah, this is a copy/paste mistake. I accidentally put tabindex on the `<header>` element below it. Updated. It looks like it's correct in the guide though.

guides/user-experience/navigation-drawer/guide.md
philipwalton: I'm not sure. It'd be great to get an a11y person's take on this.

I did test this on my phone (Android and iOS), and at least on the Apple and Google native apps I tested, sheets and drawers still slide in/out with the reduce motion setting on. Other motion is reduced though.

i don't have a good sense for what is considered acceptable motion bs. unacceptable motion.

guides/user-experience/navigation-drawer/guide.md
philipwalton: My understanding is that passive event listeners aren't needed on scroll events (just touch/wheel event) since scroll events cannot be cancelled.

Also, this audit was removed from Lighthouse last year: https://developer.chrome.com/docs/lighthouse/best-practices/uses-passive-event-listeners

philipwalton:
Sorry for the delay, I missed the notification for this...

I've addressed your feedback. I'm still not sure about the prefers-reduced-motion one, but it was easy enough to implement that, and we can always revert if this is considered "acceptable" motion.

Also, Gemini decided to update the grader file on its own, so I included that here. It looks like `gd dev` is passing, so hopefully that works.

rviscomi:
Thanks those changes LGTM. I ran the evals and both guided and unguided scored 0% so I'll look into fine tuning it.

rviscomi (approved):
LGTM with fine tuning

I think this was just a case of tool evasion; the agent thought it knew how to make a navigation drawer and failed to use the skill. I rephrased the prompt a bit without being overly prescriptive about the solution, and the guided evals are now passing at 100%.

FWIW I did have to change my test agent from Gemini Pro to Flash due to API issues

# Evaluation Results


| Group | Pass Rate | Test Runs |
|---|---|---|
| **Unguided** | 6% (1/17) | 1 |
| **Guided** | 100% (17/17) | 1 |

## TASK - NAVIGATION-DRAWER - UNGUIDED (Median: 6%)

**Pass rates across runs:** 6%

### Run 1 Details (1/17)

| Status | Expectation |
|---|---|
| ✅ | The drawer is not visible when the page first loads |
| ❌ | The menu trigger button has aria-expanded=&quot;false&quot; when the drawer is closed |
| ❌ | Clicking the menu trigger button makes the drawer visible |
| ❌ | When the drawer is open, the menu trigger button&#039;s aria-expanded attribute is true |
| ❌ | The drawer element uses the popover API with popover=&quot;manual&quot; |
| ❌ | When the drawer is open, a dimmed backdrop (the ::backdrop pseudo-element) is visible behind it that covers the rest of the page |
| ❌ | The drawer&#039;s navigation sheet does not span the full width of the viewport when open, leaving a portion of the dimmed backdrop visible alongside it |
| ❌ | Clicking the dimmed backdrop area outside the navigation sheet closes the drawer |
| ❌ | Pressing the Escape key while the drawer is open closes the drawer |
| ❌ | The main page content has the inert attribute applied while the drawer is open |
| ❌ | The main page content no longer has the inert attribute after the drawer is closed |
| ❌ | Keyboard focus is moved inside the drawer after it opens |
| ❌ | After the drawer is closed, the menu trigger button&#039;s aria-expanded attribute returns to false |
| ❌ | The drawer&#039;s horizontal scroll container uses CSS scroll snap with scroll-snap-type: x mandatory |
| ❌ | The drawer&#039;s horizontal scroll container uses scroll-behavior: smooth only when the user has not requested reduced motion (e.g., via @media (prefers-reduced-motion: no-preference)) |
| ❌ | Horizontally scrolling the drawer&#039;s scroll container all the way to its end position closes the drawer |
| ❌ | The drawer&#039;s backdrop opacity decreases as the drawer&#039;s scroll container is scrolled toward the closed position |

## TASK - NAVIGATION-DRAWER - GUIDED (Median: 100%)

**Pass rates across runs:** 100%

### Run 1 Details (17/17)

| Status | Expectation |
|---|---|
| ✅ | The drawer is not visible when the page first loads |
| ✅ | The menu trigger button has aria-expanded=&quot;false&quot; when the drawer is closed |
| ✅ | Clicking the menu trigger button makes the drawer visible |
| ✅ | When the drawer is open, the menu trigger button&#039;s aria-expanded attribute is true |
| ✅ | The drawer element uses the popover API with popover=&quot;manual&quot; |
| ✅ | When the drawer is open, a dimmed backdrop (the ::backdrop pseudo-element) is visible behind it that covers the rest of the page |
| ✅ | The drawer&#039;s navigation sheet does not span the full width of the viewport when open, leaving a portion of the dimmed backdrop visible alongside it |
| ✅ | Clicking the dimmed backdrop area outside the navigation sheet closes the drawer |
| ✅ | Pressing the Escape key while the drawer is open closes the drawer |
| ✅ | The main page content has the inert attribute applied while the drawer is open |
| ✅ | The main page content no longer has the inert attribute after the drawer is closed |
| ✅ | Keyboard focus is moved inside the drawer after it opens |
| ✅ | After the drawer is closed, the menu trigger button&#039;s aria-expanded attribute returns to false |
| ✅ | The drawer&#039;s horizontal scroll container uses CSS scroll snap with scroll-snap-type: x mandatory |
| ✅ | The drawer&#039;s horizontal scroll container uses scroll-behavior: smooth only when the user has not requested reduced motion (e.g., via @media (prefers-reduced-motion: no-preference)) |
| ✅ | Horizontally scrolling the drawer&#039;s scroll container all the way to its end position closes the drawer |
| ✅ | The drawer&#039;s backdrop opacity decreases as the drawer&#039;s scroll container is scrolled toward the closed position |

PR #602, opened 2026-04-18
docs: update CSS skill with review feedback

patrickkettner:
Addresses review comments on skills-drafts/css/SKILL.md: Clarified :focus-visible and rem usage; added container queries, DVH/DVW, aspect-ratio, and smoother animations.

### DIFF
--- a/skills-drafts/css/SKILL.md
+++ b/skills-drafts/css/SKILL.md
@@ -232,7 +232,7 @@ Use fluid and accessible typography that respects viewport widths and browser zo
 - **DO** use `@container` queries to create component-driven responsive layouts that adapt to their parent container's size rather than the viewport.
 - **DO** use dynamic viewport units (`dvh`, `dvw`) instead of `vh`/`vw` to prevent layout breakage when mobile browser UI elements (like address bars) appear or disappear.
 - **DO** use `aspect-ratio` for media elements (like `<img>` and `<video>`) to reserve space during loading and prevent Cumulative Layout Shift (CLS).
-- **DON'T** use `px` for font-size. Use `rem` to honor the user's browser font-size preferences (root font size), as `px` values are absolute and may ignore user-defined defaults.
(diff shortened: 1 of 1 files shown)


skills-drafts/css/SKILL.md
```diff
@@ -228,7 +229,10 @@ Use fluid and accessible typography that respects viewport widths and browser zo
 - **DO** use standalone `min()` and `max()` to constrain values responsively.
 - **DO** use `text-wrap: balance` for short headlines (up to 4-6 lines) to prevent uneven orphans.
 - **DO** use `overflow-wrap: break-word` (or `anywhere`) to contain long URLs.
-- **DON'T** use `px` for font-size. Use `rem` to honor user default zoom.
+- **DO** use `@container` queries to create component-driven responsive layouts that adapt to their parent container's size rather than the viewport.
+- **DO** use dynamic viewport units (`dvh`, `dvw`) instead of `vh`/`vw` to prevent layout breakage when mobile browser UI elements (like address bars) appear or disappear.
+- **DO** use `aspect-ratio` for media elements (like `<img>` and `<video>`) to reserve space during loading and prevent Cumulative Layout Shift (CLS).
+- **DON'T** use `px` for font-size. Use `rem` to honor the user's browser font-size preferences (root font size), as `px` values are absolute and may ignore user-defined defaults.
```
LeaVerou: Agreed for guidance _against_ `px` but would soften the guidance _for_ `rem`  a bit to also allow for relative units like percentages or ems or fancier techniques (such as those expressions combining viewport units with (r)ems). 

Basically, use rems unless you have a good reason to use another unit. Perhaps "Prefer" instead of "Use"?

```suggestion
- **DON'T** use `px` for font-size. Prefer `rem` to honor the user's browser font-size preferences (root font size), as `px` values are absolute and may ignore user-defined defaults.
```

LeaVerou (approved):

PR #603, opened 2026-04-18
Add guide and evals for state-aware sticky headers

### LINKED ISSUE #256: Create guide and evals for the context-sensitive-sticky-headers use case
Build sticky section headers or navbars that visually transform when they're actually "stuck" at the top, collapsing, changing their color scheme, gaining a shadow, or switching to a more compact layout.

Affected web-feature IDs: [container-scroll-state-queries](https://webstatus.dev/features/container-scroll-state-queries)

Use case subdir: [guides/ui-atoms/context-sensitive-sticky-headers](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-atoms/context-sensitive-sticky-headers)

Related features: #47

**Required files:**
- [x] demo.html
- [ ] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

patrickkettner:
Fixes #256


rviscomi:
Results of local `gd eval` run:

<details>
<summary>Pass Rate - Unguided: 0%, Guided: 0%</summary>

```json
{
  "summary": {
    "unguidedMedian": 0,
    "guidedMedian": 0,
    "unguidedPassRate": 0,
    "guidedPassRate": 0,
    "unguidedPassed": 0,
    "unguidedTotal": 0,
    "guidedPassed": 0,
    "guidedTotal": 0,
    "runsPerTest": 1,
    "expectedTotalRuns": 1,
    "taskCount": 1,
    "runCountPerTask": 1,
    "guideUsageRate": 0,
    "guideUsageCount": 0,
    "totalGuidedRuns": 1,
    "totalGuidedNonDisciplineRuns": 1,
    "toolActivationRate": 0,
    "toolActivationCount": 0,
    "unguidedEarlyFailures": 0,
    "unguidedEarlyFailureRate": 0,
    "guidedEarlyFailures": 0,
    "guidedEarlyFailureRate": 0,
    "guidedNonDisciplineEarlyFailures": 0
  },
  "results": {
    "task - context-sensitive-sticky-headers - guided": [
      {
        "runNumber": 1,
        "results": [],
        "guidesUsed": [],
        "retrievedGuides": [],
        "fileReadGuides": [],
        "guidanceToolsUsed": [],
        "discipline": "user-experience",
        "isSkill": false,
        "expectedToolPrefixes": [
          "modern-web"
        ],
        "guideName": "context-sensitive-sticky-headers",
        "taskName": "task",
        "baseApp": "daily-grind",
        "prompt": "Add new section headers inside the main content area of the page (e.g., within the cards or sections). These headers should stick to the top when scrolling."
      }
    ],
    "task - context-sensitive-sticky-headers - unguided": [
      {
        "runNumber": 1,
        "results": [],
        "guidesUsed": [],
        "retrievedGuides": [],
        "fileReadGuides": [],
        "guidanceToolsUsed": [],
        "discipline": "user-experience",
        "isSkill": false,
        "expectedToolPrefixes": [
          "modern-web"
        ],
        "guideName": "context-sensitive-sticky-headers",
        "taskName": "task",
        "baseApp": "daily-grind",
        "prompt": "Add new section headers inside the main content area of the page (e.g., within the cards or sections). These headers should stick to the top when scrolling."
      }
    ]
  },
  "stats": {
    "task - context-sensitive-sticky-headers - unguided": {
      "medianPassRate": 0,
      "runPassRates": [
        0
      ],
      "runCount": 1,
      "isSkill": false,
      "passedChecks": 0,
      "totalChecks": 0,
      "earlyFailures": 0
    },
    "task - context-sensitive-sticky-headers - guided": {
      "medianPassRate": 0,
      "runPassRates": [
        0
      ],
      "runsUsingGuide": 0,
      "runsWithToolActivation": 0,
      "runCount": 1,
      "isSkill": false,
      "passedChecks": 0,
      "totalChecks": 0,
      "earlyFailures": 0
    }
  },
  "timestamp": "2026-04-22T19:40:06.347Z",
  "runCount": 1,
  "agent": "gemini_cli",
  "serving": "skills_cli",
  "model": "gemini-pro-latest"
}
```
</details>

Given the 0% guided pass rate, this use case would require some additional investigation and fine tuning. Let's remove the eval files from this PR to unblock getting the guidance in and eng will follow up on the evals separately.

rviscomi:
@LeaVerou are you able to take a look at this one?

LeaVerou:
> @LeaVerou are you able to take a look at this one?

Just saw this, will take a look in a bit!

guides/user-experience/context-sensitive-sticky-headers/guide.md
```diff
@@ -3,4 +3,108 @@ name: context-sensitive-sticky-headers
...
+  .sticky-header {
+    background: #0056b3;
+    color: white;
+    box-shadow: 0 4px 6px rgba(0,0,0,0.15);
+    padding: 10px 20px; /* Switch to a more compact layout */
+  }
+}
+```
+
+> [!TIP]
```
LeaVerou: For skills, we have [guidance against GitHub alerts](https://github.com/GoogleChrome/guidance/blob/main/.agents/skills/project-skills-research/SKILL.md?plain=1#L95): 

> **Avoid GitHub Alerts**: Do not use GitHub-style alerts (e.g., `> [!TIP]`, `> [!IMPORTANT]`) in skills files. These files are consumed by AI agents, not humans. Keep the formatting minimal and dense.

Presumably this applies to guides too?

guides/user-experience/context-sensitive-sticky-headers/guide.md
```diff
@@ -3,4 +3,108 @@ name: context-sensitive-sticky-headers
...
+### 3. Target the Stuck State
+Use the `@container scroll-state(...)` query to apply styles when the header is stuck.
+
+```css
+/* Apply styles when the named container is stuck at the top */
+@container section-header scroll-state(stuck: top) {
+  .sticky-header {
+    background: #0056b3;
+    color: white;
+    box-shadow: 0 4px 6px rgba(0,0,0,0.15);
```
LeaVerou: Nit: Comma-separated `rgba()` is a legacy format. Instead, use:


```suggestion
    box-shadow: 0 4px 6px rgb(0 0 0 / 0.15);
```

LeaVerou:
## SME Review

### P0: Flashing when slightly scrolled

The guide downplays the problem quite a bit:

> Changing the stuck element's box size (padding, height, font-size) is a common and valid pattern. Be aware that the browser performs a two-pass rendering update to resolve scroll-state queries, so a size change on the stuck state may produce a one-frame settle. Using `transition` on the changing properties smooths this visually. Avoid changes large enough to un-stick the element (e.g., collapsing to `height: 0`), which would cause the query to oscillate.

This is what the demo page looks like if the container is only slightly scrolled:

https://github.com/user-attachments/assets/095922b8-48c1-4569-a7ff-f8a4bf26aaaf

It gets way worse if you also reduce the font-size (as is often needed). Sometimes the CSS will fight against the scrolling even:

https://github.com/user-attachments/assets/936cc6e9-d4cc-416c-aa1b-f90aa0a1cd11

And even worse if you move the transition inside the CQ (so that it only applies when you go from unstuck → stuck but NOT when you go stuck → unstuck):


https://github.com/user-attachments/assets/7083c052-961c-4b15-822b-e5a0c0c744fa

You can trigger the issue reliably by adding this snippet to the demo and playing with `offsetY`, though to experience it in its full glory you need to actually scroll slowly yourself:

```html
<script>
let firstHeader = document.querySelector(".section > .sticky-container");
firstHeader.scrollIntoView({behavior: "smooth"});
setTimeout(() => {
	// Adjust this number for different versions of the problem. Any number between 1-43 seems to trigger it.
	const offsetY = 10;
	document.scrollingElement.scrollTop += offsetY;
}, 1000);
</script>
```

Avoiding "large enough" changes does not fix this, it just reduces the interval where this is observable. I could not find *any* box-model affecting change where the problem was not present at all. Even going from `font-size: 100%` to `font-size: 99%` triggered it!

I did end up getting badly [nerd sniped](https://xkcd.com/356/) by this and doing original research to come up with workarounds 😅, which I'll publish in a separate blog post (I can drop a link here when it's out if it's of interest), but so far all workarounds I found have their flaws. 

The best one so far is adding an `::after` in the stuck state to counteract the height increase so that `.sticky-container` does not reduce in height, but then you get empty space you have to scroll past to get the element "unstuck":

https://github.com/user-attachments/assets/9d8bdb48-f6fe-4a60-9754-b3411576c98c

For this particular case, a `min-height` on `.sticky-container` would also work equally well, with the same downside:

https://github.com/user-attachments/assets/7c0b361c-967d-4bfe-8090-172ee87b0894

But a `min-height` doesn't generalize well: usually your headings have varying content → different height in different viewports, and if you undershoot even by a bit, the flickering comes back.

In both cases you'd want `pointer-events: none` on `.sticky-container` and `pointer-events: auto` on `.sticky-header` so that the extra height doesn't swallow pointer events.

But we probably want to be recommending tried and tested techniques for this, not brand new workarounds… So, I'm not quite sure what's the best way forwards. 😞 

#### P2: DO always specify container name

Given that these CQs may be nested, we should emphasize that these queries should always include a container name.

#### P2: Not sure what "context-sensitive" refers to

I don't see anything context sensitive about this? The description doesn't help either:

> description: Build sticky section headers or navbars that visually transform when they're actually "stuck" at the top, collapsing, changing their color scheme, gaining a shadow, or switching to a more compact layout.

### Fallback strategies



#### P0: No fallback in the demo page

The fallback is not present in the demo page. It should not only be there, but it should be verified that it's there via the expectations!

#### P1: Discuss PE as the default and frame the `IntersectionObserver` fallback 

Currently, the guidance recommends an `IntersectionObserver` workaround if these CQs are not supported. However, in many cases, the styling or even the sticking altogether is a nice-to-have, and would not be worth this. E.g. if you're using this to stick the current heading to provide the user some additional context cue (like the demo), IMO this falls squarely under progressive enhancement.

If the default styling works and the `scroll-state()` query just improves things, perhaps no fallback is needed.

If the stuck styling is absolutely necessary to make the stuck content readable (e.g. adds a background that can't be there in the normal state or makes the element smaller when it would normally occupy too much space), perhaps it could be enough to just gate the `position: sticky` itself under a `@supports (container-type: scroll-state)` feature query.

I'd recommend the `IntersectionObserver` fallback only if the feature is critical, not by default.

#### P2: Make polyfill more generic

P2: This is one of these cases where the fidelity of the polyfill matters, since the feature is only supported in Chrome currently, and while technically nothing breaks when it's not there, it's also not a small difference.

Hardcoding it works if all you have is one example, but ideally, you don't want to have multiple instances of this with different values. You can instead detect these, for example:
- use `getComputedStyle()` to read the actual value 
- Depending on the value, find the scroll parent that's scrollable in that direction by traversing up and comparing `clientHeight` with `scrollHeight` (or -Width), assume viewport if none found. In fact, this could be part of a `scroll-state(scrollable)` polyfill too…

#### P2: Fallback duplication

As you will see once the fallback is actually added to the demo, there is no way to maintain the styles in one place, they need to be duplicated. Unfortunately, I have no good solution for this either, besides framing this as PE-first, as I recommended above.

LeaVerou (changes requested):
See [review](https://github.com/GoogleChrome/guidance/pull/603#issuecomment-4371565769).

patrickkettner:
@LeaVerou would love to hear about the blog post once you post it!

@rviscomi redid the evals - let me know if you want them removed from this commit

LeaVerou:
Just saw the review request — GitHub didn't bother notifying me for some reason. I'll try to take a look soon.

> @LeaVerou would love to hear about the blog post once you post it!

I held off on it because after talking to @tabatkins he thinks it's either a browser bug or a spec bug, so I filed https://github.com/w3c/csswg-drafts/issues/13898 for now so we can figure that out first.

guides/ui-atoms/state-aware-sticky-headers/guide.md
```diff
@@ -0,0 +1,130 @@
...
+
+.sticky-header {
+  /* Base styles for the header */
+  background: #f0f0f0;
+  padding: 20px;
+  transition: background-color 0.3s, box-shadow 0.3s;
+}
+```
+
+**Important:** The scroll state is queried by **descendants** of the scroll-state container. The styles in the `@container` query will not apply to the `.sticky-container` itself, but to its children (like `.sticky-header`). You cannot style the container element itself with its own scroll-state query.
```
LeaVerou, line 54: A failure mode I saw in Claude today (it decided to use scroll-state(stuck) queries entirely by itself! Unguided!) was that it originally applied `container-type: scroll-state` to the `body` (i.e. the scrollable element) instead of the stuck element. We should emphasize that unlike other `scroll-state` queries, `container-type: scroll-state` should be applied on the element that has `position: sticky`.

I think the guidance here would probably avoid this, but not sure if we want to emphasize it even more.

guides/user-experience/state-aware-sticky-headers/guide.md
```diff
@@ -0,0 +1,130 @@
...
+  }
+}
+```
+
+**Tip:** You can also use logical properties like `stuck: inset-block-start` or `stuck: inset-inline-start` to better support internationalization (i18n) and right-to-left (RTL) layouts by querying flow-relative edges rather than physical ones.
+
+### Notes on Dimension Changes
+
+**MANDATORY:** **DO NOT change box-model properties (height, padding, border, font-size) when an element becomes stuck.**
+Because the browser performs a two-pass rendering update to resolve scroll-state queries, modifying any property that affects layout (even changing `font-size` from `100%` to `99%`) can cause the element to immediately become unstuck. This results in an infinite layout loop and severe visual flickering when users scroll slowly near the intersection point (specifically, within a 1-43px offset window).
```
LeaVerou: The 1-43px offset window was for this specific case! I'm sure it varies per layout depending on the exact shrinkage.

(Also having to recommend against changing box-model properties is quite unfortunate, but also probably the right thing to do… :/)

guides/user-experience/state-aware-sticky-headers/guide.md
```diff
@@ -0,0 +1,130 @@
...
+
+Adding a `transition` does not fix this flashing; in fact, putting the `transition` inside the `@container` query makes it even worse. Only change non-layout affecting properties such as `background-color`, `color`, `opacity`, and `box-shadow`.
```
LeaVerou: `transform` (and individual transform properties) too! E.g. see https://codepen.io/leaverou/pen/xbROqRo for an example on how it can be used to fake shrinking.

guides/user-experience/state-aware-sticky-headers/guide.md
LeaVerou: I _think_ we can relax this a bit. It seems that *some* types of dimension changes are ok, e.g. adding a bottom border, even if thick: https://codepen.io/leaverou/pen/ogYLZwX
Can you get it to flicker here? I can't.

Horizontal dimension changes are also ok (e.g. `width`, `padding-inline`) since they don't affect `(stuck: top)` (though they would affect other axes).

The worst offenders seem to be `padding-block` and `font-size`.

LeaVerou (changes requested):
See comments!

patrickkettner:
> I held off on it because after talking to @tabatkins he thinks it's either a browser bug or a spec bug, so I filed https://github.com/w3c/csswg-drafts/issues/13898 for now so we can figure that out first.

 [Looks like `overflow-anchor: none` lands this for us](https://github.com/w3c/csswg-drafts/issues/13898#issuecomment-4428150670). Should I update the guide to recommend that on the scroll container instead of the broad "no box-model changes" rule? Then I could reframe the flicker section around scroll anchoring as the cause. (Still keen on the blog post whenever it happens :D)

guides/ui-atoms/state-aware-sticky-headers/guide.md
patrickkettner: Great catch! Added an explicit callout at the top of Step 2 so it's the first thing an agent reads in that section, rather than something they have to infer from the example:

>   Set `container-type: scroll-state` on the `position: sticky` element, not on its scrollable ancestor.

guides/user-experience/state-aware-sticky-headers/guide.md
patrickkettner: good news :D with `overflow-anchor: none` on the scroll container, the box-model restriction goes away.

LeaVerou:
> > I held off on it because after talking to @tabatkins he thinks it's either a browser bug or a spec bug, so I filed [w3c/csswg-drafts#13898](https://github.com/w3c/csswg-drafts/issues/13898) for now so we can figure that out first.
> 
> [Looks like `overflow-anchor: none` lands this for us](https://github.com/w3c/csswg-drafts/issues/13898#issuecomment-4428150670). Should I update the guide to recommend that on the scroll container instead of the broad "no box-model changes" rule? Then I could reframe the flicker section around scroll anchoring as the cause. (Still keen on the blog post whenever it happens :D)

Yes!! I was unsure about it being applied to `:root` but it appears it works when applied to `section` too! https://codepen.io/leaverou/pen/JobKgJQ

It still worries me that it's disabling a supposedly positive behavior, and the lack of Safari support, but overall this does look very promising!

The issue is also slotted to be discussed tomorrow, so I might have more data by then.

guides/user-experience/state-aware-sticky-headers/guide.md
patrickkettner: (addressed by the rewrite) 

with `overflow-anchor: none` in place, transforms work fine alongside box-model changes. The new section explicitly calls out transforms as one of the things you can now safely animate in the stuck state.

guides/user-experience/state-aware-sticky-headers/guide.md
patrickkettner: rewritten

patrickkettner:
awesome! Ill hold off on updating anything due to it until after I hear back

LeaVerou:
> awesome! Ill hold off on updating anything due to it until after I hear back

Bottom line was: we need to fix this problem in CSS, unsure how we'll do it.

I think recommending `overflow-anchor: none` on the parent of the stuck item (not the root) is fine for now. 

Also, another point that @tabatkins brought up in the call was that if an element is bottom stuck, `overflow-anchor` would do nothing to fix the flickering. Perhaps we should recommend against adjusting box model on elements that are stuck to the bottom? Though these are so rare, I'm not sure it's worth it.

rviscomi:
@LeaVerou are you still waiting on changes?

guides/user-experience/state-aware-sticky-headers/guide.md
```diff
@@ -0,0 +1,131 @@
...
+    box-shadow: 0 4px 6px rgb(0 0 0 / 0.15);
+  }
+}
+```
+
+**Tip:** You can also use logical properties like `stuck: inset-block-start` or `stuck: inset-inline-start` to better support internationalization (i18n) and right-to-left (RTL) layouts by querying flow-relative edges rather than physical ones.
+
+### Notes on Dimension Changes
+
+Without scroll anchoring disabled, changing layout-affecting properties (height, padding, font-size, transforms) when stuck can cause visual flickering. Scroll anchoring on the in-flow content below the sticky element adjusts the scroll offset to compensate for the layout change, which pushes the element back out of its stuck position and triggers an oscillation.
```
LeaVerou:
```suggestion
Without scroll anchoring disabled, changing layout-affecting properties (height, padding, font-size) when stuck can cause visual flickering. Scroll anchoring on the in-flow content below the sticky element adjusts the scroll offset to compensate for the layout change, which pushes the element back out of its stuck position and triggers an oscillation.
```

transforms are not layout-affecting.

Really nice explanation!

guides/user-experience/state-aware-sticky-headers/guide.md
```diff
@@ -0,0 +1,131 @@
...
+
+Disable scroll anchoring on the scroll container to avoid this:
+
+```css
+:root {
+  overflow-anchor: none;
+}
+```
+
+Apply it to whichever element is the scroll container (typically `:root` for the document scroller). With this in place, you can freely change any property in the stuck state, including box-model properties and transforms.
```
LeaVerou: You can apply on the parent of the `position: sticky` element and it seems to work equally well. Did that work less well in your testing? If not, I think the narrower, the better, we don't need to disable scroll anchoring everywhere in the document (it's usually a good thing!) because one table somewhere has a sticky header.

guides/ui-atoms/state-aware-sticky-headers/guide.md
```diff
@@ -0,0 +1,131 @@
...
+```css
+:root {
+  overflow-anchor: none;
+}
+```
+
+Apply it to whichever element is the scroll container (typically `:root` for the document scroller). With this in place, you can freely change any property in the stuck state, including box-model properties and transforms.
+
+### Fallback strategies
+{{ BASELINE_STATUS("container-scroll-state-queries") }}
```
LeaVerou, line 90: How does `BASELINE_STATUS` work when different bcd keys have different support? Is it worth specifying a bcd key?

guides/ui-atoms/state-aware-sticky-headers/guide.md
```diff
@@ -0,0 +1,131 @@
...
+
+Scroll state queries are a progressive enhancement. In browsers that do not support `container-type: scroll-state`, the header will still stick to the top (due to `position: sticky`), but it will not visually transform. For most use cases, this is the recommended approach.
+
```
LeaVerou, line 93: I was expecting something like this to come after:

````suggestion

If stickiness is not essential, but `position: sticky` without different styling would be a worse experience than no stickiness at all, you can wrap the `position: sticky` and related declaration in an `@supports` query as well:

```css
.sticky-container {
  @supports (container-type: scroll-state) {
	  position: sticky;
	  top: 0;
	  container-type: scroll-state;
	  container-name: section-header;
	  z-index: 10;
	}
}
```
````

File: guides/ui-atoms/state-aware-sticky-headers/guide.md [hunk 10]
```diff
@@ -0,0 +1,131 @@
...
+**Tip:** If your "stuck" styling requires a different background color for readability, consider setting your layout up so that the default styling works everywhere (Progressive Enhancement). If you must use a fallback, you can gate the `position: sticky` behaviour itself inside an `@supports (container-type: scroll-state)` query.
```
LeaVerou, line 107: Ah, just noticed it's mentioned here, I missed it in my first skim, apologies. Leaving the suggestion above in case you prefer that wording, but feel free to resolve if not!

LeaVerou (changes requested):
Only thing I'd consider blocking is the point about `overflow-anchor` — I'd really rather avoid applying it to root unless we really need to. In my testing it seems fine applied to the sticky parent: https://codepen.io/leaverou/pen/JobKgJQ
@patrickkettner curious if you came across cases that need it on the root?

rviscomi:
@patrickkettner could you take a look at @LeaVerou's feedback?

patrickkettner:
@rviscomi the latest push here is guide-only. I narrowed `overflow-anchor: none` to the sticky element's parent instead of `:root` and added a caveat about bottom-stuck elements, per Lea's feedback. None of it touches `demo.html` or the grader, so the existing calibration is unaffected, right?

I'll re-run `gd eval` and post the updated guided/unguided numbers here before flipping this to ready for review. Still happy to pull the eval artifacts into a separate follow-up if you'd prefer to keep this PR focused on the guidance, lemme know.

guides/ui-atoms/state-aware-sticky-headers/guide.md
patrickkettner: Good question. `BASELINE_STATUS` takes an optional second argument for a specific BCD key. Without it (which is what I've got here) it reports the aggregate status for the whole `container-scroll-state-queries` feature.

I left it on the feature ID because, as far as I can tell, all the scroll-state query types (`stuck`, `snapped`, `scrollable`) shipped together in the same Chrome release, so the aggregate status matches the support window for the `stuck` query this guide actually relies on. If we want to scope it tighter later we can pass the specific BCD key as the second arg, but today it'd resolve to the same status.

patrickkettner:
re-ran the evals on the current branch. 3 runs, gemini_cli (gemini-3.1-pro-preview) with skills_cli serving. all six agent executions completed, no early failures:

- guided: 100% (21/21 checks), guide retrieved and tool activated in every run
- unguided: 81% (17/21). two runs passed everything; the third made the headers sticky but added no stuck-state styling at all (no `container-type: scroll-state`, no query, no fallback), which accounts for all four failed checks

so current models will reach for scroll-state() unprompted most of the time; with the guide it was all three runs.

Let me know if youd rather this be guidance only. Not sure why I was getting such different numbers from you before, but would love to compare evals.json if you arent seeing the same results.

@rviscomi otherwise this is ready for another look. @LeaVerou the overflow-anchor scoping you flagged is in (sticky parent, not root, plus the bottom-stuck caveat).

guides/scroll/state-aware-sticky-headers/expectations.md
```diff
@@ -0,0 +1,7 @@
+- The header container uses `position: sticky` to remain at the top of the scroller.
+- The header container defines `container-type: scroll-state` to enable scroll state queries.
+- The header container defines `container-name: section-header` to avoid collisions.
+- The header visual style changes (e.g., background color or box-shadow) when it is stuck at the top.
+- The visual style changes are implemented using the `@container scroll-state(stuck: top)` query.
+- A functional JavaScript fallback using `IntersectionObserver` is provided for browsers that do not support scroll-state queries.
```
LeaVerou: The guide says to do this only if the styling is essential, but nothing in `task.md` indicates this, so I'd say this is an anti-expectation. If we want to keep this, we should tweak the tasks to say the styling is essential. But ideally we should be able to test different requirements. It's a shame that we can only have one set of requirements, even when having multiple tasks…

guides/ui-atoms/state-aware-sticky-headers/expectations.md
```diff
@@ -0,0 +1,7 @@
+- The header container uses `position: sticky` to remain at the top of the scroller.
```
LeaVerou, line 1:
```suggestion
- The header container uses `position: sticky` to remain at the top of the scroller.
- If any of the stuck styles change layout, `overflow-anchor: none` is applied to the parent of the header container.
- If none of the stuck styles change layout, `overflow-anchor: none` should NOT be applied to the parent of the header container.
```

guides/ui-atoms/state-aware-sticky-headers/expectations.md
```diff
@@ -0,0 +1,7 @@
...
+- The header container defines `container-type: scroll-state` to enable scroll state queries.
+- The header container defines `container-name: section-header` to avoid collisions.
+- The header visual style changes (e.g., background color or box-shadow) when it is stuck at the top.
```
LeaVerou, line 6: Should we tweak one of the tasks to also change layout so we can test the `overflow-anchor: none` thing?

LeaVerou (approved):
The guide looks great! I added some suggestions/comments to the expectations, but I don't think they're blocking.

Also this may need rebasing, I see some unrelated changes that have crept in.

guides/scroll/state-aware-sticky-headers/expectations.md
patrickkettner: yep, dropped both fallback bullets. the guide frames the IntersectionObserver fallback as only-if-critical so expecting it was backwards.

guides/ui-atoms/state-aware-sticky-headers/expectations.md
patrickkettner: done. the first task now has the headers lose padding when stuck, and the grader checks overflow-anchor both ways (i.e. required when the stuck styles change layout, explicitly not when they dont). Ran the tests again, and every unguided run hit that exact issue, so seems like a good thing to add

patrickkettner:
Thanks Lea! Updated, re-ran the evals with the layout-changing task, guided 24/24, unguided 21/24 with all three misses being the missing overflow-anchor.

rviscomi:
We shouldn't need to update the README in this PR. Removing that should resolve the merge conflict, then we can push this change through.

_Edit: sorry I didn't realize the build script changed to auto-generate the README_

Issue #168, opened 2026-02-27
Create guide and evals for the size-aware-styling use case


paulirish:
Build a component whose styles can be conditionally dependent on its own width or height, rather than the width or height of the viewport. For example a card component that can change its layouts depending on how large it is, or a call-to-action button that can conditionally display helper text based on its width.

Affected web-feature IDs: [container-queries](https://webstatus.dev/features/container-queries)

Use case subdir: [guides/css/size-aware-styling](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/css/size-aware-styling)

Related features: #46

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

Issue #164, opened 2026-02-27
Create guide and evals for the fluid-scaling use case


paulirish:
Scale items like font size, spacing, and media sizes smoothly based on the parent container's size rather than using fixed breakpoints

Affected web-feature IDs: [container-queries](https://webstatus.dev/features/container-queries)

Use case subdir: [guides/css/fluid-scaling](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/css/fluid-scaling)

Related features: #46

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

PR #608, opened 2026-04-20
Add guidance for `size-aware-styling` and `fluid-scaling`

taraojo:
#164 and #168 - Guidance, demos, and expectations for size-aware styling and fluid scaling using container queries and units.

### DIFF
--- a/guides/user-experience/fluid-scaling/demo.html
+++ b/guides/user-experience/fluid-scaling/demo.html
@@ -23,7 +23,6 @@
       resize: horizontal;
       overflow: auto;
       min-width: 200px;
-      max-width: 800px;
       width: 500px;
     }
 
@@ -39,6 +38,9 @@
       .card-title {
         /* Use container units where supported */
         font-size: clamp(1.2rem, 6cqi, 3rem);
+
+        /* Scale padding based on container width */
+        padding: 0 5cqi;
       }
     }
 
@@ -46,6 +48,13 @@
       font-size: 1rem;
       color: #4a4a4a;
       line-height: 1.5;
+    }
+
+    @supports (padding: 5cqi) {
+      .card-content {
+        /* Scale padding based on container width */
+        padding: 0 5cqi;
+      }
     }
 
     .instruction {
@@ -57,13 +66,20 @@
 <body>
 
   <div class="instruction">
-    Drag the bottom-right corner of the card to resize it and see the title font size scale smoothly.
+    Notice how the two cards below have different title sizes, even though they share the same CSS declaration, because they have different widths. Drag their bottom-right corners to resize them individually.
   </div>
 
-  <div class="card-wrapper">
+  <div class="card-wrapper" style="width: 350px; margin-bottom: 2rem;">
     <h1 class="card-title">This Title Scales Fluently</h1>
     <p class="card-content">
-      The font size of the title above is determined by the width of this white card container, not the width of the entire screen. Resize the card to see it in action.
+      This card is narrow (350px width), so the title is small.
+    </p>
+  </div>
+
+  <div class="card-wrapper" style="width: 600px;">
+    <h1 class="card-title">This Title Scales Fluently</h1>
+    <p class="card-content">
+      This card is wide (600px width), so the title is larger using the exact same CSS rules.
     </p>
   </div>
 
--- a/guides/user-experience/fluid-scaling/demo.html
+++ b/guides/user-experience/fluid-scaling/demo.html
@@ -70,14 +70,14 @@
   </div>
 
   <div class="card-wrapper" style="width: 350px; margin-bottom: 2rem;">
-    <h1 class="card-title">This Title Scales Fluently</h1>
+    <h1 class="card-title">This Title Scales Fluidly</h1>
     <p class="card-content">
       This card is narrow (350px width), so the title is small.
     </p>
   </div>
 
   <div class="card-wrapper" style="width: 600px;">
-    <h1 class="card-title">This Title Scales Fluently</h1>
+    <h1 class="card-title">This Title Scales Fluidly</h1>
     <p class="card-content">
       This card is wide (600px width), so the title is larger using the exact same CSS rules.
     </p>
--- a/guides/user-experience/fluid-scaling/demo.html
+++ b/guides/user-experience/fluid-scaling/demo.html
@@ -1,1 +1,71 @@
-<!-- TODO -->
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Fluid Scaling Demo</title>
+  <style>
+    body {
+      font-family: system-ui, -apple-system, sans-serif;
+      padding: 2rem;
+      background: #f0f2f5;
+    }
+
+    /* The container that we want to measure */
+    .card-wrapper {
+      container-type: inline-size;
+      background: white;
+      padding: 2rem;
+      border-radius: 8px;
+      box-shadow: 0 2px 4px rgba(0,0,0,0.1);
+      
+      /* Make the container resizable for the demo */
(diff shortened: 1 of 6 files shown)


rviscomi:
@jamesnw two more guides to review if you're able

guides/user-experience/fluid-scaling/guide.md
```diff
@@ -3,4 +3,90 @@ name: fluid-scaling
...
   - container-queries
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries
+  - https://css-tricks.com/css-container-queries/
+  - https://www.joshwcomeau.com/css/container-queries-introduction/
+  - https://web.dev/learn/css/container-queries
+  - https://www.joshwcomeau.com/css/container-queries-unleashed/
+  - https://frontendmasters.com/blog/container-queries-and-units/
+  - https://www.smashingmagazine.com/2024/06/what-are-css-container-style-queries-good-for/
+  - https://www.oddbird.net/2026/02/18/queries-units-actions/
```
jamesnw:
```suggestion
  - https://web.dev/articles/baseline-in-action-fluid-type
```
The OddBird link just points to a web.dev post, but this one is more directly applicable. If there's a risk of too many sources, I would remove several of the generic "What is a container query" source and focus on ones that discuss type specifically.

guides/user-experience/fluid-scaling/guide.md
```diff
@@ -3,4 +3,90 @@ name: fluid-scaling
...
+  - https://web.dev/blog/how-to-use-container-queries-now
 ---
+
+## Overview
+
+Fluid scaling allows components to adapt their internal proportions (like font sizes and spacing) smoothly as their container resizes. This creates a more cohesive design than jumping between fixed breakpoints. This approach is recommended because it ensures components look good regardless of where they are placed in a layout, promoting component isolation and reusability.
```
jamesnw: This is conflating fluid scaling and container queries. Could it introduce fluid scaling, and then say container queries allow the scale to be based on the container rather than the viewport?

Also, potentially a nit, but containers are rarely resized by users (although web devs do it constantly). I'd say "adjust their internal proportions based on the current dimensions" rather than "smoothly as their container resizes".

Would it make sense to have 2 cards in the demo with different widths and the same font size declaration? They can still be resizable, but that's the more likely use case.

guides/user-experience/fluid-scaling/guide.md
```diff
@@ -3,4 +3,90 @@ name: fluid-scaling
...
+
+.component-body {
+  /* Scale padding based on container width */
+  padding: 5cqi;
+}
+```
+
+### 3. Constrain values with `clamp()`
+
+To prevent sizes from becoming too small or too large, use the CSS `clamp()` function.
```
jamesnw:
```suggestion
To prevent sizes from becoming too small or too large, use the CSS `clamp()` function. This impacts the user's ability to zoom or adjust their base font size. To ensure text meets accessibility guidelines, the maximum size must not be more than 2.5 times the minimum size.
```
This should be added to the expectations as well, and the examples adjusted.

guides/user-experience/fluid-scaling/guide.md
```diff
@@ -3,4 +3,90 @@ name: fluid-scaling
...
+```css
+.component-title {
+  /* Scale font size based on container width */
+  /* 10cqi means 10% of the container's width */
+  font-size: 10cqi;
+}
+
+.component-body {
+  /* Scale padding based on container width */
+  padding: 5cqi;
```
jamesnw, line 59: Could we add this to the demo as well?

guides/user-experience/fluid-scaling/guide.md
```diff
@@ -3,4 +3,90 @@ name: fluid-scaling
...
+
+@supports (font-size: 1cqi) {
+  .component-title {
+    /* Use container units where supported */
+    font-size: clamp(1rem, 5cqi, 3rem);
+  }
+}
+```
+
+This fallback ensures that the text still scales, but it will be based on the screen width rather than the component's width. This is usually a safe fallback.
```
jamesnw:
```suggestion
This fallback ensures that the text still scales, but it will be based on the screen width rather than the component's width. This should be tested to verify it works in your use case.
```

guides/user-experience/fluid-scaling/demo.html
```diff
@@ -1 +1,71 @@
...
+      background: white;
+      padding: 2rem;
+      border-radius: 8px;
+      box-shadow: 0 2px 4px rgba(0,0,0,0.1);
+      
+      /* Make the container resizable for the demo */
+      resize: horizontal;
+      overflow: auto;
+      min-width: 200px;
+      max-width: 800px;
```
jamesnw: I think this max-width is preventing the max value of the clamp from being hit, and it could probably be removed.

guides/user-experience/size-aware-styling/guide.md
```diff
@@ -3,4 +3,118 @@ name: size-aware-styling
...
   - container-queries
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries
+  - https://css-tricks.com/css-container-queries/
+  - https://www.joshwcomeau.com/css/container-queries-introduction/
+  - https://web.dev/learn/css/container-queries
+  - https://www.joshwcomeau.com/css/container-queries-unleashed/
+  - https://frontendmasters.com/blog/container-queries-and-units/
+  - https://www.smashingmagazine.com/2024/06/what-are-css-container-style-queries-good-for/
+  - https://www.oddbird.net/2026/02/18/queries-units-actions/
```
jamesnw:
```suggestion
  - https://web.dev/articles/baseline-in-action-container-queries
```
The OddBird post is just a link to this canonical post.

jamesnw (changes requested):
I think this is looking good, but the main change needed is that since container queries are baseline, this doesn't need fallbacks. I think that will simplify the demos significantly, so I'll hold off on reviewing those more thoroughly until this is ready for re-review!

guides/user-experience/fluid-scaling/guide.md
taraojo: Added!

guides/user-experience/fluid-scaling/guide.md
taraojo: Good to know, I've added this to the guide, updated the expectations with this requirement and updated the examples in the guide. If I understand correctly the values in the demo are ok.

guides/user-experience/fluid-scaling/guide.md
taraojo: I've updated the demo and framing in the guide, let me know what you think

taraojo:
Thanks @jamesnw!

> the main change needed is that since container queries are baseline, this doesn't need fallbacks

Oh interesting I was using the `project-guides` skill and was under the impression we should have fallbacks in all cases to cover older browsers. Do we already have examples where we're skipping the fallbacks?

guides/user-experience/fluid-scaling/guide.md
jamesnw: Looks great!

guides/user-experience/fluid-scaling/guide.md
jamesnw: I think so too! Thanks!

jamesnw (approved):
You're correct- fallbacks are required. Thanks for catching me on that! LGTM!

I had to fix the coauthor field, so i fixed a [typo](https://github.com/GoogleChrome/guidance/pull/608/commits/0bf0369f09cc820eb83ddcde0121ae706362da80) since I was pushing anyways.

jamesnw:
Assigning to @malchata, since that's what I normally do after I'm done, but let me know if something else should happen.

rviscomi (approved):
LGTM! Tested in Chrome

Thanks @jamesnw for the review!

PR #610, opened 2026-04-20
Early version of high-level WebMCP skill.

Kulikowski:
Adding first version of WebMCP skill (skill-draft). 

WebMCP is proposed web standard that is still being discussed in W3C group https://github.com/webmachinelearning/webmcp. 

We want to keep it lean for now as spec will definitely change going forward. Current early version of skill consists of implementation guides for Imperative and Declarative APIs, good and bad practices, and implementation status.

### DIFF
--- a/skills-drafts/webmcp/references/declarative.md
+++ b/skills-drafts/webmcp/references/declarative.md
@@ -1,0 +1,85 @@
+# WebMCP Declarative API
+
+The Declarative API transforms standard HTML `<form>` elements into WebMCP tools via attributes. The browser synthesizes a JSON Schema from the form inputs and handles agent interactions.
+
+## Form Attributes
+
+*   `toolname`: Unique name for the tool.
+*   `tooldescription`: Purpose of the tool.
+*   `toolautosubmit`: (Optional) If present, the agent can submit the form without waiting for user interaction. 
+*   `toolparamdescription`: (Optional, on input) Maps to the property description. Defaults to the `<label>` text.
+    *   *Note*: For `<input type="radio">` groups, place this on the *first* `<input>` element.
+
+### Example
+
+```html
+<form toolname="search-cars" 
+      tooldescription="Perform a car make/model search" 
+      toolautosubmit>
+  <label for="make">Vehicle Make</label>
+  <input type="text" id="make" name="make" required>
+  
+  <label for="model">Vehicle Model</label>
+  <input type="text" id="model" name="model" toolparamdescription="e.g., 330i, F-150" required>
+  
+  <button type="submit">Search</button>
+</form>
+```
+
+## Handling Submissions in JavaScript
+
+When an agent submits the form, the `SubmitEvent` includes `agentInvoked` (boolean) and `respondWith(promise)`.
+
+```javascript
+document.querySelector('form').addEventListener('submit', (event) => {
+  event.preventDefault();
+
+  // Validate the form
+  if (!myFormIsValid()) {
+    if (event.agentInvoked) { 
+      event.respondWith(Promise.resolve({ error: "Validation failed" })); 
+    }
+    return;
+  }
+
+  const resultPromise = performAsyncSearch(new FormData(event.target));
+  
+  // Return the result directly to the agent without navigation
+  if (event.agentInvoked) {
+    event.respondWith(resultPromise);
+  }
+});
+```
+
+## Lifecycle Events
+
+The window emits events when agents start or stop interacting with a tool:
+
+```javascript
+window.addEventListener('toolactivated', ({ toolName }) => {
+  console.log(`Tool "${toolName}" was activated by the agent.`);
+});
+
+window.addEventListener('toolcancel', ({ toolName }) => {
+  console.log(`Tool "${toolName}" interaction was cancelled.`);
+});
+```
+
+## Visual Feedback (CSS)
+
+Use pseudo-classes to highlight forms when an agent interacts with them:
+
+*   `:tool-form-active`: Applied to the `<form>` element actively used by the agent.
+*   `:tool-submit-active`: Applied to the submit button when the browser pauses for user review (if `toolautosubmit` is omitted).
+
+```css
+form:tool-form-active {
+  outline: 2px dashed blue;
+  background-color: rgba(0, 0, 255, 0.05);
+}
+
+button:tool-submit-active {
+  outline: 2px dashed red;
+  animation: pulse 2s infinite;
+}
+```
--- a/skills-drafts/webmcp/references/declarative.md
+++ b/skills-drafts/webmcp/references/declarative.md
@@ -84,3 +84,9 @@ button:tool-submit-active {
   animation: pulse 2s infinite;
 }
 ```
+
+## Browser Support
+
+{{ BASELINE_STATUS("declarative-webmcp") }}
+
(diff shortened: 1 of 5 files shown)


andreban (approved):
LGTM from the WebMCP side.

skills-drafts/webmcp/references/imperative.md
```diff
@@ -0,0 +1,105 @@
...
+    annotations: { readOnlyHint: true }
+  };
+}
+```
+
+## API Notes
+
+*   **annotations**: (Optional) A dictionary for tool metadata.
+    *   **readOnlyHint**: (Optional) Set to `true` if the tool does not modify any state and only reads data. This helps agents decide when it is safe to call the tool.
+*   **Return Format**: The `execute` function can return any JSON-serializable value (object, array, string, number, boolean).
```
rviscomi: In the absence of an `outputSchema`, are there any conventions on how to most effectively structure a response?

skills-drafts/webmcp/references/declarative.md
```diff
@@ -0,0 +1,86 @@
...
+When an agent submits the form, the `SubmitEvent` includes `agentInvoked` (boolean) and `respondWith(promise)`.
+
+```javascript
+document.querySelector('form').addEventListener('submit', (event) => {
+  event.preventDefault();
+
+  // Validate the form
+  if (!myFormIsValid()) {
+    if (event.agentInvoked) { 
+      event.respondWith(Promise.resolve({ error: "Validation failed" }));
```
rviscomi: One of the best practices is to return descriptive errors. Should this also show how to provide more info about what exactly is invalid?

skills-drafts/webmcp/references/declarative.md
```diff
@@ -0,0 +1,86 @@
+# WebMCP Declarative API
+
+The Declarative API transforms standard HTML `<form>` elements into WebMCP tools via attributes. The browser synthesizes a JSON Schema from the form inputs and handles agent interactions.
```
rviscomi, line 3: Would it be worth describing how more complex form elements are represented in the schema? Things like `input[type=file]` or `select[multiple]` for example.

rviscomi (commented):
Thanks! Left a few specific comments. Other questions:

I see `requestUserInteraction` in the spec but it's not covered in the skill. Is that still under development, or is there any guidance we can provide for it?

How can a coding agent verify its work? For example is there an API it could call to see what the registered tools are?

Is there any more guidance we could provide to help agents decide when to use WebMCP? The imperative API guide lists a few examples, but are there situations where you would discourage using it? Is _every_ form a good candidate for the declarative API?

Somewhere in the skill we need to talk about Baseline compatibility. I assume the narrative is "this is a progressive enhancement, it's safe to use in production today with feature detection" but that needs to be explicit. Could you add that detail?

Kulikowski:
> Thanks! Left a few specific comments. Other questions:
> 
> I see `requestUserInteraction` in the spec but it's not covered in the skill. Is that still under development, or is there any guidance we can provide for it?
> 
> How can a coding agent verify its work? For example is there an API it could call to see what the registered tools are?
> 
> Is there any more guidance we could provide to help agents decide when to use WebMCP? The imperative API guide lists a few examples, but are there situations where you would discourage using it? Is _every_ form a good candidate for the declarative API?
> 
> Somewhere in the skill we need to talk about Baseline compatibility. I assume the narrative is "this is a progressive enhancement, it's safe to use in production today with feature detection" but that needs to be explicit. Could you add that detail?

For context: currently only Chrome has experimental support for WebMCP so my comments only apply to Chrome - imho there is no use of putting things into SKILL that will not work anywhere

requestUserInteraction is not yet implemented in Chrome, API is still evolving, so we need to keep an eye on it.

Regarding verifying tools: There is just a workaround introduced by Eng Team in Chrome with modelContextTesting with listTools() and executeTool(...) - as this will go away pretty soon we decided to not include it in the SKILL yet.

I think we could add warnings about being careful with exposing destructive actions (permanently delete XYZ) and/or action that influence finances, health data, PII.

I am confused about Baseline bit here - it's an experimental support, and based on above points you can see the complexities we are dealing with - still evolving and still limited support, can you please tell me more what would you expect to see in the SKILL in this context?

rviscomi:
Thanks!

>I am confused about Baseline bit here - it's an experimental support, and based on above points you can see the complexities we are dealing with - still evolving and still limited support, can you please tell me more what would you expect to see in the SKILL in this context?

Sure. Agents are instructed to adhere to any fallback strategies necessary to preserve compatibility with the project's Baseline target, or Widely Available by default. So if an agent reads this skill, we need to provide it with the feature's Baseline status and a strategy for handling unsupported browsers. Here's an example from the LoAF guide:

https://github.com/GoogleChrome/guidance/blob/035b0c7d53b32d91d376ccd5ebed77fbb83f5821/guides/performance/identify-heavy-scripts/guide.md?plain=1#L76-L80

"can be safely used without fallbacks" might apply to the declarative API since they're just harmless HTML attributes. But usage like this from the imperative API needs to be used defensively to prevent runtime errors:

```js
navigator.modelContext.registerTool()
```

For example, here's the `Fallback strategies` section from one of the view transitions guides showing how to safely use feature detection:

https://github.com/GoogleChrome/guidance/blob/035b0c7d53b32d91d376ccd5ebed77fbb83f5821/guides/user-experience/directional-navigation-transitions/guide.md?plain=1#L110-L136

So for WebMCP, we would need these sections added to the declarative and imperative API guides, using the `BASELINE_STATUS` macro to render the feature status at build time and guidance to either do nothing because it's safe to use or to use feature detection.

skills-drafts/webmcp/references/imperative.md
Kulikowski: We have no conventions yet for structured output, it's just a plain string - natural language or javascript object is what we've seen people doing. Best practice would be to evaluate it to understand what is the best "format" for you particular use case.

andreban: I think the point to keep in mind is that the output will make it back to the LLM. Those can be data, error messages or even further instructions to the LLM.

beaufortfrancois:
@rviscomi I've added `BASELINE_STATUS` macros as suggested.

skills-drafts/webmcp/references/declarative.md
beaufortfrancois: `input[type=file]` WebMCP support is gated behind a dedicated testing flag (`WebMCPDeclarativeFileInput`) for now.

Re: `select[multiple]` vs `select`, I'm not sure about the following diff value for the agent:

image

skills-drafts/webmcp/references/declarative.md
beaufortfrancois: You're right. I've updated it.

skills-drafts/webmcp/references/imperative.md
beaufortfrancois: I've updated it with your feedback.

Kulikowski:
@rviscomi - @beaufortfrancois and I addressed your feedback (Baseline status, form validation error messages, format for return message in imperative, when to discourage WebMCP and when to avoid declarative API), please take a look :)

Kulikowski:
We will continue to update it as API stabilizes and as we have more clarity

rviscomi (approved):
Thanks everyone. LGTM to merge.

skills-drafts/webmcp/references/declarative.md
rviscomi: Gotcha thanks

Issue #193, opened 2026-03-02
Create guide and evals for the visually-texture-content use case


paulirish:
Apply realistic weathering and texture patterns to elements to give them an organic, aged, or physical material appearance.

Affected web-feature IDs: [masks](https://webstatus.dev/features/masks)

Use case subdir: [guides/visual-design/visually-texture-content](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/visual-design/visually-texture-content)

Related features: #65

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

Issue #192, opened 2026-03-02
Create guide and evals for the stencil-cutouts use case


paulirish:
Combine multiple shapes to create complex cutouts or 'knockout' effects in elements, such as adding a notch to an element.

Affected web-feature IDs: [masks](https://webstatus.dev/features/masks)

Use case subdir: [guides/user-experience/stencil-cutouts](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/user-experience/stencil-cutouts)

Related features: #65

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [ ] Use case guidance (guide.md)
- [ ] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

Issue #191, opened 2026-03-02
Create guide and evals for the soft-edge-content-fade use case


paulirish:
Apply a transparency gradient to content edges to indicate further scrollable areas or to obscure payment-walled text.

Affected web-feature IDs: [masks](https://webstatus.dev/features/masks)

Use case subdir: [guides/visual-design/soft-edge-content-fade](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/visual-design/soft-edge-content-fade)

Related features: #65

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

Issue #189, opened 2026-03-02
Create guide and evals for the complex-shapes use case


paulirish:
Clip elements and their content to any free-form shape, like a symbol, brush stroke, or organic texture for more expressive designs.

Affected web-feature IDs: [masks](https://webstatus.dev/features/masks), [svg](https://webstatus.dev/features/svg)

Use case subdir: [guides/visual-design/complex-shapes](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/visual-design/complex-shapes)

Related features: #65

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

PR #613, opened 2026-04-20
Add guides for the CSS masking use cases

taraojo:
Guides, expectations and generated demo.html updates for all CSS masking use cases:

- complex-shapes #189 
- soft-edge-content-fade #191 
- stencil-cutouts #192 
- visually-texture-content #193


rviscomi (commented):
Did a quick scan of the use cases, and it seems not all demo files may be using the required fallbacks. Could you double check?

taraojo:
> Did a quick scan of the use cases, and it seems not all demo files may be using the required fallbacks. Could you double check?

Totally missed that - updated!

guides/user-experience/complex-shapes/guide.md
```diff
@@ -4,4 +4,83 @@ description: Clip elements and their content to any free-form shape, like a symb
...
+  - https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/mask
+  - https://web.dev/articles/building/a-theme-switch-component
+  - https://web.dev/articles/css-masking
+  - https://web.dev/articles/how-we-built-designcember
+  - https://web.dev/learn/css/paths-shapes-clipping-masking
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Masking/Introduction
+  - https://css-tricks.com/clipping-masking-css/
+  - https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorials/SVG_from_scratch/SVG_and_CSS
+  - https://www.joshwcomeau.com/svg/friendly-introduction-to-svg/
+  - https://css-tricks.com/svg-properties-and-css/
```
jamesnw: I don't know that generic SVG articles are useful here- would it be better to keep this scoped to aricles on masking?

guides/user-experience/complex-shapes/guide.md
```diff
@@ -4,4 +4,83 @@ description: Clip elements and their content to any free-form shape, like a symb
...
+
+  /* Standard property for modern browsers */
+  mask-image: url('mask.png');
+  mask-size: cover;
+  mask-repeat: no-repeat;
+}
+```
+
+### Using an SVG element in HTML
+You can also reference a `<mask>` element defined in your HTML.
```
jamesnw:
```suggestion
You can also reference a `<mask>` element defined in an inline SVG in your page's HTML.
```

guides/user-experience/complex-shapes/demo.html
```diff
@@ -1,41 +1,77 @@
...
+    /* Fallback for browsers that support clip-path but not mask-image */
+    @supports (not (mask-image: url(x))) and (clip-path: inset(0)) {
+      .shaped-element {
+        /* Use a simple rounded rectangle as fallback */
+        clip-path: inset(5% round 15px);
+      }
+    }
+
+    /* Hide the SVG from view */
+    svg {
```
jamesnw: I'd recommend setting the width and height in HTML markup instead, so other SVGs aren't impacted, and the knowledge about the svg's use as just for masking is on the svg.

guides/user-experience/complex-shapes/demo.html
```diff
@@ -1,41 +1,77 @@
...
+  <title>Complex Shapes Demo</title>
+  <style>
+    body {
+      font-family: sans-serif;
+      display: flex;
+      justify-content: center;
+      align-items: center;
+      height: 100vh;
+      margin: 0;
+      background-color: #f0f0f0;
```
jamesnw:
```suggestion
      background-color: black;
```

This makes it clearer that the text is being clipped, and not just white-on-white.

guides/user-experience/complex-shapes/demo.html
```diff
@@ -1,41 +1,77 @@
...
+
 <body>
-  <div class="custom-shape">
-    This text is clipped by the heart!
+  <!-- SVG Mask Definition -->
+  <svg>
+    <defs>
+      <!-- Use objectBoundingBox to make it scale -->
+      <mask id="star-mask" maskContentUnits="objectBoundingBox">
+        <!-- White reveals, black hides -->
```
jamesnw: Optional: Since the guide mentions partial transparency as a motivating factor for `mask-image` over `clip-path`, it would be nice if there was a gray shape in the mask to demonstrate that.

guides/user-experience/complex-shapes/demo.html
```diff
@@ -1,41 +1,77 @@
...
+        <!-- A simple star shape path in normalized coordinates (0 to 1) -->
+        <path
+          d="M 0.5,0 L 0.625,0.375 L 1,0.375 L 0.7,0.625 L 0.8125,1 L 0.5,0.75 L 0.1875,1 L 0.3,0.625 L 0,0.375 L 0.375,0.375 Z"
+          fill="white" />
+      </mask>
+    </defs>
+  </svg>
+  
+  <div class="shaped-element">
+    This text is clipped by the star!
```
jamesnw:
```suggestion
    This element (including text) is clipped by the star!
```

guides/user-experience/complex-shapes/expectations.md
```diff
@@ -0,0 +1,5 @@
+- The element with class `.shaped-element` has the `mask-image` property applied.
+- The element with class `.shaped-element` has the `-webkit-mask-image` property applied for older browser support.
+- The mask references a valid image URL or an SVG mask ID.
```
jamesnw:
```suggestion
- The mask references a valid image URL or an ID of a `<mask>` element in an `<svg>`.
```

guides/user-experience/complex-shapes/demo.html
```diff
@@ -1,41 +1,77 @@
...
+      align-items: center;
+      color: white;
+      text-align: center;
+      font-weight: bold;
+      padding: 40px 0;
+      box-sizing: border-box;
+    }
+
+    /* Fallback for browsers that support clip-path but not mask-image */
+    @supports (not (mask-image: url(x))) and (clip-path: inset(0)) {
```
jamesnw: `clip-path` is widely available, I don't think we need to check support here. Also, this will fallback in browsers that support the prefixed version- should that be included in the check?

Also update the guide to match.

guides/user-experience/complex-shapes/guide.md
```diff
@@ -4,4 +4,83 @@ description: Clip elements and their content to any free-form shape, like a symb
...
+
+```html
+<!-- Opaque areas reveal content, transparent areas hide it -->
+<svg width="0" height="0">
+  <defs>
+    <mask id="custom-shape">
```
jamesnw:
```suggestion
    <mask id="custom-shape" maskContentUnits="objectBoundingBox">
```

I think you probably want to add a comment or prose saying why this is needed.

guides/user-experience/complex-shapes/guide.md
```diff
@@ -4,4 +4,83 @@ description: Clip elements and their content to any free-form shape, like a symb
...
+  - https://css-tricks.com/svg-properties-and-css/
 ---
+
+## Overview
+To clip elements to complex, free-form shapes like brush strokes or organic textures, use CSS Masking (`mask-image`). While `clip-path` is excellent for geometric shapes or vector paths, `mask-image` allows you to use images (like PNGs with transparency) or SVGs to define the visible area of an element. This approach is more expressive because it supports semi-transparency, allowing for soft edges and complex textures that are difficult or impossible to achieve with `clip-path`.
+
+## Implementation
+To implement complex shapes using CSS masks:
+
+### Using a raster image or SVG file
```
jamesnw: It would be nice to see this in the demo, but I'm guessing it isn't easy to have additional files.  

```suggestion
### Using transparency from an image
```

guides/user-experience/complex-shapes/guide.md
```diff
@@ -4,4 +4,83 @@ description: Clip elements and their content to any free-form shape, like a symb
...
+
+```css
+.shaped-element {
+  /* MANDATORY: Use vendor prefix for wider support in older browsers */
+  -webkit-mask-image: url('mask.png');
+  -webkit-mask-size: cover; /* Scale mask to cover element */
+  -webkit-mask-repeat: no-repeat; /* Do not tile the mask */
+
+  /* Standard property for modern browsers */
+  mask-image: url('mask.png');
```
jamesnw: Perhaps use an .svg as an example to help differentiate from the inline HTML below?

guides/user-experience/complex-shapes/guide.md
```diff
@@ -4,4 +4,83 @@ description: Clip elements and their content to any free-form shape, like a symb
...
+  mask-size: cover;
+  mask-repeat: no-repeat;
+}
+```
+
+### Using an SVG element in HTML
+You can also reference a `<mask>` element defined in your HTML.
+
+```html
+<!-- Opaque areas reveal content, transparent areas hide it -->
```
jamesnw:
```suggestion
<!-- White areas reveal content, black or transparent areas hide it -->
```

guides/user-experience/complex-shapes/guide.md
```diff
@@ -4,4 +4,83 @@ description: Clip elements and their content to any free-form shape, like a symb
...
 ---
+
+## Overview
+To clip elements to complex, free-form shapes like brush strokes or organic textures, use CSS Masking (`mask-image`). While `clip-path` is excellent for geometric shapes or vector paths, `mask-image` allows you to use images (like PNGs with transparency) or SVGs to define the visible area of an element. This approach is more expressive because it supports semi-transparency, allowing for soft edges and complex textures that are difficult or impossible to achieve with `clip-path`.
+
+## Implementation
+To implement complex shapes using CSS masks:
+
+### Using a raster image or SVG file
+You can use a URL to reference a PNG with transparency or an SVG file.
```
jamesnw:
```suggestion
You can use the transparency of an image as a mask, with opaque parts visible and transparent parts hidden. This can be a PNG, SVG, or other image with transparency, or a generated image, like a CSS gradient.
```

guides/user-experience/soft-edge-content-fade/guide.md
```diff
@@ -3,4 +3,74 @@ name: soft-edge-content-fade
...
 web-feature-ids:
   - masks
+sources:
+  - https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/mask
+  - https://web.dev/articles/building/a-theme-switch-component
+  - https://web.dev/articles/css-masking
+  - https://web.dev/articles/how-we-built-designcember
+  - https://web.dev/learn/css/paths-shapes-clipping-masking
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Masking/Introduction
+  - https://css-tricks.com/clipping-masking-css/
```
jamesnw:
```suggestion
  - https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/mask
  - https://web.dev/articles/css-masking
  - https://web.dev/learn/css/paths-shapes-clipping-masking
  - https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Masking/Introduction
  - https://css-tricks.com/clipping-masking-css/
```
While the others mention masking, they are focused on other techniques.

guides/user-experience/soft-edge-content-fade/demo.html
```diff
@@ -1,52 +1,65 @@
...
 </head>
-
 <body>
-  <div class="fade-container">
-    <h2>Premium Article</h2>
-    <p>This is the start of an incredibly interesting article. Read on to find out more. But wait, you need to be a
-      premium member to read the rest! The text slowly fades out into nothingness to entice you to subscribe. This
-      content continues but is hidden by the mask.</p>
+
+  <div class="scroll-container">
```
jamesnw: It struck me as odd initially that the whole card fades out- I _think_ I generally see just the text fade out, not its container. I'm not sure if there's a standard or right way on this.

guides/user-experience/soft-edge-content-fade/demo.html
```diff
@@ -1,52 +1,65 @@
...
-      overflow: hidden;
-      line-height: 1.5;
-      background: white;
+      height: 200px;
+      background-color: white;
       padding: 20px;
-      border-radius: 8px;
-      box-shadow: 0 2px 10px rgba(0, 0, 0, 0.1);
-      mask-image: linear-gradient(to bottom, black 60%, transparent 100%);
+      overflow-y: auto;
```
jamesnw: If this scrolls, I think the expectation is that there is sufficient padding at the bottom of the scroll container so that when you reach the bottom, the text isn't faded out. Alternatively, you could demonstrate the paywall use case, and disable scroll.

guides/user-experience/stencil-cutouts/guide.md
```diff
@@ -3,4 +3,74 @@ name: stencil-cutouts
 description: Combine multiple shapes to create complex cutouts or 'knockout' effects in elements, such as adding a notch to an element.
 web-feature-ids:
   - masks
+sources:
+  - https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/mask
+  - https://web.dev/articles/building/a-theme-switch-component
+  - https://web.dev/articles/css-masking
+  - https://web.dev/articles/how-we-built-designcember
```
jamesnw:
```suggestion
```
I don't think the designcember article is relevant here, but the theme-switch-component is fairly relevant.

guides/user-experience/stencil-cutouts/guide.md
jamesnw: I'm not quite sure what a "Stencil" means here, and am not finding many examples of that term on CodePen. I would assume stencil means text, so I'm wondering if there is a better name for this guide?

guides/user-experience/stencil-cutouts/guide.md
```diff
@@ -3,4 +3,74 @@ name: stencil-cutouts
...
+  - https://web.dev/learn/css/paths-shapes-clipping-masking
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Masking/Introduction
+  - https://css-tricks.com/clipping-masking-css/
 ---
+
+## Overview
+To create complex cutouts or "knockout" effects (like adding a notch to a card or creating a stencil effect with a hole), use CSS Masking. The most robust and widely supported method for complex cutouts is using an SVG mask, where you can combine shapes using standard SVG elements. For simpler cutouts, like a circular hole or a simple notch, you can use a CSS gradient that includes transparent areas.
```
jamesnw:
```suggestion
To create complex cutouts or "knockout" effects (like adding a notch to a card or creating a stencil effect with a hole), use CSS Masking. For complex cutouts, use an SVG mask, where you can create a mask by combining shapes using standard SVG elements. For simpler cutouts, like a circular hole or a simple notch, you can use a CSS gradient that includes transparent areas.
```

guides/user-experience/shaped-cutouts/demo.html
jamesnw, line 1: I think this demo should either use a gradient mask (`mask-image: radial-gradient(circle at center, transparent 20px, black 20px);` or find a more complex use case for use with SVG. 

I also think that the cutout should be on an element that doesn't contain text, or on a ::before pseudo element potentially. That way if the mask fails or stretches unexpectedly, content is not lost. That should be mentioned in the guide as well.

As it is, if the mask fails (can't be loaded), the screen is blank. If the browser doesn't support mask-image, the mask won't be applied, and we don't need to swap the colors 

Another way to strengthen the demo would be to add a body background that demonstrates that you aren't just overlaying a white circle for the effect, but that's minor and unlikely to be seen.

guides/user-experience/stencil-cutouts/guide.md
```diff
@@ -3,4 +3,74 @@ name: stencil-cutouts
...
+
+If a browser does not support `mask-image` or the prefixed version:
+- The element will not have the cutout and will display as a solid shape.
+- Ensure the content is still readable and the layout does not break without the cutout (progressive enhancement).
+- For critical UI shapes, consider using an inline SVG instead of CSS masking to ensure the shape is rendered correctly in all browsers.
+
+```css
+/* Fallback: Ensure content is visible even if mask fails */
+.notched-card {
+  background-color: #f0f0f0; /* Fallback background */
```
jamesnw: This would put white text on a white background, right?

guides/user-experience/visually-texture-content/demo.html
```diff
@@ -1,83 +1,90 @@
...
 
-    .badge-content {
-      text-align: center;
-      padding: 20px;
-      mask-image: url('grunge-texture.svg');
-      mask-mode: luminance;
-      mask-size: cover;
+    /* Fallback using CSS pattern for browsers without mask support */
+    @supports not (mask-image: url(x)) {
+      .weathered-bg {
```
jamesnw: This needs size- 
```suggestion
      .weathered-bg {
        height:100%;
        width:100%;
        position: absolute;
```

guides/user-experience/visually-texture-content/guide.md
```diff
@@ -3,4 +3,78 @@ name: visually-texture-content
 description: Apply realistic weathering and texture patterns to elements to give them an organic, aged, or physical material appearance.
 web-feature-ids:
   - masks
+sources:
+  - https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/mask
+  - https://web.dev/articles/building/a-theme-switch-component
+  - https://web.dev/articles/css-masking
+  - https://web.dev/articles/how-we-built-designcember
```
jamesnw:
```suggestion
  - https://web.dev/articles/css-masking
```

guides/user-experience/visually-texture-content/guide.md
```diff
@@ -3,4 +3,78 @@ name: visually-texture-content
...
+  mask-repeat: repeat;
+  mask-size: 300px;
+}
+```
+
+### Method 2: Using CSS Gradients for geometric patterns
+You can generate patterns using CSS gradients. This is self-contained and does not require external image files.
+
+```css
+.patterned-element {
```
jamesnw, line 35: I think this would be clearer if you create a `--checkerboard-gradient: linear-gradient...` property and then reuse it where needed.

guides/user-experience/visually-texture-content/demo.html
jamesnw, line 1: Would it be possible to demo with a grunge texture as well, since that's the primary use case listed?

jamesnw (changes requested):
Overall, this looks really good! There are a number of suggestions repeated across multiple use cases, but I didn't want to overwhelm, so there are likely some suggestions that apply across use cases that I just mentioned on one.

A big question I have is around baseline support. `masks` hits baseline in 5 weeks, so I wonder about putting the prefix in the fallback? As it stands, the fallback in the @supports` section ignores the prefixed version, and the fallback is applied to browsers that support it. 

Should we mention `mask-type: alpha` vs `mask-type: luminance` more explicitly somewhere? 

Let me know if you have any questions!

guides/user-experience/complex-shapes/guide.md
taraojo: I added these as `svg` is in the list of `web-feature-ids` above. I thought we would want to map sources to the feature list. Happy to remove if you think best?

guides/user-experience/complex-shapes/guide.md
taraojo: Added a sentence and comment to highlight - does this make sense?

guides/user-experience/soft-edge-content-fade/demo.html
taraojo: Opted for disabling the scroll here

guides/user-experience/soft-edge-content-fade/demo.html
taraojo: I've updated this so just the text fades out as part of the "paywall" demo. I can imagine people have different kinds of use cases for this so I haven't added specific best practice for this, but let me know if you think otherwise.

guides/user-experience/stencil-cutouts/guide.md
taraojo: I've changed to `shaped-cutouts` though they seem the same to me :)

taraojo (commented):
I think I've addressed everything, let me know what you think!

guides/user-experience/complex-shapes/guide.md
```diff
@@ -4,4 +4,88 @@ description: Clip elements and their content to any free-form shape, like a symb
...
+sources:
+  - https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/mask
+  - https://web.dev/articles/building/a-theme-switch-component
+  - https://web.dev/articles/css-masking
+  - https://web.dev/articles/how-we-built-designcember
+  - https://web.dev/learn/css/paths-shapes-clipping-masking
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Masking/Introduction
+  - https://css-tricks.com/clipping-masking-css/
+  - https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorials/SVG_from_scratch/SVG_and_CSS
+  - https://www.joshwcomeau.com/svg/friendly-introduction-to-svg/
```
jamesnw:
```suggestion
```
I think this can be removed per the conversation in the Web Features Guidance chat.

guides/user-experience/complex-shapes/guide.md
jamesnw: I don't have a strong enough sense on how these are used, so I'll defer to @malchata perhaps? I'm fine either way.

guides/user-experience/shaped-cutouts/demo.html
jamesnw: The change isn't quite what I had in mind here, and I think it's because we're trying to get this svg mask to work.

- The ::before should be stacked on top of the .notched-card element, and each should provide their own background
- You'll need to move the border radius for the top to the ::before

This CSS I think works, at least to show the stacking and separation. It uses linear-gradient rather than svg, but you may find a way for svg to work here- you can't rely on the height of the container to stretch the mask anymore.

```css
    .notched-card {
      position: relative;
      width: 300px;
      color: white;
      padding: 20px;
      padding-top: 0;
      box-sizing: border-box;
      background-color: #3498db;
      border-radius: 0 0 8px 8px;
    }

    /* Apply the mask to a pseudo-element to protect text legibility and progressive enhancement */
    .notched-card::before {
      content: '';
      position: absolute;
      left: 0;
      right: 0;
      --offset: 20px;
      top: calc(-1 * var(--offset));
      height:var(--offset);
      background-color: #3498db;
      border-radius: 8px 8px 0 0;
      
      
      /* Apply the mask to the background shape only */
      -webkit-mask-image: radial-gradient(circle at top center, transparent 20px, black 20px);
      mask-image: radial-gradient(circle at top center, transparent 20px, black 20px);
      
    }
```

guides/user-experience/visually-texture-content/demo.html
```diff
@@ -3,79 +3,135 @@
...
+      /* Apply the grunge SVG mask */
+      -webkit-mask-image: url('grunge-texture.svg');
+      -webkit-mask-repeat: repeat;
+      -webkit-mask-size: 200px;
+
       mask-image: url('grunge-texture.svg');
-      mask-mode: luminance;
-      mask-size: cover;
+      mask-repeat: repeat;
+      mask-size: 200px;
```
jamesnw, line 82: I think you're missing `mask-mode: luminance` here. SVG images default to alpha, like other images, where inline SVG masks default to luminance. Alternatively you could just inline the grunge svg (and maybe turn it into a mask?

guides/user-experience/visually-texture-content/guide.md
```diff
@@ -3,4 +3,76 @@ name: visually-texture-content
...
+  /* Apply a checkerboard pattern as a mask */
+  -webkit-mask-image: var(--checkerboard-gradient);
+  -webkit-mask-size: 20px 20px;
+  -webkit-mask-position: 0 0, 0 10px, 10px -10px, -10px 0px;
+  
+  mask-image: var(--checkerboard-gradient);
+  mask-size: 20px 20px;
+  mask-position: 0 0, 0 10px, 10px -10px, -10px 0px;
+}
+```
```
jamesnw, line 51: It might be worth including a decision matrix or table on when alpha and luminance are the default modes?

jamesnw (commented):
Overall, lots of great improvements! I flagged a couple additional items, but this is close!

guides/user-experience/shaped-cutouts/demo.html
taraojo: Thanks James, I've been digging into this more and haven't found a nice way to make it work with an svg. Apparently referencing internal IDs (url(#id)) on pseudo-elements is unreliable. Given this is the case, do you think it makes sense to rework the guide? I'm thinking specifically for complex cutouts.

We're recommending pseudo-elements for a11y when elements contain text and I assume that will not always be the case for some users. So maybe we:

1. Keep "Using a single CSS Gradient for simple cutouts"
2. Keep "Using an SVG Mask (Recommended for complex cutouts)" - *without* pseudo-elements, only when there is no important content in the element like text
3. Recommend using using a Data-URI SVG (For complex cutouts) - *with* pseudo-elements, when there is text involved. This is the one I'm not sure about, but the only way I've been able to get your approach to work at the moment with an svg. I don't know if we want to recommend this as best practice.

What do you think?

guides/user-experience/visually-texture-content/demo.html
taraojo: Oh nice, I've added `mask-mode: luminance`

guides/user-experience/shaped-cutouts/demo.html
jamesnw: TIL about internal ids and pseudo elements!

I think keep number 1, with the radial gradient.

For number 2, I think either find a use case without text, or alternatively apply it to a <div> positioned relative to the text box, rather than a pseudo-element. A pseudo element is useful for a11y here not because of its impact of the AT, but because it is a different element, and applying a mask to it won't visually hide text. You should be able to get the same effect by stacking an empty `<div>` above the element with the text.

guides/user-experience/complex-shapes/guide.md
rviscomi: @taraojo were these sources actually used to help generate the guide, or are they intended more as "additional reading" for agents?

taraojo: I used them to generate the initial draft of the guides, they're probably not needed for additional reading

rviscomi:
@taraojo is this ready for another look?

taraojo:
Yes, please review!

guides/user-experience/shaped-cutouts/demo.html
```diff
@@ -1,70 +1,307 @@
...
+      flex-direction: column;
       align-items: center;
-      height: 100vh;
-      margin: 0;
-      /* Grid background to demonstrate the transparent cutout */
-      background-color: #e5e9f0;
-      background-image: radial-gradient(#cbd5e1 1px, transparent 1px);
-      background-size: 20px 20px;
+      /* Subtle micro-pattern background */
+      background-image: radial-gradient(rgba(99, 102, 241, 0.15) 1px, transparent 1px);
```
jamesnw, line 31: Not really related, but the muted text on this background isn't accessible.

guides/user-experience/shaped-cutouts/guide.md
```diff
@@ -0,0 +1,159 @@
...
+2. An **empty decorative `div`** (`.card-accent`) placed next to it, which receives the custom SVG mask.
+
+By assigning both parts the same background color, they visually merge into a single, custom-shaped component.
+
+```html
+<!-- 1. Define the SVG mask (hidden from view) -->
+<svg width="0" height="0" style="position: absolute;" aria-hidden="true">
+  <defs>
+    <!-- Use objectBoundingBox to make the mask scale with the element -->
+    <mask id="card-accent-mask" maskContentUnits="objectBoundingBox">
```
jamesnw:
```suggestion
    <mask id="accent-stencil" maskContentUnits="objectBoundingBox">
```
To match the guide and expectations

jamesnw (changes requested):
This looks great! Just a couple small items.

I do recommend removing the hover animations and `pointer: cursor` - I don't think they're relevant to the use case, and get into accessibility questions regarding reduced motion and divs appearing interactive.

rviscomi (approved):
LGTM. Tested in Chrome.

Waiting for @jamesnw's LGTM before merging.

guides/user-experience/shaped-cutouts/guide.md
rviscomi: Updated

guides/user-experience/shaped-cutouts/demo.html
rviscomi: While not ideal I think this is safe because the demo is only used for calibration

rviscomi (commented):
Addressed @jamesnw's two comments

rviscomi:
> This looks great! Just a couple small items.

I'll take this as a conditional LGTM, so I'll go ahead and merge this now that the items have been addressed. @jamesnw feel free to open a PR if there's anything major.

PR #616, opened 2026-04-20
#336 shared-element-transitions

### LINKED ISSUE #336: Create guide and evals for the shared-element-transitions use case
Visually connect persisting elements across different page states or navigations (e.g. expanding a product thumbnail into a full-bleed hero image) by smoothly morphing their size, position, or other styling properties.

Affected web-feature IDs: [view-transitions](https://webstatus.dev/features/view-transitions)

Use case subdir: [guides/user-experience/shared-element-transitions](https://github.com/GoogleChrome/guidance/tree/main/guides/user-experience/shared-element-transitions)

Related features: #87

**Required files:**
- [x] demo.html
- [ ] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

jamesnw:
This could potentially be fleshed out later with more guides on transitioning cleanly, especially with text and images.


guides/user-experience/shared-element-transitions/guide.md
```diff
@@ -3,4 +3,162 @@ name: shared-element-transitions
...
+// Function called when a thumbnail is clicked
+function goFromListToDetail(e){
+  e.currentTarget.classList.add("selected");
+  const hero = document.getElementById("hero");
+  const bgColor = getComputedStyle(e.currentTarget).backgroundColor;
+  hero.style.background = bgColor;
+
+  // Trigger the transition, checking for support
+  if (!document.startViewTransition) {
+    document.body.classList.remove("detail");
```
malchata: Do you think the function should end right here if view transitions aren't supported?

```suggestion
    document.body.classList.remove("detail");
    return; // MANDATORY: End function execution if view transitions are not supported.
```

guides/user-experience/shared-element-transitions/guide.md
```diff
@@ -3,4 +3,162 @@ name: shared-element-transitions
...
+  }
+  const transition = document.startViewTransition(() => {
+    document.body.classList.remove("detail");
+  });
+  // Clean up the list view
+  transition.finished.finally(() => {
+    // Remove selected classList to remove view-transition-names
+    document.querySelectorAll("selected").forEach(
```
malchata: Missing the class dot here:

```suggestion
    document.querySelectorAll(".selected").forEach(
```

guides/user-experience/shared-element-transitions/guide.md
malchata: If we're transitioning from the list _to_ the detail, should this be `classList.add` instead?

guides/user-experience/shared-element-transitions/guide.md
```diff
@@ -3,4 +3,162 @@ name: shared-element-transitions
...
+  const hero = document.getElementById("hero");
+  const bgColor = getComputedStyle(e.currentTarget).backgroundColor;
+  hero.style.background = bgColor;
+
+  // Trigger the transition, checking for support
+  if (!document.startViewTransition) {
+    document.body.classList.remove("detail");
+  }
+  const transition = document.startViewTransition(() => {
+    document.body.classList.remove("detail");
```
malchata: Similar comment as from above re: `classList.add` instead.

guides/user-experience/same-document-transitions/guide.md
```diff
@@ -3,4 +3,162 @@ name: shared-element-transitions
...
+For Single-Page Applications (SPAs) or simple state changes, wrap the logic that updates the DOM in `document.startViewTransition`. The browser captures a snapshot of the current state, runs the update, and then captures the new state.
+
+```javascript
+function navigate(view) {
+  // MANDATORY: Wrap the update in startViewTransition
+  document.startViewTransition(() => updateDOM(view));
+}
+```
+
+### Step 2: Assign Shared Transition Names
```
malchata, line 36: Should it be emphasized that `view-transition-name` should be unique per element? And elaborate on what happens if there are naming collisions?

guides/user-experience/same-document-transitions/guide.md
```diff
@@ -3,4 +3,162 @@ name: shared-element-transitions
...
+  - https://web.dev/learn/css/view-transitions-spas
 ---
+
+# Shared Element Transitions
+
+## The Problem
+
+Web sites often provide multiple views of an object, for instance a list of products, and then a detail page for each product. Navigating between the two views often feels disconnected. When a user clicks a product thumbnail to view its details, the thumbnail disappears and a new, larger image appears instantly elsewhere on the screen. This lack of continuity makes it harder for users to track relationships between elements.
+
+## The Solution
```
malchata, line 19: Sort of a top level comment, not sure quite where it goes, but it might be good to mention that the transitions occur in the top layer, and thus will occur even above elements with a high `z-index` value.

guides/user-experience/shared-element-transitions/guide.md
```diff
@@ -3,4 +3,162 @@ name: shared-element-transitions
...
+
+## The Solution
+
+The **View Transitions API** allows you to specify element pairs that exist in different states before and after a transition. When a transition occurs, the browser identifies these shared elements by their shared unique `view-transition-name`. It then automatically calculates the difference in their position, size, and styling, and animates them smoothly from the old state to the new state. 
+
+## Implementation Guide
+
+### Step 1: Wrap State Changes in `startViewTransition`
+
+For Single-Page Applications (SPAs) or simple state changes, wrap the logic that updates the DOM in `document.startViewTransition`. The browser captures a snapshot of the current state, runs the update, and then captures the new state.
```
malchata: Worth mentioning here that the transition behavior is async, as `startViewTransition` returns a `Promise`.

malchata (changes requested):

guides/user-experience/shared-element-transitions/guide.md
jamesnw: It actually returns an object with `ready` and `finished` Promises, as well as a few other items. The `finished` Promise is used in an example later- is that sufficient?

guides/user-experience/shared-element-transitions/demo.html
```diff
@@ -1,115 +1,197 @@
...
+        if (!document.startViewTransition) {
+          document.body.classList.remove("detail");
+          return;
+        }
+        const transition = document.startViewTransition(() => {
+          document.body.classList.remove("detail");
+        });
+        transition.finished.finally(() => {
+          // Remove selected classList to remove view-transition-names
+          document.querySelectorAll("selected").forEach(
```
rviscomi:
```suggestion
          document.querySelectorAll(".selected").forEach(
```

guides/user-experience/shared-element-transitions/guide.md
```diff
@@ -3,4 +3,165 @@ name: shared-element-transitions
 description: Visually connect persisting elements across different page states or navigations (e.g. expanding a product thumbnail into a full-bleed hero image) by smoothly morphing their size, position, or other styling properties.
```
rviscomi: So we have [cross-document-transitions](https://github.com/GoogleChrome/guidance/blob/main/guides/user-experience/cross-document-transitions/guide.md?plain=1) and IIUC this use case is effectively the same-document equivalent. WDYT about renaming this to `same-document-transitions` and calling that out more clearly in the description? That way coding agents should know never to fetch this guide if they have a cross-document use case.

guides/user-experience/shared-element-transitions/guide.md
malchata: I think so!

guides/user-experience/shared-element-transitions/guide.md
jamesnw: Good idea, renamed!

guides/user-experience/same-document-transitions/guide.md
```diff
@@ -0,0 +1,167 @@
...
+.thumbnail.selected {
+  view-transition-name: hero;
+}
+```
+
+When a thumbnail is clicked, we need to prepare the list view by assigning the `view-transition-name` using the `.selected` class selector, and making any changes to the DOM before starting the transition.
+
+Then, you can call `document.startViewTransition()`, and apply the changes to transition the page from the detail to list view.
+
+After navigating back to the list view, you must clean up the view transition classes to prevent the next navigation from erroring. You can perform this cleanup after the transition's `finished` promise resolves.
```
rviscomi, line 77: The example code below shows the cleanup happening in `goFromListToDetail`, but the demo shows it happening in `goFromDetailToList`. The demo works well, so I think maybe the code below might have a bug. WDYT?

jamesnw: Good catch- I fixed the guide. Thanks!

rviscomi (approved):
LGTM

PR #617, opened 2026-04-20
Recommend `:is()` instead of duplicate CSS rules

LeaVerou:
Fix one of the [review](https://github.com/GoogleChrome/guidance/issues/332#issuecomment-4283244943) points, namely the one about needing duplicate CSS rules (recommend `:is()` instead).

Also removed some trailing whitespace in these files.

`grader.ts` still needs to be regenerated.


dvdherron (approved):
LGTM! Thanks!

PR #620, opened 2026-04-21
CSS skill rewrite

### LINKED ISSUE #532: Create a high-level CSS skill
https://github.com/GoogleChrome/guidance/blob/main/skills-drafts/css/SKILL.md

LeaVerou:
See #532


skills-drafts/css/SKILL.md
```diff
@@ -7,47 +7,125 @@ description: Action-oriented guidelines for modern CSS architecture, layouts, an
...
 
-## 1. Architectural Foundations and The Cascade
-Effective CSS architecture focuses on managing the global namespace and controlling specificity natively.
+## 1. Foundations
+
+Be allergic to knowledge duplication. Prefer variables over repetition, but whenever possible, prefer built-in conventions such as:
+- `currentColor` instead of defining a variable and setting `color` to it
+- The `inherit` keyword instead of defining a variable on the parent and using it on the same property across parent and child.
+- `em` units instead of `font-size: var(--size)`
+- `cqw`/`cqh` units instead of repeating box model values.
```
jamesnw: Since we recommend logical properties, do we want to say `cqi`/`cqb`?

skills-drafts/css/SKILL.md
```diff
@@ -7,47 +7,125 @@ description: Action-oriented guidelines for modern CSS architecture, layouts, an
...
+- Do not use logical properties indiscriminately — ask yourself "would I want this to flip in RTL?" — if the answer is no, use the physical property instead.
+- Consider different viewing modes (dark mode, high contrast mode), different viewport sizes, and different input modes (touch, keyboard, pointer).
+
+## 2. Inheritance and The Cascade
+
+**Avoid** introducing BEM naming conventions to manage specificity.
+Instead, use modern CSS features such as cascade layers and `:where()` to make cascade behavior predictable and follow author intent.
+
+Use cascade layers (`@layer`) to define explicit priority zones (e.g., `reset`, `base`, `theme`, `components`, `utilities`), and declare their order upfront (e.g. `@layer reset, base, theme, components, utilities;`).
+Within each layer, use `:where()` to make selectors only compete based on meaningful signals, not incidental filters (`:not()` edge cases, remote ancestors, etc.) or for one-off easily overridable defaults..
```
jamesnw:
```suggestion
Within each layer, use `:where()` to make selectors only compete based on meaningful signals, not incidental filters (`:not()` edge cases, remote ancestors, etc.) or for one-off easily overridable defaults.
```

skills-drafts/css/SKILL.md
```diff
@@ -7,47 +7,125 @@ description: Action-oriented guidelines for modern CSS architecture, layouts, an
...
+
+## 3. Selectors and scoping
+
+Modern browser-native selectors reduce the need for preprocessors and complex state-tracking in JS.
+
+### Prefer CSS selectors over JS for complex element targeting
+
+- **DO** use `:has()` to style parents based on child state instead of managing classes in JS (e.g. `label:has(:checked)` instead of a manual `label.has-checked` class)
+- **DO NOT** nest `:has()` or use pseudo-elements inside it (browser API limitation)
+- Use `:nth-child(<An+B> of <selector>)` when you need to style every n-th element of a certain type. E.g. `details:nth-child(1 of [open])` will style the first open `<details>` element it finds, whereas `details[open]:first-child` would style only the first child iff it was open.
```
jamesnw:
```suggestion
- Use `:nth-child(<An+B> of <selector>)` when you need to style every n-th element of a certain type. E.g. `details:nth-child(1 of [open])` will style the first open `<details>` element it finds, whereas `details[open]:first-child` would style only the first child if it was open.
```
Or is this `if and only if`?

guides/css/SKILL.md
```diff
@@ -7,47 +7,125 @@ description: Action-oriented guidelines for modern CSS architecture, layouts, an
...
+Use cascade layers (`@layer`) to define explicit priority zones (e.g., `reset`, `base`, `theme`, `components`, `utilities`), and declare their order upfront (e.g. `@layer reset, base, theme, components, utilities;`).
+Within each layer, use `:where()` to make selectors only compete based on meaningful signals, not incidental filters (`:not()` edge cases, remote ancestors, etc.) or for one-off easily overridable defaults..
+
+<!-- - **DO** use keywords like `inherit`, `initial`, `unset`, or `revert` to explicitly control inheritance.
+TODO too vague move this elsewhere and ground it on use cases
+-->
+
+## 3. Selectors and scoping
+
+Modern browser-native selectors reduce the need for preprocessors and complex state-tracking in JS.
```
jamesnw, line 67: Do we need to be as elementary as specifying when to use element, class, and id selectors?

skills-drafts/css/SKILL.md
jamesnw:
```suggestion
- **DON'T** use gradient functions (`linear-gradient()`, `radial-gradient()`, or `conic-gradient()`) without a color space override if you are transitioning between colors that are highly saturated or very different.
```

jamesnw: There are no Don'ts in this section.

guides/css/SKILL.md
```diff
@@ -7,47 +7,125 @@ description: Action-oriented guidelines for modern CSS architecture, layouts, an
...
+Be allergic to knowledge duplication. Prefer variables over repetition, but whenever possible, prefer built-in conventions such as:
+- `currentColor` instead of defining a variable and setting `color` to it
+- The `inherit` keyword instead of defining a variable on the parent and using it on the same property across parent and child.
+- `em` units instead of `font-size: var(--size)`
+- `cqw`/`cqh` units instead of repeating box model values.
+- Code duplication is not knowledge duplication. The goal is robustness and maintainability, not saving characters.
+- Prefer **logical properties and values** over physical ones (e.g. `margin-inline-start` instead of `margin-left`) so that styles adapt to different writing modes and orientations. Even if the page author does not plan to localize, external translation tools often display translated text in context.
+- Do not use logical properties indiscriminately — ask yourself "would I want this to flip in RTL?" — if the answer is no, use the physical property instead.
+- Consider different viewing modes (dark mode, high contrast mode), different viewport sizes, and different input modes (touch, keyboard, pointer).
+
```
jamesnw, line 51: Other potential foundations- (wording could be improved)

- Don't over-declare CSS properties. Set values for what is needed, but no more. 
- Work with the user's preferences and the browser defaults, rather than completely overriding. 
- Use CSS to describe the intended styles, rather than prescribe specific solutions. (Not sure an agent can follow this, but it is a different way of reasoning)

jamesnw (commented):
Lots of good changes already! Some potential additions-
- custom properties
- styling lists
- counters
- styling form elements

skills-drafts/css/SKILL.md
LeaVerou: It was! I figured an LLM would know what it means, but after a little research that would be unlikely — even though it _does_ know what it means, it would deem this context as not sufficiently precise/mathematical and would be more likely to interpret it as a typo.

guides/css/SKILL.md
LeaVerou: I don't think so, except for cases where they are used incorrectly in the wild.

skills-drafts/css/SKILL.md
LeaVerou: This entire section needs a lot of rework — unfortunately the lack of gamut mapping makes the current guidance not work well in practice even though it would have been fine otherwise. 😢 

I've been avoiding it because I'm frankly at a bit of a loss about what to recommend. Using `oklch` as an interpolation space risks you getting colors wildly out of gamut but _does_ preserve chroma better, whereas `oklab` is better but can still give you muted colors in the middle. I'll probably outline a decision framework depending on the use case and the colors… (unless you have any better ideas!)

guides/css/SKILL.md
LeaVerou: All good ideas! As for the latter, they tend to need examples for these kinds of abstract things, but if you give them good examples they can be surprisingly good at it.

skills-drafts/css/SKILL.md
jamesnw: I think a decision framework makes a lot of sense here. There doesn't seem to be a single solution.

guides/css/SKILL.md
```diff
@@ -7,198 +7,226 @@ description: Action-oriented guidelines for modern CSS architecture, layouts, an
...
+}
+[popover].\:popover-open {
+  /* same styles again, for polyfilled popovers */
 }
 
-/* Unlayered styles (Highest priority outside of !important) */
-.emergency-override {
-  display: block;
+/* GOOD */
+[popover]:where(:popover-open, .\:popover-open) {
```
jamesnw, line 89: Nice- I'll add this to the popover polyfill docs as an option!

jamesnw:
Some additional ideas that came up in the overlap with other skills-

Should there be a section on pseudo elements, selectors and state-based attributes? The HTML skill had `DOs` to style with `details[open]` and `::backdrop`, but that might belong here. Another potentially missing piece here is alternate text in `content` values.

LeaVerou:
> Some additional ideas that came up in the overlap with other skills-
> 
> Should there be a section on pseudo elements, selectors and state-based attributes? The HTML skill had `DOs` to style with `details[open]` and `::backdrop`, but that might belong here. Another potentially missing piece here is alternate text in `content` values.

Alternate text for `content` definitely. Less sure about the others, I'll need to think more about it. Offhand, they seem more well suited to guides (a guide for dialogs and a guide for different use cases around details)

guides/css/SKILL.md
```diff
@@ -7,183 +7,448 @@ description: Action-oriented guidelines for modern CSS architecture, layouts, an
...
-- **DO** use a global `box-sizing: border-box` reset to ensure padding doesn't break layouts.
-- **DO** use keywords like `inherit`, `initial`, `unset`, or `revert` to explicitly control inheritance.
-- **DO** use **Cascade Layers** (`@layer`) to define explicit priority zones (e.g., `reset`, `base`, `theme`, `components`, `utilities`).
-- **DO** declare the layer order upfront in your main CSS file: `@layer reset, base, theme, components, utilities;`.
-- **DO** understand that `!important` declarations in a lower-priority layer override those in a higher-priority layer (precedence is reversed).
-- **DO** use specialized naming conventions like **BEM** (`.block__element--modifier`) to avoid namespace collisions.
-- **DO** use `isolation: isolate` to create a fresh stacking context without side-effects.
-
-### Code Example: Foundations & Reset
+1. [1. Foundations](#1-foundations)
```
castastrophe, line 10: This is an interesting idea of creating an internalized table of contents. Does that speed up processing for the models and is this something a macro could do consistently across all the skills? If it's helpful too, perhaps the model's guidance should note that every skill includes these internal links.

Another idea here, what if the internal links included a very small summary, making it easier for the model to figure out where to jump to. Or perhaps each heading in the toc could include keywords or common search terms?  Almost like an optimized index?

castastrophe: Can the models stop reading in a document once it has begun? I've read that ideally models should receive a small amount of highly-relevant content because there's less for them to parse and hold in-memory. So would it be more efficient to have smaller, more focused documents and an entry document with headings and keywords?

LeaVerou: See https://github.com/GoogleChrome/guidance/issues/684 for the ToC rationale (spoiler: It's part of Anthropic's guidance for skills > 100 loc)

> Another idea here, what if the internal links included a very small summary, making it easier for the model to figure out where to jump to. Or perhaps each heading in the toc could include keywords or common search terms? Almost like an optimized index?

I think the latter is the way to go, adding too much detail kinda defeats the point, but we should definitely try to maximize signal in the headings!

> Can the models stop reading in a document once it has begun?

My understanding is that they do not read it sequentially in the first place, and have different strategies for what to scan and how depending on the specifics.

> I've read that ideally models should receive a small amount of highly-relevant content because there's less for them to parse and hold in-memory.

Yes, for resources stored in the context window every token should matter and add value. For skills, generally only their metadata is loaded upfront, but the whole skill is loaded once it is seen as relevant. For guides, once they come up in a use case search, but I assume they'd be more or less loaded in the context window after that point. 

> So would it be more efficient to have smaller, more focused documents and an entry document with headings and keywords?

Yes and no. Anthropic does [recommend](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices#progressive-disclosure-patterns) breaking skills down and linking beyond a certain point of length, but [also](https://platform.claude.com/docs/en/agents-and-tools/agent-skills/best-practices#avoid-deeply-nested-references) against overly deep hierarchies.
Anecdotally, as another data point @rviscomi said he has experienced agents often not following links when they should.

rviscomi:
@LeaVerou are you still making changes or is this ready to open up for broader review?

LeaVerou:
> @LeaVerou are you still making changes or is this ready to open up for broader review?

I'm still making changes, but I'll try to wrap it up by EOD. Thanks for the ping!

guides/css/SKILL.md
```diff
@@ -7,183 +7,449 @@ description: Action-oriented guidelines for modern CSS architecture, layouts, an
...
+
+**Avoid** introducing BEM naming conventions to manage specificity.
+Instead, use modern CSS features such as cascade layers and `:where()` to make cascade behavior predictable and follow author intent.
+
+Use cascade layers (`@layer`) to define explicit priority zones (e.g., `reset`, `base`, `theme`, `components`, `utilities`), and declare their order upfront (e.g. `@layer reset, base, theme, components, utilities;`).
+Within each layer, use `:where()` to make selectors only compete based on meaningful signals, not incidental filters (`:not()` edge cases, remote ancestors, etc.) or for one-off easily overridable defaults.
+
+<!-- - **DO** use keywords like `inherit`, `initial`, `unset`, or `revert` to explicitly control inheritance.
+TODO too vague move this elsewhere and ground it on use cases
+-->
```
rviscomi: What's the resolution on this, should it be removed?

guides/css/SKILL.md
```diff
@@ -7,183 +7,449 @@ description: Action-oriented guidelines for modern CSS architecture, layouts, an
...
+
+For most styling purposes (e.g. colors, borders, backgrounds, typography, etc) treat these elements as normal text containers.
+
+- Use `:placeholder-shown` and `::placeholder` to style input placeholders.
+- Use `field-sizing: content` to make text fields size to content.
+- For `<textarea>` elements, use `resize: vertical` to disable horizontal resizing or `resize: none` to disable all resizing.
+
+#### Multiple choice controls (select, radios, checkboxes)
+
+- To select one among many options presented in a dropdown: Use a `<select>` + `appearance: base-select` + `::picker(select)`
```
rviscomi: Would this benefit from a `BASELINE_STATUS` note about browser support? Alternatively, maybe a `GUIDE_REF` link to the `branded-select-styling` guide?

rviscomi (commented):
Overall looks great. Just a couple of comments. Also directly committed a couple of typo fixes.

guides/css/SKILL.md
LeaVerou: It's not harmful, I just don't think it's going to affect much. We can uncomment it for now, until we have more practical guidance.

guides/css/SKILL.md
LeaVerou: I think a guide ref makes total sense! Re:`BASELINE_STATUS`, the guidance is still the same, since it's PE.

guides/css/SKILL.md
castastrophe, line 77: Do `:is()` and `:where()` fall within the baseline support for this project's purposes or would they require polyfill details as well?

LeaVerou: They are Baseline Widely Available, yes (since 2023!): 
- https://web-platform-dx.github.io/web-features-explorer/features/is/
- https://web-platform-dx.github.io/web-features-explorer/features/where/

guides/css/SKILL.md
LeaVerou: Done! I ended up writing a bit more to make it a little more concrete.

guides/css/SKILL.md
LeaVerou: Done!

rviscomi (approved):
LGTM! Let's wait for another LGTM from @castastrophe or @jamesnw before merging.

jamesnw (approved):
I've read this through a couple times, and it looks good to me!

LeaVerou:
> I've read this through a couple times, and it looks good to me!

Thanks! As I told @rviscomi there are lots of things I want to add to it (including some of the things you proposed here) but I think we can do those as incremental improvements after.

rviscomi:
Great! Let's merge.

cc @micahjo7 for #767

PR #621, opened 2026-04-21
tag version in modern-web-guidance release repo

connorjclark:
This should make GitHub tag + generate a zip file for us in the Releases page.


paulirish (approved):
well isn't that delightful.

i do really like that ghpages package.

paulirish (approved):
well isn't that delightful.

i do really like that ghpages package.

PR #632, opened 2026-04-22
revert gh-pages deploy but keep the good bits

paulirish:
Reverting the approach of deploying dashboard results to gh-pages. It ended up being a bad idea after a few follow-up fixes. SORRY MICAH.

But we're keeping the good stuff we discovered along the way:

- **`STATIC=true` mode** in [server.js](./eval-view/server.js): Still useful for testing static deployment structure locally.
- **Optimizations in dashboard**: Like checking `run.files` first and handling remote trajectories safely in [dashboard.js](./eval-view/dashboard.js).
- **mTLS handling** for opening app source for remote results in [dashboard.js](./eval-view/dashboard.js).
- **Harness upgrades** in [collection.ts](./harness/lib/collection.ts) to avoid greedy matching and support fallback runtime estimation for historical results (ayy!).

We also moved the `dist` folder for static serving up to the project root [dist/dashboard](./dist/dashboard) to keep `eval-view` cleaner.


micahjo7 (approved):
thank you for doing all of this!!

So, is the purpose of the STATIC mode essentially to validate that changes will work on the remote dashboard?

When running gd dashboard with STATIC=true, it creates:

eval-view/grouped-tasks.gen.json
eval-view/suites.gen.json

should we put these in gitignore?

paulirish:
> So, is the purpose of the STATIC mode essentially to validate that changes will work on the remote dashboard?

yup bingo. no more weird behavior when its deployed that we can't repro locally.

> should we put these in gitignore?

good call. that got lost.   restored it.

PR #633, opened 2026-04-23
Add `optimize-scroll-stability` guide

### LINKED ISSUE #311: Create guide and evals for the optimize-scroll-stability use case
Keep the page layout, scroll bar, and scroll position as stable as possible as users scroll through content-heavy pages (feeds, dashboards, long lists) where elements transition from hidden to visible.

Affected web-feature IDs: [contain-intrinsic-size](https://webstatus.dev/features/contain-intrinsic-size)

Use case subdir: [guides/performance/optimize-scroll-stability](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/performance/optimize-scroll-stability)

Related features: #44

**Required files:**
- [x] demo.html
- [ ] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

knowler:
Towards #311 

This adds the guide for `optimize-scroll-stability` which uses `contain-intrinsic-size` (and friends) for `content-visibility: auto`. I went with the simplest use case as we can include more advanced ones after I/O.

### DIFF
--- a/guides/performance/optimize-scroll-stability/guide.md
+++ b/guides/performance/optimize-scroll-stability/guide.md
@@ -3,4 +3,44 @@ name: optimize-scroll-stability
 description: Keep the page layout, scroll bar, and scroll position as stable as possible as users scroll through content-heavy pages (feeds, dashboards, long lists) where elements transition from hidden to visible.
 web-feature-ids:
   - contain-intrinsic-size
+  - content-visibility
+sources:
+  - https://web.dev/articles/content-visibility#specify_the_natural_size_of_an_element_with_contain-intrinsic-size
+  - https://12daysofweb.dev/2024/css-content-visibility/#pairing-with-contain-intrinsic-size
 ---
+`content-visibility: auto` is a useful tool for improving rendering performance as irrelevant content can be skipped for rendering.
+
+Since `content-visibility: auto` applies size containment to an element when its contents are being skipped, the dimensions of the element will be as if it did not have any content: only its padding and border size will contribute. This can have undesirable effects on an element’s surrounding context: when an element becomes relevant it can cause reflow, within a scrollport it can cause the scrollbar to grow or even jitter, or it can cause the scroll position to be unstable.
+
+## `contain-intrinsic-size`
+
+The `contain-intrinsic-size` property, as well as its longhand and logical variants, can be used to help mitigate these issues.
+
+- `contain-intrinsic-size` is a shorthand for `contain-intrinsic-width` and `contain-intrinsic-height`.
+- `contain-intrinsic-block-size` and `contain-intrinsic-inline-size` are logical variants of the longhand properties.
+
+The value syntax for the longhand properties is as follows:
+
+- `none` means the element has no intrinsic size for the respective dimension.
+- A `<length>` value can be provided to set the intrinsic size.
+- `auto <length>` can be provided to set the intrinsic size when it is unknown, otherwise, it will use the last known size of the contents while skipped for rendering.
+
+The default value is `auto none` which means initially there will be no intrinsic size for the contents, but it will use the last known size once that becomes available.
+
+### Determining an intrinsic size
+
+While in some scenarios the size of the content can vary significantly, in others where the size of the content is consistent, it can become possible to provide values that are accurate even across browsers or at different zoom levels. CSS provides numerous units that can be useful for this, such as the `lh` unit or `cap` units.
+
+For example, if we know that the contents will only be 2 lines of text, then we can use a value like `2lh`:
+
+```css
+contain-intrinsic-block-size: auto 2lh;
+```
+
+For scenarios that use different layouts, we can add the sizes of the elements as well as any spacing between them.
+
+When content size varies, one can use an average value or make a guess at what the could be. This won’t totally prevent a scrollbar from growing or shrinking, but it can reduce the negative effect.
+
+## Fallback strategies
+
+`content-visibility` can be applied as a progressive enhancement, so there is no fallback strategy necessary. Consequently, this means there is no fallback necessary for `contain-intrinsic-size` as the element will not have size containment applied to them.
--- a/guides/performance/optimize-scroll-stability/guide.md
+++ b/guides/performance/optimize-scroll-stability/guide.md
@@ -40,11 +40,8 @@ contain-intrinsic-block-size: auto 2lh;
 For scenarios that use different layouts, we can add the sizes of the elements as well as any spacing between them.
 
 ```css
-/* Label block-size: font-size multiplied by its line-height */
---label--block-size: 1rem * 1.5;
-/* Description block-size: font-size multiplied by its line-height */
---description-block-size: 0.8rem * 1.2;
-/* Gap between the label and description */
+--label--block-size: 1lh;
+--description-block-size: 0.8lh;
 --gap: 8px;
 /* Intrinsic size adds all the values together */
 contain-intrinsic-block-size: calc(
--- a/guides/performance/optimize-scroll-stability/guide.md
+++ b/guides/performance/optimize-scroll-stability/guide.md
@@ -42,7 +42,7 @@ contain-intrinsic-block-size: auto 2lh;
 For scenarios that use different layouts, we can add the sizes of the elements as well as any spacing between them.
 
 ```css
---label--block-size: 1lh;
+--label-block-size: 1lh;
 --description-block-size: 0.8lh;
 --gap: 8px;
 /* Intrinsic size adds all the values together */
--- a/guides/performance/optimize-scroll-stability/guide.md
+++ b/guides/performance/optimize-scroll-stability/guide.md
@@ -4,11 +4,6 @@ description: Keep the page layout, scroll bar, and scroll position as stable as 
 web-feature-ids:
   - contain-intrinsic-size
   - content-visibility
-sources:
-  - https://web.dev/articles/content-visibility#specify_the_natural_size_of_an_element_with_contain-intrinsic-size
-  - https://12daysofweb.dev/2024/css-content-visibility/#pairing-with-contain-intrinsic-size
-  - https://developer.mozilla.org/docs/Web/CSS/Reference/Properties/contain-intrinsic-size
-  - https://developer.mozilla.org/docs/Web/CSS/Guides/Containment/Using#size_containment
 ---
 `content-visibility: auto` is a useful tool for improving rendering performance as irrelevant content can be skipped for rendering.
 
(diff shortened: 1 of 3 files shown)


knowler:
@malchata do you think the guide should include the following which could affect scroll stability? 

1. Should it cover reflow use cases (e.g. items in a wrapping flex layout). The current guide focuses on height/block-size for intrinsic size. For wrapping layout, an item’s width can affect when subsequent content wraps which in turn can also affect scroll. This can be harder to guess a good intrinsic size for.
2. Should it over cover advanced [use cases like Alex Russell’s](https://infrequently.org/2020/12/resize-resilient-deferred-rendering/) which leverages ResizeObserver and IntersectionObserver to set the exact intrinsic size for elements inline.

I guess you could think of the first as problems that arise in two-dimensional layouts. The technique Alex demonstrates is like “what to do when you need the exact intrinsic size, but can’t infer that from the content,” which could be relevant for both one and two dimension situations.

malchata:
I would consider an advanced use case as a "nice to have", and the more I think about it, it might be something to land after I/O. `ResizeObserver` is tricky in part because it runs its callback at a time where it could delay the next frame from being painted. So let's hold off on the advanced stuff for now.

rviscomi:
>Should it cover reflow use cases (e.g. items in a wrapping flex layout). The current guide focuses on height/block-size for intrinsic size. For wrapping layout, an item’s width can affect when subsequent content wraps which in turn can also affect scroll. This can be harder to guess a good intrinsic size for.

Is there a workaround for this, or is the guidance effectively "make an educated guess and good luck"? Asking as someone who's given up on a `content-visibility` optimization before for this exact reason, so I would actually be very eager to see this covered if there's anything actionable.

I would almost rather hold off on publishing this until we have more time to smooth over these advanced use cases and rough edges with more guidance. IMO it's fine to tackle that after I/O rather than rush out a more basic version of this guide, which might trip some developers up.

WDYT @knowler @malchata?

rviscomi (commented):
This is missing an expectations file

knowler:
@rviscomi Unless the exact size of the contents is known ahead of time, it’s very difficult to avoid the reflow issues for two-dimensional layouts, since even for good guesses, an error has more significant effects that don’t get smoothed over as well compared to one-dimensional layouts. With that said, if you can leverage more information about the content ahead of time, like a character count, then you could improve your guesses significantly. This is a some what contrived example (since I’m using the same character to fill the items): https://knowler.dev/demos/hO9UCFj?codepen

The more advanced techniques of caching the sizes after an initial layout seem a bit too tricky to include as instructions for an agent, as there could be a lot of contextual variables necessary to factor into that. There’d also be some performance tradeoffs as well.

LeaVerou:
(Apologies if you got an email with a half-written version of this, I accidentally hit the keyboard shortcut and posted too soon)

My high level concern is that "optimize scroll stability" is in the same vein as "improve text legibility", i.e. not a concrete use case that an agent is likely to search for. It's good guidance, but it's more useful referenced from elsewhere (i.e. like the new guide-skills) instead of expecting it to come up in a use case search. 

But the description already kind of hints at the actual use cases:

>  Keep the page layout, scroll bar, and scroll position as stable as possible as users scroll through content-heavy pages (feeds, dashboards, long lists) where elements transition from hidden to visible.

It just needs flipping: When we're implementing content-heavy pages (feeds, dashboards, long lists), especially lazy loaded, keeping the page layout and scrollbar stable is one of the things that improve user experience. Are there other things? Can we make a guide that's focused on the concerns that one needs to have in mind when implementing these kinds of long lazy loaded feeds? Scroll stability is only one factor. What are the others?

I would suggest:
- Move the guidance that's the same for all uses of `content-visibility` and `contain-intrinsic-size` to a `features/` guide and include it here. E.g. the advice about determining good sizes for this is good, and I can't imagine how other uses of `contain-intrinsic-size` wouldn't benefit from it (but I haven't yet looked at the 6-7 guides we have on these features)
- Focus this around use cases rather than scroll stability and include only the delta that's different from other uses of these features.

From a quick look, we also repeat guidance about how to pick a good size in [performance/efficient-background-processing](https://github.com/GoogleChrome/guidance/blob/main/guides/performance/efficient-background-processing/guide.md) and in [performance/defer-rendering-heavy-content](https://github.com/GoogleChrome/guidance/blob/main/guides/performance/defer-rendering-heavy-content/guide.md):

> - 'auto' is optional and enables the browser to remember the actual size
>   once rendered. It must be paired with a <length> value to be used for
>   the first render.
> - 'none' tells the browser not to apply any intrinsic width to this element.
>   It can be used for either the height or the width value.
> - '500px' is the estimated height of this element. This can be any valid
>   CSS <length> value. Replace it with the expected height of your
>   component.

Whereas in other places where `contain-intrinsic-size` is mentioned, it would benefit from this guidance, but it isn't there. E.g. the [CSS guide](https://github.com/GoogleChrome/guidance/blob/main/guides/css/css/guide.md#performance):

> - Always pair `content-visibility` with `contain-intrinsic-size` to prevent scrollbar jumps (CLS).
> 
> #### Code Example: Render Optimization
> 
> ```css
> .large-section {
>   content-visibility: auto;
>   contain-intrinsic-size: auto 800px;
> }
> 
> .popover-reveal {
>   /* Allow discrete animations for display transitions */
>   transition: display 0.2s allow-discrete;
> }
> ```

Demo: It's very hard to see what the demo is demonstrating. Commenting `content-visibility` and `contain-intrinsic-size` out seems to produce no difference. Also note that [scroll anchoring](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Scroll_anchoring/Overview) makes the baseline behavior better (though that only affects scroll position, not size). I think it's because all 1000 items are rendered synchronously in one for loop, including their contents.

Last, there's a section at the start with reference-style docs — these shouldn't be needed, this hasn't changed recently and is not super brand new.

_(@rviscomi one caveat about this review is that that like I said, perf-related things are **not** my forte, so I can review what's here, but I'm not the right person to tell you what's missing, if anything!)_

guides/performance/optimize-scroll-stability/guide.md
```diff
@@ -3,4 +3,59 @@ name: optimize-scroll-stability
...
+- `contain-intrinsic-size` is a shorthand for `contain-intrinsic-width` and `contain-intrinsic-height`.
+- `contain-intrinsic-block-size` and `contain-intrinsic-inline-size` are logical variants of the longhand properties.
+
+The value syntax for the longhand properties is as follows:
+
+- `none` means the element has no intrinsic size for the respective dimension.
+- A `<length>` value can be provided to set the intrinsic size.
+- `auto <length>` can be provided to set the intrinsic size when it is unknown, otherwise, it will use the last known size of the contents while skipped for rendering.
+
+The default value is `auto none` which means initially there will be no intrinsic size for the contents, but it will use the last known size once that becomes available.
```
LeaVerou, line 25: This reads like reference documentation for `contain-intrinsic-size`, which the model shouldn't need (the property is Baseline Widely Available, so its syntax is well within its training data)

knowler:
@LeaVerou Thank you for the review! I agree that including this guidance along with `content-visibility: auto` is a good idea — as I was just writing the expectations that became apparent. Also, TIL about scroll anchoring!

@rviscomi Does it seem like a good idea to defer this guide until after I/O? Perhaps before then, there could be some minor additions to the CSS skills that suggest almost always pairing `content-visibility: auto` with `contain-intrinsic-size`, as Lea suggested?

LeaVerou:
> @rviscomi Does it seem like a good idea to defer this guide until after I/O? Perhaps before then, there could be some minor additions to the CSS skills that suggest almost always pairing `content-visibility: auto` with `contain-intrinsic-size`, as Lea suggested?

We already have this guidance — that it was a direct quote from the CSS skill! Though it's missing the guidance about the CSS units that produce better estimates, which is great!

rviscomi:
>Does it seem like a good idea to defer this guide until after I/O?

Yeah let's pause this for now and pick it back up after I/O. I'll move it to draft.

rviscomi:
@knowler let's pick this back up. Where did we leave off and what are the next steps?

knowler:
@rviscomi Sorry for the delay.

I agree with @LeaVerou that flipping the framing and focusing on a more concrete use case will help — especially for the scope of this guide, which I think is what was blocking us before (i.e. needing to add the more advanced stuff: see my second point in this comment https://github.com/GoogleChrome/modern-web-guidance-src/pull/633#issuecomment-4346531922). If we all agree on that, I can make those changes.

For guidance around using `content-visibility` for non-lists of things that cause scroll instability, we should come up with more concrete use cases for those.

- - -

Since we put the `contain-instrinsic-size` guidance in the CSS guide, do we still need that here? Will it pick up that guide if we mention `contain-instrinsic-size` or `content-visibility`?

rviscomi:
Ok thanks @knowler. So IIUC there really isn't a concrete use case here. The obvious one is already covered by `defer-rendering-heavy-content`, but there are some guidance gaps that are only in this PR, namely the "Determining an intrinsic size" section. I like @LeaVerou's suggestion to move that into its own `features/contain-intrinsic-size.md` file that can be transcluded across the relevant guides. We can do that in this PR or close it and open a new one.

knowler:
@rviscomi ya that would work. We did already move the `contain-intrinsic-size` guidance into the CSS guide in #831. So I guess we'd be splitting that out into that separate feature guide you suggested?

I am curious if models would overly favour virtualization solutions instead of `content-visibility: auto` for the long list of deferred contents use cases? Not that they're meant to be exclusive — cv:auto can work in tandem with a virtualization solution (leveraging the `contentvisibilityautostatechange` event). Maybe the other cv:auto guides will suffice.

rviscomi:
> We did already move the `contain-intrinsic-size` guidance into the CSS guide in #831. So I guess we'd be splitting that out into that separate feature guide you suggested?

I'll defer to @LeaVerou, but IMO we could keep what we have and add a `GUIDE_REF` to the `defer-rendering-heavy-content` guide, which would transclude the "Determining an intrinsic size" content from the feature guide.

guides/performance/optimize-scroll-stability/guide.md
```diff
@@ -3,4 +3,54 @@ name: optimize-scroll-stability
 description: Keep the page layout, scroll bar, and scroll position as stable as possible as users scroll through content-heavy pages (feeds, dashboards, long lists) where elements transition from hidden to visible.
 web-feature-ids:
   - contain-intrinsic-size
+  - content-visibility
 ---
+`content-visibility: auto` is a useful tool for improving rendering performance as irrelevant content can be skipped for rendering.
```
malchata, line 8: I think this would be more accurate:

```suggestion
`content-visibility: auto` useful for improving rendering performance as rendering of off-screen content can be deferred.
```

guides/performance/optimize-scroll-stability/guide.md
```diff
@@ -3,4 +3,54 @@ name: optimize-scroll-stability
...
+
+Since `content-visibility: auto` applies size containment to an element when its contents are being skipped, the dimensions of the element will be as if it did not have any content: only its padding and border size will contribute. This can have undesirable effects on an element’s surrounding context: when an element becomes relevant it can cause reflow, within a scrollport it can cause the scrollbar to grow or even jitter, or it can cause the scroll position to be unstable.
```
malchata, line 10: Wouldn't the reflow possibly cause even more rendering work in the long run? If so, is that a downside worth mentioning?

PR #634, opened 2026-04-23
Adds additional test-results patterns to .gitignore.

malchata:
I noticed a couple of `test-results` folders that were not being ignored. Not sure if this is by design, but in case it's not, this PR adds two additional `test-results` folder patterns to `.gitignore`.


micahjo7 (approved):

malchata:
I do believe some others aren't being captured, but I'll merge this one and add others as I come across them.

PR #635, opened 2026-04-23
Adds detect initial visibility state guide

malchata:
Progress on https://github.com/GoogleChrome/guidance/issues/366

Adds the guide and other artifacts for the detect-initial-visibility use case.


guides/performance/detect-initial-visibility-state/guide.md
```diff
@@ -2,7 +2,95 @@
...
+
+MANDATORY: Use `performance.getEntriesByType('visibility-state')` to access the exact visibility history. Do not rely solely on checking `document.visibilityState` at execution time, as it is susceptible to race conditions.
+
+```javascript
+/**
+ * Accurately determines visibility state history using the Performance API.
+ */
+function getVisibilityInfo() {
+  // Feature detect the VisibilityStateEntry API on the window object.
+  if ('VisibilityStateEntry' in window && 'getEntriesByType' in performance) {
```
rviscomi: Could you update this so that the body of the guide shows the modern approach (no fallbacks or feature detection) while the fallback section shows how to do the feature detection and what to fall back to?

Also `getEntriesByType` has been around for 10+ years so probably safe to skip the feature detection.

malchata: I updated the guide to implement this feedback. Give 'er another pass and let me know if it works!

guides/performance/detect-initial-visibility-state/guide.md
```diff
@@ -2,7 +2,107 @@
...
+{{ BASELINE_STATUS("page-visibility-state") }}
+
+The `VisibilityStateEntry` API is not universally supported in older browser versions. For unsupported environments, you may fall back to checking the `document.visibilityState` property and listening for the `visibilitychange` event.
+
+**MANDATORY:** You must understand that this fallback approach is often **highly inaccurate for determining initial background state**. Because scripts can load and execute asynchronously, a page could be opened in a background tab and then foregrounded by the user *before* your script has finished downloading and executing. When your script finally runs, `document.visibilityState` will synchronously read as `'visible'`, and you will incorrectly assume the page was loaded in the foreground, completely missing its initial hidden state. Furthermore, the fallback timestamp lacks the internal precision of the Performance API. If precision is a high priority, do not use the fallback.
+
+```javascript
+/**
+ * Fallback implementation using document.visibilityState.
+ * <!-- This approach is prone to race conditions if the script loads asynchronously. -->
```
rviscomi: It's strange to see HTML comment in a JS comment. Could you fix?

(this happens in a couple of other places)

guides/performance/detect-initial-visibility-state/guide.md
```diff
@@ -2,7 +2,107 @@
...
+    };
+  }
+}
+```
+
+### Fallback strategies
+
+{{ BASELINE_STATUS("page-visibility-state") }}
+
+The `VisibilityStateEntry` API is not universally supported in older browser versions. For unsupported environments, you may fall back to checking the `document.visibilityState` property and listening for the `visibilitychange` event.
```
rviscomi: A couple of suggestions:

- remove the browser support note, since that's handled by the `BASELINE_STATUS` macro and kept up to date as support changes
- clarify `visibilityState` and `visibilitychange` as "A or B" not "A and B"

I assume the second suggestion is correct because the demo doesn't listen to the `visibilitychange` event but it still works

```suggestion
For unsupported environments, you may fall back to checking the `document.visibilityState` property or listening for the `visibilitychange` event.
```

guides/performance/detect-initial-visibility-state/guide.md
malchata: Weird thing for the agent to do 😅  fixed in commit ddadffef.

rviscomi (approved):
LGTM!

Issue #326, opened 2026-03-17
HTML-in-canvas


ThomasTheDane:
HTML allows you to render HTML / DOM content directly into a canvas through context2D, WebGL, or WebGPU, while also maintaining interactivity and browser hookups like accessibility, translate, and more! 

Explainer: 
https://github.com/WICG/html-in-canvas

Chrome Documentation: 
[notes / beyond-early draft state] https://docs.google.com/document/d/1LfLHO6_NEJf9rTenh3ZxJHQ8ECG06ENUobY9iKP9wEk/edit?resourcekey=0-h2CCgTQ7NESyQ7ewSk3Neg&tab=t.baxhkb3xjrnr#heading=h.fn7guf4ujsfe 

Feature ID: canvas-html

**Use cases:**
- [x] #659
- [x] #660
- [x] #661
- [x] #662

paulirish:
I keep thinking about this feature cuz is 🔥 so hot 🔥 right now.. and having coverage at launch would be pretty amaze.

PR #637, opened 2026-04-23
Add HTML-in-Canvas use cases

ewewraw:
Progress on #326 

Add HTML-in-Canvas use cases


guides/user-experience/visually-texture-content/guide.md
```diff
@@ -3,4 +3,5 @@ name: visually-texture-content
 description: Apply realistic weathering and texture patterns to elements to give them an organic, aged, or physical material appearance.
 web-feature-ids:
   - masks
+  - html-in-canvas
```
rviscomi: Could you show how this would affect the demo?

guides/accessibility/accessible-canvas-content/guide.md
rviscomi, line 1: I wonder if this use case is actually the same as `expose-canvas-content-to-browser-features`, as the technical solutions appear to be identical. What's your feeling on whether they should be kept separate or merged?

rviscomi (commented):
Just a couple of comments. Could you also point me to any docs or articles on how to test this in Chrome?

ewewraw:
Thanks @rviscomi ,
You can [test](https://github.com/WICG/html-in-canvas#developer-trial-dev-trial-information) it in Canary with the `chrome://flags/#canvas-draw-element` flag enabled.

I've merged the accessibility use case into more general browser features support use case. I've also removed html-in-canvas from the visually-texture-content use case, since it's too specific, and the webgl-shaders use case is a broader one for this.

rviscomi (approved):
LGTM

Thanks @ewewraw! Tested all demos in Canary with the flag enabled and they work as expected.

PR #638, opened 2026-04-23
feat: Surface token usage in dashboard runs

natorion:
This surfaces token consumed to the dashboard, if available in the data. 

A red coin is displayed in the bar chart, if the guided run is more expensive to spot where we have to optimize.

### DIFF
--- a/harness/lib/collection.ts
+++ b/harness/lib/collection.ts
@@ -278,12 +278,12 @@ export async function collectResults(resultsDir: string, suiteConfig: SuiteConfi
           const content = fs.readFileSync(filePath, 'utf8');
           const session = JSON.parse(content);
           if (session.messages) {
-            for (const msg of session.messages) {
-              if (msg.tokens) {
-                totalTokens += msg.tokens.total || 0;
-                cachedTokens += msg.tokens.cached || 0;
-                hasTokenData = true;
-              }
+            const messagesWithTokens = session.messages.filter((m: any) => m.tokens);
+            const lastMsg = messagesWithTokens[messagesWithTokens.length - 1];
+            if (lastMsg) {
+              totalTokens += lastMsg.tokens.total || 0;
+              cachedTokens += lastMsg.tokens.cached || 0;
+              hasTokenData = true;
             }
           }
         }
--- a/harness/lib/collection.ts
+++ b/harness/lib/collection.ts
@@ -264,6 +264,53 @@ export async function collectResults(resultsDir: string, suiteConfig: SuiteConfi
       const testName = `${taskName} - ${guide} - ${runType}`;
       const actualBaseApp = taskInfo.baseApp;
 
+      let totalTokens = 0;
+      let cachedTokens = 0;
+      let hasTokenData = false;
+
+      try {
+        const files = fs.readdirSync(dir);
+        
+        // Gemini sessions
+        const geminiSessions = files.filter(f => f.startsWith('session-') && f.endsWith('.json'));
+        for (const file of geminiSessions) {
+          const filePath = path.join(dir, file);
+          const content = fs.readFileSync(filePath, 'utf8');
+          const session = JSON.parse(content);
+          if (session.messages) {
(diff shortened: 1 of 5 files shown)


harness/lib/collection.ts
```diff
@@ -264,6 +264,53 @@ export async function collectResults(resultsDir: string, suiteConfig: SuiteConfi
...
+          const filePath = path.join(dir, file);
+          const content = fs.readFileSync(filePath, 'utf8');
+          const session = JSON.parse(content);
+          if (session.messages) {
+            for (const msg of session.messages) {
+              if (msg.tokens) {
+                totalTokens += msg.tokens.total || 0;
+                cachedTokens += msg.tokens.cached || 0;
+                hasTokenData = true;
+              }
```
paulirish: the msg level token stats are actually cumulative already so..  you end up double counting here.

see https://docs.cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/GenerateContentResponse#UsageMetadata

this code (from another project) should help:

```js
    messages.forEach(item => {
        const m = /** @type {{type: string, tokens?: {input: number, output: number, cached?: number}, model?: string}} */ (item);
        if (m.type === 'gemini' && m.tokens) {
            const modelTier = getModelTier(m.model || '');
            if (!statsByTier[modelTier]) {
                statsByTier[modelTier] = { calls: 0, input: 0, cached: 0, output: 0, cost: 0 };
            }
            const s = statsByTier[modelTier];
            s.calls++;

            const input = m.tokens.input || 0;
            const output = m.tokens.output || 0;
            const cached = m.tokens.cached || 0;
            const uncached = input - cached;

            s.input += uncached;
            s.cached += cached;
            s.output += output;
        }
    });```

File: harness/lib/collection.ts [hunk 2]
```diff
@@ -264,6 +264,53 @@ export async function collectResults(resultsDir: string, suiteConfig: SuiteConfi
...
+        const codexSessions = files.filter(f => f.startsWith('session-') && f.endsWith('.jsonl'));
+        for (const file of codexSessions) {
+          const filePath = path.join(dir, file);
+          const content = fs.readFileSync(filePath, 'utf8');
+          const lines = content.split('\n');
+          for (const line of lines) {
+            if (!line.trim()) continue;
+            try {
+              const obj = JSON.parse(line);
+              if (obj.type === 'event_msg' && obj.payload && obj.payload.type === 'token_count') {
```
paulirish: https://developers.openai.com/api/reference/resources/responses#(resource)%20responses%20%3E%20(model)%20response_usage%20%3E%20(schema) i thinkkkkkkkkk this is the relevant API response type for this.

havent spent much time with tokens and codex.

paulirish (commented):
This is awesome. Thanks so much for the contribution

harness/lib/collection.ts
```diff
@@ -264,6 +264,53 @@ export async function collectResults(resultsDir: string, suiteConfig: SuiteConfi
...
+            const lastMsg = messagesWithTokens[messagesWithTokens.length - 1];
+            if (lastMsg) {
+              totalTokens += lastMsg.tokens.total || 0;
+              cachedTokens += lastMsg.tokens.cached || 0;
+              hasTokenData = true;
+            }
+          }
+        }
+
+        // Codex sessions
```
micahjo7: this should be the same filename structure as Claude (session-... .jsonl).

were you able to test this with Claude Code too?

harness/lib/collection.ts
natorion: Thanks, addressed

harness/lib/collection.ts
natorion: Thanks, done. I will double-check

harness/lib/collection.ts
natorion: yes, tested with codex and claude files we have in the repo. format is a bit different.

natorion:
PTAL again

micahjo7 (approved):

Issue #211, opened 2026-03-04
Create guide and evals for the interactions-in-complex-layouts use case


paulirish:
Make interactions snappier and more responsive (reducing Interaction to Next Paint (INP) scores) by avoiding layout re-calculations in complex layouts, such as data-heavy dashboards or spreadsheet-style grids.

Affected web-feature IDs: [content-visibility](https://webstatus.dev/features/content-visibility)

Use case subdir: [guides/performance/interactions-in-complex-layouts](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/performance/interactions-in-complex-layouts)

Related features: #49

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

Issue #210, opened 2026-03-04
Create guide and evals for the faster-spa-view-transitions use case


paulirish:
Enable faster transitions back to previously visited views in a Single-Page Application (SPA) by preserving their structural DOM state instead of destroying and rebuilding them on every navigation.

Affected web-feature IDs: [content-visibility](https://webstatus.dev/features/content-visibility)

Use case subdir: [guides/performance/faster-spa-view-transitions](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/performance/faster-spa-view-transitions)

Related features: #49

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [ ] negative-demo.html
- [x] grader.ts

Issue #209, opened 2026-03-04
Create guide and evals for the efficient-background-processing use case


paulirish:
Conserve system resources and battery life by pausing background JavaScript execution (such as `<canvas>` animations, WebGL rendering, or high-frequency WebSocket data polling) when the component is off-screen and then resume them just-in-time when they scroll back into view.

Affected web-feature IDs: [content-visibility](https://webstatus.dev/features/content-visibility)

Use case subdir: [guides/performance/efficient-background-processing](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/performance/efficient-background-processing)

Related features: #49

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

Issue #208, opened 2026-03-04
Create guide and evals for the defer-rendering-heavy-content use case


paulirish:
Reduce rendering times in content-heavy web pages (e.g. pages with long feeds, lots of articles, or complex dashboards), by deferring rendering for any content that is not immediately visible to the user.

Affected web-feature IDs: [content-visibility](https://webstatus.dev/features/content-visibility)

Use case subdir: [guides/performance/defer-rendering-heavy-content](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/performance/defer-rendering-heavy-content)

Related features: #49

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

PR #643, opened 2026-04-24
Guides for `content-visibility`

taraojo:
Note: There is also #208 which is another `content-visibility` guide already completed in conjunction with `hidden=until-found`.

**`efficient-background-processing` #209** 
Includes updates to the description to help selection with more vague dev prompts. `demo.html` has been updated to include the recommended fallback in the guide. Also extra information on when to use the event listener vs the Intersection Observer API to ensure the agent uses this approach for the right reasons.

**`faster-spa-view-transitions` #210** 
Includes updates to the description to match updates in the `guide.md`. `demo.html` CSS fallback has been refactored for clarity

**`interactions-in-complex-layouts` #211** 
`demo.html` has been updated to include fallbacks. The `guide.md` has an additional step for managing interactions. Perhaps we don't need this?


guides/performance/interactions-in-complex-layouts/guide.md
```diff
@@ -3,4 +3,86 @@ name: interactions-in-complex-layouts
...
+}
+```
+
+### Fallback strategies
+
+{{ BASELINE_STATUS("content-visibility") }}
+
+The property degrades gracefully. In unsupported browsers:
+*   The property is ignored, and mutations will cause the standard global reflow.
+*   To achieve a similar isolation effect in older browsers, you can fall back to applying strict containment manually:
```
knowler: I wonder if we should avoid “strict” here to avoid conflation with the `strict` keyword which also includes size containment.

guides/performance/efficient-background-processing/guide.md
knowler, line 1: The demo avoids this, but there is an interop issue for the `contentvisibilityautostatechange` event where it doesn’t bubble, so either:

- The event listener should be added to the element with `content-visibility: auto` applied to it (the demo is doing this).
- A capturing event listener should be used. Necessary in situations where event delegation is happening.

It could be helpful to note this somewhere and reference the following issues:

- Spec issue: https://github.com/w3c/csswg-drafts/issues/11310
- Chrome issue: https://issues.chromium.org/issues/365168180

guides/performance/faster-spa-view-transitions/guide.md
```diff
@@ -1,6 +1,92 @@
...
+### 1. Configure View Containers
+
+Set `content-visibility: hidden` on views that are not currently active.
+
+```css
+.spa-view {
+  /* Default state for active views */
+  content-visibility: visible;
+}
+
```
knowler: `content-visibility: visible` is the default, so I don’t think we need to be explicit about it. It could be worth noting somewhere that it’s default though.

```suggestion
```

guides/performance/faster-spa-view-transitions/guide.md
```diff
@@ -1,6 +1,92 @@
...
+.spa-view.inactive {
+  /* MANDATORY: Use hidden to cache the rendering state of inactive views */
+  content-visibility: hidden;
+  
+  /* Optional: Prevent hidden views from taking up physical space in layout flow */
+  position: absolute;
```
knowler, line 49: The element with the hidden content itself could be styled (i.e. its background, border, padding would be visible), so I wonder if it could be worth noting something about that?

guides/performance/faster-spa-view-transitions/guide.md
```diff
@@ -1,6 +1,92 @@
...
+
+# Faster SPA View Transitions via State Caching
+
+Enable instant navigation between views in a Single-Page Application (SPA) by caching the rendered state of inactive views instead of destroying them.
+
+## Overview
+
+Traditionally, when a user navigates between tabs or views in an SPA, developers either destroy the old view or hide it using `display: none`. Both approaches require the browser to recreate or recalculate the full layout and paint when the user returns to that view.
+
+By using `content-visibility: hidden` on inactive views, the browser removes the element from the layout flow and stops painting it, but *retains* its cached rendering state in memory. When the user switches back, the view restores nearly instantly.
```
knowler:
```suggestion
By using `content-visibility: hidden` on inactive views, the browser removes the element’s contents from the layout flow and stops painting it, but *retains* its cached rendering state in memory. When the user switches back, the view restores nearly instantly.
```

Updates this wording to reflect that the contents are what’s hidden/removed and not the element itself.

guides/performance/interactions-in-complex-layouts/guide.md
```diff
@@ -3,4 +3,86 @@ name: interactions-in-complex-layouts
...
+In complex layouts, performing a minor change to a single item—such as dragging a card or editing a cell—can trigger a chain reaction of style and layout calculations that forces the browser to reflow the entire page. This results in dropped frames and high Interaction to Next Paint (INP) latency.
+
+By applying `content-visibility: auto` to self-contained layout regions (like columns in a Kanban board), you can isolate rendering work.
+
+### Mechanism for On-Screen Elements
+
+It is important to understand how `content-visibility: auto` benefits elements that are **already visible on the screen**:
+
+*   For visible elements, the browser **does not** skip rendering.
+*   Instead, the performance benefit comes entirely from strict **CSS Containment** (`contain: layout style paint size`) that the property automatically enforces.
```
knowler:
```suggestion
*   Instead, the performance benefit comes entirely from the **CSS containments** that the property automatically enforces (i.e., layout, style, and paint).
```

For on-screen element whose contents aren’t skipped for rendering, size containment will not be applied. That only applies when the contents are skipped for rendering.

Some relevant bits from the [css-contain-2 spec](https://www.w3.org/TR/css-contain-2/):

> …[content-visibility: auto](https://www.w3.org/TR/css-contain-2/#propdef-content-visibility) imposes [size containment](https://www.w3.org/TR/css-contain-2/#size-containment) when the element [skips its contents](https://www.w3.org/TR/css-contain-2/#skips-its-contents)…

> Note that in the [content-visibility: auto](https://www.w3.org/TR/css-contain-2/#propdef-content-visibility) case, [layout containment](https://www.w3.org/TR/css-contain-2/#layout-containment), [style containment](https://www.w3.org/TR/css-contain-2/#style-containment), and [paint containment](https://www.w3.org/TR/css-contain-2/#paint-containment) persist even if the element is not [skipped](https://www.w3.org/TR/css-contain-2/#skips-its-contents).

knowler (commented):
It’s looking good. ~~This is a first pass of reviewing. I’ll continue tomorrow for perhaps a final pass.~~ (No need for a second pass)

guides/performance/faster-spa-view-transitions/guide.md
taraojo: I've added a note in the guide

guides/performance/efficient-background-processing/guide.md
taraojo: @knowler thanks for the review! On this one I've added a note and referenced the issues in the resources list. I also added another code example for the capturing event listener you mentioned. Let me know what you think.

Also, there seems to be a problem with your CLA causing one of the PR checks to fail. Can you take a look?

knowler: > Also, there seems to be a problem with your CLA causing one of the PR checks to fail. Can you take a look?

@taraojo, this should work for the Co-authored-by commits:

```
Co-authored-by: Nathan Knowler <nathan@oddbird.net>
```

I signed the CLA with my OddBird account which both wasn’t linked to my GitHub and isn’t my primary email on GitHub.

taraojo: I've updated those commits!

guides/performance/faster-spa-view-transitions/demo.html
rviscomi: There are two expectations around accessibility that don't seem to be implemented in the demo:

>- The implementation must use aria-hidden="true" on inactive view elements to ensure they are removed from the accessibility tree.
>- The implementation should maintain focus management by moving focus to the active view container upon transition.

(here and on the other `section`)

```suggestion
    <section id="dashboard-page" class="page-container active" tabindex="-1" aria-hidden="false">
```

rviscomi, line 323: I think this is also needed, along with the other comment

```suggestion
if (page.id === targetId) {
  page.classList.add('active');
  page.setAttribute('aria-hidden', 'false');
  page.focus();
} else {
  page.classList.remove('active');
  page.setAttribute('aria-hidden', 'true');
}
```

rviscomi (approved):
Overall LGTM. Just a couple of comments about the `faster-spa-view-transitions` demo

knowler (approved):
LGTM

rviscomi (approved):
LGTM with just a couple of minor fixes

PR #647, opened 2026-04-24
#252 position-aware-tooltips

### LINKED ISSUE #252: Create guide and evals for the position-aware-tooltips use case
Build tooltips and popovers with directional arrows (or other visual styling) that automatically point the correct way when the element flips to a fallback position.

Affected web-feature IDs: [container-anchor-position-queries](https://webstatus.dev/features/container-anchor-position-queries), [popover](https://webstatus.dev/features/popover)

Use case subdir: [guides/ui-atoms/position-aware-tooltips](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-atoms/position-aware-tooltips)

Related features: #45 #32

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

jamesnw:
Addresses #252 

This was a tricky one to keep scoped, I hope it didn't expand too much.


guides/user-experience/position-aware-tooltips/guide.md
```diff
@@ -5,5 +5,160 @@ web-feature-ids:
   - container-anchor-position-queries
```
malchata, line 5: Since this guide relies quite a bit on popover, should we add it here?

```suggestion
  - container-anchor-position-queries
  - popover
```

guides/user-experience/position-aware-tooltips/guide.md
```diff
@@ -5,5 +5,160 @@ web-feature-ids:
...
 sources:
   - https://developer.chrome.com/blog/anchored-container-queries
-  - https://github.com/mdn/content/pull/42710/changes
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Anchor_positioning/Anchored_container_queries
+  - https://oidaisdes.org/blog/anchored-container-queries/
```
malchata, line 10: If we add the `popover` feature ID, let's add a source URL here for it:

```suggestion
  - https://oidaisdes.org/blog/anchored-container-queries/
  - https://developer.mozilla.org/en-US/docs/Web/API/Popover_API
```

guides/user-experience/position-aware-tooltips/guide.md
```diff
@@ -5,5 +5,160 @@ web-feature-ids:
...
 ---
+
+# Position-aware tooltips
+
+When building tooltips or popovers with [CSS Anchor Positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Anchor_Positioning), the browser can automatically "flip" the element to a fallback position if it would otherwise overflow the viewport. When this happens, you may want to adjust the style of the positioned content, for instance to reposition an arrow that points from the positioned content to the anchor.
```
malchata: Links shouldn't be in the markdown body, per the `project-guides` skill. If this is a necessary API, include its feature ID under `web-feature-ids` and the removed URL to the `sources` key in the front matter.

```suggestion
When building tooltips or popovers with CSS Anchor Positioning, the browser can automatically "flip" the element to a fallback position if it would otherwise overflow the viewport. When this happens, you may want to adjust the style of the positioned content, for instance to reposition an arrow that points from the positioned content to the anchor.
```

guides/user-experience/position-aware-tooltips/demo.html
```diff
@@ -4,72 +4,101 @@
...
-        top: 100%;
-        left: 1rem;
+      @supports (container-type: anchored) {
+        .tooltip-content::before {
+          content: "▼";
+          position: absolute;
+          inset-block-end: 0;
+          inset-inline-start: 1rem;
+          color: black;
+          text-box: trim-both ex alphabetic;
```
malchata: I'd recommend removing this to focus on the core functionality.

```suggestion
```

guides/user-experience/position-aware-tooltips/demo.html
```diff
@@ -4,72 +4,101 @@
...
+      document.getElementById("anchor").click();
+
+      // Fallback strategy: Dynamically import the popover-polyfill for browsers that do not support the popover attribute natively.
+      if (!("popover" in HTMLElement.prototype)) {
+        import("https://unpkg.com/@oddbird/popover-polyfill@latest/dist/popover-fn.js").then(
+          ({ apply }) => {
+            apply();
+          },
+        );
+      }
```
malchata: Avoids a potential race condition on browsers not supporting popover, though there may be a better way to do it (possibly using top-level `await`).

```suggestion
      // Fallback strategy: Dynamically import the popover-polyfill for browsers that do not support the popover attribute natively.
      if (!("popover" in HTMLElement.prototype)) {
        import("https://unpkg.com/@oddbird/popover-polyfill@latest/dist/popover-fn.js").then(
          ({ apply }) => {
            apply();
            document.getElementById("anchor").click();
          },
        );
      } else {
        document.getElementById("anchor").click();
      }
```

malchata (changes requested):

jamesnw:
@malchata I made those adjustments. I also realized the guide was falling back to a nonexistent browser version where popover was not supported, but anchor positioning was. I tweaked the guide a bit to make it more realistic.

guides/user-experience/position-aware-tooltips/guide.md
```diff
@@ -3,7 +3,162 @@ name: position-aware-tooltips
...
+### 3. Style based on the fallback
+
+Use `@container anchored(fallback: <value>)` to apply styles when a specific fallback is active.
+
+Like all container queries, `@container` can only style **descendants** of the container. A common strategy to create the arrows is with the `::before` and `::after` pseudo-elements, which are treated as descendants and can be styled directly. However, to style the tooltip itself (as seen in step 4), we will add a child element to the tooltip, and create the arrow in its `::before` pseudo-element.
+
+```html
+<div id="tooltip" popover>
+  <div class="tooltip-content">Tooltip</div>
+</div>
```
malchata: Would this make sense from an accessibility standpoint?

```suggestion
<div id="tooltip" popover role="tooltip">
  <div class="tooltip-content">Tooltip</div>
</div>
```

guides/user-experience/position-aware-tooltips/guide.md
```diff
@@ -3,7 +3,162 @@ name: position-aware-tooltips
...
+
+By setting `container-type: anchored` on your positioned element, you turn it into a query container that knows about its own anchor-positioned state. You can then use the `@container anchored()` query to update its descendants or pseudo-elements.
+
+### 1. Create the tooltip and trigger
+
+Use the Popover API to create a tooltip. This creates an implicit anchor connection that can be used for positioning.
+
+```html
+<button popovertarget="tooltip" id="anchor">anchor</button>
+<div id="tooltip" popover></div>
```
malchata: Accessibility suggestion. If it doesn't matter, feel free to reject.

```suggestion
<button popovertarget="tooltip" id="anchor" aria-describedby="tooltip">anchor</button>
<div id="tooltip" popover role="tooltip"></div>
```

guides/user-experience/position-aware-tooltips/guide.md
```diff
@@ -3,7 +3,162 @@ name: position-aware-tooltips
...
+}
+```
+
+## Best practices
+
+- **Prefer logical fallbacks**: Use keywords like `flip-block` and `flip-inline` in `position-try-fallbacks` for simpler queries that handle RTL and different writing modes automatically.
+- **Use pseudo-elements for arrows**: Tooltip arrows are purely decorative and are perfect candidates for `::before` or `::after`, which can be styled via anchored container queries without extra DOM.
+
+
+## Fallbacks
```
malchata:
```suggestion
## Fallback strategies
```

guides/user-experience/position-aware-tooltips/expectations.md
```diff
@@ -0,0 +1,9 @@
+- The implementation MUST the Popover API with `popovertarget` and the `popover` attribute.
```
malchata: Caught a stray typo here.

```suggestion
- The implementation MUST use the Popover API with `popovertarget` and the `popover` attribute.
```

malchata (changes requested):

malchata:
> @malchata I made those adjustments. I also realized the guide was falling back to a nonexistent browser version where popover was not supported, but anchor positioning was. I tweaked the guide a bit to make it more realistic.

I'm not seeing the applied suggestions in this version of the review, and I don't believe my view is out of date and I did update to the branch to HEAD. I did make a couple of other comments in a re-review though.

guides/user-experience/position-aware-tooltips/guide.md
jamesnw: `role="tooltip"` is a good addition, but I think the `aria-details` attribute that is implicitly added to popover triggers (when the popover doesn't immediately follow it in the DOM) is sufficient.

jamesnw:
@malchata Thanks for the review! I've pushed the changes- hopefully you see everything now!

guides/user-experience/position-aware-tooltips/guide.md
```diff
@@ -3,7 +3,162 @@ name: position-aware-tooltips
...
+
+- **Prefer logical fallbacks**: Use keywords like `flip-block` and `flip-inline` in `position-try-fallbacks` for simpler queries that handle RTL and different writing modes automatically.
+- **Use pseudo-elements for arrows**: Tooltip arrows are purely decorative and are perfect candidates for `::before` or `::after`, which can be styled via anchored container queries without extra DOM.
+
+
+## Fallback strategies
+
+{{ BASELINE_STATUS("container-anchor-position-queries") }}
+
+Positioning the arrow based on the applied fallback is a progressive enhancement, and there is not another way of reacting to the fallback position. To hide the arrow in browsers that don't support anchor position container queries, test for CSS support with `@supports (container-type: anchored)`.
```
rviscomi, line 119: Besides the arrow, any asymmetric tooltip styles would also appear inverted.

Chrome:
Image

Firefox:
Image

I know one of the expectations emphasizes not using JS to detect the flip, but would that be the only viable way to fix the style issues in unsupported browsers? And at that point we could bring the arrow back, right?

Or is the JS fallback prohibitively complex?

jamesnw: I moved the border radius inside the supports statement to fix the first part. For fallbacks, I'm not sure we have great options here.

1. Drop anchor positioning API completely, and use Floating UI library, at which point I would recommend only implementing with Floating UI, even in modern browsers
2. Use custom `@position-try` fallbacks with a clip-path/margin as detailed [here](https://goulet.dev/posts/tooltip-with-popover-anchor-positioning/). It feels hacky, and could be tricky to implement alongside anchored containers.
3. Potentially there would be some observer I could use? I'd have to think what would be possible, but it would likely be a fairly heavy hit performance wise to work consistently with the nondeterminism of fallbacks.

#2 seems like the least bad option, but is there something else I'm missing?

guides/user-experience/position-aware-tooltips/guide.md
```diff
@@ -3,7 +3,162 @@ name: position-aware-tooltips
...
+
+```css
+@supports (container-type: anchored) {
+  #tooltip::before {
```
malchata: Should this be the same as the selector in the examples above for consistency?

```suggestion
  .tooltip-content:: {
```

malchata (approved):
Small nit and I think this is okay to go!

guides/user-experience/position-aware-tooltips/guide.md
jamesnw: Good catch, fixed!

rviscomi (approved):
LGTM

PR #650, opened 2026-04-27
Publish extension and webstore guides

sebastianbenz:
* Requires disabling checks for demo.html when no web-feature-id is
  given to pass build
* Mention Chrome Extensions in the modern-web-use-cases skill for LLMs
  to use the skill when building extensions


guides/chrome-extensions/add-context-menu/guide.md
```diff
@@ -0,0 +1,123 @@
+---
+name: add-context-menu
+description: Add custom items to the browser right-click context menu that act on selected text, links, images, or the current page.
```
rviscomi, line 3: Seeing this, I'm a little nervous about agents mistakenly picking a guide for Chrome extensions when it's not actively building one. That would really throw it off. I know that `category: chrome-extensions` will be in the search results, but I wonder if we should be even more unambiguous here in the description.

@paulirish @atopal @philipwalton any other thoughts on if or how we should compartmentalize extension guidance? For example, making it opt-in somehow.

```suggestion
description: Add custom items to the browser right-click context menu that act on selected text, links, images, or the current page. Only applicable when building Chrome extensions.
```

rviscomi:
@paulirish can you comment if these guides are set up the way you expect?

There are also several infrastructural changes worth a closer look. Some of it has to do with extension guides not having feature IDs or demo files. I assume the expectation is for us not to run evals on them?

There's another change to the modern-web skill description. I'm not sure we want to do that if this category of guides will be opt-in only.

cc @micahjo7

sebastianbenz:
Closing in favour of  https://github.com/GoogleChrome/guidance/pull/780

PR #651, opened 2026-04-27
[#255] carousel-snap-highlights

### LINKED ISSUE #255: Create guide and evals for the carousel-snap-highlights use case
Visually highlight the currently snapped non-interactive item in scroll-snapping carousels, galleries, or full-page swipe experiences. For example, expanding a card when snapped, or revealing hidden content.

Affected web-feature IDs: [container-scroll-state-queries](https://webstatus.dev/features/container-scroll-state-queries)

Use case subdir: [guides/ui-behaviors/carousel-snap-highlights](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-behaviors/carousel-snap-highlights)

Related features: #47

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

jamesnw:
I went with `overflow-x` rather than the logical equivalent, as `overflow-inline` is baseline newly available.


guides/user-experience/carousel-snap-highlights/guide.md
```diff
@@ -3,4 +3,114 @@ name: carousel-snap-highlights
 description: Visually highlight the currently snapped item in scroll-snapping carousels, galleries, or full-page swipe experiences. For example, expanding a card when snapped, or revealing hidden content.
 web-feature-ids:
   - container-scroll-state-queries
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Conditional_rules/Container_scroll-state_queries
+  - https://developer.chrome.com/blog/scroll-state-queries/
```
rviscomi:
```suggestion
  - https://developer.chrome.com/blog/css-scroll-state-queries
```

guides/user-experience/carousel-snap-highlights/guide.md
```diff
@@ -3,4 +3,114 @@ name: carousel-snap-highlights
...
+  }
+}
+```
+
+The `snapped` descriptor can query specific axes: `x`, `y`, `inline`, `block`, or `both`.
+
+
+### Accessibility
+
+**AVOID**: using `scroll-state` with interactive elements.
```
rviscomi, line 78: Do you think this is worth noting in the use case description somehow?

Maybe something like:

>Visually highlight the currently snapped **non-interactive** item...

rviscomi (approved):
LGTM just two minor suggestions

Tested in Chrome and Firefox

guides/user-experience/carousel-snap-highlights/guide.md
jamesnw: Good call. I also removed the phrase "highly interactive" from the intro description.

guides/user-experience/carousel-snap-highlights/guide.md
```diff
@@ -1,6 +1,116 @@
...
+
+```css
+/* 
+Only show the effect for users not requesting reduced motion. Disable completely, including the color change, as it causes a flash that may be problematic.
+*/
+@media (prefers-reduced-motion: no-preference) {
+  /* Style the content when its parent .carousel-item is snapped on the x axis */
+  @container scroll-state(snapped: x) {
+    .card {
+      background: #007bff;
```
malchata, line 64: Would it make sense to pull any base styles out of this `.card` class to be outside the `@container` block? Additionally, I think the `transition` property would not animate when the element is "unsnapped".

jamesnw: I don't think so- These are all the style overrides that only apply when it is snapped. 

Can you say more about `transition` not animating? In browser, it appears to work- do you see something different, or why would you expect it not to work?

rviscomi: Ah yeah I agree with Jeremy about the `transition` thing. The demo file (which works) doesn't do it in the same way it's shown in the guide:

https://github.com/GoogleChrome/guidance/blob/12a61219985542443f9e1a6ed03bc94f27b5fbc2/guides/user-experience/carousel-snap-highlights/demo.html#L51-L55

By moving the `transition` style out of the `@container` block, the card smoothly transitions from blue to gray. However, if implemented the same way as shown in the guide, the card immediately goes from blue to gray without the smooth transition.

Here's a screen recording showing both approaches, first the way shown in the guide, then the demo:

https://github.com/user-attachments/assets/66da34e8-9829-4dd5-bb84-570e3ee3343b

guides/user-experience/carousel-snap-highlights/guide.md
```diff
@@ -1,6 +1,116 @@
...
+```
+
+#### 3. Query the `snapped` state
+
+Because container queries style **descendants**, you must apply the highlight styles to an element *inside* the snap target.  Because the scroll container is set to overflow on the x axis, use the `scroll-state(snapped: x)` query.
+
+**MANDATORY**: Wrap the styles in ` @media (prefers-reduced-motion: no-preference)` to only show the effect to users who have not requested reduced motion. Depending on your use case, you may retain portions of the effect, but in this case, the cards flash from white to blue in a way that may cause problems for some users, so we disable it completely.
+
+```css
+/*
```
rviscomi, line 57:
```suggestion
.card {
  transition:
    scale 0.4s cubic-bezier(0.25, 0.8, 0.25, 1),
    background-color 0.4s,
    color 0.4s,
    box-shadow 0.4s;
}

/* 
```

guides/user-experience/carousel-snap-highlights/guide.md
```diff
@@ -1,6 +1,116 @@
...
+  @container scroll-state(snapped: x) {
+    .card {
+      background: #007bff;
+      color: white;
+      scale: 1.15;
+      transition:
+        scale 0.4s cubic-bezier(0.25, 0.8, 0.25, 1),
+        background-color 0.4s,
+        color 0.4s,
+        box-shadow 0.4s;
```
rviscomi:
```suggestion
      box-shadow: 0 10px 25px rgba(0, 123, 255, 0.3);
```

guides/user-experience/carousel-snap-highlights/guide.md
jamesnw: I see- apologies for misunderstanding. Fixed!

rviscomi (approved):

PR #653, opened 2026-04-27
Guide for `overflow-clipping-control` use case for `overflow:clip` and `overflow-clip-margin`

### LINKED ISSUE #341: Create guide and evals for the overflow-clipping-control use case
Adjust the visible clipping boundary of an element to align with the content edge, padding edge, or border edge—or a specified offset from any of these—offering finer-grained control over how content is clipped.

Affected web-feature IDs: [overflow-clip](https://webstatus.dev/features/overflow-clip), [overflow-clip-margin](https://webstatus.dev/features/overflow-clip-margin)

Use case subdir: [guides/css/overflow-clipping-control](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/css/overflow-clipping-control)

Related features: #68

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

pattishin:
**Refer to:** #341


guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,117 @@ web-feature-ids:
...
+- **DO** use `overflow: clip` when you want to ensure content is strictly contained on non-replaced elements, as it is **mandatory** for `overflow-clip-margin` to take effect.
+- **DO** remember that replaced elements default to `overflow: clip`, so declaring it explicitly is **optional** but recommended for cross-browser consistency and readability.
+- **DO** use `overflow-clip-margin` to allow small visual elements (like box shadows or decorative elements) to overflow slightly without triggering scrollbars or being cut off abruptly.
+- **DO NOT** assume that setting `overflow: visible` on replaced elements like `<img>` will always behave as it did prior to Chrome 108. It may cause unexpected layout shifts if the image aspect ratio doesn't match the container.
+- **DO NOT** use `overflow: scroll` or `overflow: auto` on the target containers when the intent is to strictly clip content without scrolling.
+- **DO** account for the **Precedence Hierarchy**: If `overflow-clip-margin` is specified along with `overflow: hidden`, the margin is ignored.
+
+## Fallback Strategy
+
+{{ BASELINE_STATUS("overflow-clip") }}
```
rviscomi, line 116: `overflow-clip` has been Baseline since 2022, but `-margin` is Limited Availability.

Looks like Chrome also only has partial support for it: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/overflow-clip-margin#browser_compatibility

Might be worth adding a note about that to avoid unexpected results

```suggestion
{{ BASELINE_STATUS("overflow-clip-margin") }}
```

rviscomi:
@jamesnw I've tagged this with the OddBird label, in case any of you are available to review it. This one seems like there might be some tricky edge cases.

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,117 @@ web-feature-ids:
...
+
+To implement a complete fallback experience:
+1. **Provide a fallback class**: Define a class that applies `overflow: hidden !important` to override the modern styles.
+2. **Create a notification banner**: Add a banner to inform the user that a fallback is active.
+3. **Feature detect with JavaScript**: Use `CSS.supports('overflow', 'clip')` to detect support and apply the fallback class and show the banner if unsupported.
+
+```css
+.box {
+  /* Base property for modern browsers */
+  overflow: clip;
```
jamesnw: Is there a reason it wouldn't work to put the older syntax first, so if the newer syntax doesn't parse, if uses the already set value?
This would remove the need for the !important and the utility class.

```suggestion
  overflow: hidden;
  overflow: clip;
```

jamesnw (commented):
There's a lot of good parts here, but I'm wondering about making this more of a focused use case rather than a demo of the various values. 

I think more compelling than preventing programmatic scroll is the ability to clip on a single axis, and allow the other axis to be visible. Could we focus the demo on that? The downside of that isn't baseline yet to use `overflow-clip-margin` in that case, so we may need a different use case for `overflow-clip-margin`. 

I think what you mention as a use case of "allowing visual effects like box shadows to bleed slightly outside a clipped container without triggering scrollbars" is a compelling option here. 

Another overall point is that this somewhat overemphasizes replaced elements- I think the fact that they default to `overflow: clip` is a good thing to mention, but I don't think it needs to be repeated.

Let me know what you think- if it makes sense to stick with this demo style, I can do a more thorough review of what's here.

rviscomi:
@pattishin is this ready for another look?

pattishin:
> @pattishin is this ready for another look?

So sorry, it's almost there - I'm double-checking some items in Safari. Will ping when ready!

rviscomi:
No worries! Just wanted to make sure.

guides/user-experience/overflow-clipping-control/guide.md
pattishin: Ty for the wait! I updated it so the default will be the fallback that mimics `overflow-clip-margin` if it isn't supported. Do let me know what you think!

pattishin:
Ready for review again!

guides/user-experience/overflow-clipping-control/expectations.md
```diff
@@ -0,0 +1,8 @@
+- The implementation MUST demonstrate multiple values for `overflow-clip-margin`, including at least one box edge (e.g., `content-box`) and one length value (e.g., `20px`).
+- The implementation MUST feature-detect support for `overflow: clip` using JavaScript (`CSS.supports`).
+- The implementation MUST use `overflow: hidden` as the default fallback style on target containers if `overflow: clip` is not supported.
+- The implementation MUST document or demonstrate simulating `overflow-clip-margin` fallback using `clip-path: inset()` combined with `overflow: visible`.
+- The implementation MUST feature-detect support for `overflow:clip` or `overflow-clip-margin` using JavaScript (`CSS.supports`).
+- The implementation MUST display a notification banner if `overflow:clip` or `overflow-clip-margin` are not supported.
```
jamesnw: I don't think the banner should be an expectation for the use case- a person wanting to clip content wouldn't want a banner on their site. I think this expectations, and the JavaScript CSS.supports expectations above, should be removed.

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,136 @@ web-feature-ids:
...
+  }
+}
+```
+
+## Strategic Implementation & Best Practices
+
+- **DO** use `overflow: clip` when you want to ensure content is strictly contained on target containers, as it is **mandatory** for `overflow-clip-margin` to take effect.
+- **DO** use `overflow-clip-margin` to allow small visual elements (like box shadows or decorative elements) to overflow slightly without triggering scrollbars or being cut off abruptly.
+- **DO NOT** use `overflow: scroll` or `overflow: auto` on target containers when the intent is to strictly clip content without scrolling.
+- **DO** account for the **Precedence Hierarchy**: If `overflow-clip-margin` is specified along with `overflow: hidden`, the margin is ignored.
```
jamesnw: Is this different than the first point? I'm also not sure what the "Precedence Hierarchy" is referring to here.

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,136 @@ web-feature-ids:
...
+{{ BASELINE_STATUS("overflow-clip-margin") }}
+
+For browsers that do not support `overflow: clip`, the standard fallback is `overflow: hidden`, which clips content to the padding box but permits programmatic scrolling. 
+
+Using `clip-path: inset()` is the most powerful way to simulate `overflow-clip-margin` for older browsers because it allows you to define exactly where the cut-off boundary is located. To make this work as a fallback, set `overflow: visible` on the container (so browsers do not preemptively clip at the padding box) and use `clip-path` to perform the clipping manually. Modern browsers can then override these styles using `@supports`.
+
+To implement a complete fallback experience:
+1. **Simulate margins via `clip-path`**: Author base rules using `clip-path: inset()` to simulate precise box edges or length margins, paired with `overflow: visible`.
+2. **Create notification banners**: Add specific banners to inform users when fallback clipping or margin simulations are active.
+3. **Feature detect with JavaScript**: Use `CSS.supports()` to verify native support, conditionally displaying notification banners and applying core fallback classes as needed.
```
jamesnw: I don't think this should be part of the guide.

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,136 @@ web-feature-ids:
...
+    clip-path: none;
+  }
+}
+```
+
+## Strategic Implementation & Best Practices
+
+- **DO** use `overflow: clip` when you want to ensure content is strictly contained on target containers, as it is **mandatory** for `overflow-clip-margin` to take effect.
+- **DO** use `overflow-clip-margin` to allow small visual elements (like box shadows or decorative elements) to overflow slightly without triggering scrollbars or being cut off abruptly.
+- **DO NOT** use `overflow: scroll` or `overflow: auto` on target containers when the intent is to strictly clip content without scrolling.
```
jamesnw: I wouldn't expect users to try these values when that is the intent, but I would expect`overflow: hidden` to be a common mistake. Should that replace `scroll` and `auto`?

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,136 @@ web-feature-ids:
...
+  position: relative;
+  background: #fff;
+  /* Level 1 Fallback: Baseline */
+  overflow: hidden;
+}
+
+@supports (overflow: clip) {
+  .box {
+    /* Level 2: Modern Baseline (Safari 16+) */
+    overflow: clip;
```
jamesnw, line 62: This doesn't need to be in a supports block. This will be hidden when clip isn't supported-

```css
.box{
  overflow: hidden;
  overflow: clip;
}
```

You do need it for the fallbacks where you need to use `overflow: visible`.

guides/user-experience/overflow-clipping-control/demo.html
```diff
@@ -1,32 +1,258 @@
...
+      .box {
+        /* Level 2: Modern Baseline (Safari 16+) */
+        overflow: clip;
+      }
+    }
+
+    /* 1. Content-box Fallback */
+    .box.clip-content {
+      overflow: visible;
+      clip-path: inset(calc(var(--box-padding) + 5px));
```
jamesnw: This is different than the main case, as it clips the border. (This also happens in the other `content-box` fallback.)

I think you could either find a variant on the solution so they match, or note it as a difference.

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,136 @@ web-feature-ids:
...
+/**
+ * Target container variant applying an extended clipping margin.
+ * Demonstrates modern support overrides paired with manual (fallback) clip-path simulation.
+ */
+.clip-20px {
+  overflow: visible;
+  clip-path: inset(-20px);
+}
+
+@supports (overflow-clip-margin: 20px) {
```
jamesnw: The current direction is that main section of the guide should not have the fallbacks- it should just have the ideal way of doing it. The guide and the fallbacks section should show the complete interoperable solution.

jamesnw (changes requested):
Thanks for the changes- I think it's heading in the right direction! I haven't done a thorough review of the demo fallbacks yet. My main concern at this point is whether agents will be able to pick up and apply the patterns when requested. The prompt seems to be "create a demo on how to use overflow-clip-margin", but I would expect people to prompt something like "I want this element to overflow on just the x axis",  or "I want the shadow to extend outside".

I think more focused demos would help, with each example mapped to the reason to use it- does that make sense?

guides/user-experience/overflow-clipping-control/guide.md
pattishin: 10-4, I can reduce this significantly then. I over-invested on how it would render in the non-supported browsers! Alrighty, stand by. :)

pattishin:
@jamesnw Ty for waiting! I focused in on replace elements for images for the demo + guide.
I did keep the `overflow: clip` within the `@supports` only because it's applied upon `overflow-margin-clip` detection.

Let me know what you think! Screenshot below. Also, more than happy to chat over video if more alignment is necessary. Ty for reviewing!

 Screenshot 2026-05-13 at 4 40 19 PM

jamesnw:
> @jamesnw Ty for waiting! I focused in on replace elements for images for the demo + guide. I did keep the `overflow: clip` within the `@supports` only because it's applied upon `overflow-margin-clip` detection.
> 
> Let me know what you think! Screenshot below. Also, more than happy to chat over video if more alignment is necessary. Ty for reviewing!

Thanks for the updates! I want to do a quick pass on the demo to get some feedback to you, but note that I'm at this point I'm just pointing out issues, and don't (yet) have proposed changes. 

Example 1-
The element isn't actually clipped, as it doesn't have a height set. If we actually clip it, I don't think this is the effect we're going for-
image
Also, you probably wouldn't want to clip an element with text in it, in most cases, to prevent text from being hidden.


Example 2-
I'm a bit unclear on what overflow: clip is doing here. The image edges are coming from the padding and the border radius. I think if overflow: clip is having an effect on subpixel rendering (which overflow: hidden would too I think?) it would only happen on the box edges, not on the circle from the border radius.

Example 3-
The `overflow-clip-margin` isn't doing what the description says here- the drop shadow on an element is outside of its clipped bounds. I think what you're wanting to show is if a parent element is `overflow: clip` and the child element has a drop shadow.

My Thursday is quite packed, otherwise I'd love to have a call on this. I think a good next step is identifying a more realistic use case, and building that. [This post](https://kilianvalkhof.com/2022/css-html/do-you-know-about-overflow-clip/) may have some good ideas?  The clip on one axis, visible on the other axis examples would work, but overflow-clip-margin doesn't work in Chrome in those cases. 

Or perhaps the use case is you have an image inside a clipped box with padding- if you set `overflow-clip-margin: content-box;` on the box, the content is clipped to inside the padding- this is set on the second box below.

image

https://codepen.io/jamessw/pen/LEbRBQr/d072c2f1692449b5328439fb675b6de9

Let me know if you have any questions about this, or other ideas about a use case!

pattishin:
@jamesnw Ah nice, those are helpful! Listing out per example:
1. I swapped this one for a use-case from Adam Argyle where it solves [nested curved components](https://nerdy.dev/perfect-nested-radius-with-overflow-clip-margin).
2. Relatively the same - I made the fix so that overflow clip wasn't on the image
3. You are right on about this case, so I made sure the drop shadow is on the button and overflow-clip-margin declaration was on its parent element to prevent premature clipping of the shadow.

Let me know what you think!

Screenshot 2026-05-14 at 8 41 06 AM

guides/user-experience/overflow-clipping-control/demo.html
```diff
@@ -1,32 +1,281 @@
...
+        <div class="nested-curve-parent">
+          <img src="test_pattern.svg" alt="Nested Curve Demo">
+          <div class="nested-curve-footer">Card Footer</div>
+        </div>
+      </div>
+    </div>
+
+    <div class="demo-item">
+      <h3>Gutter Protection</h3>
+      <p class="description">Applies directly to the parent element of an <code>&lt;img&gt;</code> with padding to create a
```
jamesnw: I'm still not entirely following the subpixel image leakage part of this example. I do think you're close to a good use case there- if you remove the border-radius, the `overflow-clip-margin: content-box` moves the clip boundary to where you'd expect-

**Without**
Image 

**With**

Image

Alternatively, you already have 2 use cases here, and I think you would be able to just use those.

guides/user-experience/overflow-clipping-control/demo.html
```diff
@@ -1,32 +1,281 @@
...
+      <h3>Nested Rounded Curves</h3>
+      <p class="description">Applies <code>overflow-clip-margin: content-box</code> to a parent container with rounded corners
+        and custom padding. Automatically applies similar rounded corners on inner child media and footer components along the
+        concentric inner content box boundary, solving awkward nesting curves without custom <code>calc()</code> logic.
+      </p>
+    
+      <div class="nested-curve-wrapper">
+        <div class="nested-curve-parent">
+          <img src="test_pattern.svg" alt="Nested Curve Demo">
+          <div class="nested-curve-footer">Card Footer</div>
```
jamesnw, line 157: ~~I don't think the Card Footer is doing anything useful here, and the demo is a bit stronger if it is removed.~~

Oh, I see what's happening. This would be more apparent if the card footer didn't have its own border-radius. 

Image

guides/user-experience/overflow-clipping-control/demo.html
```diff
@@ -1,32 +1,281 @@
...
+      width: 100%;
+      padding: 10px 16px;
+      background: #e52e71;
+      color: #fff;
+      border: none;
+      border-radius: 8px;
+      font-weight: bold;
+      cursor: pointer;
+      /* Prominent box-shadow rendered perfectly unclipped inside the visible safety zone */
+      box-shadow: 0 8px 20px rgba(229, 46, 113, 0.65);
```
jamesnw: I think the demo is slightly more clear with a less blurred box shadow, and slightly smaller padding on the `.safety-zone-parent`. It's not as aesthetic, but it's clearer what's happening.

guides/user-experience/overflow-clipping-control/demo.html
```diff
@@ -1,32 +1,281 @@
...
+    }
+
+    /* --- 3. CHILD ELEMENT SHADOW BLEED STYLES --- */
+    .safety-zone-wrapper {
+      display: flex;
+      justify-content: center;
+      align-items: center;
+      padding: 1.5rem 0;
+    }
+
```
jamesnw: I think you can remove the `safety-zone-wrapper`, and then just remove the `max-width` from the `safety-zone-parent`.

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,351 @@ web-feature-ids:
...
+  - https://web.dev/en/learn/css/overflow
   - https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/overflow-clip-margin
   - https://css-tricks.com/almanac/properties/o/overflow-clip-margin/
 ---
+
+# Overflow Clipping Control
+
+While `overflow: hidden` is a "blunt instrument" that almost always clips content strictly at the padding-box, `overflow: clip` combined with `overflow-clip-margin` provides the "scalpel" for fine-grained layout control across both standard block containers and replaced elements.
+
+By using `overflow: clip` and `overflow-clip-margin`, developers can specify exactly where clipping occurs—aligning the boundary precisely with inner box-model edges—or extend the clipping boundary beyond the element's box by a specified offset (a safety margin). This modern approach is highly performant and eliminates the legacy requirement of adding extra wrapper elements with custom padding and negative margins just to let visual effects (like drop-shadows or absolute decorative badges) bleed outside a container.
```
jamesnw:
```suggestion
Specify exactly where clipping occurs with `overflow: clip` and `overflow-clip-margin`. You can align the boundary precisely with inner box-model edges. or extend the clipping boundary beyond the element's box by a specified offset (a safety margin). This modern approach is highly performant and eliminates the legacy requirement of adding extra wrapper elements with custom padding and negative margins just to let visual effects (like drop-shadows or absolute decorative badges) bleed outside a container.
```
I found the dashes confusing here, especially when mixed with hyphenated terms.

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,351 @@ web-feature-ids:
...
+
+As of Chrome 108, `overflow: clip` and `overflow-clip-margin: content-box` are the default user-agent styles for replaced elements (`<img>`, `<video>`, `<canvas>`), making this pattern essential for cleanly containing images that use `object-fit`, `border-radius`, or CSS filters without sub-pixel leakage. On standard block containers, explicitly declaring `overflow: clip` enables high-performance containment while unlocking custom offset clip margins.
```
jamesnw:
```suggestion
Replaced elements (`<img>`, `<video>`, `<canvas>`, etc.) default to`overflow: clip` and `overflow-clip-margin: content-box`, giving you control to cleanly contain images that use `object-fit`, `border-radius`, or CSS filters without sub-pixel leakage. On standard block containers, explicitly declaring `overflow: clip` enables high-performance containment while unlocking custom offset clip margins.
```

I don't think we need to specify the Chrome 108 change.

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,351 @@ web-feature-ids:
...
+2. **Align to a Box-Edge**: Use keywords to precisely align the clipping boundary to inner box-model edges:
+   - `content-box`: Clips content exactly where the content area begins, leaving the padding area completely clean. Image or content stops right at the padding's edge. Excellent for replaced elements with padding frames.
+   - `padding-box` (Default): Clips content at the inner edge of the border.
+   - `border-box`: Clips content at the outer edge of the border, allowing content to sit under or partially overlap a translucent border.
+3. **Define a Specified Offset (The Bleed)**: Provide a length value (e.g., `15px` or `5px`) to create a safety zone before cutting pixels. This allows decorative glows, absolute badges, or shadows to stick out past the edge without expanding layout geometry.
+4. **Combine Box-Edge and Offset**: Specify both a box edge and a length offset simultaneously (e.g., `content-box 15px`) if multi-boundary targeting is required.
```
jamesnw:
```suggestion
4. **Combine Box-Edge and Offset**: Specify both a box edge and a length offset simultaneously (e.g., `content-box 15px`) to offset from a specific box edge.
```

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,351 @@ web-feature-ids:
...
+
+## Example Code
+
+The following examples demonstrate controlled clipping on standard block layout containers and replaced element frames, showcasing inner content curve nested framing, parent wrapper frame protection, and drop-shadow alongside progressive enhancement fallbacks.
+
+### Block Containers: Nested Rounded Curves
+
+Applies `overflow-clip-margin: content-box` to a parent container with rounded corners and custom padding. This also applies similar rounded corners on inner child media and footer components along the concentric inner content box boundary, solving awkward nesting curves without custom `calc()` logic.
```
jamesnw:
```suggestion
Apply `overflow-clip-margin: content-box` to a parent container with rounded corners and custom padding. This also applies similar rounded corners on inner child media and footer components along the concentric inner content box boundary, solving awkward nesting curves without custom `calc()` logic.
```

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,351 @@ web-feature-ids:
...
+</div>
+```
+
+```css
+/**
+ * Standard block layout container with outer corner radii and padding.
+ * Keeps base level 1 fallback clipping roughly at the inner padding box.
+ */
+.nested-curve-parent {
+  display: flex;
```
jamesnw: I'd recommend removing a lot of the presentation styles from the guide, and just showing the parts that are relevant to the use case. If a person requests that this is implemented, they care that there is a padding, but not a border, background, box-shadow, etc.

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,351 @@ web-feature-ids:
...
+    overflow: clip;
+    /* Automatically curves clipping edge to match inner content-box radius */
+    overflow-clip-margin: content-box;
+  }
+}
+```
+
+### Parent Wrapper Frames: Gutter Protection
+
+Applies directly to the parent element of an `<img>` with padding to create a clean gutter frame, preventing sub-pixel image leakage over the inner padding edge. Setting `overflow-clip-margin: content-box` reinforces strict inner containment across all environments.
```
jamesnw: This will need to be updated if you decide to update the demo to show the padded image use case.

Also, I'd avoid the phrase "strict inner containment across all environments" - "containment" is a separate CSS concept that doesn't apply here, and it's not "all environments" since it's Baseline Limited.

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,351 @@ web-feature-ids:
...
+  }
+}
+```
+
+## Strategic Implementation & Best Practices
+
+- **DO** apply `overflow: clip` on target elements when utilizing `overflow-clip-margin`, as setting `overflow` strictly to `clip` is **mandatory** on standard block layout containers to activate custom or inner curved clip margins.
+- **DO** set `overflow-clip-margin: content-box` on padded containers with rounded corners to automatically clip unrounded internal child elements into mathematically perfect nested border curves without manual padding subtraction logic.
+- **DO** configure `overflow-clip-margin` with a specified length offset when applying external visual effects (like `filter: drop-shadow()`) to prevent sharp bounding box truncation without altering or expanding layout geometry.
+- **DO** utilize `overflow: clip` instead of `overflow: hidden` on complex layouts or rounded media galleries to optimize rendering performance, as `clip` avoids computationally expensive scrolling stencils by simply painting pixels inside the designated clip margin.
```
jamesnw: I'm not sure what this part means- "as `clip` avoids computationally expensive scrolling stencils by simply painting pixels inside the designated clip margin." I think it can just be removed.

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,351 @@ web-feature-ids:
...
+- **DO NOT** apply `overflow: clip` if the container requires programmatic scroll manipulation via JavaScript or serves as the immediate layout context for `position: sticky` elements, as `clip` completely disables all scrollability and locks containment across both axes simultaneously.
```
jamesnw:
```suggestion
- **DO NOT** apply `overflow: clip` if the container requires programmatic scroll manipulation via JavaScript or serves as the immediate layout context for `position: sticky` elements, as `clip` completely disables scrolling.
```

Avoiding "containment" and needing to get into the less-supported `overflow-x` and `overflow-y`.

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,351 @@ web-feature-ids:
...
+
+## Fallback Strategies
+
+{{ BASELINE_STATUS("overflow-clip-margin") }}
```
jamesnw, line 117:
```suggestion
{{ BASELINE_STATUS("overflow-clip") }}
{{ BASELINE_STATUS("overflow-clip-margin") }}
```

guides/user-experience/overflow-clipping-control/guide.md
```diff
@@ -5,6 +5,351 @@ web-feature-ids:
...
+
+For target environments lacking native support for `overflow: clip` or `overflow-clip-margin`, progressive enhancement fallback strategies depend directly on the visual intent:
+- **Strict Containment**: Fallback to `overflow: hidden` as the base experience to guarantee core boundaries are maintained.
```
jamesnw: Can we find another term instead of "Strict Containment"? I think we want it to convey that "users can't see what's outside the box"

guides/user-experience/overflow-clipping-control/expectations.md
```diff
@@ -0,0 +1,8 @@
+- The implementation MUST demonstrate controlled clipping on block layout containers using `content-box` alignment to automatically apply concentric curved boundaries on inner child media and footer components.
+- The implementation MUST demonstrate controlled clipping on parent wrapper elements of images using inner box-edge alignment to prevent sub-pixel frame leakage.
+- The implementation MUST apply a box-edge keyword (e.g., `content-box`) to configure inner curved clip margins on target layout containers or padded frames.
+- The implementation MUST apply a length value (e.g., `15px`) to configure extended offset clip margins for external shadow ink bleeds.
+- The implementation MUST feature-detect runtime support for `overflow: clip` and `overflow-clip-margin` using JavaScript (`CSS.supports`).
```
jamesnw: Is CSS.supports still needed as an expectation?

jamesnw (changes requested):
Thanks for the quick updates! This is getting much closer! I think the major blocker is the second demo, which could be tweaked or removed.

Besides that, I think the demo functionally is in a good place, but could be cleaned up for clarity. I don't think the paragraphs of explanation are needed, and there's a lot of styles that are not directly related to the use case. 

Also, in the guide, reduce the code to the minimum needed to reproduce the use case- the presentational styles related to the demo (but not the user's use case) aren't useful here.

Sidenote, a useful prompt I've found after making large content shifts like this is something like "Review the 3 files, and give a high level overview of suggested improvements, especially around consistency. Don't make any changes"

guides/user-experience/overflow-clipping-control/expectations.md
pattishin: Nope, removed!

guides/user-experience/overflow-clipping-control/demo.html
pattishin: Screenshot 2026-05-14 at 12 21 27 PM

Made the shadow a little less diffused and decreased the padding!

guides/user-experience/overflow-clipping-control/demo.html
pattishin: Roger that, removed the extra `border-radius`.

guides/user-experience/overflow-clipping-control/demo.html
pattishin: Entirely removed this demo per recommendation.

pattishin:
@jamesnw ⚡ You reviewed so fast! ⚡  I believe I got to all of your recommendations and also took out most of the frills! Screenshot below.

Screenshot 2026-05-14 at 12 43 49 PM

rviscomi (approved):
LGTM

rviscomi:
@jamesnw does this one look ready to you?

jamesnw (approved):
LGTM! 🚀 

Thanks, @pattishin!

jamesnw:
> @jamesnw does this one look ready to you?

@rviscomi Sorry for missing this- this looks good to go!

PR #655, opened 2026-04-27
revise project skills based on human reviews

paulirish:
Had some agents analyze ALLLLLLLLLL the human reviewer feedback on PRs to this repo (from rick, phil, jeremy). 

synthesized it to some points and then it turned into these edits.

Also adding in the deepresearch script change which means we'll include the feature deepresearch results in the repo.


paulirish:
@rviscomi plz

.agents/skills/project-guides/SKILL.md
```diff
@@ -48,28 +48,34 @@ sources:
...
-* **Focus:** Keep it abstract but short. No fluff. No conversational text. Include a brief overview of the use case and explanation of why the solution outlined in the guide is the recommended approach.
+* **Focus:** Keep the guidance focused on the specific use case and short. No fluff. No conversational text. Include a brief overview of the use case and explanation of why the solution outlined in the guide is the recommended approach.
 * **Self-Contained:** DO NOT include any external links in the markdown body (`[link text](url)`). All required knowledge to use the feature MUST be fully synthesized into the markdown body. Agents must not be slowed down or require additional resources to implement the guidance.
+* **Strict Structure & Headings**: Use exact standard headings (like `### Fallback strategies`) exactly as specified in this project. Our downstream AI tooling and parsers depend on these exact strings to locate content.
```
rviscomi: I think we can remove this. There's a fallback strategies section below. It's also not that strict.

.agents/skills/project-guides/SKILL.md
```diff
@@ -48,28 +48,34 @@ sources:
...
 
 ### 3. Code Snippets
 * Include short, heavily commented code snippets.
-* Put directives directly in code comments so they are impossible to miss (e.g., `<!-- Always use the required attribute -->`).
+* Put directives directly in code comments using explicit keywords so they are impossible to miss (e.g., `<!-- MANDATORY: Always use the required attribute -->` or `/* DO NOT override this value */`).
```
rviscomi: We've been going overboard on the explicit keyword thing, so I'd prefer not to re-emphasize that here.

.agents/skills/project-guides/SKILL.md
```diff
@@ -48,28 +48,34 @@ sources:
...
 * Code comments MUST explain why a value or approach is chosen, not just what the code does. An agent that copies magic values without understanding them will apply them incorrectly. If a value is context-dependent (e.g., a threshold that should vary by use case), say so explicitly.
+* **Modern Standards**: Exclusively use ES modules (`import`/`export`) in JavaScript code examples; avoid CommonJS (`require`).
+* **Clarifying Arbitrary Values**: Explicitly identify placeholder values (like `2rem` or `50ms`) as example-only in comments to avoid them being mistaken for strict technical constraints.
+* **Modern API Prioritization**: When documenting performance or async patterns, prioritize modern APIs like **Speculation Rules** for prefetching, the `priority` attribute for `fetch()`, and `scheduler.yield()` for yielding to the main thread, rather than older hacks or complex wrappers.
```
rviscomi: This is too feature specific

.agents/skills/project-guides/SKILL.md
```diff
@@ -48,28 +48,34 @@ sources:
...
 
 ### 4. Implementation Steps
 * Only mark steps as `MANDATORY` if they are truly required for the feature to function. Optional steps (e.g., adding scroll snap, adding an event listener for progressive enhancement) must be labeled as optional. Incorrect use of `MANDATORY` causes agents to implement unnecessary complexity.
 * The guide is the agent's **only** source of truth. DO NOT reference `demo.html` or any other file — agents won't have access to them. Everything the agent needs to implement the use case must be in `guide.md`.
 
 ### 5. Fallback Strategies
-* You **MUST** include a `### Fallback strategies` section regardless of Baseline status, as developers may have older baseline targets.
+* You **MUST** include a `### Fallback strategies` section regardless of Baseline status. However, prioritize describing fallbacks for modern environments and avoid teaching legacy fallbacks for browsers released before 2021 to keep the guide focused and clean.
```
rviscomi: Not sure about this one. The only hard requirement is that the fallback needs to support Widely Available. We don't have any guidelines about not supporting too far back.

One change I would suggest is removing the `###` because sometimes the heading structure is different and the agent thinks it MUST be an H3.

.agents/skills/project-guides/SKILL.md
```diff
@@ -48,28 +48,34 @@ sources:
...
 * **MANDATORY**: The `{{ BASELINE_STATUS("feature-id") }}` macro must *always* be placed as the first, standalone line inside the `### Fallback strategies` section. Do not place it at the top of the document.
 * **OPTIONAL** provide an optional second argument for specific BCD keys: `{{ BASELINE_STATUS("feature-id", "bcd.key") }}`. This is useful when a critical sub-feature's status differs from the overall feature status.
 * **MANDATORY**: You MUST explicitly describe the fallback experience for unsupported browsers. Explain if the feature is a progressive enhancement (and what the base experience looks like), or show explicit code for feature detection (e.g., `CSS.supports()`, `if ('feature' in window)`) and graceful degradation techniques.
+* **Realistic Fallbacks**: Mandate that fallback recommendations be based on real-world **Baseline** support states, rather than arbitrary "legacy" assumptions.
```
rviscomi: Not sure what this means?

.agents/skills/project-use-cases/SKILL.md
```diff
@@ -33,11 +33,15 @@ A "use case" in this project is not a description of a feature; it's a task that
...
 * **Merge rather than split**: If two proposed use cases would result in guides that are 99% identical, combine them into one, more general use case. Duplicate guides bloat context windows and create confusing contradictions.
+* **Break down complex features**: Conversely, do not cram multi-step, intricate features (like passkeys) into a single generic guide. Split them into logical, detailed use cases.
+* **UX-Driven, Not Feature-Driven**: Do not simply list every method, property, or option of an API as a separate use case. A use case must represent a distinct user experience goal or a distinct developer problem, not just a variation in API usage. If the implementation across proposed use cases is 90% identical, consolidate them.
+* **Avoid Forcing Use Cases on Low-Level Utilities**: If a feature is a low-level utility (like a new Promise method or a general object cloning function) that primarily acts as a drop-in replacement for legacy patterns, avoid forcing it into multiple outcome-oriented use cases. Instead, consider recommending a single 'Fundamental Guide' (e.g., "Deep cloning complex objects") or placing it in a top-level discipline skill file.
```
rviscomi, line 43: I think it's fine to have use cases for these things as needed. The plan is to use the high-level skills to assist with discoverability.

.agents/skills/project-guides/SKILL.md
paulirish: Yeah.. clearly as this point i wish i had the provenance chain so i can point to the PR review that inspired this. I definitely have that somewhere. just... not convenient. :)

.agents/skills/project-use-cases/SKILL.md
paulirish: So.. this one comes from me. :) 

i was working on `Promise.withResolvers()` and `structuredClone()` and honestly i couldnt come up with usecases for those that were outcome-oriented but still representative of like.. 80% of general uses.

And if our usecases dont cover 80% then i feel like its not gonna do well in retrieval. 

in this case i'm thinking prompts like

> I need a better pattern to manage the promise resolution better here is there an established pattern for hooking up these existing variables to my resolve and reject callbacks?

and 

> Is there a faster way to duplicate a JavaScript object than JSON Stringify and Parse?

paulirish:
@rviscomi big thx. agree with all.. except the last one where i left thoughts.

.agents/skills/project-use-cases/SKILL.md
rviscomi: I don't feel strongly about it so it's fine if you'd prefer to keep it in

rviscomi (approved):

Issue #258, opened 2026-03-09
Create guide and evals for the scrollability-affordance-hints use case


paulirish:
Build scroll-shadow overlays, gradient fades, or directional arrow indicators that appear only when there's actually more content to scroll to in that direction.

Affected web-feature IDs: [container-scroll-state-queries](https://webstatus.dev/features/container-scroll-state-queries)

Use case subdir: [guides/ui-atoms/scrollability-affordance-hints](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-atoms/scrollability-affordance-hints)

Related features: #47

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

Issue #257, opened 2026-03-09
Create guide and evals for the scroll-position-aware-elements use case


paulirish:
Build floating buttons or widgets (back-to-top, scroll-to-bottom, chat launchers, etc.) that appear and disappear based on whether the user has scrolled at all.

Affected web-feature IDs: [container-scroll-state-queries](https://webstatus.dev/features/container-scroll-state-queries)

Use case subdir: [guides/ui-atoms/scroll-position-aware-elements](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-atoms/scroll-position-aware-elements)

Related features: #47

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

Issue #47, opened 2026-02-19
Container scroll-state queries


rviscomi:
Container scroll-state queries with the @container scroll-state(...) at-rule apply styles to an element based on the sticky positioning, snapped, and scrollable state of the container.

### MDN Documentation
- [Using container scroll-state queries](https://developer.mozilla.org/docs/Web/CSS/Guides/Conditional_rules/Container_scroll-state_queries)

### Chrome Documentation

<details>
<summary>View 8 articles</summary>

_This list was generated with AI assistance. While the URLs are real, please manually verify that they are relevant to the feature described in this issue._

- https://developer.chrome.com/en/blog/accessible-carousel
- https://developer.chrome.com/en/blog/carousels-with-css
- https://developer.chrome.com/en/blog/css-scroll-state-queries
- https://developer.chrome.com/en/blog/css-wrapped-2023
- https://developer.chrome.com/en/blog/new-in-web-ui-io-2025-recap
- https://developer.chrome.com/en/docs/devtools/accessibility/navigation
- https://web.dev/en/blog/next-gen-css-2019
- https://web.dev/en/shows/gui-challenges/jmLdZY_Lo1k
</details>

---
Feature ID: container-scroll-state-queries

**Use cases:**
- [x] #257
- [x] #258
- [x] #1196
- [x] #255
- [ ] #1228
- [ ] #1238
- [ ] #1244

PR #658, opened 2026-04-28
Add guides for container scroll state queries

taraojo:
For issue #47 

- Guide for `scrollability-affordance-hints` #258 
- Guide for `scroll-position-aware-elements` #257


rviscomi:
@jamesnw a couple more 🙏

guides/user-experience/scroll-position-aware-elements/guide.md
```diff
@@ -3,4 +3,120 @@ name: scroll-position-aware-elements
 description: Build floating buttons or widgets (back-to-top, scroll-to-bottom, chat launchers, etc.) that appear and disappear based on whether the user has scrolled at all.
 web-feature-ids:
   - container-scroll-state-queries
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Conditional_rules/Container_scroll-state_queries
+  - https://developer.chrome.com/blog/css-scroll-state-queries
+  - https://css-tricks.com/scrollytelling-on-steroids-with-scroll-state-queries/
+  - https://una.im/scroll-state-scrolled
+  - https://nerdy.dev/scroll-state-container-queries
```
jamesnw: This is just a link to his developer.chrome.com post above.
```suggestion
```

guides/user-experience/scroll-position-aware-elements/guide.md
```diff
@@ -3,4 +3,120 @@ name: scroll-position-aware-elements
...
 ---
+
+## Overview
+
+Floating buttons, like a "Back to Top" link, improve user experience by appearing only when they are useful. This guide shows how to build these elements using CSS `container-scroll-state-queries`, which allows styling elements based on the scroll position of their container without relying on JavaScript scroll listeners.
```
jamesnw: Slight tweak to what's getting improved.
```suggestion
Improve the user experience of floating buttons, like a "Back to Top" link, by showing them only when they are useful. This guide shows how to build these elements using CSS `container-scroll-state-queries`, which allows styling elements based on the scroll position of their container without relying on JavaScript scroll listeners or observers.
```

guides/user-experience/scroll-position-aware-elements/demo.html
```diff
@@ -1,106 +1,150 @@
...
-</html>
+    <!-- Floating button wrapper stuck to the bottom of the scroll container -->
+    <div class="sticky-wrapper">
+      <a href="#" class="back-to-top">↑ Top</a>
+    </div>
+  </div>
+
+  <script>
+    // Just a bit of JS to make the link functionally scroll us back to the top of our local container
+    document.querySelector('.back-to-top').addEventListener('click', (e) => {
```
jamesnw: Instead of JS, could you add an id to the h2, and have the button link to that fragment?

I think you'll want something like this on `.content` to get it to be smooth and scroll where you want-
```css
scroll-margin-top: 1em;
scroll-behavior: smooth;
```

Somewhat related, do we want to move the focus to the header, instead of leaving it on the (now invisible) button? I'm trying to think of cases where that wouldn't be desirable, but it seems like in most cases that's what I would expect.

guides/user-experience/scroll-position-aware-elements/demo.html
```diff
@@ -1,106 +1,150 @@
...
-        pointer-events: auto; /* Re-enable clicks on the button */
-
-        /* Hidden by default */
-        opacity: 0;
-        translate: 0 20px;
-        transition:
-          opacity 0.3s ease,
-          translate 0.3s ease;
+      /* Hidden by default */
+      opacity: 0;
```
jamesnw, line 67: Opacity doesn't remove it from the tab order, so it could still be focused and triggered, even while invisible- what about `display:none`, and using starting-style? This makes it a much more involved demo, but I think it's relevant because it impacts what properties are set in the container query.

Maybe an easier way would be adding and removing `aria-hidden=true`, but the CSS-only option would be nice. Thoughts?

guides/user-experience/scrollability-affordance-hints/demo.html
```diff
@@ -40,19 +37,41 @@
...
+      opacity: 1;
+    }
+
+    .scroller.can-scroll-down .shadow-bottom {
+      opacity: 1;
+    }
+
+    /* Sentinel elements styling */
+    .sentinel-top, .sentinel-bottom {
+      height: 0;
```
jamesnw:
```suggestion
      height: 1px;
```
I think this needs a height in order for the bottom one at least to be triggered.

guides/user-experience/scrollability-affordance-hints/demo.html
```diff
@@ -75,7 +97,32 @@
...
+      const bottomSentinel = document.querySelector('.sentinel-bottom');
+      const scroller = document.querySelector('.scroller');
+
+      const observer = new IntersectionObserver((entries) => {
+        entries.forEach(entry => {
+          if (entry.target === topSentinel) {
+            scroller.classList.toggle('scrolled-down', !entry.isIntersecting);
+          }
+          if (entry.target === bottomSentinel) {
+            scroller.classList.toggle('can-scroll-down', entry.isIntersecting);
```
jamesnw:
```suggestion
            scroller.classList.toggle('can-scroll-down', !entry.isIntersecting);
```
We can scroll down until the bottom sentinel is visible, I think? (Update the guide as well.)

guides/user-experience/scrollability-affordance-hints/guide.md
```diff
@@ -3,4 +3,133 @@ name: scrollability-affordance-hints
 description: Build scroll-shadow overlays, gradient fades, or directional arrow indicators that appear only when there's actually more content to scroll to in that direction.
 web-feature-ids:
   - container-scroll-state-queries
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Conditional_rules/Container_scroll-state_queries
+  - https://developer.chrome.com/blog/css-scroll-state-queries
+  - https://css-tricks.com/scrollytelling-on-steroids-with-scroll-state-queries/
+  - https://una.im/scroll-state-scrolled
+  - https://nerdy.dev/scroll-state-container-queries
```
jamesnw:
```suggestion
```
Just a link to the developer.chrome.com post

guides/user-experience/scrollability-affordance-hints/guide.md
```diff
@@ -3,4 +3,133 @@ name: scrollability-affordance-hints
...
 ---
+
+## Overview
+
+Visual hints, like shadows or gradients, help users understand that they can scroll to see more content. This guide shows how to build these hints using CSS `container-scroll-state-queries`, which allows styling elements based on the scrollable state of their container without relying on JavaScript scroll listeners.
```
jamesnw:
```suggestion
Visual hints, like shadows or gradients, help users understand that they can scroll to see more content. This guide shows how to build these hints using CSS `container-scroll-state-queries`, which allows styling elements based on the scrollable state of their container without relying on JavaScript scroll listeners or observers.
```

guides/user-experience/scroll-position-aware-elements/guide.md
```diff
@@ -3,4 +3,120 @@ name: scroll-position-aware-elements
...
+  visibility: hidden;
+}
+
+.scrolled .back-to-top {
+  opacity: 1;
+}
+```
+
+```javascript
+const sentinel = document.querySelector('.scroll-sentinel');
```
jamesnw: I think it makes sense to include the CSS.supports wrapper here for copyability.

guides/user-experience/scrollability-affordance-hints/guide.md
```diff
@@ -3,4 +3,133 @@ name: scrollability-affordance-hints
...
+  overflow-y: auto;
+  /* Establish this element as a scroll-state query container */
+  container-type: scroll-state;
+  position: relative;
+}
+```
+
+### 2. Style the Indicators
+
+Place the indicator elements (like shadows, gradients, or arrows) inside the container and style them. They should be hidden by default.
```
jamesnw:
```suggestion
Place the indicator elements (like shadows, gradients, or arrows) inside the container and style them. By default, they should not be visible. When they are shown, they should not be interactive, by setting `pointer-events: none`.
```

guides/user-experience/scroll-position-aware-elements/guide.md
```diff
@@ -3,4 +3,120 @@ name: scroll-position-aware-elements
...
+.scroller {
+  overflow-y: auto;
+  /* Establish this element as a scroll-state query container */
+  container-type: scroll-state;
+}
+```
+
+### 2. Style the Floating Element
+
+Place the element inside the container and style it. By default, it should be hidden.
```
jamesnw, line 33: It's somewhat orthogonal, but is it worth noting that the styles must be such that the button can't hide the content at the bottom of the scroll container? Thoughts?

guides/user-experience/scroll-position-aware-elements/guide.md
```diff
@@ -3,4 +3,120 @@ name: scroll-position-aware-elements
...
+```css
+.back-to-top {
+  position: sticky;
+  bottom: 20px;
+  opacity: 0;
+  transition: opacity 0.3s ease;
+}
+```
+
+### 3. Query the Scroll State
```
jamesnw:
```suggestion
### 3. Respond to the Scroll State
```

I'm unsure about this, but it might make it easier for the agent to discover?

jamesnw (changes requested):
Nice work!

guides/user-experience/scroll-position-aware-elements/guide.md
taraojo: I'm happy to update, but if I understand correctly, the guide will be discovered on just the info in the frontmatter.

guides/user-experience/scroll-position-aware-elements/demo.html
taraojo: TIL! Gemini recommended adding `tabindex="-1"` to the `<h2>` in the demo to enable the browser to move focus - does this sound like good practice to you?

guides/user-experience/scroll-position-aware-elements/demo.html
taraojo: Oh interesting I didn't spot that. The CSS-only options feels better. I'm new to `@starting-style` but I've updated the guide and demo to include it as another feature for this use case as it seems like it should be the default. Let me know if you think this is too much!

guides/user-experience/scroll-position-aware-elements/guide.md
taraojo: That makes sense, I've added a note below

guides/user-experience/scrollability-affordance-hints/demo.html
taraojo: Good spot! Updated

guides/user-experience/scroll-position-aware-elements/demo.html
jamesnw: I think this looks great!

guides/user-experience/scroll-position-aware-elements/demo.html
jamesnw: Good question- In testing your demo, I realized navigating to #title already accomplishes what we want. If there is a button in the content, and the user clicks the scroll to top link, then tabs, the button in the content is focused. Before, tabbing took the focus out of the content. So, I don't think the `tabindex="-1"` is needed. We want to move the focus to the title, but not actually focus the title. I updated this while fixing my git co-authorship.

jamesnw (approved):
LGTM!

guides/user-experience/scroll-position-aware-elements/demo.html
rviscomi, line 1: @taraojo @jamesnw the fallback behavior in FF and Safari feels a bit buggy.

- The "Top" button appears on page load, when the container is in its default position. If I attempt to scroll up, it disappears.
- Scrolling down and clicking the button takes me back to the top, but the button is still visible. (Probably the for same reason)
- The button animates in but not out

guides/user-experience/scroll-position-aware-elements/demo.html
rviscomi: I think this change caused the fade out issue I'm seeing in unsupported browsers.

Would using `visibility: hidden` fix this?

IIUC that would also remove the need for `@starting-style`.

guides/user-experience/scroll-position-aware-elements/demo.html
jamesnw: The `scroll-margin-top` needs to be equal to the scroll container's padding, so the sentinel is in view. That fixes 1 and 2, and switching from `display` to `visibility` fixes the third. @taraojo I pushed those changes to the demo since I made them while verifying- can you update the guide and expectations? (Thanks and sorry for the wrong direction)

taraojo: Thanks for fixing those! I've updated the other docs

rviscomi (approved):
LGTM

Tested in Chrome, Firefox, and Safari

Issue #545, opened 2026-04-10
Create a high-level Web AI skill


rviscomi:
General best practices for things like built-in AI and WebMCP

Related: https://github.com/GoogleChrome/guidance/issues/91 https://github.com/GoogleChrome/guidance/issues/90

andreban:
I don't think Built-in AI and WebMCP belong in the same bucket, as they serve quite different purposes. Here's how I'd organize this:

- Optimizing site for AI Agents (ie: Agentic Web)
  - Best practices for agents to understand a site
  - WebMCP 
- Building AI features into web products (Web AI)
  - Built-in AI
  - Potentially - WASM, WebGPU, WebNN for AI, but give those are mostly going to be used via abstractions it might not  make sense to put guidance here. @tomayac for thoughts.

gilbertococchi:
Hi @andreban, +1 to what you said, could it be you swapped the categories in ()?

andreban:
> Hi [@andreban](https://github.com/andreban), +1 to what you said, could it be you swapped the categories in ()?

Indeed, i just fixed the comment :)

rviscomi:
To clarify, this project supports two tiers of guidance: high-level skills and low-level guides. Skill files are intended to cover broadly scoped web disciplines (think performance, security, etc). We're aiming to have only about a dozen of these files, and each one should focus on breadth over depth. OTOH guides cover the implementation details for individual use cases (think tooltips, image optimization, etc). We expect to have hundreds of guides, focusing on depth over breadth.

I understand the distinction between using AI on your site and making your site usable by AI. Conceptually though, it can all be described as parts of the Web AI discipline. So given the need to minimize the number of high-level skills and cover a broad range of topics, I think it makes the most sense to have a single Web AI skill file for things like built-in AI and WebMCP. It could be structured exactly as @andreban outlined in https://github.com/GoogleChrome/guidance/issues/545#issuecomment-4266804798.

matthiasrohmer:
Thought about this a little, based on a conversation in chat. I wonder if built-in AI should move into the user-experience category, instead of the high-level Web AI skill proposed here.

Instead I would suggest the new category proposed in this issue to become an "Agentic browsing" skill, following the new related Lighthouse audit with the same scope. So next to WebMCP a valid part of this skill could be how to (and when to) generate an llms.txt for example.

[...]

Issue #91, opened 2026-02-20
WebMCP


beaufortfrancois:
Following up on https://github.com/GoogleChrome/guidance/issues/90, we've also started that effort on the WebMCP side at https://github.com/GoogleChromeLabs/webmcp-tools/pull/12

cc @andreban 

Feature ID: navigator-modelcontext

**Use cases:**
- [x] #676

natorion:
@Kulikowski is this something you can work on?

Kulikowski:
Yes, we are handling that already :)

rviscomi:
@Kulikowski to confirm, are you working on writing the use case-specific guidance for WebMCP? Agent skills? Both?

Kulikowski:
We are working on Agent SKILL for WebMCP, including best practices for implementing tools.

By use-case specific do you mean "Implement in React" or "Implement for shopping CUJ" ?

rviscomi:
By use case I mean something task-oriented, a solution that addresses a specific problem statement. For WebMCP maybe there's really just one use case like "give end-users' AI assistants direct hooks into the web app to automate core interactions" and that guide contains all of the best practices that you would otherwise put in a skill.

For now, starting by writing the skill SGTM and we can reorganize as needed. See [go/xuwik](http://go/xuwik) for more info about creating skills.

PR #666, opened 2026-04-28
Publish the WebMCP skill and guides

rviscomi:
Splits the WebMCP skill's declarative and imperative API subresources into standalone guides and combines the rest of the subresources into `guides/webmcp/SKILL.md` to make everything production-ready.

Also creates demo and expectations files for both guides so they can be eval'd.

Progress on #91 and #545


guides/webmcp/agentic-javascript-tools/expectations.md
```diff
@@ -0,0 +1,7 @@
+- The code checks for `modelContext` in `navigator` before registering a tool.
+- `navigator.modelContext.registerTool` is called with a tool definition object.
+- The tool definition includes a `name`, `description`, and `inputSchema`.
+- The `inputSchema` is a valid JSON Schema object with property descriptions.
+- The `execute` function returns a JSON-serializable value.
```
beaufortfrancois: We made a mistake earlier: the output of the `execute` function is actually `any`, not necessarily a JSON-serializable value. See https://webmachinelearning.github.io/webmcp/#callbackdef-toolexecutecallback. We should update `guides/webmcp/agentic-javascript-tools/guide.md` file :

> *   **Return Format**: The `execute` function can return any ~~JSON-serializable~~ value (object, array, string, number, boolean). Select a structure that best serves your specific use case while ensuring the content is optimized for the LLM to process. The output may encompass raw data, specific error logs, or direct instructions to influence the agent's next action.

guides/webmcp/agentic-forms/expectations.md
```diff
@@ -0,0 +1,7 @@
+- The form element has both `toolname` and `tooldescription` attributes.
+- Input elements have associated labels or `toolparamdescription` attributes.
+- The `submit` event listener uses `event.preventDefault()`.
+- The `submit` event listener checks `event.agentInvoked`.
+- The `submit` event listener calls `event.respondWith()` with a Promise.
+- The response provided to `event.respondWith()` is a JSON-serializable object.
```
beaufortfrancois: Same here. It is `any` value: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/html/forms/submit_event.idl;l=13;drc=68b03e5264cdd8606b3e846c837b2882f0b48b8e

guides/webmcp/agentic-javascript-tools/expectations.md
```diff
@@ -0,0 +1,7 @@
+- The code checks for `modelContext` in `navigator` before registering a tool.
+- `navigator.modelContext.registerTool` is called with a tool definition object.
+- The tool definition includes a `name`, `description`, and `inputSchema`.
```
beaufortfrancois: It should also have `execute`.
Since `annotations` is optional, we don't have to include but it in the `demo.html` file. WDYT?

guides/webmcp/agentic-javascript-tools/demo.html
```diff
@@ -0,0 +1,54 @@
...
+                },
+                execute(input) {
+                    console.log('Tool invoked with:', input);
+                    return input.a + input.b;
+                },
+                annotations: { readOnlyHint: true }
+            }, { signal: controller.signal });
+
+            // Mandatory: Cleanup on page unload
+            window.addEventListener('beforeunload', () => {
```
beaufortfrancois: I'm not sure this reflects what developers would do. Shall we have a react app instead that shows how `controller.abort()` would get called instead? See https://github.com/GoogleChromeLabs/webmcp-tools/blob/94d09c9facf7dbd010ab0f77df6b96a7243e4fc5/demos/analytics-dashboard/src/hooks/useWebMCPQueryTool.js#L82 for instance.

Or maybe it's not worth in a demo file. WDYT?

guides/webmcp/agentic-javascript-tools/demo.html
```diff
@@ -0,0 +1,54 @@
...
+<body>
+    <h1>WebMCP Imperative Demo</h1>
+    <p>Check the console or use a WebMCP-enabled browser assistant to interact with the <code>calculate_sum</code> tool.</p>
+    <div id="status">Checking WebMCP support...</div>
+
+    <script>
+        const status = document.getElementById('status');
+
+        if ('modelContext' in navigator && 'registerTool' in navigator.modelContext) {
+            status.textContent = 'WebMCP Imperative API is supported! Tool "calculate_sum" registered.';
```
beaufortfrancois: `Tool "calculate_sum" registered` is wrong. It should set only after `registerTool()` gets called.

guides/webmcp/agentic-forms/demo.html
```diff
@@ -0,0 +1,72 @@
+<!DOCTYPE html>
```
beaufortfrancois, line 1: `negative-demo.html` and `grader.ts` files are missing. Is that intended?

guides/webmcp/agentic-forms/expectations.md
```diff
@@ -0,0 +1,7 @@
...
+- The `:tool-form-active` pseudo-class is used to provide visual feedback.
```
beaufortfrancois, line 6: Should we also have `tool-submit-active` as well?

guides/webmcp/agentic-forms/demo.html
```diff
@@ -0,0 +1,72 @@
...
+        #result { margin-top: 1rem; padding: 1rem; background: #eee; display: none; }
+    </style>
+</head>
+<body>
+    <h1>WebMCP Declarative Demo</h1>
+    
+    <form id="demo-form" 
+          toolname="get_weather" 
+          tooldescription="Get current weather for a location" 
+          toolautosubmit>
```
beaufortfrancois: If this specific case, since we have `form:tool-form-active` and `button:tool-submit-active` defined, we shouldn't have toolautosubmit otherwise we won't use them. I'd lean towards removing `toolautosubmit` then.

guides/webmcp/agentic-forms/demo.html
```diff
@@ -0,0 +1,72 @@
...
+            const resultText = `Weather in ${city}: 22°C, Sunny`;
+            resultDiv.textContent = resultText;
+            resultDiv.style.display = 'block';
+            resultDiv.style.color = 'black';
+
+            if (event.agentInvoked) {
+                event.respondWith(Promise.resolve(weatherResult));
+            }
+        });
+    </script>
```
beaufortfrancois, line 77: Nit: Shall we test this as well?

```js
window.addEventListener('toolactivated', ({ toolName }) => {
  console.log(`Tool "${toolName}" was activated by the agent.`);
});

window.addEventListener('toolcancel', ({ toolName }) => {
  console.log(`Tool "${toolName}" interaction was cancelled.`);
});
```

guides/webmcp/agentic-javascript-tools/guide.md
```diff
@@ -1,4 +1,10 @@
-# WebMCP Imperative API
+---
+name: agentic-javascript-tools
+description: Programmatically register client-side JavaScript functions as tools for AI agents using the WebMCP Imperative API.
+web-feature-ids:
+  - navigator-modelcontext
+sources: []
```
beaufortfrancois: Shall we link to the spec for now?

beaufortfrancois (changes requested):
Thank you for this PR! I've left some comments.

guides/webmcp/agentic-javascript-tools/guide.md
andreban: Could the fact that the specs and actual implementation have differences cause the agent to pull the specs and then get confused by the differences?

guides/webmcp/agentic-javascript-tools/demo.html
andreban: How are we dealing with React best practices in MWG? I definitely think React guidance to agents is valuable, but wondering how it aligns more broadly with the project.

guides/webmcp/agentic-javascript-tools/guide.md
beaufortfrancois: it's not clear even with https://github.com/GoogleChrome/guidance/blob/b328190da3cc0eb34af76294bc386ff588cae5c2/.agents/skills/project-guides/SKILL.md?plain=1#L47 if sources are used for this.

guides/webmcp/agentic-forms/demo.html
rviscomi: Yes, we're splitting out authoring guidance from generating the evals to unblock the guidance

guides/webmcp/agentic-forms/demo.html
rviscomi: Done

guides/webmcp/agentic-forms/demo.html
rviscomi: Added

guides/webmcp/agentic-forms/expectations.md
rviscomi: I'll remove this. If the response can be any value, there's no failure case for us to check.

guides/webmcp/agentic-forms/expectations.md
rviscomi: Added

guides/webmcp/agentic-javascript-tools/demo.html
rviscomi: Fixed

guides/webmcp/agentic-javascript-tools/demo.html
rviscomi: Yeah the demo file is never shown to agents, it's just a reference implementation for calibrating our evals. The only requirements are for it to adhere to all of the best practices in the guide and meet all of the expectations. Other than that, it should be as minimal as possible, so React would be overkill.

The only expectation related to this is:

> An `AbortController` is created and its signal is passed to `registerTool`.

There's nothing about how to call `abort()` so if you'd prefer, we can remove this event handler and it should still work.

guides/webmcp/agentic-javascript-tools/expectations.md
rviscomi: Updated

guides/webmcp/agentic-javascript-tools/expectations.md
rviscomi: Removed this expectation and updated the guide

guides/webmcp/agentic-javascript-tools/guide.md
rviscomi: Added links to the spec in both guides.

> Could the fact that the specs and actual implementation have differences cause the agent to pull the specs and then get confused by the differences?

No, the source URLs are not exposed to agents

rviscomi:
Thanks @beaufortfrancois! Addressed all feedback. This is ready for another look.

guides/webmcp/agentic-forms/demo.html
beaufortfrancois: Shall we add expectations for those in `guides/webmcp/agentic-javascript-tools/expectations.md`?

guides/webmcp/agentic-javascript-tools/demo.html
beaufortfrancois: Let's remove this event handler then.

beaufortfrancois (approved):
LGTM with nits

guides/webmcp/agentic-forms/demo.html
rviscomi: It comes down to whether you consider handlers for these events to be requirements for the correct implementation. Sorry I should have clarified that before adding it to the demo.

If so, the guide may need to be rephrased to more strongly emphasize the importance of listening to these events, and what to do in their handlers, or if console logs are enough.

Otherwise if they're optional implementation details, it's ok to talk about them in the guide and use them in the demo, but not include them in the expectations.

guides/webmcp/agentic-javascript-tools/demo.html
rviscomi: Removed

rviscomi:
Thanks for the LGTM @beaufortfrancois. One remaining open question here: https://github.com/GoogleChrome/guidance/pull/666#discussion_r3161946214

Any other feedback @andreban or @Kulikowski before we merge this?

guides/webmcp/agentic-javascript-tools/demo.html
```diff
@@ -1,54 +1,61 @@
...
-            status.style.color = 'red';
+  <h1>WebMCP Imperative Demo</h1>
+  <p>Check the console or use a WebMCP-enabled browser assistant to interact with the <code>calculate_sum</code> tool.
+  </p>
+  <div id="status">Checking WebMCP support...</div>
+
+  <script>
+    const status = document.getElementById('status');
+
+      if ('modelContext' in navigator && 'registerTool' in navigator.modelContext) {
```
beaufortfrancois: Formatting seems off in this script. Can you run again?

guides/webmcp/agentic-forms/demo.html
beaufortfrancois: I think we're good then as they're not mandatory.

guides/webmcp/agentic-javascript-tools/demo.html
rviscomi: Weird. Should be fixed now, thanks!

rviscomi:
Ok merging! This should be available for testing in the [modern-web-guidance](https://github.com/GoogleChrome/modern-web-guidance) repo shortly. Feel free to raise a PR if there's anything else.

Issue #537, opened 2026-04-10
Create a high-level security skill


philipwalton:
Initial generated draft is here:
https://github.com/GoogleChrome/guidance/blob/main/skills-drafts/security/SKILL.md

This needs to be reviewed for accuracy and comprehensiveness.

agektmr:
I'd be cautious about launching this skill without sufficient review by the security team (not the guide, but the outcome by using the skill).

camillelamy:
I had a look at the content. It would be helpful to understand what the intended use cases for the skill package are, because as it is written right now, I think it would just result in a lot of breakages in any existing application that does not deploy the security policies already. A lot of the security policies mentioned in the package are complex to deploy, and require to look at the application holistically for the deployment to be successful. If deploying CSP was a simple as just send a strict CSP header, developers would not struggle with deploying it as they are now.

Overall, I am somewhat skeptical that we can write a skill package that just says deploy this and that policy. I'm even more skeptical of any agent ability to deploy a security policy without breaking the app, but that is a different topic.

Assuming the agent could to that, the first step is to check the constraints on the application. Then based on this, you will look at what is actually possible to deploy. We also need to integrate reporting in the deployment as much as possible. While you mention reporting, none of your header's examples integrate the reporting endpoint parameters.

Another thing that would be worth mentioning is avoiding DOM XSS sinks in the first place. I think this is a case where you are more likely to see success with an LLM, because we can say stuff like don't use innerHTML of setHTMLUnsafe, always use setHTML instead. This is much easier for the LLM to follow successfully and without breakages.

rviscomi:
Thank you @camillelamy! I've opened PR #677 to restructure the skill. Any additional feedback welcome!

PR #677, opened 2026-04-29
Security skill updates

rviscomi:
Progress on #537

apply feedback, 3-phase process


skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,208 @@
...
+### Core enforcement (data-driven rollouts)
 
-## Security Headers Quick Guide
+#### 3.1 Analyzing CSP Reports
+- **Scenario**: Many violations for inline scripts.
+  - **Condition**: The app uses a framework that relies on inline scripts.
+  - **Decision**: Implement Nonces (server-rendered) or Hashes (static) before enforcing.
+- **Scenario**: Violations for third-party analytics scripts.
+  - **Condition**: The scripts are required.
+  - **Decision**: Allow the specific domains, or use `'strict-dynamic'` if they load dependencies.
```
camillelamy: This is not correct. You should not use domain based allowlists as they are not secure. So you should use Nonce + strict dynamic. There are some cases where nonce is not possible. For those, Chrome plans to release an extension of CSP, but it is still in Origin Trial. Once it releases, we could update the guidance here.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,208 @@
...
+- **Scenario**: Trusted Types violations on specific sinks.
+  - **Condition**: Legacy code paths still write strings to `innerHTML` etc.
+  - **Decision**: Refactor those sinks (per §1.1) or route them through a Trusted Types policy (§3.3) before enforcing.
 
-### X-Content-Type-Options
-- **DO**: Set `X-Content-Type-Options: nosniff` to block MIME-type sniffing.
+#### 3.2 Transitioning to CSP Enforcement
+Only move to enforced mode when:
+1. Violations in the report-only logs have dropped to near zero or are accounted for.
+2. You have refactored unsafe sinks identified in Phase 1.
```
camillelamy: I don't think this is needed for CSP enforcement, unless you also do TrustedTypes enforcement. But then this should show up in violations.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,208 @@
...
-- **DO**: Use `Strict-Transport-Security: max-age=31536000; includeSubDomains; preload` for force-HTTPS.
-- **TIP**: In production rollout, start with a short `max-age` (e.g., 300 seconds) and incrementally increase it to 1 year to avoid rendering the site permanently inaccessible due to misconfiguration.
+HTML for nonce-based CSP:
+```html
+<script nonce="{RANDOM}" src="https://example.com/script.js"></script>
+```
 
-### Referrer Policy
-- **DO**: Use `strict-origin-when-cross-origin` as a safe default.
+For static/cached HTML (SPAs) where a per-response nonce is not possible, use hash-based CSP: hash each inline script and list the hashes in `script-src`.
```
camillelamy, line 222: Note: we will be able to update this once the extension to CSP hashes in Chrome moves out of Origin Trial.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,208 @@
...
-- **DO**: Combine with `crossorigin="anonymous"`.
-- **DO NOT**: Use SRI for dynamic or unversioned assets, as silent updates will cause script execution to fail. It is strictly for immutable, versioned assets.
-- **DO**: Ensure the **server/CDN** sends `Access-Control-Allow-Origin: *` or specific origin for the browser to successfully compute the hash.
+**Avoid**: URL allowlists like `script-src https://cdn.example.com` — they are easily bypassed by open redirects, JSONP endpoints, and dependency injection on the allowed origin.
 
-```html
-<script src="https://cdn.example.com/lib.js" integrity="sha384-H8df...39v" crossorigin="anonymous"></script>
-```
+#### 3.3 Trusted Types Enforcement
+Trusted Types enforces the §1.1 source-level guidance at runtime: once enabled, the browser blocks string assignments to dangerous sinks unless they pass through a named policy.
```
camillelamy: Note: if the websites is relying on a framework that doesn't support TrustedTypes, TrustedTypes will not be deployable.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,208 @@
...
+    createHTML: str => str.replace(/</g, '&lt;').replace(/>/g, '&gt;')
+  });
+  el.innerHTML = policy.createHTML(untrustedString);
+}
 ```
 
-```http
-Permissions-Policy: camera=(), geolocation=(), microphone=()
-```
+#### 3.4 Handling Advanced Isolation (COOP / COEP / CORP)
```
camillelamy: Ok so this would be worth breaking down into three subsections.

One for CrossOriginIsolation, one for COOP by itself and one for CORP by itself.

For crossOriginIsolation, you only need to look at enabling it if you need to access SharedArrayBuffers. If not, you don't need to care. If you need SharedArrayBuffers, on Chrome you can use DocumentIsolationPolicy which is much simpler to deploy. It will still break all non-CORS cross-origin fetches. So you need an audit of subresources. Document-Isolation-Policy: isolate-and-credentialless might be easier to deploy than isolate-and-require-corp, since it will strip credentials from those Fetches instead of just blocking them if they don't have CORP. For non-Chrome browsers, you do need to deploy COOP same-origin and COEP require-corp. They are very hard to deploy, so I am not sure it even makes sense to put it in the skill package.

Now, for COOP itself:
If you don't expect that you need to be reachable from an opener (ie you're not an OAuth or paiement system), then it is a good idea to deploy COOP same-origin-allow-popups, which prevents from a bunch of XS-leaks attacks that a malicious website could perform by opening you. It also doesn't break OAuth or payment flows initiated by the website, so it's much more deployable.

For CORP itself
Ideally, for each resource you would set a CORP policy explicitly, based on whether they should be embeddable in other contexts or not. This both protects them, and ensure they can be loaded in COEP/DIP contexts. I am not sure of what the policies web application actually use for that though.

skills-drafts/security/SKILL.md
```diff
@@ -154,34 +219,73 @@ app.use((req, res, next) => {
...
+```
+
+```html
+<iframe src="https://trusted-video.com/player" allow="fullscreen; camera"></iframe>
+```
+
+#### Subresource Integrity (SRI)
+- **DO**: Use the `integrity` attribute with a cryptographic hash when loading third-party scripts, combined with `crossorigin="anonymous"`.
+- **DO**: Ensure the server/CDN sends an appropriate `Access-Control-Allow-Origin` header so the browser can compute the hash.
+- **DO NOT**: Use SRI for dynamic or unversioned assets — silent updates will cause script execution to fail. SRI is strictly for immutable, versioned assets.
```
camillelamy, line 332: For those, we are introducing signature based SRI. But it might still be in Origin Trial.

camillelamy (commented):
Thanks! This looks better. I have added a few comments.

I would also suggest that you reach out to ISE Web, which has a lot of experience rolling out these policies on Google properties. Their skillset is probably closer to the skill you're trying to codify here, compared to me who is building the implementation of the security policies in the browser. aaj@ is a point of contact for the team.

rviscomi:
Thank you @camillelamy! I've reached out to Artur and addressed your feedback, except for the notes about upcoming origin trials, as that guidance is not as clear-cut for agents to follow. We will definitely need to revisit this doc often as the landscape changes and features stabilize.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,225 @@
...
-- **DO**: Include `'strict-dynamic'` to allow trusted scripts to load dependencies automatically.
-- **DO**: Restrict `object-src 'none'` (block plugins) and `base-uri 'none'` (block `<base>` hijacking).
-- **DO NOT**: Rely on URL allowlists (e.g., `script-src https://cdn.example.com`), as they are easily bypassed.
+Trusted Types can enforce this pattern at runtime by blocking string assignments to dangerous sinks. Deploying it is a CSP enforcement step with real breakage risk — see §3.3.
 
-### Code Pattern (Nonce-based)
+### 1.2 Secure Cookies
+Ensure new cookies are configured securely by default.
+- **DO**: Explicitly set `SameSite=Lax` for standard first-party cookies.
+- **DO**: Set `SameSite=None; Secure; Partitioned` for third-party contexts (iframes).
```
arturjanc: "... if your application needs to be embedded as an iframe in third-party contexts."

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,225 @@
...
+- **Enumerate inline scripts and styles** in server-rendered templates and static HTML — these will need nonces, hashes, or refactoring.
+- **List third-party script origins** loaded by the app (analytics, ads, tag managers, CDNs). These dictate what `script-src` must allow or whether `'strict-dynamic'` is viable.
+- **Identify popup-dependent flows**: OAuth, payment gateways, SSO. These constrain COOP choices.
+- **Identify cross-origin embeds and embedders**: iframes the app loads, and sites that embed the app. These constrain COEP/CORP/`frame-ancestors`.
+- **Enumerate required browser features**: List any features (camera, geolocation, microphone, fullscreen) used by the app or embedded third-party widgets to inform `Permissions-Policy`.
+- **Identify dynamic dependencies**: Check if third-party scripts are versioned or if they receive silent updates, determining if `SRI` can be used.
+- **Map cross-site integrations**: List all incoming Webhooks, cross-site APIs, or SSO redirect endpoints so `Fetch Metadata` resource isolation policies don't break them.
 
-### Code Pattern
+### 2.2 Deploy Report-Only Policies
```
arturjanc, line 142: I think here it could be useful to list which headers have report-only policies, e.g. CSP, COOP, etc.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,225 @@
...
+- **Scenario**: Violations for third-party analytics scripts.
+  - **Condition**: The scripts are required.
+  - **Decision**: Use `'strict-dynamic'` with a per-request nonce so the analytics loader can attach its dependencies. Do **not** add the analytics origin to a URL allowlist — domain allowlists are bypassable via open redirects, JSONP, and dependency injection on the listed origin.
+- **Scenario**: Trusted Types violations on specific sinks.
+  - **Condition**: Legacy code paths still write strings to `innerHTML` etc.
+  - **Decision**: Refactor those sinks (per §1.1) or route them through a Trusted Types policy (§3.3) before enforcing.
+
+#### 3.2 Transitioning to CSP Enforcement
+Only move to enforced mode when:
+1. Violations in the report-only logs have dropped to near zero or are accounted for.
```
arturjanc, line 201: This is good guidance, but in practice reporting endpoints receive a lot of false-positive violations from older browsers, extensions, etc. I'm not sure how to address it, it would likely get fairly complex if we wanted to try to give more concrete guidance on how to decide whether a given violation report is likely a false positive.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,225 @@
...
 
-## Advanced Cross-Origin Security
+#### 3.7 Fetch Metadata (Resource Isolation)
+Server-side enforcement that uses `Sec-Fetch-*` request headers to reject suspicious cross-site requests. Requires the cross-site integration mapping from §2.1 before enforcing.
 
-### Fetch Metadata
-- **DO**: Implement a Resource Isolation Policy on the server check `Sec-Fetch-*` headers.
-- **DO**: Reject `cross-site` requests for non-navigational endpoints.
+- **DO**: Implement a server-side resource isolation policy that checks `Sec-Fetch-*` headers and rejects `cross-site` requests for non-navigational endpoints.
 - **DO**: Include `Vary: Sec-Fetch-Dest, Sec-Fetch-Mode, Sec-Fetch-Site` to prevent intermediate caches (CDNs) from serving cached responses to attackers.
```
arturjanc, line 278: Nit: In this context we're not worried not so much about shared caches, but about an attacker-controlled document on evil.com receiving (e.g. navigating to) a resource cached in a same-origin context. So I'd simplify this to: "... to prevent serving cached responses to attackers".

skills-drafts/security/SKILL.md
```diff
@@ -154,34 +236,73 @@ app.use((req, res, next) => {
...
+These carry significantly lower breakage risk than the core enforcement track. They can be deployed alongside — or before — the CSP and isolation rollouts.
+
+#### HTTP Strict Transport Security (HSTS)
+- **DO**: `Strict-Transport-Security: max-age=31536000; includeSubDomains; preload` to force HTTPS.
+- **TIP**: In production rollout, start with a short `max-age` (e.g., 300 seconds) and incrementally increase to 1 year. A misconfigured HSTS with a long max-age can render the site permanently inaccessible until the cache expires in every browser that saw it.
+
+#### X-Content-Type-Options
+- **DO**: Set `X-Content-Type-Options: nosniff` to block MIME-type sniffing.
+
+#### Clickjacking: `frame-ancestors`, X-Frame-Options, iframe `sandbox`
```
arturjanc: This is just an opinion, but IMHO setting `X-Frame-Options: SAMEORIGIN` (or frame-ancestors in CSP, but XFO is simpler) is a good bit of advice that could be moved to Phase 1 because it's usually relatively simple to enable and provides fairly comprehensive clickjacking protection.

arturjanc (commented):
This looks pretty nice overall - I reviewed the skill and the guidance seems in line with what we're recommending to developers elsewhere.

The main question is what work to emphasize as the most actionable bit of guidance (low cost vs. high reward). For reference, we've traditionally divided web security work into three buckets:
1. Encryption: fixed through the use of HTTPS and HSTS.
2. Isolation: fixed by SameSite=Lax cookies, X-Frame-Options/frame-ancestors, policies based on Sec-Fetch-* headers, COOP, etc.
3. Injections/XSS: mitigated through CSP, Trusted Types, and avoiding DOM sinks.

In practice, based on data from bug bounties (3) is the largest threat, but deploying CSP & TT is also fairly involved, compared to (1) and (2) for which it's generally relatively easy-to-deploy defenses. 

So maybe there could be some emphasis on XSS defenses as an important goal in practice to make sure the guidance tool prioritizes deploying either CSP or Trusted Types. Just a thought...

skills-drafts/security/SKILL.md
rviscomi, line 1: As discussed with @arturjanc, we should also have guidance for secure window messaging.

> Guidance for window.postMessage is missing. High-maturity apps must strictly validate the origin of incoming messages and avoid using the * wildcard when sending data.

rviscomi:
Thanks @arturjanc! All of your feedback should be addressed.

skills-drafts/security/SKILL.md
```diff
@@ -154,34 +278,61 @@ app.use((req, res, next) => {
...
+
+#### HTTP Strict Transport Security (HSTS)
+- **DO**: `Strict-Transport-Security: max-age=31536000; includeSubDomains; preload` to force HTTPS.
+- **TIP**: In production rollout, start with a short `max-age` (e.g., 300 seconds) and incrementally increase to 1 year. A misconfigured HSTS with a long max-age can render the site permanently inaccessible until the cache expires in every browser that saw it.
+
+#### X-Content-Type-Options
+- **DO**: Set `X-Content-Type-Options: nosniff` to block MIME-type sniffing.
+
+#### Referrer Policy
+- **DO**: Use `strict-origin-when-cross-origin` as a safe default.
```
mikewest:
```suggestion
- **DO**: Use `Referrer-Policy: strict-origin-when-cross-origin` as a safe default.
```

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,267 @@
...
-description: Preventative security guidelines for web developers (XSS, CSP, Cookies, Cross-Origin Isolation). Use this skill when implementing standard HTTP security headers, configuring secure cookies, or enforcing cross-origin isolation.
+description: Preventative security guidelines for web developers (XSS, CSP, Cookies, Cross-Origin Isolation). Use this skill to guide the process of auditing, testing, and deploying security policies safely.
 ---
 
 # Web Security
 
-Guidelines for implementing preventative security measures on the web.
+Guidelines for implementing preventative security measures on the web safely and incrementally.
+
+**NOTE**: This skill covers standard web platform defenses but does **not** replace comprehensive backend security, authorization models, or server-side input validation.
```
mikewest: Nit: This reads a little more clearly and specifically to me:

```suggestion
**NOTE**: This skill covers standard web platform defenses, focusing mostly on the browser. Applications still require comprehensive server-side security, authorization models, and input validation.
```

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,267 @@
...
+// Unsafe
+element.innerHTML = `Hello, ${untrustedName}!`;
 
-**NOTE**: This skill covers standard web platform defenses (XSS, CSRF, Clickjacking) but does **not** replace comprehensive backend security, authorization models (Access Control), or server-side input validation.
+// Safe
+element.textContent = `Hello, ${untrustedName}!`;
+```
 
-## Content Security Policy (CSP)
+Trusted Types can enforce this pattern at runtime by blocking string assignments to dangerous sinks. Deploying it is a CSP enforcement step with real breakage risk — see §3.3.
```
mikewest, line 82: I think I disagree with the last sentence. _Enforcing_ trusted types does entail real breakage risk, but there's no need to do everything at once. Rolling out a policy that you use in a small portion of an app that you're refactoring and then slowly spreading out the usage as you find more sinks to replace seems like a totally viable approach that we could be more positive and aggressive about. That has the benefit of making eventual enforcement simpler, while not breaking anything in the short-term.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,267 @@
...
+
+- **Retrofitting an existing app**: Always start at Phase 1. Strict policies applied without discovery will break the app. Treat Phase 2 (report-only) as a prerequisite for any Phase 3 enforcement.
+- **Greenfield app or new feature**: You can adopt Phase 3 enforced policies directly, but still wire up reporting from day one.
+- **SaaS template / framework defaults**: Ship Phase 1 hygiene and Phase 3 policies enabled by default, with Phase 2 reporting on so downstream users can detect regressions.
+
+If you are unsure which case applies, default to Phase 1 → 2 → 3 in order.
+
+**Focusing on Leverage**: While Phase 1 and 2 establish baseline hygiene and data gathering, Phase 3 core enforcement represents the highest-leverage security work. Specifically, Injection/XSS mitigation through CSP (§3.2) and Trusted Types (§3.3) addresses the largest practical threat, while companion policies and isolation defenses provide important defense-in-depth.
+
+## Phase 1: Quick Wins & Obvious Anti-Patterns
```
mikewest, line 56: I'd suggest adding a note about encryption in the quick wins section: you talk about HTTPS a few times in the document, and note `Strict-Transport-Security` at the bottom, but it seems necessary to specifically advise developers to deploy their sites (prod and test) via HTTPS. Perhaps you could bring the HSTS section up here, a la:

```
### 1.0 Secure Contexts

- **DO**: Deliver resources over HTTPS to protect against both passive and active network attackers.
- **DO**: Serve a header like `Strict-Transport-Security: max-age=31536000; includeSubDomains; preload` to force HTTPS whenever possible.
- **TIP**: In production rollout, start with a short `max-age` (e.g., 300 seconds) and incrementally increase to 1 year. A misconfigured HSTS with a long max-age can render the site permanently inaccessible until the cache expires in every browser that saw it.
```

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,267 @@
...
+Trusted Types can enforce this pattern at runtime by blocking string assignments to dangerous sinks. Deploying it is a CSP enforcement step with real breakage risk — see §3.3.
 
-Mitigate XSS risks by restricting executable scripts and resources. Prefer **Strict CSP** over allowlists.
+### 1.2 Secure Cookies
+Ensure new cookies are configured securely by default.
+- **DO**: Explicitly set `SameSite=Lax` for standard first-party cookies.
+- **DO**: Set `SameSite=None; Secure; Partitioned` if your application needs to be embedded as an iframe in third-party contexts.
+- **DO**: Use the `__Host-` prefix for cookies scoped to a single origin (requires `Secure`, `Path=/`, and no `Domain` attribute) — the browser rejects writes that don't meet those constraints, preventing subdomain injection.
+- **DO**: Use the `__Secure-` prefix for any cookie that must only be set over HTTPS.
+- **DO NOT**: Rely on unpartitioned `SameSite=None` — these are being systematically blocked for tracking prevention.
```
mikewest: I have a few suggestions here, but nothing actually material. Just shifting the emphasis a bit in certain places in ways that I think might be helpful:

```suggestion
### 1.2 Secure Cookies
Ensure new cookies are configured securely by default.
- **DO**: Prefer naming cookies with the `__Host-` prefix when they'll only be used by one domain. This requires the `Secure` and `Path=/` attributes to be set, and the `Domain` attribute to be omitted. This protects against same-site and network attackers.
- **DO**: Prefer naming cookies with the `__Secure-` prefix when `__Host-` isn't appropriate. This requires the `Secure` attribute, and protects against network attackers.
- **DO**: Explicitly set `SameSite=Lax` for standard first-party cookies.
- **DO**: If your application will be embedded as an iframe in third-party contexts, use `SameSite=None; Secure; Partitioned`.
- **DO NOT**: Rely on unpartitioned `SameSite=None` — these are being systematically blocked for tracking prevention.
```

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,267 @@
...
+Set-Cookie: __Host-session_id=value; SameSite=Lax; HttpOnly; Secure; Path=/
+Set-Cookie: third_party_var=value; SameSite=None; Secure; Partitioned
+```
 
-### Code Pattern (Nonce-based)
+### 1.3 Clickjacking Protection (Frame-Ancestors & X-Frame-Options)
+Clickjacking protection is easy to deploy, carries extremely low risk of breaking legitimate functionality, and provides immediate defense against malicious UI redressing.
+- **DO**: Set the `X-Frame-Options: SAMEORIGIN` header to prevent other sites from embedding your pages in an iframe (or use `DENY` if you should never be embedded).
+- **DO**: For fine-grained control in modern browsers, use `frame-ancestors 'self'` (or specified trusted domains) in your CSP header.
+- **DO**: When *you* embed untrusted content, always use the iframe `sandbox` attribute to restrict its capabilities unless explicitly allowed (e.g., `sandbox="allow-scripts"`).
```
mikewest: This is reasonable advice, but seems misplaced in this section. It also seems somewhat less-likely to be "Phase 1", given that `sandbox="allow-scripts"` can break things like `localStorage`, which often breaks pages.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,267 @@
...
+```http
+Set-Cookie: __Host-session_id=value; SameSite=Lax; HttpOnly; Secure; Path=/
+Set-Cookie: third_party_var=value; SameSite=None; Secure; Partitioned
+```
 
-### Code Pattern (Nonce-based)
+### 1.3 Clickjacking Protection (Frame-Ancestors & X-Frame-Options)
+Clickjacking protection is easy to deploy, carries extremely low risk of breaking legitimate functionality, and provides immediate defense against malicious UI redressing.
+- **DO**: Set the `X-Frame-Options: SAMEORIGIN` header to prevent other sites from embedding your pages in an iframe (or use `DENY` if you should never be embedded).
+- **DO**: For fine-grained control in modern browsers, use `frame-ancestors 'self'` (or specified trusted domains) in your CSP header.
```
mikewest: "in modern browsers" is a little outdated. `frame-ancestors` has been available for a decade+ at this point. If you have a CSP already, it's perfectly reasonable to rely on this exclusively.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,267 @@
...
+X-Frame-Options: SAMEORIGIN
+Content-Security-Policy: frame-ancestors 'self' https://trusted-partner.com;
 ```
 
-HTML:
-```html
-<script nonce="{RANDOM}" src="https://example.com/script.js"></script>
+### 1.4 Secure Window Messaging (postMessage)
+If your application communicates with other origins using `window.postMessage`, you must strictly validate the sender and receiver.
+- **DO**: Always validate the `event.origin` of incoming messages on the receiver side using strict equality against a list of trusted origins. Do **not** trust wildcards (`*`) or unverified payloads.
```
mikewest, line 109: It would be ideal to suggest using the `Origin` interface for these comparisons when available, e.g. `Origin.from(event).isSameOrigin(trustedOrigin)`.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,267 @@
...
+- **DO**: Always specify a target origin (rather than the wildcard `*`) when calling `postMessage` to send sensitive data, ensuring only the intended origin can receive it.
+- **DO**: Parse and sanitize message payloads (e.g., using `JSON.parse` and strict schema validation) before performing operations or writing to DOM sinks.
+
+```javascript
+// Receiver (Safe)
+window.addEventListener('message', (event) => {
+  if (event.origin !== 'https://trusted-origin.com') return;
+  try {
+    const data = JSON.parse(event.data);
```
mikewest: `postMessage` will do serialization and deserialization internally for objects that one side passes to the other. It seems a little odd to me to recommend that folks serialize manually on the one side and deserialize on the other rather than just passing an object around.

Certainly the properties of that object should be verified, but is there additional value in asking the developer to do more work here? It seems superfluous to me.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,267 @@
...
 ```
 
-## Trusted Types
+## Phase 2: Discovery & Data Collection (Prerequisites)
+
+Do not blindly apply strict policies to an existing application. You must first understand the application's constraints by collecting data.
+
+### 2.1 Inspect the Application
+Before turning anything on, gather facts:
+- **Grep for existing security headers** in server config, middleware, CDN/edge config, and meta tags: `Content-Security-Policy`, `Strict-Transport-Security`, `X-Frame-Options`, `Permissions-Policy`, `Cross-Origin-*`, `Reporting-Endpoints`.
```
mikewest: I'd suggest adding `X-Content-Type-Options`, CORS headers like `Access-Control-*`, `Timing-Allow-Origin`.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,267 @@
...
+- **DO**: Run report-only for long enough to cover real traffic patterns (typically days to weeks), not just synthetic testing.
+
+**Example headers:**
+```http
+Reporting-Endpoints: main-endpoint="https://reports.example/main"
+Content-Security-Policy-Report-Only: script-src 'nonce-{RANDOM}' 'strict-dynamic' https: 'unsafe-inline'; object-src 'none'; base-uri 'none'; report-to main-endpoint;
+```
 
-Prevent DOM-based XSS by enforcing sanitization on dangerous sinks.
+The `'strict-dynamic'`, `https:`, and `'unsafe-inline'` tokens together form a backwards-compatibility ladder: modern browsers honor `'strict-dynamic'` (nonce-propagating) and ignore the others; older browsers fall back to `https:`; very old browsers fall back to `'unsafe-inline'`. The fallbacks are harmless on any browser that supports a stricter token.
```
mikewest, line 158: I think we should probably move away from this advice. Maybe even right now. `strict-dynamic` has been widely supported for ~3-4 years at this point, and I think we can start telling folks to send the simpler `script-src 'nonce-...' 'strict-dynamic'` without the confusing `https: 'unsafe-inline'`.

@arturjanc, et al might have data to suggest otherwise, but it seems like we can get rid of some baggage given https://caniuse.com/mdn-http_headers_content-security-policy_strict-dynamic.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,267 @@
...
 
-- **DO**: Enable Trusted Types via CSP.
-- **DO**: Use Trusted Types policies to sanitize data before passing to sinks like `innerHTML`, `eval()`, `script.src`.
-- **Dangerous Sinks**: `innerHTML`, `outerHTML`, `document.write`, `eval`, `setTimeout` (with string), `script.src`.
+**Managing report false-positives**: Reporting endpoints receive a significant volume of false-positive violation reports caused by aggressive browser extensions, ancient browsers, web crawlers, or antivirus scanners. When analyzing report-only logs, focus on high-frequency patterns from modern user-agents and filter out noise before making deployment decisions.
```
mikewest: @arturjanc: Does ISE have any public documentation on strategies here? The advice here seems reasonable but very light, and it might be helpful to have more-specific advice about this large body of work elsewhere.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,267 @@
...
-### Code Pattern
+### 2.3 Data Hygiene for Reports
+- **DO NOT**: Include sensitive data (PII, authentication tokens, session identifiers, query strings with secrets) in logs or violation reports. Mask or omit them at the edge before they reach the reporting endpoint.
 
-Header:
-```http
-Content-Security-Policy: require-trusted-types-for 'script'
-```
+### 2.4 Automated Discovery via Browser APIs and DevTools
+- **Reporting API**: Use `Reporting-Endpoints` in combination with report-only headers (e.g., `Content-Security-Policy-Report-Only`, `Document-Policy-Report-Only`) to have the browser automatically post violations to your server.
```
mikewest, line 169: It might be helpful to recommend that folks set an endpoint named `default`, which will capture deprecation reports and crash reports. Maybe other stuff I'm forgetting too?

mikewest: (We should probably document this better... https://developer.mozilla.org/en-US/docs/Web/API/DeprecationReport#:~:text=Deprecation%20reports%20are%20also%20sent%20as%20a%20JSON%20object%20in%20a%20POST%20request%20to%20the%20reporting%20server%20endpoint%20named%20%22default%22%2C%20if%20it%20is%20defined. is all I found.)

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,267 @@
...
-```
+#### 3.2 Transitioning to CSP Enforcement
+Only move to enforced mode when:
+1. Violations in the report-only logs have dropped to near zero or are accounted for.
+2. Reporting remains wired up after the switch — keep `report-to` on the enforced header so regressions are visible.
 
-### Clear-Site-Data (Logout)
-- **DO**: Use `Clear-Site-Data: "cache", "cookies", "storage"` on logout endpoints to ensure complete session termination.
+**Key directives to set:**
+- `default-src 'self'` as a fallback for unspecified fetch directives — anchors the policy so any new resource type defaults to a safe value.
```
mikewest: I think I'd actually suggest that we narrow the advice here to something more closely hewing to https://web.dev/articles/strict-csp. If developers want to constrain the sources from which they can obtain images or fonts or etc, great, CSP can help them do that. But from a security perspective, the core value is the `script-src` set of directives and there's little value in anything beyond them (along with `object-src` and `base-uri` (and honestly I think we can drop `object-src` at this point, post-Flash)). Other directives (particularly `default-src`) dramatically complicate deployment.

I think it's mostly non-harmful to point towards `form-action`, but deploying it seems more likely to lead to breakage, and getting folks over the line of having a robust XSS defense seems like the primary goal we ought to aim for.

There's little harm in adding `upgrade-insecure-requests`, but also little benefit given auto-upgrading mixed content.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,267 @@
...
-- **DO**: Disable unused browser features (camera, geolocation) for the page and iframes using Structured Fields syntax.
-- **DO**: When delegating features to an iframe, use the `allow` attribute in HTML *in addition* to the header.
+Set CORP explicitly on each response based on whether it should be embeddable in other contexts. Two core benefits: it protects resources from malicious cross-origin reads, and ensures compatibility when pages request stronger client-side isolation.
 
-```html
-<iframe src="https://trusted-video.com/player" allow="fullscreen; camera"></iframe>
-```
+- **DO**: Default to `Cross-Origin-Resource-Policy: same-origin` for app-internal resources (authenticated data, user session JSON, restricted internal scripts).
+- **DO**: Use `same-site` for endpoints utilized across subdomains of the same eTLD+1.
+- **DO**: Provide `cross-origin` exclusively for resources created for generic embedding or widely cached delivery (e.g., static shared assets or public CDNs).
```
mikewest, line 258: I'd like to find a way to bring in some of the guidance from https://www.w3.org/TR/post-spectre-webdev/ here. The patterns sketched out there seem quite viable to automatically deploy based on an understanding of the resource type involved, and would have real value in mitigating some isolation risks.

skills-drafts/security/SKILL.md
```diff
@@ -1,143 +1,267 @@
...
 ```http
-Access-Control-Allow-Origin: https://trusted-app.com
-Access-Control-Allow-Credentials: true
+Cross-Origin-Opener-Policy: same-origin
+Cross-Origin-Embedder-Policy: require-corp
+Cross-Origin-Resource-Policy: same-origin
 ```
 
-## Advanced Cross-Origin Security
+#### 3.7 Fetch Metadata (Resource Isolation)
```
mikewest, line 273: It's unclear to me what the scope of a skill is, but it might be helpful to link to documents like https://web.dev/articles/fetch-metadata that explain best practice here in much more detail.

skills-drafts/security/SKILL.md
```diff
@@ -154,34 +278,61 @@ app.use((req, res, next) => {
...
+### Companion policies (deploy in parallel)
+
+These carry significantly lower breakage risk than the core enforcement track. They can be deployed alongside — or before — the CSP and isolation rollouts.
+
+#### HTTP Strict Transport Security (HSTS)
+- **DO**: `Strict-Transport-Security: max-age=31536000; includeSubDomains; preload` to force HTTPS.
+- **TIP**: In production rollout, start with a short `max-age` (e.g., 300 seconds) and incrementally increase to 1 year. A misconfigured HSTS with a long max-age can render the site permanently inaccessible until the cache expires in every browser that saw it.
+
+#### X-Content-Type-Options
+- **DO**: Set `X-Content-Type-Options: nosniff` to block MIME-type sniffing.
```
mikewest, line 310: Perhaps note that this also requires setting correct `Content-Type` headers for your resources (`application/javascript` for script, `application/json` for API endpoints, `text/html` for documents, etc).

skills-drafts/security/SKILL.md
mikewest: Chrome shipped this in ~142. No other vendor yet seems interested.

skills-drafts/security/SKILL.md
```diff
@@ -154,34 +278,61 @@ app.use((req, res, next) => {
...
-- **DO NOT**: Forget CORP for subresources (images, scripts) when using `COEP: require-corp`; otherwise, they will fail to load.
+#### Subresource Integrity (SRI)
+- **DO**: Use the `integrity` attribute with a cryptographic hash when loading third-party scripts, combined with `crossorigin="anonymous"`.
+- **DO**: Ensure the server/CDN sends an appropriate `Access-Control-Allow-Origin` header so the browser can compute the hash.
+- **DO NOT**: Use SRI for dynamic or unversioned assets — silent updates will cause script execution to fail. SRI is strictly for immutable, versioned assets.
 
-### Operational Constraints
-- **CAUTION**: COOP severs `window.opener` references, which will break standard pop-up OAuth flows and payment gateways relying on `postMessage()`. Use `credentialless` if supported, or isolate OAuth to subdomains.
+```html
+<script src="https://cdn.example.com/lib.js" integrity="sha384-H8df...39v" crossorigin="anonymous"></script>
```
mikewest: Nit: IMO, we should push folks towards sha-256 or sha-512. sha-384 is a middle ground with a very specific threat model that we probably shouldn't have included in the spec in the first place.

```suggestion
<script src="https://cdn.example.com/lib.js" integrity="sha256-H8df...39v" crossorigin="anonymous"></script>
```

mikewest (commented):
Hi! Apologies for the belated response, but this looks quite good already. Thanks for pulling it together.

skills-drafts/security/SKILL.md
rviscomi: Since the feature is so new, I think we'll need a standalone guide to get into the browser support and fallback strategies. I filed https://github.com/GoogleChrome/guidance/issues/725 to track this.

I'll leave this for now, but I've made a note to revisit it.

skills-drafts/security/SKILL.md
rviscomi: We usually try to keep our guidance self-contained so that we give agents pretty much everything they need in one compact bundle.

I had an agent look at the coverage gaps between our skill and that article and it came up with a couple of small things that would be easy to inline:

```md
- **DO**: Reject disallowed requests *before* authentication or authorization checks, so the response does not leak timing information about whether a resource or session exists.
- **CAUTION**: `same-site` trusts every subdomain under your eTLD+1. If any subdomain hosts user-generated content, a legacy app, or otherwise untrusted code, drop `same-site` from the allowlist and accept only `same-origin` and `none`.
```

skills-drafts/security/SKILL.md
rviscomi: Similar comment to the one about fetch metadata, I looked at the coverage gap and it's not immediately clear if we're missing anything substantial.

Here's the agent's response FWIW:

>The Post-Spectre link in §3.6 mostly points to a header matrix per resource type. The skill already covers the practical pieces (COOP, COEP, CORP per-response defaults, `nosniff`, `Vary: Sec-Fetch-*`, CORS specificity, frame-ancestors).
>
>One item from the W3C doc that isn't in the skill and is mildly worth keeping: **`Content-Security-Policy: sandbox` on non-HTML data responses** (JSON APIs, etc.) — neutralizes script execution if an attacker tricks a browser into rendering the response as a document. Niche defense-in-depth.
>
>My suggestion: just drop the link. The skill's existing per-response CORP guidance (§3.5) and the §3.6 isolation block already cover the actionable defaults; the W3C doc is more of a reference matrix than additional critical guidance.

WDYT?

rviscomi:
@camillelamy @arturjanc @mikewest thank you all for your thorough feedback! Most comments have been addressed, but there are a couple of minor unresolved open questions. Could you please take another look and give your LGTM if you're happy to ship it?

skills-drafts/security/SKILL.md
arturjanc: We unfortunately don't have anything very actionable published externally about this :(

There's a short internal doc (https://docs.google.com/document/d/1PiZLUd6U6Z066B0lr2C0RgaH62aqzk1l9X4cS4rokFQ/edit?tab=t.0) that covers the basic types of false positives. The relevant 2 sections (slightly modified to possibly be useful here) are:

"""
The heuristics to discard spurious CSP violation reports in the CSP Frontend are constantly evolving; however, they generally hinge upon filtering one of the following types of violations:

- Reports sent by old browsers with known bugs triggering spurious violations.
- Reports for markup which is known to be injected by popular browser extensions.
- Identical reports seen across many distinct applications, indicating that the report is not caused by any particular application but by client-side middleware.
- Reports which don't contain enough information to allow for debugging.
- Reports which, for any other reason, are unlikely to point to a legitimate application bug. 

These rules do not guarantee that the violations displayed in the frontend will be completely free of false positive (or false negative) reports, but represent the best-effort attempt to maximize the signal-to-noise ratio and make it possible to identify any legitimate application incompatibilities.
"""

and

"""
For violations which appear to be caused an incompatibility in your application (usually those where the "Sample" or "Blocked URI" seem like legitimate scripts which might be present in your markup), investigate the problem. There are two common ways to do this:
- Code search for the offending script or URL to see if it's present in your code or HTML templates.
- Open the page which sent the violation (the "Document URI" column), ideally in the same browser as the one which sent the report, and check the JS console for CSP violations. Note that some violations may be caused by interacting with a certain element on the page, so exercise as much page functionality as possible.

Note: If your policy is enabled in production, a low violation volume often indicates that the violation might be a false positive and can be ignored.
"""

I think we could paraphrase this general guidance here (e.g. "filter out reports from old browsers", "ignore violations with low counts"). Also, one actionable thing would be to add 'report-sample' to the policy, i.e.

```
Content-Security-Policy-Report-Only: script-src 'nonce-{RANDOM}' 'strict-dynamic' 'report-sample'; object-src 'none'; base-uri 'none'; report-to main-endpoint;
```

arturjanc (approved):
LGTM overall; I think we could improve the section about CSP violations as suggested below, but it's still relatively high-level guidance, so it only makes that section a bit more actionable.

skills-drafts/security/SKILL.md
rviscomi: Updated PTAL

rviscomi:
I'll go ahead and merge this but feel free to comment on #537 or open a PR if you have any other feedback. Thanks again everyone!

PR #678, opened 2026-04-29
Review and revise css-layout SKILL.md (#533)

### LINKED ISSUE #533: Create a high-level CSS Layout skill
https://github.com/GoogleChrome/guidance/blob/main/skills-drafts/css-layout/SKILL.md

castastrophe:
Closes #533.

Audit pass on `skills-drafts/css-layout/SKILL.md` for accuracy, comprehensiveness, modern-feature coverage, and citations.

## Bug fixes

- Replace invalid `grid-span: 3` with `grid-column: span 3`. The original property does not exist; the example would have silently failed.
- Replace invalid unit `100i` with `100vi` / `100dvi`. `100i` is not a valid CSS length; `vi` is the inline-axis viewport unit.
- Correct claim that `align-self` is unsupported in Flexbox. `align-self` IS supported on the cross axis; only `justify-self` is ignored on flex items. Auto-margin guidance kept for the main axis.
- Fix duplicated word in flexbox example ("Push push button" → "Push the action to the end of the main axis").

## Coverage additions

- `aspect-ratio` and the `place-*` shorthands in layout fundamentals.
- New **Masonry** section that flags it as not-Baseline (Firefox flag-only), with a multicol stand-in and an `@supports (grid-template-rows: masonry)` progressive-enhancement example.
- Container queries: style queries (`@container style()`), `container-name`, and the block-size requirement for `container-type: size`.
- Popover: documented `popover="hint"` value and added an `@supports (anchor-name: --x)` fallback pattern (anchor positioning is still Limited Availability — Chromium-only).
- Note that `position-area` was previously `inset-area` in Chromium.
- `overscroll-behavior` for scroll-chain containment.
- Unprefixed `line-clamp` shorthand with feature detection, while keeping the `-webkit-box` + `-webkit-box-orient: vertical` + `-webkit-line-clamp` triad as the cross-browser default (this combo is fully specified, not deprecated).
- Explicit subgrid fallback pattern via `@supports (grid-template-rows: subgrid)` for projects whose Baseline target predates Sep 2023.

## Removals

- `shape-outside` + `float` — visual/typographic concern, not a layout primitive. Belongs in a typography or visual-effects skill.
- `list-style-position: inside` — list rendering, not layout. Modern alternatives like `::marker` and `padding-inline-start` would belong in a separate skill.

## Other changes

- Every modern feature is now annotated with current Baseline status (Widely Available / Newly Available / Limited) so the agent knows when fallbacks are required vs. unnecessary.
- Replaced the original subgrid example, which had `grid-span: 3` and a parent grid that didn't actually demonstrate track inheritance, with a card-grid where children span and inherit parent rows (the canonical "ragged edge" use case).
- Switched examples to logical properties (`inline-size`, `block-size`, `max-block-size`) for consistency with the logical-properties DO.
- Added a **Sources** section linking MDN, W3C specs (CSS Grid 3, Anchor Positioning 1, Containment 3), web.dev, Chrome for Developers, OddBird's anchor-positioning polyfill, and webstatus.dev for Baseline lookup.

## Reviewer notes

- The skill description in frontmatter was lightly expanded to mention Subgrid and anchor positioning so the agent is more likely to surface this skill for those topics.
- Citations live in a single Sources block at the bottom rather than inline, matching the style of the neighboring `skills-drafts/css/SKILL.md` and `skills-drafts/html/SKILL.md` (which have no citations) while still satisfying the issue's request for valid sources.
- All Baseline assertions were cross-checked against MDN reference pages during the review.


skills-drafts/css-layout/SKILL.md
```diff
@@ -1,76 +1,85 @@
...
 | Approach | Dimensions | Main Strength | Logic Type |
 | --- | --- | --- | --- |
-| **Flexbox** | 1D | Content distribution | Content-first |
-| **Grid** | 2D | Structural integrity | Layout-first |
-| **Subgrid** | 2D (Inherited) | Grandchild alignment | Relationship-first |
-| **Masonry** | 1.5D / 2D | Eliminating dead space | Flow-first |
+| **Flexbox** | 1D | Content distribution along a single axis | Content-first |
+| **Grid** | 2D | Structural integrity across rows and columns | Layout-first |
+| **Subgrid** | 2D (inherited tracks) | Aligning grandchildren to an outer grid | Relationship-first |
+| **Multicol** (`columns`) | 1D flow | Newspaper-style content reflow | Flow-first |
```
jamesnw: Is Multicol a widely used name? I would have assumed Multicolumn.

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,76 +1,85 @@
...
+| **Masonry** | 2D | Packing items of varied sizes without dead space | Flow-first (limited availability) |
```
jamesnw: Mention the term grid-lanes here maybe?

skills-drafts/css-layout/SKILL.md
```diff
@@ -82,29 +91,36 @@ Laying out items in a single dimension (row or column).
...
+- Use the `fr` unit to distribute available fractional space proportionally.
+- Use `minmax(min, max)` to make tracks flexible but bounded (e.g., `repeat(auto-fill, minmax(200px, 1fr))`).
+- Use `grid-template-areas` for complex, readable layout definitions. Areas are self-documenting.
+- Use **Subgrid** on either axis — `grid-template-columns: subgrid`, `grid-template-rows: subgrid`, or both — to align nested children to the outer grid's tracks.
+- Use Subgrid to solve the "ragged edge" problem in card lists where internal elements (titles, metadata, CTAs) must line up across siblings.
+- Use `span <n>` (e.g., `grid-column: span 3`) to size an item across multiple tracks. Subgrid only inherits parent tracks for the cells the child actually spans.
+- Use `grid-auto-flow: dense` deliberately, not by default — it can reorder items visually and break the logical DOM order that assistive technology and keyboard users follow.
+- Subgrid only one axis when item count is variable. Subgridding both axes when the child has more items than the parent has cells piles extras into the last track; use `grid-auto-rows`/`grid-auto-columns` for the implicit axis instead.
```
jamesnw:
```suggestion
- Subgrid only one axis when item count is variable. When the child has more items than the parent has cells, using subgrid on both axis places extras into the last track; use `grid-auto-rows`/`grid-auto-columns` for the implicit axis instead.
```
I found this hard to parse, maybe this is better?

guides/css-layout/css-layout/guide.md
```diff
@@ -113,153 +129,284 @@ Two-dimensional layout control.
   display: grid;
   grid-template-columns: repeat(3, 1fr);
   grid-template-areas:
-    "header header header"
-    "sidebar main main"
-    "footer footer footer";
+    "header  header  header"
+    "sidebar main    main"
+    "footer  footer  footer";
```
jamesnw, line 141: Nice- is it worth adding prose that describes grid-template-areas as ASCII art, and should be formatted to be visually parseable?

skills-drafts/css-layout/SKILL.md
```diff
@@ -113,153 +129,284 @@ Two-dimensional layout control.
...
-## 4. Container Queries and Advanced Sizing
-Querying the size of a parent container rather than the viewport for component-level adaptability.
+## 4. Container Queries and Component-Level Responsiveness
+
+Query the size (or computed style) of a parent container rather than the viewport. Container queries are **Baseline Widely Available** (cross-browser since Safari 16, February 2023).
+
+- Establish a containment context with `container-type: inline-size` on a wrapper to enable inline-axis size queries on its descendants.
+- Name containers with `container-name` (or the `container` shorthand) when nested contexts could collide; queries can target a specific name.
+- Use the full set of container query units: `cqi` (inline), `cqb` (block), `cqw` (width), `cqh` (height), `cqmin`, and `cqmax`.
+- Use **style queries** (`@container style(--state: active)`) to react to custom property values cascaded down from a parent — useful for theming subtrees without restructuring DOM.
```
jamesnw: This isn't baseline yet- I think either it should be omitted here, or noted as separate support from size queries.

skills-drafts/css-layout/SKILL.md
```diff
@@ -113,153 +129,284 @@ Two-dimensional layout control.
...
-### DOs and DON'Ts
-- **DO** use the `popover` attribute for menus and tooltips (non-modal content).
-- **DO** use `<dialog>` and `.showModal()` for modal interactions that require focus entrapment.
-- **DO** use `anchor-size()` to size popovers dynamically relative to their anchor dimensions.
-- **DO** use `position-try-fallbacks: flip-block` to let the browser automatically reposition when an overlay overflows viewport edges.
-- **DO** use `pointer-events: none` to let clicks pass through decorative overlays or ignore inert layers.
+Baseline status:
+- `<dialog>`: **Baseline Widely Available** (cross-browser since March 2022).
+- `popover` attribute: **Baseline Newly Available** (April 2024).
+- CSS Anchor Positioning (`anchor-name`, `position-anchor`, `anchor()`, `anchor-size()`, `position-area`, `position-try-fallbacks`): **Limited Availability** — Chromium-only at the time of writing. Always feature-detect.
```
jamesnw: This isn't correct- it is available (but buggy) in all browsers.

skills-drafts/css-layout/SKILL.md
```diff
@@ -113,153 +129,284 @@ Two-dimensional layout control.
...
+
+- Use the `popover` attribute for menus, tooltips, and other non-modal flyouts — popovers live in the top layer, so layering doesn't need `z-index` management.
+  - `popover="auto"` (or bare `popover`): light-dismissable; only one `auto` popover open at a time.
+  - `popover="manual"`: programmatic show/hide only; multiple can coexist.
+  - `popover="hint"`: like `auto` but does not close other `auto` popovers — designed for hover/focus tooltips that should layer over an open menu.
```
jamesnw:
```suggestion
  - `popover="hint"`: like `auto` but closes other `hint` popovers without closing `auto` popovers — designed for hover/focus tooltips that should layer over an open menu.
```

skills-drafts/css-layout/SKILL.md
```diff
@@ -113,153 +129,284 @@ Two-dimensional layout control.
...
-  width: anchor-size(width);
-  position-try-fallbacks: flip-block; /* Native overflow handling */
+  inset-block-start: 100%;
+  inset-inline-start: 0;
+}
+
+@supports (anchor-name: --x) {
+  [popover] {
+    position-anchor: --trigger;
+    position-area: bottom span-inline-end;
```
jamesnw: You can't mix physical and logical keywords here, so this is invalid.
```suggestion
    position-area: block-end span-inline-end;
```

skills-drafts/css-layout/SKILL.md
```diff
@@ -113,153 +129,284 @@ Two-dimensional layout control.
...
+- `<dialog>`: **Baseline Widely Available** (cross-browser since March 2022).
+- `popover` attribute: **Baseline Newly Available** (April 2024).
+- CSS Anchor Positioning (`anchor-name`, `position-anchor`, `anchor()`, `anchor-size()`, `position-area`, `position-try-fallbacks`): **Limited Availability** — Chromium-only at the time of writing. Always feature-detect.
+
+- Use the `popover` attribute for menus, tooltips, and other non-modal flyouts — popovers live in the top layer, so layering doesn't need `z-index` management.
+  - `popover="auto"` (or bare `popover`): light-dismissable; only one `auto` popover open at a time.
+  - `popover="manual"`: programmatic show/hide only; multiple can coexist.
+  - `popover="hint"`: like `auto` but does not close other `auto` popovers — designed for hover/focus tooltips that should layer over an open menu.
+- Use `<dialog>` and `.showModal()` for modal interactions that require focus trapping and an inert backdrop.
+- Use `anchor-size()` to size a popover to its trigger (e.g., a dropdown matching the input width) where supported.
```
jamesnw:
```suggestion
- Use `anchor-size()` to size a popover to its trigger (e.g., a dropdown matching the input width) where supported.
- Use `position-area` or `anchor()` on insets to position an overlay relative to its trigger.
```

skills-drafts/css-layout/SKILL.md
```diff
@@ -113,153 +129,284 @@ Two-dimensional layout control.
...
   -webkit-line-clamp: 3;
   -webkit-box-orient: vertical;
-  overflow: hidden; /* Multi-line truncation */
+  overflow: hidden;
+}
+
+@supports (line-clamp: 3) {
+  .snippet {
+    /* Modern shorthand sets display, box-orient, and line-clamp together. */
+    line-clamp: 3;
```
jamesnw: I'm surprised it needs to be wrapped in a supports block- won't it just be ignored if not supported?

skills-drafts/css-layout/SKILL.md
```diff
@@ -113,153 +129,284 @@ Two-dimensional layout control.
...
-## 7. Advanced Layout Metrics and Viewport Mechanics
-Manage edge cases involving mobile viewport shifts and track distributions.
+## 7. Viewport Mechanics and Track Distribution
+
+Handle the edge cases of mobile viewport shifts and decide between `auto-fit` vs `auto-fill`.
+
+- Use **dynamic viewport units** (`dvh`, `dvw`, `dvi`, `dvb`) for mobile layout containers that need to account for browser UI shifting (URL bar collapse/expand). All `dv*`/`sv*`/`lv*` units are Baseline Widely Available since 2024.
+- Understand the difference between `auto-fill` and `auto-fit` in CSS Grid:
+  - `auto-fill` creates as many tracks as fit, **including empty ones** (placeholder space is preserved).
+  - `auto-fit` creates tracks but **collapses empty ones to 0**, allowing filled tracks to stretch and fill the container.
```
jamesnw: Should this be in the grid section instead?

skills-drafts/css-layout/SKILL.md
```diff
@@ -113,153 +129,284 @@ Two-dimensional layout control.
...
+
+/* Logical-axis full inline width — safe in vertical writing modes too. */
+.full-bleed {
+  inline-size: 100dvi;
+}
+```
+
+## 8. Masonry — Status and Today's Workaround
+
+Native CSS masonry is **not yet Baseline**. The W3C is still finalizing syntax (current proposals include `display: grid-lanes` / `display: inline-grid-lanes` and `grid-template-rows: masonry`). Firefox supports an earlier `grid-template-rows: masonry` syntax behind the `layout.css.grid-template-masonry-value.enabled` flag; no other engines ship it in stable as of this writing.
```
jamesnw: I think this is settled?

jamesnw (commented):
This looks largely really good, with some nits and questions in this pass. The largest question I have is that this incorporates support much differently than the other skills. I was assuming this should be limited to baseline widely available, but that may be an incorrect assumption.

guides/css-layout/css-layout/guide.md
knowler, line 1: Should we coordinate some guidance about `display: contents` here and within the accessibility skill? It can be very handy in some layouts and I know a lot of folks even use it for the host element when styling their custom elements, but the effects it can have on semantics are still quite concerning, so a word of caution could be helpful (maybe: avoid using `display: contents` with interactive elements or even elements with meaningful semantics, e.g. headings). I don’t see `display: contents` mentioned here yet and we haven’t mentioned it in the accessibility skill, but I figure it could be beneficial to have some guidance for it.

Some resources:

- Adrian Roselli’s frequently updated [“It’s Mid-2022 and Browsers (Mostly Safari) Still Break Accessibility via Display Properties”](https://adrianroselli.com/2022/07/its-mid-2022-and-browsers-mostly-safari-still-break-accessibility-via-display-properties.html)
- Alice Boxhall’s [“A threat model for accessibility on the web”](https://alice.boxhall.au/articles/a-threat-model-for-accessibility-on-the-web/#:~:text=display%3A%20contents)

skills-drafts/css-layout/SKILL.md
castastrophe: I think there's still some open discussions to bikeshed some of the properties. I can double check though. I'm not sure I'd say the masonry syntax is finalized yet though because the spec is still in draft.

skills-drafts/css-layout/SKILL.md
castastrophe: Great point - best not to use vernacular! I'll update.

skills-drafts/css-layout/SKILL.md
castastrophe: Good idea, more keywords, the better.

guides/css-layout/css-layout/guide.md
castastrophe: That feels like a great fit for the accessiblity skill. UAs actually assign the host of custom elements as display: contents so it's definitely pertinent to the web components skill as well.  I'm not sure I know where to squeeze this into layouts though. It almost feels like an anti-layout tool.

guides/css-layout/css-layout/guide.md
castastrophe: I might not relate it to ASCII art per-say, that seems like maybe a niche reference, but I added a note about the positive developer experience of aligning the content in readable rows and columns.

skills-drafts/css-layout/SKILL.md
castastrophe: Yes, great point. I get so excited by the prospect of baseline style queries!

skills-drafts/css-layout/SKILL.md
castastrophe: Yes good call, moved that up.

skills-drafts/css-layout/SKILL.md
```diff
@@ -82,184 +167,409 @@ Laying out items in a single dimension (row or column).
...
+    display: grid;
+    grid-template-columns: repeat(auto-fill, minmax(200px, 1fr));
+    grid-template-rows: masonry;
+    gap: 1rem;
+    columns: unset;
+  }
 }
 ```
+
+## 9 Resources
```
rviscomi: I think we can remove this. We typically don't include links to external resources.

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
-
 # CSS Layouts and Responsive Design
 
-This skill defines standard practices for modern CSS layouts, focusing on Flexbox, Grid, Container Queries, and intrinsic sizing.
-
-## 1. Modern Layout Fundamentals
-Leverage the browser's layout engine by using intrinsic sizing before resorting to hardcoded dimensions.
-
-### Layout Decision Matrix
+This skill defines standard practices for modern CSS layouts. Coverage focuses on flex-box, grid, subgrid, container queries, anchor positioning, intrinsic sizing, and modern viewport units. Each feature is annotated with its current Baseline status so that fallbacks can be applied only where they're actually needed — modern CSS rarely needs a polyfill, but it does need feature detection at the bleeding edges (e.g., subgrid, anchor positioning, masonry).
```
rviscomi: >Each feature is annotated with its current Baseline status so that fallbacks can be applied only where they're actually needed

I noticed the `BASELINE_STATUS` macro is used inconsistently. In many places it's not used at all, which could lead to stale statuses. Avoid hard-coding status info if possible.

skills-drafts/css-layout/SKILL.md
```diff
@@ -82,184 +167,409 @@ Laying out items in a single dimension (row or column).
...
+
+## 3 Grid and subgrid
+
+CSS grid facilitates the creation of 2-dimensional layouts where you can either define both rows and columns explicitly or let the engine derive them. This is a powerful tool for creating complex layouts and is a key part of the modern CSS layout system.
+
+Subgrid is a feature of CSS grid that allows you to create nested grids that inherit the tracks of their parent grid.
+
+The most common use cases for CSS grid are:
+
+- Defining regions of a page or section where you want the children elements to be distributed according to the defined layout settings (e.g. `grid-areas`).
```
rviscomi:
```suggestion
- Defining regions of a page or section where you want the children elements to be distributed according to the defined layout settings (e.g. `grid-template-areas`).
```

skills-drafts/css-layout/SKILL.md
```diff
@@ -82,184 +167,409 @@ Laying out items in a single dimension (row or column).
...
+  /* Stop scroll from chaining into the page on rubber-band gestures. */
+  overscroll-behavior: contain;
 }
 
 .snippet {
   display: -webkit-box;
   -webkit-line-clamp: 3;
   -webkit-box-orient: vertical;
-  overflow: hidden; /* Multi-line truncation */
+  /* Browsers will ignore the property if not yet supported. */
```
rviscomi: The guidance above says "feature-detect with `@supports`". Should we remove that part if it's harmlessly ignored?

rviscomi (commented):
Overall this is great. Left a few suggestions. Also tagging @LeaVerou if she has time to review.

I have one high-level comment about the file size. At 500+ lines, this is getting pretty long. I think the suggested limit is around 300 lines, which helps to conserve tokens and improve reliability. This is all good content and I don't think we should cut anything (except the Resources section, I left a comment on that) but there are a few things we could look out for to reduce the size:

- If we already have a dedicated use case-level guide for anything, avoid repeating the same best practices and code examples and use the `GUIDE_REF` macro to suggest the agent goes to read that guide.
- Or, if something _could_ work as a use case-level guide, let's break it out of the skill file and write a new guide for it.
- Avoid long prose or extraneous background info. Keeping the guidance short and sweet helps agents follow instructions more reliably.
- Split it up into two smaller guides if there's a clean breakpoint.

LeaVerou:
> Overall this is great. Left a few suggestions. Also tagging @LeaVerou if she has time to review.

Yup, will take a look tomorrow!

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
 ---
 name: css-layout
-description: Action-oriented guidelines for modern CSS layouts, including Flexbox, Grid, Container Queries, and Subgrid. Use this skill when architecting responsive UI components, page layouts, or component-driven designs.
+description: Modern CSS layouts such as flex-box, grid, subgrid, container queries, anchor positioning, and intrinsic sizing. Use this skill when architecting responsive UI components, page layouts, or component-driven designs.
```
LeaVerou: What are "component-driven designs"? The way I think about it, "Component-driven" is an architecture, not a design style, but I could be missing something. Or perhaps there's a clearer way to word what we're trying to convey here. Perhaps it's a reference to context-aware components?

skills-drafts/css-layout/SKILL.md
LeaVerou: Do we need this blurb? It seems to mainly be listing the features the guide discusses, which is bound to get stale, and is already covered by the way the RAG does retrieval + the ToC.
And the bit about baseline is already discussed  by the `modern-web` skill, which is read before we get here.

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
-This skill defines standard practices for modern CSS layouts, focusing on Flexbox, Grid, Container Queries, and intrinsic sizing.
-
-## 1. Modern Layout Fundamentals
-Leverage the browser's layout engine by using intrinsic sizing before resorting to hardcoded dimensions.
-
-### Layout Decision Matrix
+This skill defines standard practices for modern CSS layouts. Coverage focuses on flex-box, grid, subgrid, container queries, anchor positioning, intrinsic sizing, and modern viewport units. Each feature is annotated with its current Baseline status so that fallbacks can be applied only where they're actually needed — modern CSS rarely needs a polyfill, but it does need feature detection at the bleeding edges (e.g., subgrid, anchor positioning, masonry).
+
+[1 Fundamentals](#1-fundamentals)
+[1.1 Decision matrix](#11-decision-matrix)
```
LeaVerou: If we include a ToC, we should try to keep [information scent](https://www.nngroup.com/articles/information-scent/) high in the headings. "Decision matrix" tells me (or the agent) nothing about whether this content is relevant. I'd frame it around **the decision being made, rather than the methodology of making the decision**.

E.g. perhaps something like:

```suggestion
[1.1 Which layout mode to use?](#11-which-layout-mode-to-use)
```

(just an example to make the suggestion concrete, not necessarily the best wording)

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
+[1.2 Quick selection guide](#12-quick-selection-guide)
```
LeaVerou: Same above — _what_ is being selected?

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
-### Code Example: Shapes and Intrinsic Sizing
-```css
-.profile-pic {
-  float: left;
-  width: 150px;
-  height: 150px;
-  shape-outside: circle(50%); /* Text wraps around circle */
-}
-```
+| **flex-box** | 1D | Content distribution along a single axis | Content-first |
```
LeaVerou: Nit: "flexbox" seems to be more common than "flex-box"

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
+[7.1 Code example: viewport units and grid distribution](#71-code-example-viewport-units-and-grid-distribution)
+[8 Grid lanes (aka masonry)](#8-grid-lanes-aka-masonry)
+[8.1 Code example: multi-column as a masonry stand-in](#81-code-example-multi-column-as-a-masonry-stand-in)
+[9 Resources](#9-resources)
+
+## 1 Fundamentals
+
+Lean on the browser's layout engine when possible for better performance. Reach for intrinsic sizing, logical properties, and `aspect-ratio` before resorting to hardcoded dimensions or complicated media-queries.
+
+### 1.1 Decision matrix
```
LeaVerou: I’d merge 1.1, 1.2, and 1.3 into a single high-level decision tree. Currently the decision logic is split across three sections, making it hard to follow and reconcile the different parameters.

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
+
+1. **Is it a simple row OR column?** Use **flex-box**.
+2. **Is it a complex page structure with rows AND columns?** Use **grid**.
+3. **Does a nested element need to line up with the outer grid?** Use **subgrid**.
+4. **Is the content a long flow of prose that should split into balanced columns?** Use **multi-column**.
+5. **Are items different heights but need to be packed tightly?** Use a CSS Grid with `grid-auto-flow: dense` today; reach for native masonry only when it ships in your Baseline target (see [§3](#3-grid-and-subgrid)).
+
+### 1.3 Working principles
+
+- Use logical properties for content-focused layouts where the directional flow of the content should impact the order or directionality of the layout: `margin-inline`, `padding-block`, `inset-inline-start`, `border-inline-end`, etc. They flip automatically under `dir="rtl"` and vertical writing modes.
```
LeaVerou: We already have this guidance in the general CSS skill.

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
+| **multi-column** | 1D flow | Newspaper-style content reflow | Flow-first |
+| **grid-lanes** (aka masonry) | 2D | Packing items of varied sizes without dead space | Flow-first (limited availability) |
+
+### 1.2 Quick selection guide
+
+1. **Is it a simple row OR column?** Use **flex-box**.
+2. **Is it a complex page structure with rows AND columns?** Use **grid**.
+3. **Does a nested element need to line up with the outer grid?** Use **subgrid**.
+4. **Is the content a long flow of prose that should split into balanced columns?** Use **multi-column**.
+5. **Are items different heights but need to be packed tightly?** Use a CSS Grid with `grid-auto-flow: dense` today; reach for native masonry only when it ships in your Baseline target (see [§3](#3-grid-and-subgrid)).
```
LeaVerou: This seems too low-level for this section, I'd remove it. Also, it seems use-casey enough that perhaps we need a guide for it.

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
+
+### 1.3 Working principles
+
+- Use logical properties for content-focused layouts where the directional flow of the content should impact the order or directionality of the layout: `margin-inline`, `padding-block`, `inset-inline-start`, `border-inline-end`, etc. They flip automatically under `dir="rtl"` and vertical writing modes.
+- Match the display type to the dimensionality of the layout: `flex` for one-dimensional layouts, `grid` for two-dimensional layouts.
+- Apply the **content-first vs layout-first** mental model. Use flex-box when items determine the layout (content dictates flow). Use grid when you want to define the skeleton first (layout dictates placement).
+- Prefer **flex-box** for linear components (navbars, item lists, toolbars) and small components.
+- Prefer **grid** for two-dimensional page structures or complex component grids.
+- Use the `place-*` shorthands (`place-content`, `place-items`, `place-self`) to align across both axes simultaneously in a single declaration.
```
LeaVerou: This and the following bullets seem very distinct from the guidance around layout mode. I'd add subheadings, e.g.:

- 1.1 Overall layout (with the merged 1.1 + 1.2 + part of 1.3)
- 1.2 Sizing

LeaVerou: This is more valuable when the value is the same, otherwise its value is questionable: it saves a few characters but removes some clarity.

I'd also add an example here. E.g.

```suggestion
- Use the `place-*` shorthands (`place-content`, `place-items`, `place-self`) when alignment across both axes is the same. E.g. use `place-items: center` instead of `align-items: center; justify-items: center;`
```

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
+- Reach for **intrinsic sizing** (`min-content`, `max-content`, `fit-content()`) and flexible tracks (`fr`, `minmax()`) before fixed `width`/`height` — let content or track logic express the constraint where it can. This leads to more maintainable and responsive layouts with fewer media queries and less reliance on hardcoded dimensions.
```
LeaVerou: This is a really good argument. I'd make it even stronger:

```suggestion
- **DON'T** use fixed `width`/`height`/`inline-size`/`block-size`. They are fragile, and require more overrides to adapt to different conditions (viewport sizes, etc) which makes code less maintainable. Instead reach for content or context aware sizing, such as intrinsic sizing keywords (`min-content`, `max-content`, `fit-content()`), flexible grid tracks (`fr`, `minmax()`), or `flex` and its constituent properties. If you must set dimensions explicitly, prefer relative units (viewport-relative or font-size-relative, depending on what is likely to need fewer overrides) over absolute units like `px`.
```

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
+- Use `aspect-ratio` to reserve space for media and prevent layout shift before assets load.
+
+**Code example** - intrinsic sizing and `aspect-ratio`:
```
LeaVerou: I think in skills, it works better to keep code snippets minimal and close to the guidance they're illustrating (which also saves us the need from having to insert comments that repeat the guidance) rather than one big blob of code. And not every bit of guidance needs a code block, only those where the code helps illustrate something that the prose does not.

In this particular case, I think the only part that illustrates something the prose doesn't is the one about media. Though I wonder if it should be more specific than `.media`, e.g. you don't want that for `<audio>` elements.

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
+  display: flex;
+  /* Row is the main axis, allow items to wrap to the next line if the container is too narrow. */
+  flex-flow: row wrap;
+  gap: 1rem;
+  /* Evenly lay out items horizontally. */
+  justify-content: space-between;
+  /* Center items vertically. */
+  align-items: center;
+}
+```
```
LeaVerou: I'd remove this.
- It's not actionable. Deciding whether to use flexbox is covered by the decision tree in section 1, this section should be about how to use it well.
- The code example is just showing the basics of how flexbox works, which the agent already knows.

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
 
-### Code Example: Flexbox Component
+Be sure the container or its children have a definite size. When the flexbox container is meant to resize to contain its children and those children are also resizing to fill the container, the rendered result will likely lead to overflow issues and/or unexpected layout results. This is a common issue when using flexbox to create a responsive layout.
```
LeaVerou: This is golden. This is exactly the type of gotchas the skill should focus on, not exhaustive reference-style info that agents already have (for a feature as old as flexbox). I'd expand this a little more and add a little DO/DON'T code snippet.

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
+
+### 2.1 Available properties for the container
+
+Available properties and their usage (for the container element):
+
+- MANDATORY: `display: flex`: must be present on the container element to establish a flex context.
+- `flex-wrap`: sets whether the items should wrap to the next line if the container is too narrow. Valid values are `nowrap` (default), `wrap`, and `wrap-reverse`.
```
LeaVerou: Delete. The agent knows this.

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
+  - IMPORTANT: Use `flex-wrap: wrap` to prevent horizontal overflow on small viewports. When using `nowrap` without `overflow: hidden` or `overflow: auto`, the items will overflow the container.
+- `flex-direction`: sets the direction of the **main** axis. This property will determine the directionality of the `justify-content` and `align-items` properties. Valid values are `row` (default), `row-reverse`, `column`, and `column-reverse`.
+  - Note that `*-reverse` values reverse the directionality of the **main** axis, but not the **cross** axis.
+- `row-gap` and `column-gap`: sets the spacing between items along their respective axis. Valid values are a length, a percentage, or a calc() expression.
+- `justify-content`: sets the alignment of items along the **main** axis. Note that this directionality is dependent on the `flex-direction` property. Valid values are `flex-start` (default), `flex-end`, `center`, `space-between`, `space-around`, and `space-evenly`.
```
LeaVerou: Delete. The agent knows this. The guidance should be about WHEN to use these, rather than what properties exist and what values they take (unless they're new or it's targeted based on actual known agent weaknesses)

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
-### Code Example: Flexbox Component
+Be sure the container or its children have a definite size. When the flexbox container is meant to resize to contain its children and those children are also resizing to fill the container, the rendered result will likely lead to overflow issues and/or unexpected layout results. This is a common issue when using flexbox to create a responsive layout.
+
+### 2.1 Available properties for the container
+
+Available properties and their usage (for the container element):
+
+- MANDATORY: `display: flex`: must be present on the container element to establish a flex context.
+- `flex-wrap`: sets whether the items should wrap to the next line if the container is too narrow. Valid values are `nowrap` (default), `wrap`, and `wrap-reverse`.
+  - IMPORTANT: Use `flex-wrap: wrap` to prevent horizontal overflow on small viewports. When using `nowrap` without `overflow: hidden` or `overflow: auto`, the items will overflow the container.
```
LeaVerou: I'd remove `IMPORTANT:`. It's useful, but in the large scheme of things, it's not THE top priority. And it's not a universal fix, there are cases where `flex-wrap` produces very poor results, and it's better to just switch to `flex-flow: column` with a CQ.

That said, having **some** kind of strategy for what happens in small viewports _would_ warrant an "IMPORTANT" IMO. I.e. don't just specify `flex-flow: row` and yolo it.

Perhaps something like: 

```suggestion
  - IMPORTANT: When using a horizontal flex container, ensure items never overflow their container when narrow. For example, `flex-wrap: wrap`, a container or media query that sets `flex-flow: column`, or `overflow: auto` if scrolling would be acceptable.
```

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,78 +1,163 @@
...
+- `flex-direction`: sets the direction of the **main** axis. This property will determine the directionality of the `justify-content` and `align-items` properties. Valid values are `row` (default), `row-reverse`, `column`, and `column-reverse`.
+  - Note that `*-reverse` values reverse the directionality of the **main** axis, but not the **cross** axis.
+- `row-gap` and `column-gap`: sets the spacing between items along their respective axis. Valid values are a length, a percentage, or a calc() expression.
+- `justify-content`: sets the alignment of items along the **main** axis. Note that this directionality is dependent on the `flex-direction` property. Valid values are `flex-start` (default), `flex-end`, `center`, `space-between`, `space-around`, and `space-evenly`.
+  - IMPORTANT: When `flex-direction` is `row` (default), `justify-content` aligns items along the inline axis (left to right). When `flex-direction` is `column`, `justify-content` aligns items along the block axis (top to bottom). This is one of the most common gotchas when using flexbox and is often misinterpreted.
```
LeaVerou: This seems part of the definition of a main and a cross axis, rather than a gotcha? _Humans_ get it wrong, but I haven't seen agents get it wrong.

LeaVerou (changes requested):
I didn't look throughout the full skill because I kept hitting the same issues and I figured a repetitive review wouldn't be very useful.

First, echoing what @rviscomi said above around conciseness, `BASELINE_STATUS` macro over hardcoded info and not repeating guidance we have elsewhere.

I left some comments, but the higher level issue IMO is that a lot of this is **exhaustively enumerating syntax, reference-style**. What properties exist, what their values are, etc, without any info about what to do with them, WHEN to use each of these values and WHY — or when not to. Example from the flex section:

> - `flex-direction`: sets the direction of the **main** axis. This property will determine the directionality of the `justify-content` and `align-items` properties. Valid values are `row` (default), `row-reverse`, `column`, and `column-reverse`.
>   - Note that `*-reverse` values reverse the directionality of the **main** axis, but not the **cross** axis.
> - `row-gap` and `column-gap`: sets the spacing between items along their respective axis. Valid values are a length, a percentage, or a calc() expression.
> - `justify-content`: sets the alignment of items along the **main** axis. Note that this directionality is dependent on the `flex-direction` property. Valid values are `flex-start` (default), `flex-end`, `center`, `space-between`, `space-around`, and `space-evenly`.

Unless a feature is either rare or very new (and neither grid nor flexbox are as a whole, though parts are, e.g. subgrid), this type of thing is unnecessary. Agents are not starting from a blank slate. They _know_ all this info at an encyclopedic level. When they don't use e.g. flexbox well, it's not because they're not aware of what values `align-items` takes or what alignment properties exist, and just giving them this info will fix things. It's because despite having allll the info, **they lack the right decision framework** for when to use what, they're just pattern matching based on what code they've seen. That's where we come in: to give them that decision framework. Think **tips & course correction, not teaching from scratch**. 

There _are_ cases where it may make sense to document stuff reference-style, but it needs to be targeted, not an exhaustive attempt to document every relevant API, which would be a Sisyphean uphill battle and would dilute the actual _nuggets of wisdom_ we have in there. 

I suspect this will also automatically fix the length issue as well!

Caveat: The solution is **not** to go in and expand this even more by adding use cases in every single existing reference-style bullet — that would be forced and likely to produce contrived use cases like "Use `flex-flow: row` to lay items out horizontally" (not a real example) that would be a very small delta from reference material. We need to start the other way: what are the mistakes we're trying to correct? What newer features are agents underusing? What older features could they be using better? If we don't have a lot of content, that's ok (and better actually!) for the guide to be short for now. We don't need to pad it. I see us building up the guidance in these over time as well, as we spot more and more recurring patterns.

----

Re:splitting, I was surprised to see popovers here, it seems they may be better suited to something around interactivity or positioning, perhaps? Not sure. I can see the argument around including them in a layout guide post-hoc, it just surprised me at first.

skills-drafts/css-layout/SKILL.md
castastrophe: I tried to only include it in places where the feature was not fully baseline yet. Would you like to see it included in more places or fewer?

skills-drafts/css-layout/SKILL.md
castastrophe: Yes, great point.

skills-drafts/css-layout/SKILL.md
castastrophe: Yes this is a good point,I've reworked the introduction.

skills-drafts/css-layout/SKILL.md
castastrophe: Yes good point, I've pushed up a rework to try to simplify this significantly.

skills-drafts/css-layout/SKILL.md
castastrophe: Yes, +1. I've worked to simplify the TOC and document in the latest push.

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,265 +1,307 @@
...
 name: css-layout
-description: Action-oriented guidelines for modern CSS layouts, including Flexbox, Grid, Container Queries, and Subgrid. Use this skill when architecting responsive UI components, page layouts, or component-driven designs.
+description: Modern CSS layouts such as flexbox, grid, subgrid, container queries, anchor positioning, and intrinsic sizing. Use this skill when architecting responsive UI components or page layouts.
 ---
 
-
 # CSS Layouts and Responsive Design
 
-This skill defines standard practices for modern CSS layouts, focusing on Flexbox, Grid, Container Queries, and intrinsic sizing.
+[1 Fundamentals](#1-fundamentals)
```
jamesnw: I like the toc format, but find this formatting confusing- can it be in the same format as the CSS skill? I think that is just making this a numbered list. The CSS skill also includes subsections in the TOC, and doesn't number subsections, but I'm not sure how consistent they need to be.

(Sidenote for the people working on tooling- I think the tooling should generate this if it's useful, and something deterministic like how Sass uses `markdown-toc`, rather than an LLM - [example](https://github.com/sass/sass/blob/main/tool/update-toc.ts))

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,265 +1,307 @@
...
-| **Subgrid** | 2D (Inherited) | Grandchild alignment | Relationship-first |
-| **Masonry** | 1.5D / 2D | Eliminating dead space | Flow-first |
+### 1.1 Which layout mode to use?
 
-### Summary: Quick Selection Guide
-1. **Is it a simple row or column?** Use **Flexbox**.
-2. **Is it a complex page structure with rows AND columns?** Use **Grid**.
-3. **Does a nested element need to line up with the outer grid?** Use **Subgrid**.
-4. **Are items different heights but need to be packed tightly?** Use **Masonry**.
+Walk the decision tree top-to-bottom and stop at the first match:
```
jamesnw: Is it worth specifying this is on an element, not the whole page? Or that sections of the page can use different layouts? Or does the agent already know this?

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,265 +1,307 @@
...
-  shape-outside: circle(50%); /* Text wraps around circle */
-}
-```
+**Do:**
 
-```css
-.sidebar {
-  width: max-content; /* Size to fit its longest word/content */
-}
+- Use logical properties (`margin-inline`, `padding-block`, `inset-inline-start`, `border-inline-end`) so layouts flip automatically under `dir="rtl"` and vertical writing modes.
```
jamesnw: Are there enough logical values (`position-area: block-end`) and units (`dvb`) to warrant mention here?

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,265 +1,307 @@
...
+- Use `aspect-ratio` to reserve space for media and prevent layout shift before assets load.
 
-.main-content {
-  width: fit-content; /* Expand up to available space but no further */
-}
+```css
+.sidebar       { inline-size: max-content; }    /* Size to longest unbreakable token. */
+.main-content  { inline-size: fit-content; }    /* Grow to available space, no further. */
+.media         { aspect-ratio: 16 / 9; inline-size: 100%; block-size: auto; }
+.centered-card { display: grid; place-content: center; min-block-size: 100dvh; }
```
jamesnw:
```suggestion
.center-card { display: grid; place-content: center; min-block-size: 100dvb; }
```

I'm not certain if it's an improvement, but `centered-card` made me think the targeted element is centered, not that it has a centered card.

Also, logical properties need logical units (learned that one when something didn't work during a livestream 😊 )

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,265 +1,307 @@
...
-Querying the size of a parent container rather than the viewport for component-level adaptability.
+## 4 Container queries
+
+Query the size (or computed style) of an ancestor container rather than the viewport. Mental model: container queries = component context; media queries = global page layout and user preferences (`prefers-color-scheme`, `prefers-reduced-motion`).
+
+**Do:**
+
+- Establish a containment context with `container-type: inline-size` (width-only queries) or `container-type: size` (both axes) on a wrapper before its descendants can be queried.
+- Name containers with `container-name` (or the `container` shorthand: `container: inline-size card`) when nested contexts could collide.
+- Size fluid type and spacing with container query units: `cqi`/`cqb` (logical inline/block), `cqw`/`cqh` (physical), `cqmin`/`cqmax`.
```
jamesnw:
```suggestion
- Include container query units in calculating fluid type and spacing: `cqi`/`cqb` (logical inline/block), `cqw`/`cqh` (physical), `cqmin`/`cqmax`.
```

Optional nit- makes it clearer that you shouldn't _just_ use container query units.

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,265 +1,307 @@
...
+
+**Do not:**
 
-### DOs and DON'Ts
-- **DO** use `container-type: inline-size` on a wrapper to enable width-based queries for children.
-- **DO** use the full set of container units: `cqi` (inline), `cqb` (block), `cqw` (width), `cqh` (height), `cqmin`, and `cqmax`.
-- **DO** remember that container units default to standard viewport units (`svw`/`svh`) if no container parent is found.
-- **DON'T** read layout size inside a query that modifies the container itself (infinite loop risk).
+- Don't use `block-size` as a `container-type` value — it isn't valid. Use `size` for both axes.
+- Don't let children's intrinsic size influence the container after declaring `container-type`. The container is computed as if it has no children once containment is active.
```
jamesnw:
```suggestion
- Don't expect children's intrinsic size to influence the container after declaring `container-type`. The container is computed as if it has no children once containment is active.
```

I don't think you _can_ influence it, or am I misunderstanding?

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,265 +1,307 @@
...
-/* Adapt child based on parent width */
-@container (min-width: 400px) {
+@container card (inline-size > 400px) {
   .card-content {
     display: flex;
     gap: 2rem;
   }
 }
 
 .item-title {
```
jamesnw:
```suggestion
.card-title {
```

guides/css-layout/css-layout/guide.md
```diff
@@ -1,265 +1,307 @@
...
-| Feature | Container Queries (`@container`) | Media Queries (`@media`) |
-|---|---|---|
-| **Reference** | Nearest containment parent | Global Viewport |
-| **Reusability** | ✅ High | ❌ Low |
-| **Typical Use Case** | Adaptive cards, sidebar widgets | Page layout grids, Navigation breakpoints |
+**Do:**
 
-**Single-Sentence Mental Model**: "Container Queries = Component context, Media Queries = Global page layout."
+- Use the `popover` attribute for menus, tooltips, and other non-modal flyouts — popovers live in the top layer, so layering doesn't need `z-index` management. Values: `auto` (light-dismissable, one open at a time), `manual` (programmatic only, multiple can coexist), `hint` (like `auto` but layers above an open `auto` popover — for hover/focus tooltips).
+- Use `<dialog>` with `.showModal()` for modal interactions that require focus trapping and an inert backdrop.
```
jamesnw, line 216: Is it worth saying here (and in the intro) when to use non-modal dialogs?

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,265 +1,307 @@
...
+
+### 5.1 Code example: popover with anchor positioning and fallback
 
-### Code Example: Popovers and Anchors
 ```css
-/* Tethering a popover to match trigger width */
-#trigger { 
-  anchor-name: --my-anchor; 
+#trigger {
+  anchor-name: --trigger;
```
jamesnw: Perhaps too indepth, but this likely isn't needed- most use cases end up opening a popover with an implicit anchor.

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,265 +1,307 @@
...
-- **DO** use `-webkit-line-clamp` combined with `display: -webkit-box` for multi-line truncation.
-- **DO** use `touch-action` (e.g., `pan-y` or `pan-x`) to optimize scrolling interactions on mobile.
+- Use `overflow: auto` so scrollbars appear only when content actually overflows.
+- Use `overflow: clip` to clip content **without** establishing a scroll container; opt into spillover with `overflow-clip-margin`.
+- Use `scrollbar-gutter: stable` to reserve space for scrollbars and prevent layout shifts when content grows. {{ BASELINE_STATUS("scrollbar-gutter") }}
+- Use `overscroll-behavior: contain` (or `none`) on scrollable containers to stop scroll chains from bubbling into the parent or document.
+- Use the `-webkit-line-clamp` + `display: -webkit-box` + `-webkit-box-orient: vertical` triad for multi-line truncation — despite the prefix, this pattern is fully specified and not deprecated. Declare the unprefixed `line-clamp` shorthand alongside it; browsers that don't yet support it ignore the property harmlessly. {{ BASELINE_STATUS("line-clamp") }}
+- Use `touch-action` (e.g., `pan-y`, `pan-x`, `none`) to optimize gesture handling on touch surfaces.
```
jamesnw:
```suggestion
```

The CSS skill goes into more depth [here](https://github.com/GoogleChrome/guidance/blob/1e855aebbdf08d2397fa83cadab55f977c0bb229/guides/css/css/guide.md?plain=1#L224), and I don't think it's really part of layout.

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,265 +1,307 @@
...
+  overscroll-behavior: contain;   /* No scroll chaining into the page. */
 }
 
 .snippet {
   display: -webkit-box;
   -webkit-line-clamp: 3;
   -webkit-box-orient: vertical;
-  overflow: hidden; /* Multi-line truncation */
+  line-clamp: 3;                  /* Ignored where unsupported. */
+  overflow: hidden;
```
jamesnw:
```suggestion
  overflow: clip;
```
I think this follows the guidance above?

skills-drafts/css-layout/SKILL.md
```diff
@@ -1,265 +1,307 @@
...
-.stretching-grid {
-  display: grid;
-  grid-template-columns: repeat(auto-fit, minmax(200px, 1fr)); /* Fill full container */
-}
+## 8 Grid lanes (aka masonry)
 
-.filling-grid {
-  display: grid;
-  grid-template-columns: repeat(auto-fill, minmax(200px, 1fr)); /* Keep track gaps */
+The spec is in development. The currently agreed-upon name is "grid lanes" (e.g., `display: grid-lanes`). Firefox ships `grid-template-rows: masonry` behind a flag; no other engines ship it in stable as of this writing.
```
jamesnw: Given this, I'm wondering if grid lanes should be omitted for now, as it really isn't actionable yet?

jamesnw (changes requested):
A few nits, but this is shaping up well!

skills-drafts/css-layout/SKILL.md
castastrophe: So I opted to keep this in as a brief mention because it does solve a very common layout use-case and there is almost no polyfill without JS that solves for this requirement. Hope that makes sense!

skills-drafts/css-layout/SKILL.md
castastrophe: I updated this to be just .title because I think that demonstrates the point even better.

guides/css-layout/css-layout/guide.md
castastrophe: I ended up linking off to the more descriptive usage guides instead.

jamesnw (approved):
I think this is looking good!

guides/css-layout/css-layout/guide.md
```diff
@@ -1,265 +1,304 @@
...
-- **DO** use **Subgrid (`grid-template-rows: subgrid`)** to align nested children to the parent grid definition.
-- **DO** use **Subgrid** to solve the "ragged edge" problem in components like card lists where internal elements (like titles) must line up perfectly across columns.
+- Establish a context with `display: grid`.
+- Use `grid-template-areas` for complex page-level layouts — area names are self-documenting and the declaration can be aligned in rows and columns for at-a-glance readability.
+- Use `repeat(auto-fit, minmax(200px, 1fr))` for responsive card grids that stretch filled tracks to fill the row, or `auto-fill` to preserve empty repeated tracks at their min size.
+- Use `fr` for proportional track distribution and `minmax(min, max)` for flexible-but-bounded tracks.
+- Position items with `grid-column: span <n>` to size across tracks, `grid-column: <start> / <end>` to place at specific lines, or `grid-area: <name>` for named regions.
+- Use subgrid (`grid-template-columns: subgrid` or `grid-template-rows: subgrid`) to solve the "ragged edge" problem in card lists — internal elements like titles, metadata, and CTAs line up across siblings.
```
LeaVerou, line 118: Great use case. We should have a guide for this!

guides/css-layout/css-layout/guide.md
```diff
@@ -1,265 +1,304 @@
...
-  grid-span: 3;
+  /* Same-cascade fallback: ignored when subgrid is supported. */
+  grid-template-rows: auto 1fr;
+  grid-template-rows: subgrid;
 }
 ```
 
-## 4. Container Queries and Advanced Sizing
-Querying the size of a parent container rather than the viewport for component-level adaptability.
+## 4 Container queries
```
LeaVerou, line 166: The title is around the solution instead of the problem, which makes it less likely that agents that need it will pick it from the ToC. 

Maybe "Responsive sizing" or  "Context-aware sizing" or "Adapting to different sizes" or something?

guides/css-layout/css-layout/guide.md
```diff
@@ -1,265 +1,304 @@
...
+
+- Establish a containment context with `container-type: inline-size` (width-only queries) or `container-type: size` (both axes) on a wrapper before its descendants can be queried.
+- Name containers with `container-name` (or the `container` shorthand: `container: inline-size card`) when nested contexts could collide.
+- Include container query units in calculating fluid type and spacing: `cqi`/`cqb` (logical inline/block), `cqw`/`cqh` (physical), `cqmin`/`cqmax`.
+- Give the container a definite `block-size` whenever `container-type: size` is used — without one, descendants collapse because size containment forces the container to ignore its content.
+
+**Do not:**
+
+- Don't use `block-size` as a `container-type` value — it isn't valid. Use `size` for both axes.
+- Don't expect children's intrinsic size to influence the container after declaring `container-type`. The container is computed as if it has no children once containment is active.
```
LeaVerou, line 182:
```suggestion
- Don't expect children's intrinsic size to influence the container after declaring `container-type: inline-size` or `container-type: size`. The container is computed as if it has no children once containment is active.
```

(other `container-type` values, e.g. `scroll-state` don't have this property)

guides/css-layout/css-layout/guide.md
```diff
@@ -1,265 +1,304 @@
...
+- Don't rely on container query units inside descendants of a non-qualifying ancestor; they fall back to the small viewport (`svw`/`svh`).
+
+### 4.1 Code example: fluid typography using container query units
 
-### Code Example: Container Responsive Component
 ```css
-/* Define container */
 .card-wrapper {
-  container-type: inline-size;
+  container: inline-size / card; /* shorthand for container-type + container-name */
```
LeaVerou, line 189:
```suggestion
  container: inline-size / card;
```

The comment is not necessary, and we mention it above anyway.

skills-src/css-layout/SKILL.md
```diff
@@ -1,265 +1,304 @@
...
-.item-title {
-  font-size: 4cqi; /* Fluid based on container size */
+.title {
+  /* Fluid type bound to the container width, not the viewport. */
+  font-size: clamp(1rem, 4cqi, 2rem);
 }
 ```
 
-### Responsive Triggers Decision Matrix
+> For component-driven responsive styling patterns, see [Size-aware styling](../../guides/user-experience/size-aware-styling/) and [Fluid scaling](../../guides/user-experience/fluid-scaling/).
```
LeaVerou: Use `GUIDE_REF`, not inline links:

```suggestion
> For component-driven responsive styling patterns, see {{ GUIDE_REF("size-aware-styling") }} and {{ GUIDE_REF("fluid-scaling") }}.
```

guides/css-layout/css-layout/guide.md
```diff
@@ -1,265 +1,304 @@
...
 
-### DOs and DON'Ts
-- **DO** use the `popover` attribute for menus and tooltips (non-modal content).
-- **DO** use `<dialog>` and `.showModal()` for modal interactions that require focus entrapment.
-- **DO** use `anchor-size()` to size popovers dynamically relative to their anchor dimensions.
-- **DO** use `position-try-fallbacks: flip-block` to let the browser automatically reposition when an overlay overflows viewport edges.
-- **DO** use `pointer-events: none` to let clicks pass through decorative overlays or ignore inert layers.
+- Use `popover` for transient, non-modal UI (flyouts, toasts, tooltips) — lives in the top layer, no `z-index` management needed.
+- Use `<dialog>` with `.showModal()` for modal interactions that require focus trapping and an inert backdrop.
+- Don't combine `popover` and `.showModal()` on the same element — they're mutually exclusive runtime states.
```
LeaVerou, line 217: We have this in the HTML guide: https://github.com/GoogleChrome/guidance/blob/main/guides/html/html/guide.md#4-native-overlays-dialogs-and-popovers

skills-src/css-layout/SKILL.md
```diff
@@ -1,265 +1,304 @@
...
+- Don't mix physical and logical keywords in a single `position-area` value — pick one coordinate system.
+- Feature-detect with `@supports (anchor-name: --x)` and provide an absolute-position fallback.
 
-| Feature | Popover API (`popover`) | `<dialog>` |
-|---|---|---|
-| **Modality** | Non-modal (Doesn't block background) | Modal (Blocks background via `.showModal()`) |
-| **Focus Trapping** | ❌ Natural tab flow | ✅ Locks focus inside dialog |
-| **Light-Dismiss** | ✅ Auto-closes on click-outside / ESC | ❌ Requires JS / Form submit |
+> For full implementation detail, polyfill strategies, and `popover` value reference, see [Declarative dialog and popover control](../../guides/user-experience/declarative-dialog-popover-control/) and [Position-aware tooltips](../../guides/user-experience/position-aware-tooltips/). For anchor positioning applied to menus and tab indicators, see [Resilient context menus and nested dropdowns](../../guides/user-experience/resilient-context-menus-and-nested-dropdowns/) and [Anchor positioning tab underline](../../guides/user-experience/anchor-positioning-tab-underline/).
```
LeaVerou:
```suggestion
For full implementation detail, polyfill strategies, and `popover` value reference, see {{ GUIDE_REF("declarative-dialog-popover-control") }} and {{ GUIDE_REF("position-aware-tooltips") }}. For anchor positioning applied to menus and tab indicators, see {{ GUIDE_REF("resilient-context-menus-and-nested-dropdowns") }} and {{ GUIDE_REF("anchor-positioning-tab-underline") }}.
```

Though there's also no need to have these all together — I'd move them to be closer to the content they're about.

guides/css-layout/css-layout/guide.md
```diff
@@ -1,265 +1,304 @@
...
+- Use `overscroll-behavior: contain` (or `none`) on scrollable containers to stop scroll chains from bubbling into the parent or document.
+- Use the `-webkit-line-clamp` + `display: -webkit-box` + `-webkit-box-orient: vertical` triad for multi-line truncation — despite the prefix, this pattern is fully specified and not deprecated. Declare the unprefixed `line-clamp` shorthand alongside it; browsers that don't yet support it ignore the property harmlessly.
+**Do not:**
 
-### DON'Ts
-- **DON'T** use `overflow: scroll` habitually as it forces scrollbars even if they aren't needed.
-
-### Code Example: Layout Stability
+- Don't use `overflow: scroll` when `auto` will do — `scroll` forces scrollbars even when there's nothing to scroll.
+- Don't reach for `overflow: hidden` when you only want to clip — `hidden` establishes a scroll container that can be programmatically scrolled.
```
LeaVerou, line 245: I'd move these up and pair them with the instruction they're about. 

Instead of having a list of DOs and a list of DON'Ts, it works better to have pairs of do this, don't do that.

skills-src/css-layout/SKILL.md
```diff
@@ -1,265 +1,304 @@
...
   display: -webkit-box;
   -webkit-line-clamp: 3;
   -webkit-box-orient: vertical;
-  overflow: hidden; /* Multi-line truncation */
+  line-clamp: 3;                  /* Ignored where unsupported. */
+  overflow: clip;
 }
 ```
 
+> For `overflow: clip` and `overflow-clip-margin` in depth, see [Overflow clipping control](../../guides/user-experience/overflow-clipping-control/). For scrollbar color, sizing, and theming, see [Customize scrollbar color and thickness](../../guides/user-experience/customize-scrollbar-color-and-thickness/), [Adapt scrollbar to light/dark preferences](../../guides/user-experience/adapt-scrollbar-to-light-dark-preferences/), and [Adapt scrollbar to contrast preferences](../../guides/user-experience/adapt-scrollbar-to-contrast-preferences/).
```
LeaVerou: Same as above: use `GUIDE_REF`, move to relevant sections.

guides/css-layout/css-layout/guide.md
```diff
@@ -1,265 +1,304 @@
...
-.stretching-grid {
-  display: grid;
-  grid-template-columns: repeat(auto-fit, minmax(200px, 1fr)); /* Fill full container */
-}
+> For the full viewport unit reference (`svh`, `lvh`, `dvi`, `dvb`, etc.), see the [CSS guide](../../guides/css/css/guide.md).
 
-.filling-grid {
-  display: grid;
-  grid-template-columns: repeat(auto-fill, minmax(200px, 1fr)); /* Keep track gaps */
+## 8 Grid lanes (aka masonry)
```
LeaVerou, line 275: Not a strong opinion, but given the browser support and the specificity of this, I'd lean towards moving this content to a separate use case guide.

LeaVerou (approved):
Left some comments (top priority one I'd say is the one about guide refs) and suggestions but overall this looks _great_! Fantastic signal-to-noise ratio, night and day from the earlier version! 😍

rviscomi (approved):
LGTM! +1 to what @LeaVerou said, this is looking really good @castastrophe!

I pushed a couple of changes to address Lea's suggestions to use `GUIDE_REF` and to move this to the `guides/` dir 🚀

PR #679, opened 2026-04-29
improve output of BASELINE_STATUS macro

connorjclark:
Previous:

> Popover is Newly. It's been Baseline since 2025-01-27.

Now:

> Baseline status for Popover: Newly available. It's been Baseline since 2025-01-27.

### DIFF
--- a/serving/lib/macros.test.ts
+++ b/serving/lib/macros.test.ts
@@ -7,7 +7,8 @@ describe('replaceMacros (Functional with real data)', () => {
     it('replaces macro with widely available status', () => {
       const content = '{{ BASELINE_STATUS("grid") }}';
       const result = replaceMacros(content, 'test.md');
-      assert.ok(result.includes('2017-10-17') && result.includes('Widely available'));
+      assert.ok(result.includes('2017-10-17'));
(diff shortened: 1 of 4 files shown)


serving/lib/macros.test.ts
```diff
@@ -20,7 +20,7 @@ describe('replaceMacros (Functional with real data)', () => {
       // Accelerometer is typically limited
```
paulirish, line 21: weird... look at line 10. 

https://github.com/GoogleChrome/guidance/blob/baseline-macro/serving%2Flib%2Fmacros.test.ts#L7-L10

do you know what that was already "widely available" and not just widely?

doesnt seem to jive with the rest of this.

paulirish (approved):
thank you!  so much better

serving/lib/macros.test.ts
connorjclark: knowing absolutely nothing about any of this code, i'll just point out that assertion has an `||`. maybe the other thing is true.

paulirish: good call

PR #695, opened 2026-04-30
Change modern-web CLI to use positional arguments for search and retrieve

connorjclark:
b/508297009

### DIFF
--- a/harness/agents/claude-code-agent.ts
+++ b/harness/agents/claude-code-agent.ts
@@ -160,8 +160,8 @@ export async function collectClaudeGuidesFromTrajectory(dirPath: string, _servin
             for (const contentItem of obj.message.content) {
               if (contentItem.type === 'tool_use' && contentItem.name === 'Bash' && contentItem.input && contentItem.input.command) {
                 const command = contentItem.input.command;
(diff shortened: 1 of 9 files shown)


harness/agents/claude-code-agent.ts
```diff
@@ -160,8 +160,8 @@ export async function collectClaudeGuidesFromTrajectory(dirPath: string, _servin
             for (const contentItem of obj.message.content) {
               if (contentItem.type === 'tool_use' && contentItem.name === 'Bash' && contentItem.input && contentItem.input.command) {
                 const command = contentItem.input.command;
-                if (command.includes('modern-web') && command.includes('--retrieve')) {
-                  const match = command.match(/--retrieve\s+["']?([^"'\s]+)["']?/);
```
paulirish, line 164: we'll need these changes in harness/agents/* to match on both old and new.  (sry)

because sometimes these process historical results.

paulirish (commented):
lovely

paulirish (approved):
lovely

harness/agents/claude-code-agent.ts
connorjclark: done

PR #699, opened 2026-04-30
prepare for usage of npx and publish to npm

connorjclark:
ref b/508297009

This PR prepares the codebase for publishing the `modern-web-guidance` package to NPM by introducing a specialized build distribution and enabling the corresponding GitHub Actions workflow.

There is no impact on `GoogleChrome/modern-web-guidance` for now. Eventually, we will update the SKILL.md shipped there to use `npx`. We don't now b/c npm will only have a subset of guides until we are ready to go public.

* Updated the build process to generate a new `dist/skills-cli-npx` directory. This dist folder is specifically tailored for npx usage, replacing local execution paths with `npx -p modern-web-guidance@latest`.
* Added a --subset flag to the guide builder. This allows us to publish just a subset to npm for now.
* Enabled the OIDC-based NPM publishing step in the publish.yml workflow, targeting the new skills-cli-npx distribution.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+      - run: git fetch --tags
 
       - name: Setup pnpm
         uses: pnpm/action-setup@v5
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
+      - run: git fetch --tags
 
       - name: Setup pnpm
         uses: pnpm/action-setup@v5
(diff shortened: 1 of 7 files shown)


paulirish (commented):
can you ask yr agent to have updateReadmeWithFeaturesAndUseCases be built from build/guides  ?
that way our npm readme doesnt spill the beans on our guide coverage

otherwise, nice. 

https://www.npmjs.com/package/modern-web-guidance

connorjclark:
I can't untangle why copySkills test fails. works fine locally.  https://github.com/GoogleChrome/guidance/actions/runs/25195170788/job/73874022165?pr=699#step:6:365

The last 3 commits could probably be dropped as they didn't solve the issue.

connorjclark:
needed to fetch tags in test workflow, as now the build script also grabs the latest tag via git

paulirish (approved):

Issue #227, opened 2026-03-06
Create guide and evals for the component-specific-light-dark-theme use case


paulirish:
Force certain elements to be in light mode or dark mode (e.g. code blocks, media players, etc) independently of the page's color-scheme.

Affected web-feature IDs: [color-scheme](https://webstatus.dev/features/color-scheme), [light-dark](https://webstatus.dev/features/light-dark)

Use case subdir: [guides/ui-atoms/component-specific-light-dark-theme](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-atoms/component-specific-light-dark-theme)

Related features: #42 #62

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [x] Use case guidance (guide.md)
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

LeaVerou:
## SME review

### P0: Missing gotcha that can produce inaccessible UIs

There is a big gotcha when it comes to using `light-dark()` in inherited color properties (such as `color`), including registered custom properties, that IMO should be mentioned here: `light-dark()` **inherits as the computed value, i.e. one of the two colors**. It does not retain its dynamic switching. Unless this is cautioned against, it's very easy to happen by accident as authors see registering properties as a universally Good Thing™. It can be especially dangerous when `light-dark()` in registered custom properties or `color` is combined with unregistered ones or system colors as you can end up with completely inaccessible UIs. 

For more details, see:
- CSS WG issue: https://github.com/w3c/csswg-drafts/issues/13836
- Demo 1: https://codepen.io/leaverou/pen/gbwNvKB?editors=1100
- Demo 2: https://codepen.io/leaverou/pen/raMEdPv?editors=1100

### P1: Describes the HOW but not the WHY/WHEN (use cases)

The guide is more of a tutorial around _how_ to apply component-specific light/dark themes, but does not really describe the **use cases** around WHY/WHEN you'd want to do that. When should you use `color-scheme: dark` and when should you just apply different colors? The code examples only show a somewhat contrived `.force-light` or `[data-theme="dark"]`. 

E.g. if your primary buttons are white with blue text in dark mode and blue with white text in light mode should you define them like this:

```css
button.primary {
	background: light-dark(blue, white);
	color: light-dark(white, blue);
}
```

or like this:

```css
button.primary {
	background: light-dark(white, blue);
[...]

PR #701, opened 2026-05-01
add light-dark mention

### LINKED ISSUE #238: Create guide and evals for the browser-ui-color-theme use case
Configure built-in browser UI (e.g. scrollbars, form controls, etc) to respect the user's light/dark theme preference.

Affected web-feature IDs: [color-scheme](https://webstatus.dev/features/color-scheme), [prefers-color-scheme](https://webstatus.dev/features/prefers-color-scheme), [light-dark](https://webstatus.dev/features/light-dark), [accent-color](https://webstatus.dev/features/accent-color)

Use case subdir: [guides/user-experience/browser-ui-color-theme](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/user-experience/browser-ui-color-theme)

Related features: #42 #62 #37

**Required files:**
- [x] demo.html
- [x] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

dvdherron:
Closes #238


guides/user-experience/browser-ui-color-theme/guide.md
```diff
@@ -62,7 +93,9 @@ You can override the global theme for specific elements. This is useful for "dar
 ## Fallback strategies
 
 {{ BASELINE_STATUS("color-scheme") }}
+{{ BASELINE_STATUS("light-dark") }}
 
-- **Progressive Enhancement**: Browsers that do not support `color-scheme` will simply ignore the property and use their default light-mode UI.
+- **Progressive Enhancement**: Browsers that do not support `color-scheme` or `light-dark()` will simply ignore these properties and use their default light-mode UI.
+- **Handling `light-dark()` Support**: For browsers that support `color-scheme` but not yet `light-dark()`, provide a simple fallback color before the `light-dark()` declaration, or continue using `@media (prefers-color-scheme)` for critical theme overrides.
```
LeaVerou: In most real-world cases colors will be stored in variables, so providing a simple fallback won't work, sadly… `light-dark()` is very annoying to fallback, IIRC it's MQs + `@supports`, unfortunately. There was a good discussion on maintainability of fallback code in the `light-dark()` guide discussion issue.

guides/user-experience/browser-ui-color-theme/guide.md
```diff
@@ -53,6 +56,34 @@ You can override the global theme for specific elements. This is useful for "dar
 }
 ```
 
+### 4. Fine-grained control with `light-dark()`
+For more control over the colors of built-in UI such as `accent-color` or `scrollbar-color`, use the `light-dark()` function. This function automatically picks the correct color based on the computed `color-scheme` of the element, eliminating the need for redundant media queries.
```
LeaVerou:
```suggestion
For more control over the colors of built-in UI such as `accent-color` or `scrollbar-color`, use the `light-dark()` function. This function automatically picks the correct color based on the computed `color-scheme` of the element and eliminates the need for redundant media queries.
```

Nit: These are separate, orthogonal benefits. As written, it sounds (to me) like the primary benefit is eliminating the need for MQs. That's a QoL nice-to-have. The primary benefit is locality, that these respond to element color-scheme.

Also, perhaps add some guidance not to register this property? Otherwise you'd just get the root value regardless of the element color-scheme.

LeaVerou (changes requested):
Left some comments, LGTM otherwise!

guides/user-experience/browser-ui-color-theme/guide.md
```diff
@@ -62,7 +93,9 @@ You can override the global theme for specific elements. This is useful for "dar
 ## Fallback strategies
 
 {{ BASELINE_STATUS("color-scheme") }}
+{{ BASELINE_STATUS("light-dark") }}
 
-- **Progressive Enhancement**: Browsers that do not support `color-scheme` will simply ignore the property and use their default light-mode UI.
+- **Progressive Enhancement**: Browsers that do not support `color-scheme` or `light-dark()` will simply ignore these properties and use their default light-mode UI.
+- **Handling `light-dark()` Support**: For browsers that support `color-scheme` but not yet `light-dark()`, provide a simple fallback color before the `light-dark()` declaration, or continue using `@media (prefers-color-scheme)` for critical theme overrides.
 - **Manual Dark Mode Styling**: For older browsers, continue to use `prefers-color-scheme` media queries to provide custom styles for your own components.
```
malchata, line 177: Given @LeaVerou's comment just above, would it be best then to lead with this as the fallback? But it is super annoying to do. Is it tolerable to fallback to no color scheme management styling in the absence of `light-dark()`?

LeaVerou: The `color-mix()` fallback I mentioned [here](https://github.com/GoogleChrome/guidance/issues/227#issuecomment-4308967648) could also work (and keeps them together). 
But given the support for `light-dark()` (Baseline 2024, [90% of users](https://caniuse.com/?search=light-dark)), just the light mode color seems like a perfectly acceptable fallback to me. Others' MMV though!

guides/user-experience/browser-ui-color-theme/guide.md
dvdherron: Thanks. I worked on cleaning this up and making the fallback path more clear.

guides/user-experience/browser-ui-color-theme/guide.md
dvdherron: Updated the guidance here and added to the Best Practices section with guidance about registering tokens that rely on light-dark(). Also added examples of the correct and wrong way to use light-dark()/inheritance.

guides/user-experience/browser-ui-color-theme/guide.md
dvdherron: I showed an example of using light-dark without `prefers-color-scheme` in the implementation section and showed a version using it in the fallback. If you think that's overload for the fallback example @LeaVerou , I can remove it.

guides/user-experience/browser-ui-color-theme/guide.md
```diff
@@ -45,15 +48,97 @@ MANDATORY: Apply the `color-scheme` property to the `html` element or the `:root
 You can override the global theme for specific elements. This is useful for "dark mode" sections within a light-themed site (e.g., a dark sidebar or a code editor).
 
 ```css
-.dark-sidebar {
+.contact-form {
```
LeaVerou: Why would we want a contact form to be always in dark mode though? Unless this is a thing and I just haven't come across it?

The code example is quite good, since it's quite common to have code snippets in effectively be in dark mode even on light mode websites. So perhaps:

```suggestion
pre, code {
```

(Technically `pre` is not just for code, but `pre:has(>code)` feels a bit overkill…)

dvdherron: This makes sense. Changed the code examples and synced the expectations and demo in https://github.com/GoogleChrome/guidance/pull/701/commits/aac033af2d6e66d092e0f3e5b134d389010429c2

guides/user-experience/browser-ui-color-theme/guide.md
```diff
@@ -45,15 +48,103 @@ MANDATORY: Apply the `color-scheme` property to the `html` element or the `:root
...
+  accent-color: var(--accent-color);
+  /* Set colors for other theme styles */
+  background-color: var(--background-color);
+  color: var(--text-color);
 }
 ```
 
 ## Best Practices
+  - **Token Resolution and Inheritance**: Be aware that `light-dark()` currently resolves based on the `color-scheme` of the element where the value is **defined**, not where it is **used**.
+  - **The "Inheritance Footgun"**: If you define a theme variable on the `:root`, the browser resolves it to a specific color immediately. Descendant elements that change their `color-scheme` (e.g., a dark-themed card on a light page) will inherit the *already-resolved color* instead of the dynamic function.
```
LeaVerou: See comments about this on the other PR — this is only true if the property is registered! So NOT registering design tokens fixes this. It can still be present for built-in inherited properties, but at least that's a much narrower scope (basically `color` and `accent-color`)

LeaVerou (changes requested):
See comment about the inheritance issue!

guides/user-experience/browser-ui-color-theme/guide.md
LeaVerou: I have no strong opinion on this either way! 🤷🏽‍♀️

guides/user-experience/browser-ui-color-theme/guide.md
dvdherron: Thanks for the clarification. Tried to make this more clear with updates in this section and some of the code examples. Ready for another look.

guides/user-experience/browser-ui-color-theme/guide.md
```diff
@@ -45,15 +48,101 @@ MANDATORY: Apply the `color-scheme` property to the `html` element or the `:root
...
+  /* Set colors for other theme styles */
+  background-color: var(--background-color);
+  color: var(--text-color);
 }
 ```
 
 ## Best Practices
+  - The `@property` Risk **:
+    - If you define a theme variable with registered custom properties (via `@property`), the browser resolves it to a specific color immediately. Descendant elements that change their `color-scheme` (e.g., a dark-themed card on a light page) will inherit the *already-resolved color* instead of the dynamic function. 
+    - Since theme variables defined with `@property` have a defined `syntax: "<color>"`, they "lock" the resolved color at computed value time, making them non-reactive to local theme changes. **Mandatory**: Do not register properties meant to be design tokens that dynamically switch based on `light-dark()`. If you need to animate a color variable, use a separate property.
```
LeaVerou: Nit, non-blocking: I'd put the guidance on what to do *first* (as the most important bit) and the explanation of why after it. E.g. something like:


```suggestion
    - **Mandatory**: Do not register properties meant to be design tokens that dynamically switch based on `light-dark()` as `<color>`. If you need to animate a color variable, use a separate property. `light-dark()` resolves at computed value time, and registered properties inherit resolved, computed values, so the color would no longer be reactive to local theme changes. 
```

Also, I just realized the first bullet seems to be repeating the same explanation. I'd collapse them into one.

guides/user-experience/browser-ui-color-theme/guide.md
```diff
@@ -45,15 +48,101 @@ MANDATORY: Apply the `color-scheme` property to the `html` element or the `:root
...
+    - Built-in inherited properties such as `color` and `accent-color` can also be affected by this issue. To prevent this, always reassign the dynamic theme variable on these built-in properties when they are inside of a nested theme. 
+    - **Example: Handling Built-In Inherited Properties**
+
+      ```css
+      :root {
+        /* Define browser UI accent color for each mode */
+        --brand-accent-light: #0056b3;
+        --brand-accent-dark: #00e5ff;
+        --accent-color: var(--brand-accent-light, AccentColor);
```
LeaVerou: Fallback strategies are meant to be contained in the Fallback strategies section. Also note that this will never be used, even if `light-dark()` is not supported because declarations with `var()` references are assumed to always be valid, and fail at computed value time.

(In the fallback strategies section the overriding declaration is in a conditional)

guides/user-experience/browser-ui-color-theme/guide.md
```diff
@@ -62,7 +151,39 @@ You can override the global theme for specific elements. This is useful for "dar
 ## Fallback strategies
 
 {{ BASELINE_STATUS("color-scheme") }}
+{{ BASELINE_STATUS("light-dark") }}
+
+- **Progressive Enhancement**: Browsers that do not support `color-scheme` or `light-dark()` will simply ignore these properties and use their default light-mode UI.
```
LeaVerou: This is not completely true. `color-scheme` is indeed PE, but `light-dark()` isn't, and its use can be quite destructive if used without a fallback.

To elaborate: If a browser doesn't support `light-dark()` and we use colors with it on e.g. `color`, `background`, `accent-color` etc, the browser will only "realize" the color is invalid at the very end, the declaration will become IACVT, which is basically equivalent to applying the `unset` value which e.g. will resolve to `transparent` for `background-color`, which can make overlay surfaces entirely unreadable.

(The rest of this section is fine and does not assume `light-dark()` is PE)

guides/user-experience/browser-ui-color-theme/guide.md
```diff
@@ -62,7 +151,39 @@ You can override the global theme for specific elements. This is useful for "dar
...
+
+  /* OPTIONAL: use light-dark() for more control of built-in UI colors */
+  @supports (color: light-dark(white, black)) {
+    --accent-color: light-dark(var(--brand-accent-light), var(--brand-accent-dark));
+  }
+}
+
+textarea {
+  color-scheme: dark;
+  /* **Mandatory**: when nesting schemes, dynamic values set with light-dark() must also be re-assigned on the element where the scheme changes */
```
LeaVerou:
```suggestion
  /* **Mandatory**: when nesting schemes, dynamic values set with light-dark() used in inherited properties must also be re-assigned on the element where the scheme changes */
```

guides/user-experience/browser-ui-color-theme/guide.md
```diff
@@ -45,15 +48,101 @@ MANDATORY: Apply the `color-scheme` property to the `html` element or the `:root
...
+  --background-color: var(--brand-bg-light, Canvas);
+  --text-color: var(--brand-text-light, CanvasText);
+
+
+  /* MANDATORY: Automatically adapt native UI to user system preferences */
+  color-scheme: light dark;
+  
+
+  /* OPTIONAL: use light-dark() for more control of built-in UI colors */
+  @supports (color: light-dark(white, black)) {
```
LeaVerou: The guideline currently is to only use fallbacks in the fallbacks section, though see https://chat.google.com/room/AAQAPIjpZaI/cIRFFKSHjhI/cIRFFKSHjhI?cls=10 for a discussion about this

dvdherron: Updated with fallbacks only in the fallback section.

guides/user-experience/browser-ui-color-theme/guide.md
dvdherron: Fixed in https://github.com/GoogleChrome/guidance/pull/701/changes/58a8f5af0bfd32949b67779773f2e1a65017b163

guides/user-experience/browser-ui-color-theme/guide.md
dvdherron: Updated this guidance in https://github.com/GoogleChrome/guidance/pull/701/changes/58a8f5af0bfd32949b67779773f2e1a65017b163

guides/user-experience/browser-ui-color-theme/guide.md
dvdherron: Removed these in https://github.com/GoogleChrome/guidance/pull/701/changes/58a8f5af0bfd32949b67779773f2e1a65017b163

guides/user-experience/browser-ui-color-theme/guide.md
dvdherron: Combined these in https://github.com/GoogleChrome/guidance/pull/701/changes/58a8f5af0bfd32949b67779773f2e1a65017b163

LeaVerou (approved):
LGTM! Thank you!

guides/user-experience/browser-ui-color-theme/expectations.md
rviscomi, line 1: I'm not sure how reliably these conditional expectations would get mapped to tests, but since we were discussing possibly combining this guidance into a single "dark mode" use case, maybe this won't be an issue.

rviscomi (approved):
LGTM. One non-blocking comment on the expectations file.

PR #704, opened 2026-05-01
Telemetry client

micahjo7:
Sets up MWG telemetry client, which logs with Clearcut HTTP API. Changes in `google3` are listed in b/495796940.

## Description
This PR implements an anonymous usage statistics collection system (telemetry) for the Chrome Modern Web Guidance tool to gather insights. The telemetry pipeline operates fully asynchronously and is designed to never block user execution speed or CLI responsiveness.
### Key Highlights
1. **Detached Watchdog Process**: 
   - Spawns a background process ([watchdog/main.ts](file:///Users/micahjo/guidance/serving/skills-cli/telemetry/watchdog/main.ts)) detached from the parent CLI.
   - Communicates via a one-way JSON-IPC pipe (`stdin`).
   - Automatically performs a graceful shutdown when the parent process terminates, flushing any pending network requests before exiting.
2. **Clearcut Delivery**:
   - Formats and transmits search and retrieve telemetry events to Google Play's Clearcut API (`https://play.googleapis.com/log?format=json_proto`) using standard fetch with a `10s` timeout.
3. **Opt-Out Mechanism**:
   - Telemetry is active by default but can be fully deactivated at any time by setting the environment variable `DISABLE_TELEMETRY=1`.
4. **Clean Standard Output**:
   - All informational telemetry banners and watchdog exit logs have been routed to `stderr` (via `console.warn`).
   - Keeps `stdout` exclusively JSON-formatted, preventing syntax errors in automated test suites (e.g., `copy-skills.test.ts`) and external tool integrations.

Logged data is queried with Plx script [go/zmpsy](http://go/zmpsy) and dashboard viewed at [go/guidance-dashboard](http://go/guidance-dashboard).


serving/bin/modern-web.ts
```diff
@@ -49,53 +49,90 @@ async function main() {
 
   if (command === "search") {
     if (!arg) {
+      const logger = new ClearcutLogger();
```
paulirish: should we have a single logger instance for the whole file?  just making sure its deliberate a logger per command (which... seems unexpected)

serving/bin/modern-web.ts
```diff
@@ -49,53 +49,90 @@ async function main() {
...
+        const logger = new ClearcutLogger();
+        await logger.logSearchResult([], { latencyMs, success: true });
         console.log("[]");
       } else {
+        // Instantiate logger
+        const logger = new ClearcutLogger();
+        // Log search results
+        const searchItems = results.map(r => ({
+          guide_id: r.id,
+          score: parseFloat((1 - parseFloat(r.distance)).toFixed(4)),
```
paulirish: fwiw i'm thinking of flipping this to be "similarity" of 0-1  because i think distance is harder to parse.

serving/skills-cli/telemetry/ClearcutLogger.ts
```diff
@@ -0,0 +1,139 @@
...
+import {
+  type ChromeModernWebGuidance,
+  type SearchItem,
+  WatchdogMessageType,
+  OsType,
+} from './types.ts';
+import { getVersion } from '../../lib/version.ts';
+
+function isTelemetryEnabled(): boolean {
+  if (process.env.DISABLE_TELEMETRY === '1' || process.env.DISABLE_TELEMETRY === 'true') {
```
paulirish:
```suggestion
  if (process.env.DISABLE_TELEMETRY) {
```

serving/skills-cli/telemetry/ClearcutLogger.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export function detectOS(): OsType {
+  const platform = process.platform;
+  if (platform === 'darwin') return OsType.MACOS;
+  if (platform === 'win32') return OsType.WINDOWS;
+  if (platform === 'linux') return OsType.LINUX;
+  return OsType.UNSPECIFIED;
+}
+
+const LATENCY_BUCKETS = [50, 100, 250, 500, 1000, 2500, 5000, 10000];
```
paulirish: we def want to fill out these buckets nicely

i'd probably propose we have a 5 and 15 beneath these

serving/skills-cli/telemetry/watchdog/main.ts
```diff
@@ -0,0 +1,112 @@
...
+    clearcutIncludePidHeader,
+  } = parseWatchdogArgs();
+
+  const sender = new ClearcutSender({
+    clearcutEndpoint,
+    includePidHeader: clearcutIncludePidHeader,
+  });
+
+  let isShuttingDown = false;
+  function onParentDeath() {
```
paulirish, line 50:
```ts
  function onParentDeath() {
    if (isShuttingDown) {
      return;
    }

    isShuttingDown = true;
    sender.sendShutdownEvent()
      .then(() => process.exit(0))
      .catch(() => process.exit(1));
  }
  ```

File: serving/skills-cli/template/README.md [hunk 6]
```diff
@@ -146,3 +146,15 @@ We grade both outputs and only release guides that demonstrate a massive, quanti
 
 \# Choose which skills you want  
 `npx modern-web-guidance install --choose`
+
+## Usage Statistics & Opt-Out
+
+Google collects anonymous usage statistics (such as search queries, guide retrievals, and successful skill installations) to improve the reliability, relevance, and performance of the Modern Web Guidance tool.
```
paulirish: would be cool to link to https://github.com/GoogleChrome/guidance/blob/main/serving/bin/modern-web.ts  so people can see exactly what's collected.

serving/skills-cli/telemetry/ClearcutLogger.ts
```diff
@@ -0,0 +1,139 @@
...
+      this.#watchdog = new WatchdogClient({
+        parentPid: process.pid,
+        logFile: options.logFile,
+        clearcutEndpoint: options.clearcutEndpoint,
+        clearcutIncludePidHeader: options.clearcutIncludePidHeader,
+      });
+    }
+  }
+
+  async logSearchResult(
```
paulirish: would slightly prefer these method signatures are on one line. could also extract metrics since it's repeatd.

```ts
  type Metrics = {latencyMs: number; success?: boolean};

  async logSearchResult(searchItems: SearchItem[], metrics?: Metrics): Promise<void> {
  ```

**Review by paulirish — COMMENTED** [review]
nice stuff!

File: serving/bin/modern-web.ts (hunk 1 above)
Comment (micahjo7) [reply serving/bin/modern-web.ts]: for modern-web `--version`, `--help`, or `list` we don't want the instantiation, since it will log a "Sending telemetry event..." message and create a watchdog background process

File: serving/skills-cli/telemetry/ClearcutLogger.ts (hunk 3 above)
Comment (micahjo7) [reply serving/skills-cli/telemetry/ClearcutLogger.ts]:
```
Using if (process.env.DISABLE_TELEMETRY === '1' || process.env.DISABLE_TELEMETRY === 'true') in 

ClearcutLogger.ts#L18
 is actually more robust than a simple truthiness check.

In Node.js, process.env values are always strings. If a user explicitly tries to enable telemetry by setting DISABLE_TELEMETRY=0 or DISABLE_TELEMETRY=false, checking if (process.env.DISABLE_TELEMETRY) would evaluate "0" and "false" as truthy strings, accidentally disabling telemetry.

Explicitly matching '1' or 'true' prevents this ambiguity.
```

File: serving/bin/modern-web.ts (hunk 2 above)
Comment (micahjo7) [reply serving/bin/modern-web.ts]: updated to use similarity

File: serving/skills-cli/telemetry/watchdog/main.ts [hunk 8]
```diff
@@ -0,0 +1,108 @@
...
+    clearcutEndpoint,
+    clearcutIncludePidHeader,
+  };
+}
+
+function main() {
+  const {
+    clearcutEndpoint,
+    clearcutIncludePidHeader,
+  } = parseWatchdogArgs();
```
formgit, line 41: Looks like the 'log-file' flag is ignored here? And it seems we don't have other ways to catch stderr logs in the watchdog client.

serving/skills-cli/telemetry/ClearcutLogger.ts
paulirish: heh. yeah sure. i guess.

serving/skills-cli/telemetry/watchdog/main.ts
micahjo7: thanks, i don't think we need those two fields. removed

serving/bin/modern-web.ts
```diff
@@ -204,7 +241,7 @@ function maybeEmitUpdateMessage(callerSkillVersion: string|null): void {
     `Your version: ${callerSkillVersion}`,
     `Latest version: ${latestSkillVersion}`,
     '',
-    'To update, run: npx modern-web-guidance@latest update',
+    'To update, run: npx modern-web-guidance@latest update --auto',
```
paulirish: whats that

serving/skills-cli/telemetry/ClearcutLogger.ts
```diff
@@ -0,0 +1,128 @@
...
+export type Metrics = { latencyMs?: number; success?: boolean; skillVersion?: string | null };
+
+export class ClearcutLogger {
+  #watchdog: WatchdogClient | null = null;
+
+  constructor(options: {
+    clearcutEndpoint?: string;
+    clearcutIncludePidHeader?: boolean;
+  } = {}) {
+    if (isTelemetryEnabled()) {
```
paulirish: lets switch to a early return... just lower chance of adding something else in here mistakenly not behind the guard.

paulirish (approved):

serving/bin/modern-web.ts
```diff
@@ -84,43 +109,67 @@ async function main() {
     console.log(JSON.stringify(catalog, null, 2));
   } else if (command === "retrieve") {
     if (!arg) {
```
paulirish: also.. since we're looking at it.. these two !arg and ids.length === 0 cases are handled exactly the same no?   lets combine them.

serving/bin/modern-web.ts
```diff
@@ -49,29 +54,49 @@ async function main() {
...
     }
+    const startTime = Date.now();
     try {
       // Dynamic import to keep the CLI loading fast -- only load the embedder if needed.
       const { searchUseCases } = await import("../lib/search.ts");
       const results = await searchUseCases(arg);
+      const latencyMs = Date.now() - startTime;
+
       if (results.length === 0) {
+        const logger = new ClearcutLogger();
```
paulirish: well... okay one more thought... Is there any other way that we can kinda _lessen_ the impact on this file?  

These new added lines really add a bunch of stuff to the screen that it would be lovely if that wasn't as dramatic.  something more clean and terse.

serving/bin/modern-web.ts
micahjo7: created a flag so we know when the agent is running the update command vs when it is run manually but... that is going away so we don't need this anymore.

micahjo7: removed

serving/skills-cli/telemetry/ClearcutLogger.ts
micahjo7: thanks!

serving/bin/modern-web.ts
micahjo7: done

serving/bin/modern-web.ts
micahjo7: i've updated the file to use lazy initialization of the logger. modern-web.mjs file increases from 1306 (before this change) to 1488 now

serving/bin/modern-web.ts
```diff
@@ -49,20 +52,33 @@ async function main() {
...
       // Dynamic import to keep the CLI loading fast -- only load the embedder if needed.
       const { searchUseCases } = await import("../lib/search.ts");
       const results = await searchUseCases(arg);
+      const latencyMs = Date.now() - startTime;
+
+      const searchItems = results.map(r => ({
+        guide_id: r.id,
+        similarity: parseFloat(r.similarity),
+      }));
+      await getLogger().logSearchResult(searchItems, { latencyMs, success: true, skillVersion });
```
paulirish: you could pass `skillVersion` to the `new ClearcutLogger()` constructor... which means it happens just ones on L59 and then you dont need it repeated here in the callers.  

unless you're solving some multi-skill-version logger situation.. hmm.

serving/bin/modern-web.ts
```diff
@@ -72,6 +88,8 @@ async function main() {
         console.log("[" + jsonLines.join(",\n") + "]");
       }
     } catch (error) {
+      const latencyMs = Date.now() - startTime;
+      await getLogger().logSearchResult([], { latencyMs, success: false, skillVersion });
```
paulirish: also.. seems like latencyMs is always passed.. so.. probably should pass it as the first argument

eh. maybe not. shrug.

micahjo7: moved latency and success as first and second args

serving/bin/modern-web.ts
micahjo7: added skillVersion to constructor

serving/skills-cli/telemetry/ClearcutLogger.test.ts
```diff
@@ -52,45 +60,47 @@ describe('ClearcutLogger', () => {
     }
   });
 
-  it('logs tool command results correctly to console.warn', async () => {
-    const originalWarn = console.warn;
-    const warnings: string[] = [];
-    console.warn = (msg: string) => {
-      warnings.push(msg);
+  it('logs tool command results correctly via watchdog', async () => {
+    const originalSend = WatchdogClient.prototype.send;
```
paulirish: https://nodejs.org/api/test.html#mocking  you can use mocks for this instead.. probably will be a tad cleaner? maybe?

paulirish (approved):
very attractive. :)

PR #717, opened 2026-05-05
Add macros for selective content transclusion from other md files

### LINKED ISSUE #591: Centralize feature-specific content (fallbacks, a11y guidance, browser bugs etc)
_(Originally discussed in [chat](https://chat.google.com/room/AAQAPIjpZaI/VegL-q0QJtg/VegL-q0QJtg?cls=10), filing an issue for posterity)_
.
There is currently a lot of content listed in guides that is not use case specific, but tied to the features being used. For example:
- Fallbacks are largely the same across use cases. E.g. the fallback for `scrollbar-color` is always `::-webkit-scrollbar` regardless of the use case.
- Accessibility guidance is largely the same. E.g. guidelines for `prefers-reduced-motion` apply very similarly to all animation-related guides
- Same with information about browser bugs to avoid triggering or work around, these apply to features, not use cases (e.g. #556).

Currently these are repeated individually per guide, and are thus often inconsistent, simply due to the non-determinism of maintaining the same content in multiple places rather than any factors specific to the use case at hand.

Centralizing them would provide several benefits:
- Ensuring that we provide the best possible version of that content
- Easier human review
- Casting a wider net by triggering them individually if the feature is used or requested even if no use case matches

The centralization doesn't need to be reflected in the final CLI, it could be pulled into the guides via macros, like baseline status. And individual use cases can still provide their own fallbacks, a11y guidance etc if different.

Types of knowledge to centralize:
- Fallbacks
- a11y guidance
- Known bugs that affect a majority of use cases (more niche bugs should be documented in their respective guides)
- What else?

LeaVerou:
Closes #591. 

After seeing the Nth[^1] case of repeated content that was not fully in sync, I figured I'd get an agent to prototype this idea, just to get a feel of how difficult it would be to have this functionality. Turns out, it was much easier than I thought, kudos to whomever set up the macros infrastructure! In fact, I ended up generalizing it to _any_ kind of content transclusion, that way guides can share related content from each other as needed, without needing to set up a project wide feature for that specific type of content. I did add higher level shortcuts for features however, since that's by far the most common case.

[^1]:  Where N → +∞ — ok ok I'm exaggerating _juuust_ a tad for effect 😅

This PR uses the existing macro machinery that `BASELINE_STATUS` was already using to add several macros for including content from arbitrary or predefined Markdown files to minimize duplication (see #591).

It also introduces a `defineFeatureMacro()` abstraction for more easily defining macros that take a feature id, and rewrites `BASELINE_STATUS` to use it.

Example usage in #718 . 

## Macros added, from lowest level to highest level

For all macros below, unless otherwise specified:
- **Errors**: Generally, author errors fail loudly, missing content fails silently. For feature macros, if the feature is not found, it throws a `MacroError`. If the *file* or *section* is not found, it fails silently (just outputs `""`) so that references can be added independently of referenced content, if we anticipate certain referenced content.
- Recursive: individual macros still work within the referenced content.
- Content is trimmed. so it can become part of. a list, a table etc

Out of scope for now, but we can add easily:
- Cycle detection for recursive macros
- Proper Markdown parsing (code uses a regexp and assumes hash heading syntax)
- A slugify package (currently using a GitHub-style helper)

#### `INCLUDE("path/to/file.md")`: Include whole file

Transcludes an entire markdown file, ignoring any frontmatter and level 1 heading (so that they can also be used independently).

Path resolution is repo-root-relative for bare paths (e.g. `features/foo.md` is resolved from the repo root), but caller-relative for `./` and `../.` (so that files can include relative to themselves).

#### `INCLUDE("path/to/file.md#section")`: Include section content

Transcludes an individual section, ignoring its heading. If section is not found, outputs `""` silently.

Section IDs match either an explicit `{ #id }` suffix or the slugified heading text (slugify function implemented ad hoc to be similar to GitHub's).

### `FEATURE("feature-id", "section")`

Wrapper around `INCLUDE("features/feature-id.md", "#section")` to make the common case of feature-related content nicer. Compare:

```
{{ FEATURE("user-pseudos", "aria-invalid") }}
{{ INCLUDE("features/user-pseudos.md", "#aria-invalid") }}
```

### `FEATURE_ISSUES("feature-id")`

Prints out a heading like `### Issues to be aware of when using [feature name]` plus the content of the `#issues` section for that feature, iff there is one. If there is no issues section, the entire macro returns `""`

### `FEATURE_FALLBACKS("feature-id")`

Prints out:
1. A heading like `### Fallbacks & browser support for [Feature name]`
2. `BASELINE_STATUS("feature-id")`
3. `FEATURE("feature-id", "fallbacks")`

If 3 is empty, it only prints out 2, so that it can replace `BASELINE_STATUS` in guides even when we *anticipate* fallbacks to be documented in the future, before they exist.

### DIFF
--- a/serving/lib/include.ts
+++ b/serving/lib/include.ts
@@ -1,6 +1,7 @@
 import fs from "fs";
 import path from "path";
 import matter from "gray-matter";
+import { marked, type Token, type Tokens } from "marked";
 import { rootDir } from "../../lib/paths.ts";
 
 /**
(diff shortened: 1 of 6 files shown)


serving/lib/include.ts
```diff
@@ -0,0 +1,125 @@
...
+ * block is treated as a real heading.
+ */
+function loadSection(absolutePath: string, sectionId: string): string {
+  const record = fileCache.get(absolutePath)!;
+  if (sectionId in record) return record[sectionId];
+
+  const body = record[""];
+  let result = "";
+  if (body) {
+    const headings = [...body.matchAll(/^(#{2,})\s+(.+?)(?:\s*\{\s*#([\w-]+)\s*\})?\s*$/gm)];
```
paulirish: wowza. big boi.

i'm sure this is fine but... might as well use `marked.lexer`  for this instead.

serving/lib/macros.test.ts
```diff
@@ -51,3 +56,212 @@ describe('replaceMacros (Functional with real data)', () => {
...
+  it('strips punctuation', () => {
+    assert.equal(slugify('Hello, World!'), 'hello-world');
+  });
+
+  it('preserves hyphens', () => {
+    assert.equal(slugify('multi-word section'), 'multi-word-section');
+  });
+});
+
+describe('INCLUDE', () => {
```
paulirish, line 78: can you also add some stuff to guide-integrity.test.ts which captures that all the translation macros used correctly return data?

paulirish (approved):
very cool! appreciate you tackling this. 

i'm into it.

serving/lib/macros.test.ts
LeaVerou: Done!

serving/lib/include.ts
LeaVerou: Done! Doesn't really save anything in terms of length or complexity, but it _does_ give us more flexibility in authoring the guides. 

I'm caching the parsed tokens so we only parse each referenced file once.
Note that we're [still](https://github.com/GoogleChrome/guidance/pull/717/changes#diff-8353ce7daf610e750f2f7fda980b373b0b93932dfd1b8fd68075be7dfdf0618aR105) parsing the `{ #id }` syntax with a regex since no attribute syntax plugin is installed, but at least we're now extracting the heading text from the actual parsed tokens.

PR #718, opened 2026-05-05
Sample use of transclusion macros to centralize repeated content

LeaVerou:
Examples for #717


features/temporal.md
```diff
@@ -0,0 +1,21 @@
+# Temporal
```
paulirish, line 1: feels nice.

i think optionally we could do some yaml frontmatter that has some metadata like...  the items/bcd keys listed below.  maybe? wdyt

features/scrollbar-color.md
```diff
@@ -0,0 +1,50 @@
+# scrollbar-color
```
paulirish, line 1: last week i introduced `./features/<featureid>/research.md`

i slightly prefer the explict names.. so what if these files are instead located at `./features/<featureid>/fallback.md`

features/temporal.md
LeaVerou: We _could_, though without use cases it feels a bit premature. The code already ignores frontmatter and h1 so we can do that later, if we want to. I'm hoping we may eventually use the files independently *as well*, e.g. to point to when we reference a certain feature in a skill. But we don't need to do it all at once :) What do you think?

features/scrollbar-color.md
LeaVerou: You mean having a separate file per type of info? It's not just fallbacks. Perhaps we can reuse the `guide.md` convention?

paulirish: oh right. gotcha. hmmm.. `common.md` would probably be the idea then. 

thats .. eh.. less attractive.  aight. lets keep it as you have it.

LeaVerou: Pretty easy to change it down the line, if needed!

features/temporal.md
paulirish: I gotcha. okay yeah.

the transclusion design is pretty powerful. in my first pass i didnt appreciate that it can pull any section level. but yeah.. now see that it can pull just a small section or a larger one with subsections. 

neat.

LeaVerou: Yup! There's an example where the top-level fallback didn't fit wholesale, but I'm still including the lower-level section: https://github.com/GoogleChrome/guidance/pull/718/changes#diff-770333ce6dddce1a407ccbfdca45488342be17f87920dfc969cd2d7dbfbbc681

paulirish (approved):

Issue #719, opened 2026-05-05
Scroll snap


philipwalton:
### web-feature-id

scroll-snap

### Feature description

The CSS scroll snap module provides properties that let you control the panning and scrolling behavior by defining snap positions. Content can be snapped into position as the user scrolls overflowing content within a scroll container.

### Coding agent assessment

Coding agents frequently resort to writing complex pointer event physics logic for many gesture-based interactions that can be reliably, and more performantly, recreated with scroll snapping.

### Reference sources

- https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Scroll_snap

**Use cases:**
- [x] #263
- [x] #147
- [x] #371
- [x] #372
- [x] #818
- [ ] #1235
- [ ] #713
- [x] #817
- [ ] #1251

PR #720, opened 2026-05-05
Add swipe-to-remove guide

philipwalton:
This PR creates and implements a new `swipe-to-remove` use case for the scroll snap feature (#719).


guides/user-experience/swipe-to-remove/guide.md
```diff
@@ -0,0 +1,426 @@
...
+
+  const observer = new IntersectionObserver((entries, obs) => {
+    // Most recent entry only; older ones are stale.
+    const entry = entries.at(-1);
+    const ratio = entry.intersectionRatio;
+
+    // boundingClientRect.x > 0 means the content is offset to the right
+    // of the track (the user is swiping right, revealing the left spacer);
+    // the leading icon lives on the left in that case.
+    const direction = entry.boundingClientRect.x > 0 ? 'left' : 'right';
```
patrickkettner: The direction logic would only work if the item is all the way against the left edge, right? If the track was 200px offset (centered layout, padded container, anything other than full bleed), every left-swipe reads as a right-swipe for the first 200px, so the wrong icon pops during the activation phase. `entry.rootBounds` gives you the track-relative offset for free

```js
const relX = entry.boundingClientRect.x - entry.rootBounds.x;
const direction = relX > 0 ? 'left' : 'right';
```

guides/user-experience/swipe-to-remove/guide.md
```diff
@@ -0,0 +1,426 @@
...
+  const commitThreshold = 0.2;
+
+  const observer = new IntersectionObserver((entries, obs) => {
+    // Most recent entry only; older ones are stale.
+    const entry = entries.at(-1);
+    const ratio = entry.intersectionRatio;
+
+    // boundingClientRect.x > 0 means the content is offset to the right
+    // of the track (the user is swiping right, revealing the left spacer);
+    // the leading icon lives on the left in that case.
```
patrickkettner: for same reasons as below, I think this isn't always true.

guides/user-experience/swipe-to-remove/guide.md
```diff
@@ -0,0 +1,426 @@
...
+  observer.observe(content);
+}
+
+async function removeItem(item, content, direction, rect) {
+  const opts = { duration: 300, easing: 'ease', fill: 'forwards' };
+
+  // Pixel distance the content needs to travel to be fully out of view.
+  const translate = direction === 'left'
+    ? rect.width - rect.x
+    : -(rect.x + rect.width);
```
patrickkettner: [same comment](https://github.com/GoogleChrome/guidance/pull/720/changes#r3214299973)

guides/user-experience/swipe-to-remove/demo.html
```diff
@@ -0,0 +1,355 @@
...
+        // visual-feedback band and a commit band.
+        const observer = new IntersectionObserver((entries, observer) => {
+          // Only the most recent entry is relevant.
+          const entry = entries.at(-1);
+          const ratio = entry.intersectionRatio;
+
+          // boundingClientRect.x > 0 means the content is offset to the
+          // right of the track (the user swiped right, revealing the
+          // left spacer); the leading icon lives on the left.
+          const direction = entry.boundingClientRect.x > 0 ? 'left' : 'right';
```
patrickkettner: [same comment](https://github.com/GoogleChrome/guidance/pull/720/changes#r3214299973)

guides/user-experience/swipe-to-remove/demo.html
```diff
@@ -0,0 +1,355 @@
...
+    }
+  </style>
+
+  <script type="module">
+    async function removeItem(item, content, direction, rect) {
+      // In a real application, you could perform a different action based on
+      // the direction passed here (e.g., archive vs. snooze). For this demo,
+      // we remove the item from the list regardless of direction.
+      const opts = {duration: 300, easing: 'ease', fill: 'forwards'};
+      const translate = direction === 'left' ? rect.width - rect.x : -(rect.x + rect.width);
```
patrickkettner: [same comment ](https://github.com/GoogleChrome/guidance/pull/720/changes#r3214299973)

guides/user-experience/swipe-to-remove/expectations.md
```diff
@@ -0,0 +1,22 @@
...
+- The implementation MUST detect commit using an `IntersectionObserver` rooted at the track (`root: track`, not the viewport) and observing the content element. Vertical scrolling of the outer list must not change the intersection (because root and target move together).
+- The IntersectionObserver MUST use a low ratio threshold (e.g., `commitThreshold ≈ 0.20`) for commit, NOT `intersectionRatio === 0`. Committing before the snap fully settles is what allows the remove animation to start during the swipe gesture (rather than waiting for the snap to land).
+- The IntersectionObserver SHOULD also track a higher activate threshold (e.g., `activateThreshold ≈ 0.80`) so an "activating" visual state can be toggled on the item (e.g., scaling the action icon) while the user is dragging, before the commit point is reached.
+- The IntersectionObserver MUST update a `data-swipe-direction` attribute (or equivalent state) on the item indicating which side the user is swiping toward, derived from the sign of `entry.boundingClientRect.x`.
```
patrickkettner: [same comment](https://github.com/GoogleChrome/guidance/pull/720/changes#r3214299973)

guides/user-experience/swipe-to-remove/demo.html
```diff
@@ -0,0 +1,355 @@
...
+        // Add a class to signal that the element is ready to be interacted with.
+        item.classList.add('is-initialized');
+
+        // If the browser doesn't support `scroll-initial-target: nearest` then
+        // programmatically scroll the content into view and wait for this to
+        // finish before continuing to ensure that the Intersection Observer
+        // doesn't start observing before the content is fully in view.
+        if (!CSS.supports('scroll-initial-target', 'nearest')) {
+          track.scrollLeft = track.clientWidth;
+        }
```
patrickkettner: how would this work if its called when the list isn't visible? track.clientWdith would be 0, so this sets `scrollLeft` to 0, and then when the container _becomes_ visible the track renders scrolled to the left spacer - right? That would mean that the IntersectionObserver fires with a ratio of 0, and then every item commits the swipe action without the user touching anything. In the destructive variant of the guide every item in the list is deleted the first time the panel is opened.

guides/user-experience/swipe-to-remove/guide.md
```diff
@@ -0,0 +1,426 @@
...
+      item.classList.toggle('is-activating', ratio < activateThreshold);
+    }
+  }, {
+    // Root the observer at the track so vertical scrolling of the outer
+    // list doesn't change the intersection (root and target move together).
+    root: track,
+    threshold: [commitThreshold, activateThreshold],
+  });
+
+  observer.observe(content);
```
patrickkettner: [same hidden init issue](https://github.com/GoogleChrome/guidance/pull/720/changes#r3214312096)

guides/user-experience/swipe-to-remove/guide.md
```diff
@@ -0,0 +1,426 @@
...
+### Fallback for `scroll-initial-target`
+
+{{ BASELINE_STATUS("scroll-initial-target") }}
+
+If your Baseline target does not include `scroll-initial-target`, scroll the track to the content programmatically. Detect with `CSS.supports`:
+
+```js
+if (!CSS.supports('scroll-initial-target', 'nearest')) {
+  track.scrollLeft = track.clientWidth;
+}
```
patrickkettner: [same hidden init issue](https://github.com/GoogleChrome/guidance/pull/720/changes#r3214312096)

guides/user-experience/swipe-to-remove/guide.md
```diff
@@ -0,0 +1,426 @@
...
+- **DO** set `overscroll-behavior-x: none` on the track. Without it, an over-swipe can trigger the browser's back-navigation gesture on iOS/Android.
+- **DO** commit at a threshold *before* the snap settles (e.g., `commitThreshold ≈ 0.2`) rather than waiting for the content to be fully off-screen. This lets the remove animation start during the gesture, which feels significantly more responsive than waiting for the snap to land.
+- **DO NOT** rely on `pointerdown`/`pointermove`/`pointerup` to drive a manual transform. You'll lose momentum, snap-back, keyboard accessibility, and reduced-motion handling that the browser gives you for free.
+- **DO** confirm destructive actions when appropriate. For "remove", consider showing an undo toast after the swipe completes; the gesture is fast and easy to trigger by accident.
+- **DO** ensure the scroll track is focusable, keyboard accessible, and that there is a visual focus affordance.
+- **DO** provide accessible alternatives for any relevant actions triggered by the swipe (e.g., a visible button, context menu, or edit mode).
+
+## Fallback strategies
+
```
patrickkettner, line 463: ["The first, standalone line inside that section must be either {{ FEATURE_FALLBACKS("feature-id") }} (preferred) or {{ BASELINE_STATUS("feature-id") }}. Do not place either at the top of the document"](https://github.com/GoogleChrome/guidance/blob/main/.agents/skills/project-guides/SKILL.md?plain=1#L67)

guides/user-experience/swipe-to-remove/expectations.md
```diff
@@ -0,0 +1,22 @@
...
+- In parallel with the height collapse, the implementation MUST animate the content element's `translate` (or equivalent transform) the remaining distance off-screen for the same duration as the height animation. This ensures the content finishes sliding off-screen at (or before) the row finishes shrinking — without it the height-collapse animation can finish before the browser's smooth scroll-snap has scrolled the content the rest of the way out, leaving the row collapsing around still-visible content. The translate distance MUST be derived from the IO entry's `boundingClientRect` (e.g., `rect.width - rect.x` for a right-swipe, `-(rect.x + rect.width)` for a left-swipe), not a hardcoded percentage.
```
patrickkettner: on both of these lines - is it ok to have this two assertions in one bullet? on 16, you have animate the height to 0, AND wait for that animation to finish before removing. If each bullet becomes is supposed to become its own test (thats how I had thought this was working - happy to be corrected), should this split into two?

guides/user-experience/swipe-to-remove/expectations.md
```diff
@@ -0,0 +1,22 @@
...
+- The pre-collapse height and translate distance passed to the animations MUST come from the IO entry's `boundingClientRect` (or an equivalent measurement taken before the animation starts), not from a post-mutation reading.
+- A fallback for `scroll-initial-target` MUST be included: when `CSS.supports('scroll-initial-target', 'nearest')` is false, the script MUST programmatically scroll each track to the content column (e.g., `track.scrollLeft = track.clientWidth`) before wiring up the IntersectionObserver. The programmatic scroll MUST happen AFTER the `is-initialized` class is added (so the track is actually scrollable) and BEFORE the observer is attached (so the initial scroll past the left spacer is not interpreted as a commit).
```
patrickkettner, line 18: [same bullet comment](https://github.com/GoogleChrome/guidance/pull/720/changes#r3214323509)

guides/user-experience/swipe-to-remove/expectations.md
```diff
@@ -0,0 +1,22 @@
+- Each list item MUST contain a horizontal scroll container ("track") that holds three full-width snap points: a leading spacer, the visible content element, and a trailing spacer. The spacers MAY be implemented as `::before`/`::after` pseudo-elements on the track or as real elements.
+- The track MUST have `scroll-snap-type: x mandatory` (not `proximity`) so the row always rests on a snap point and is never partially scrolled.
+- The track MUST have `overscroll-behavior-x: none` to prevent the swipe from chaining into the page scroll or the browser's back-navigation gesture.
+- The track MUST hide its scrollbar (e.g., `scrollbar-width: none`) so the gesture itself is the only affordance.
+- The implementation MUST be accompanied by an accessible alternative (e.g., a visible button, context menu, or edit mode) to ensure keyboard and screen reader users can perform the action without a swipe gesture.
```
patrickkettner: I am all for the idea, but how would a grader actually test this? if they have a single button on the page, they would automatically pass, right? Would it make sense to have it be more like "each .SwipeableItem has a focusable button with an aria-label describing the action"?

guides/user-experience/swipe-to-remove/guide.md
```diff
@@ -0,0 +1,426 @@
...
+  // `GestureEvent` (a Safari-only API) and defer the actual DOM removal
+  // until the gesture has fully settled. The 5s delay is conservative;
+  // anything longer than the momentum tail is fine. The row is already
+  // collapsed and invisible at this point, so the delay is not visible.
+  if (globalThis.GestureEvent) {
+    setTimeout(() => item.remove(), 5000);
+  } else {
+    item.remove();
+  }
+}
```
patrickkettner, line 357: Based on your comment [over here](https://github.com/GoogleChrome/guidance/blob/b961d34db1a1a1e4b88ef04e8ae0fa35d3a1177e/guides/user-experience/swipe-to-remove/demo.html#L169), its visually collapsed but it's still in the DOM and the a11y tree for 5 seconds, right? So a screen reader will keep announcing the "deleted" item, :nth-child would be off, focus order still includes the dead row etc.

Maybe we could you set `inert` and `display: none` right after the animation finishes but before the setTimeout fires?

I hadnt heard of the safari bug you mentioned - would you happen to have a link to it?

guides/user-experience/swipe-to-remove/demo.html
```diff
@@ -0,0 +1,355 @@
...
+      // would then start scrolling. Instead, we have to transition it
+      // to `height:0` first (above) and then remove it only after the
+      // swipe/scroll gesture has likely finished (e.g. 5 seconds).
+      if (globalThis.GestureEvent) {
+        setTimeout(() => item.remove(), 5000);
+      } else {
+        item.remove();
+      }
+    }
+
```
patrickkettner: [same 5 second comment](https://github.com/GoogleChrome/guidance/pull/720/changes#r3214334755)

guides/user-experience/swipe-to-remove/demo.html
```diff
@@ -0,0 +1,355 @@
...
+      // Use a combination of CSS transitions (for declarative styles) and
+      // WAAPI animations (for computed values) to remove the element,
+      // then await the completion of all of them.
+      item.classList.add('is-removing');
+      item.animate([{height: `${rect.height}px`}, {height: '0px'}], opts);
+      content.animate([{translate: `${translate}px`}], opts);
+      await Promise.all(
+        item.getAnimations({subtree: true}).map((a) => a.finished),
+      );
```
patrickkettner, line 208: if any of these animations get cancelled `.finished` would reject with an AbortError and the function throws before item.remove() runs. so the collapsed row sticks around in the DOM. 

You could add a `.finished.catch(() => {})` on each promise so Promise.all resolves cleanly, but im not sure if we want to encourage empty catches like that?

guides/user-experience/swipe-to-remove/guide.md
```diff
@@ -0,0 +1,426 @@
...
+  // then await the completion of all of them.
+  // Note: the content translate animation is important because the
+  // height-collapse animation can otherwise finish before the browser's
+  // smooth scroll-snap has scrolled the content fully off-screen.
+  item.classList.add('is-removing');
+  item.animate([{ height: `${rect.height}px` }, { height: '0px' }], opts);
+  content.animate([{ translate: `${translate}px` }], opts);
+  await Promise.all(
+    item.getAnimations({ subtree: true }).map((a) => a.finished),
+  );
```
patrickkettner, line 341: [same `catch` comment](https://github.com/GoogleChrome/guidance/pull/720/changes#r3214339893)

guides/user-experience/swipe-to-remove/demo.html
```diff
@@ -0,0 +1,355 @@
...
+      });
+
+      // Safari will sometimes reset scroll position on resize, so update the
+      // scroll position on resize to avoid this bug.
+      if (!CSS.supports('scroll-initial-target', 'nearest')) {
+        addEventListener('resize', () => items.forEach((item) => {
+          const track = item.querySelector('.SwipeableItem-track');
+          track.scrollLeft = track.clientWidth;
+        }));
+      }
```
patrickkettner: I am not sure how to test this on JavaScriptCore, but I am pretty sure this would keep every a reference to every item that was ever in the list forever (even after theyre removed). So any detached node never get garbage collected, and every resize still iterates over them setting scrollLeft on orphans. 

could you query the live DOM inside the listener instead? something like `
```js
document.querySelectorAll('.SwipeableItem.is-initialized .SwipeableItem-track').forEach(t => t.scrollLeft = t.clientWidth)
```

guides/user-experience/swipe-to-remove/guide.md
```diff
@@ -0,0 +1,426 @@
...
+
+```js
+if (!CSS.supports('scroll-initial-target', 'nearest')) {
+  addEventListener('resize', () => {
+    document.querySelectorAll('.SwipeableItem-track').forEach((track) => {
+      track.scrollLeft = track.clientWidth;
+    });
+  });
+}
+```
```
patrickkettner, line 540: [same orphan comment](https://github.com/GoogleChrome/guidance/pull/720/changes#r3214346419)

guides/user-experience/swipe-to-remove/demo.html
```diff
@@ -0,0 +1,355 @@
...
+      // to `height:0` first (above) and then remove it only after the
+      // swipe/scroll gesture has likely finished (e.g. 5 seconds).
+      if (globalThis.GestureEvent) {
+        setTimeout(() => item.remove(), 5000);
+      } else {
+        item.remove();
+      }
+    }
+
+    async function init() {
```
patrickkettner: Should there be some kind of guard to prevent running on items that are already wired? If any items get added after page load, init() would be called again and every existing item picks up a duplicate observer.

guides/user-experience/swipe-to-remove/guide.md
```diff
@@ -0,0 +1,426 @@
...
+  // anything longer than the momentum tail is fine. The row is already
+  // collapsed and invisible at this point, so the delay is not visible.
+  if (globalThis.GestureEvent) {
+    setTimeout(() => item.remove(), 5000);
+  } else {
+    item.remove();
+  }
+}
+
+document.querySelectorAll('.SwipeableItem').forEach(wireUpSwipe);
```
patrickkettner: [same missing guard comment](https://github.com/GoogleChrome/guidance/pull/720/changes#r3214352658)

guides/user-experience/swipe-to-remove/demo.html
```diff
@@ -0,0 +1,355 @@
...
+          if (ratio < commitThreshold) {
+            removeItem(item, content, direction, entry.boundingClientRect);
+            observer.disconnect();
+            return;
+          }
+
+          if (entry.boundingClientRect.x !== 0) {
+            item.dataset.swipeDirection = direction;
+            item.classList.toggle('is-activating', ratio < activateThreshold);
+          }
```
patrickkettner: the `x !== 0` guard skips the line that removes is-activating. when something snaps the track back to rest, the callback fires with ratio === 1 and x === 0, but this guard would mean it doesn't have '.is-activating' toggled. if the class was on at that moment it stays on, and the icon stays scaled up.

guides/user-experience/swipe-to-remove/demo.html
```diff
@@ -0,0 +1,355 @@
...
+      } else {
+        item.remove();
+      }
+    }
+
+    async function init() {
+      const items = document.querySelectorAll('.SwipeableItem');
+      if (!items.length) return;
+
+      items.forEach(async (item) => {
```
patrickkettner: why is this async?

guides/user-experience/swipe-to-remove/demo.html
```diff
@@ -0,0 +1,355 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Swipe to remove demo</title>
+
+  <style>
+    @layer base {
```
patrickkettner, line 9: is the @layer base / @layer demo setup a part of the swipe pattern?

guides/user-experience/swipe-to-remove/expectations.md
philipwalton: This one is tricky because an acceptable accessible alternative could be to allow for removing/archiving the item after you've drilled down into it.

For example, the Gmail app doesn't offer any buttons in the list view, but it does offer ways to archive a message after you've clicked on it, so I wouldn't want to prescribe exactly how the agent does this.

To your point, though, I don't know if the grader will be able to reliably test this.

guides/user-experience/swipe-to-remove/guide.md
philipwalton: Good call. I updated to set it to `inert` before the timer starts. I can't set it to `display:none` because that still triggers the bug. (If you want to see the bug in action, run it in Safari without the timeout.)

I couldn't find an existing bug for this, but I plan to file one when I have some time after I/O.

guides/user-experience/swipe-to-remove/guide.md
philipwalton: Good catch. Updated.

guides/user-experience/swipe-to-remove/demo.html
philipwalton: Good point. Switched to using `Promise.allSettled()`

guides/user-experience/swipe-to-remove/demo.html
philipwalton: Left over from a previous iteration. Updated.

guides/user-experience/swipe-to-remove/demo.html
philipwalton: No, it's just a way to clearly group the styles that are relevant to the use case implementation from the styles that are just for the demo.

guides/user-experience/swipe-to-remove/demo.html
philipwalton: Updated.

guides/user-experience/swipe-to-remove/demo.html
philipwalton: Updated.

guides/user-experience/swipe-to-remove/demo.html
philipwalton: I was kinda assuming that the agent could figure that out, but if I end up implementing the outer IntersectionObserver I described in https://github.com/GoogleChrome/guidance/pull/720/changes#r3224192865 then I think that should cover this too.

guides/user-experience/swipe-to-remove/demo.html
philipwalton: Do you think this warrants more than just a comment in the guide saying to make sure `init()` is not called until the DOM is ready to be initialized?

FWIW, I did test this in the case that the page was loaded in the background, and it worked just fine. Your concern only manifest in situations where `init()` is called when the elements are `display:none` (or something like that).

One thing I considered for performance reasons is to have an outer IntersectionObserver that only observes items when they're in the viewport. I decided to forgo that complexity, but maybe that would ultimately be more robust?

WDYT?

guides/user-experience/swipe-to-remove/expectations.md
philipwalton: I had Gemini generate these from the project-guides skill, so I guess I'll see if it struggles to write evals from this (and if so I can change it).

TBH these two expectations could probably be removed because this is styling polish that's arguably not core to the use case.

guides/user-experience/swipe-to-remove/guide.md
philipwalton: This is being changed, but it looks like the updates haven't been merged yet: https://github.com/GoogleChrome/guidance/pull/708

philipwalton (commented):
@patrickkettner thanks for the extremely thorough review! I've addressed most of your comments, with a few open questions/discussion points still to work through.

guides/user-experience/swipe-to-remove/expectations.md
patrickkettner: Ah, didnt think about the gmail style thing. if there isnt a way to trivially pass on any random page button, maybe it should come out of expectations.md and live in guide.md prose instead?

guides/user-experience/swipe-to-remove/guide.md
patrickkettner: does inert handle the :nth-child selectors (striped lists, etc)? I think it would, but its edge enough to be fine imo - may be worth a comment though?

guides/user-experience/swipe-to-remove/demo.html
patrickkettner: got it, makes sense. one nit though, an agent reading this might pick up @layer as part of the pattern since the implementation styles are inside @layer demo. would it be worth leaving the swipe styles at top level without a wrapper, and then and only @layer-wrapping the chrome? I feel like no layer = required makes the a bit more clear.

patrickkettner (commented):
[regarding your IntersectionObserver question](https://github.com/GoogleChrome/guidance/pull/720#discussion_r3224192865)... (I can't directly reply since the change)

I feel like its worth the complexity, given the point of the repo. the comment only path puts the burden on the reader to remember it, and the failure mode is silent. `display: none` at init isnt exotic,  accordions, tabs, modals, all hit it.

Adding it doubles as the perf win you mentioned, and the shape is small:

```js
const visObs = new IntersectionObserver((entries) => {
  for (const entry of entries) {
    if (entry.intersectionRatio > 0) {
      setupItem(entry.target);
      visObs.unobserve(entry.target);
    }
  }
}, {threshold: 0});
document.querySelectorAll('.SwipeableItem').forEach(item => visObs.observe(item));
```

bonus, this also makes the is-initialized guard redundant. Im not gonna advocate too hard for it, but I do think adding it is the right call.

guides/user-experience/swipe-to-remove/expectations.md
philipwalton: Agreed. I removed it from expectations but kept it in the guide.

philipwalton (commented):
@patrickkettner I added an outer ResizeObserver as well as a MutationObserver to better handle dynamic lists. I think this now addresses all of your initial comments.

rviscomi (approved):
LGTM

PR #722, opened 2026-05-05
Fine-tune existing evals

micahjo7:
Updates graders, guides, tasks for the existing evals with the goal of improving the overall guided pass rate.

A guided pass rate of 95% was achieved with these changes (36% uplift), [here](https://googlechrome.github.io/guidance/dashboard.html?testId=full-2026-05-05T09-50-20&source=remote).


paulirish:
Nice stuff! How'd you do it? What was the operation?

are there any updates we can make to the eval-investigation skill?
or maybe make a eval-investigation/bulk-investigation-and-upgrade.md guide?

harness/base_apps/devtools-times/astro.config.mjs
```diff
@@ -25,28 +25,7 @@ export default defineConfig({
       },
      },
   },
-  experimental: {
```
paulirish: YUP. #656 has the same. sry for the delay on that one

harness/base_apps/devtools-times/src/content.config.ts
```diff
@@ -1,6 +1,8 @@
 import { defineCollection, z } from 'astro:content';
+import { glob } from 'astro/loaders';
 
 const articlesCollection = defineCollection({
+  loader: glob({ pattern: '**/[^_]*.{md,mdx}', base: "./src/content/articles" }),
```
paulirish: yup. exact same here too.

harness/base_apps/devtools-times/db/seed.ts
```diff
@@ -34,7 +34,7 @@ export default async function() {
       }
       
       commentsToSeed.push({
-        slug: article.slug,
```
paulirish: you'll have a conflict here but... i delete the file so thats straightforward.

guides/user-experience/declarative-button-actions/guide.md
```diff
@@ -46,14 +46,12 @@ For custom, application-specific actions, you can define your own command names.
 </button>
 
 <script>
-  // Listen for the 'command' event globally (or on a stable parent)
-  document.addEventListener('command', (event) => {
+  // Listen for the 'command' event directly on the target element
+  // (This is necessary because the native 'command' event does not bubble)
```
paulirish, line 50: wow. 

this is a HUGE find. and true!

https://open-ui.org/components/invokers.explainer/#:~:text=The%20dispatched,trusted%2C%20cancellable%20events

> Invokers will dispatch an CommandEvent on the Invokee (the element referenced by commandfor) when the element is Invoked. The CommandEvent’s type is always command. The event also contains an source property that will reference button that was enacted causing the CommandEvent to fire. The dispatched CommandEvents are always **non-bubbling**, composed, trusted, cancellable events.

paulirish: cc @dvdherron

micahjo7:
This took several iterations of running the full suite and handing the results to the agent (telling it to look at a specified result path). It worked best when I looked through the results and handpicked which tasks were not performing well, and listing them off to the agent. Asking the agent to "examine all the underperforming tasks" or "tasks that had unguided percentage under 90%" did not work well.

I asked the agent to look at the grader files FIRST (something like "make sure the grader is robust"), then the guide, then task (and only move on to the next step if the previous artifact didn't need any changes).

Additionally I had to prompt it a couple times to not care about the demo file performance but focus solely on the guided runs.

paulirish (approved):

guides/performance/batch-analytics-events/guide.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: batch-analytics-events
-description: Debounce and batch multiple analytics events together in a single beacon to minimize network contention and reduce server load, while still delivering real-time updates.
+description: Debounce and batch multiple analytics events or Core Web Vitals together in a single beacon using the web-vitals library to minimize network contention and reduce server load, while still delivering real-time updates.
```
micahjo7, line 3: @rviscomi @philipwalton wanted to double check this change with you- how closely should web-vitals be tied to this use case (the task mentions it)?

this has generally been one of the worse-performing evals and this seems to be what my agent suggested to improve the performance

rviscomi: CWV is a realistic example of a thing you'd want to batch up and send to your analytics, so it makes sense in the prompt. But the guide as a whole is totally independent from CWV and the web-vitals library.

rviscomi:
@micahjo7 @paulirish for future fine tuning work, if you identify that a use case's `guide.md` should be changed, I'd prefer a process that ensures at least one person from DevRel gives an LGTM before making changes. Ideally:

- reopen the corresponding use case issue and set the project status to "Needs investigation"
- leave a comment in the issue describing what needs investigation or if you have a suggested change
- someone from DevRel will reassign the issue to the closest SME
- or if you open a PR to fix it yourself, merging is blocked until you get an LGTM from one of the code owners

This wouldn't apply to any changes to the grader or task files.

guides/performance/batch-analytics-events/guide.md
rviscomi: @micahjo7 yeah I think we should revert this change

It's also worth taking a closer look at the other guide/expectations changes in this PR:

- [guides/user-experience/calculate-event-differentials/expectations.md](https://github.com/GoogleChrome/guidance/pull/722/changes#diff-78ee66ead65292dab9f40877b92f21dc9d0b04c4cc95437e888fa78e5e8fdf96)
- [guides/user-experience/child-state-based-styling/expectations.md](https://github.com/GoogleChrome/guidance/pull/722/changes#diff-9f3eded1b21f81b06b94fdaba8a5e15f03427a55440292fd31efc34816e7f956)
- [guides/user-experience/customize-scrollbar-color-and-thickness/expectations.md](https://github.com/GoogleChrome/guidance/pull/722/changes#diff-245631eee931c69a42937ba7f4afad94e1c8f332abe09b410ea3f7f9f107e15c)
- [guides/user-experience/declarative-button-actions/guide.md](https://github.com/GoogleChrome/guidance/pull/722/changes#diff-29a24a3749ea264b4b03ee8a389b43d4dab57ee8b1c5b171b2c96163f5b847a6)
- [guides/user-experience/visually-stable-mixed-fonts/guide.md](https://github.com/GoogleChrome/guidance/pull/722/changes#diff-ac81f7394c539cdf31486059e376f22782e283c234b0c3dc5b08e2c758513697)

micahjo7: adressed in https://github.com/GoogleChrome/guidance/pull/769

PR #738, opened 2026-05-07
feat: skill for web component

### LINKED ISSUE #714: Create a high-level Web Components skill

castastrophe:
## Summary

Aligned the output with the approach taken for passkeys - an orientation page linking to specific and focused guides.

Guides were broken into the following topics:

- accessible-web-components
- custom-elements
- declarative-shadow-dom
- form-associated-custom-elements
- shadow-dom
- styling-web-components

### DIFF
--- a/guides/html/web-components/accessible-web-components/guide.md
+++ b/guides/html/web-components/accessible-web-components/guide.md
@@ -1,0 +1,54 @@
+---
+name: accessible-web-components
+description: Make custom elements accessible by default, setting roles and ARIA state via the ElementInternals ARIA mixin, managing focus with delegatesFocus, and keeping ARIA relationships within a single shadow root.
+web-feature-ids:
+  - aria-attribute-reflection
+  - shadow-dom
+---
+
+# Accessible Web Components
+
+A shadow boundary changes how semantics and ARIA work. Two failures dominate: missing semantics inside the shadow tree, and ARIA references that silently break because they cannot cross shadow roots.
+
+Start from semantic markup: use real `<button>`, `<nav>`, `<ul>` inside the template so most semantics come for free, and reserve `ElementInternals`/ARIA for the gaps.
+
+## Default semantics via the ElementInternals ARIA mixin
+
+Set a custom element's role and ARIA state in JS through `ElementInternals`, rather than writing ARIA attributes onto the host. This keeps the semantics as defaults the component owns, while still letting a consumer override them with attributes in the Light DOM.
+
+```javascript
+class ToggleSwitch extends HTMLElement {
+  #internals;
+  constructor() {
+    super();
+    this.#internals = this.attachInternals();
+    this.#internals.role = 'switch';          // default role, no host attribute needed
+    this.#internals.ariaChecked = 'false';    // exposed to the accessibility tree
+  }
+  toggle() {
+    const on = this.#internals.ariaChecked !== 'true';
+    this.#internals.ariaChecked = String(on);
+  }
+}
+customElements.define('toggle-switch', ToggleSwitch);
+```
+
+This avoids polluting the DOM with ARIA attributes and prevents consumers from accidentally clobbering them, while the accessibility tree still sees the correct role and state.
+
+## Focus management and `delegatesFocus`
+
+- Attach the shadow root with `delegatesFocus: true` when the component wraps focusable controls. Focusing the host then forwards focus to the first focusable child, and `:focus`/label clicks behave as users expect.
+- For roving focus or moving focus into a newly revealed region, give the target `tabindex="-1"` and call `.focus()`; make programmatically-focusable targets explicit rather than relying on tab order.
+
+## ARIA relationships cannot cross shadow roots
+
+- **MANDATORY**: an `id` referenced by `aria-labelledby`, `aria-describedby`, `aria-controls`, etc. must live in the **same** tree as the referencing element. An attribute in the Light DOM cannot point at an `id` inside the shadow tree, and vice versa; the reference is silently ignored.
+- Keep each ARIA relationship within one root. If a label in the shadow tree must describe a slotted Light DOM element, move the relationship into one tree (e.g. render the label in the Light DOM, or use `aria-label` text instead of an `id` reference).
+- For element-to-element references that genuinely must cross roots, the emerging **Reference Target** mechanism is the standards direction, but it is not yet broadly available; design to keep references within a single tree today.
+- The reflected ARIA *element* properties — `ariaLabelledByElements`, `ariaDescribedByElements`, and siblings — let you set these relationships to actual element references in JS, with no `id` needed. Browser support is still limited and the rules for referencing across shadow roots are constrained and evolving, so verify before relying on them; today, treat them mainly as a cleaner way to wire *same-tree* relationships without minting ids.
+
+## Fallback strategies
+
+{{ BASELINE_STATUS("aria-attribute-reflection") }}
+
+Where the `ElementInternals` ARIA mixin is unavailable, set the role and ARIA state as attributes on the host instead (`this.setAttribute('role', 'switch')`). It works everywhere, at the cost of putting attributes in the DOM that a consumer could override.
--- a/guides/html/web-components/custom-elements/guide.md
+++ b/guides/html/web-components/custom-elements/guide.md
@@ -1,0 +1,125 @@
+---
+name: custom-elements
+description: Define, register, and manage the lifecycle of autonomous custom elements, covering constructor rules, lifecycle callbacks, upgrade timing, attribute/property reflection, and naming conventions.
+web-feature-ids:
+  - autonomous-custom-elements
+---
+
+# Custom Elements
+
+A custom element is a class registered against a tag name via `customElements.define()`. Registration and basic lifecycle are well understood; the rules below target the parts routinely implemented incorrectly.
+
+## Lifecycle: the parts that get implemented wrong
+
+The callbacks themselves are well documented; these are the misconceptions that cause bugs:
+
+- **`connectedCallback` can run more than once.** It fires on every insertion, so moving the element re-runs it. Make setup idempotent — don't assume a single first run.
+- **`connectedMoveCallback` skips the disconnect/reconnect pair on a move.** When an element is relocated with `Node.moveBefore()`, this callback fires *instead of* `disconnectedCallback` → `connectedCallback`, so state, focus, and iframe/media playback survive the move. Define it when your setup or teardown is expensive or observably re-runs.
+- **Tear down in `disconnectedCallback`.** Remove global listeners, timers, and observers or you leak; but the element may be re-inserted, so don't treat teardown as permanent.
+- **Do not depend on the full DOM being present in the constructor.** Attributes are usually (but not always) present; long lists of children may not have finished parsing. Use `connectedCallback` for one-shot initialization, `slotchange` to monitor direct children, and `attributeChangedCallback` to monitor attributes. `connectedCallback` may still fire before all children are parsed, so do not ignore subsequent `slotchange` / `attributeChangedCallback` calls.
+- **`attributeChangedCallback` fires only for `observedAttributes`,** and should stay cheap for attributes that don't affect output.
+- **Lifecycle hooks and `observedAttributes` are snapshotted at registration.** `customElements.define()` reads both once; adding a hook or extending `observedAttributes` on the class *after* the call is silently ignored. In particular, `attributeChangedCallback` never fires unless `observedAttributes` was already present at `define()` time.
+
+## Upgrade timing and `:defined`
+
+An element can exist in the DOM *before* its class is registered (in server-rendered HTML, or when the script loads late). Until defined it is an inert `HTMLElement` with none of your behavior; when `customElements.define()` runs, the browser **upgrades** every matching element already in the document, running its constructor and `connectedCallback` then.
+
+Two consequences models routinely miss:
+
+1. **Attributes and children may already be present** when your constructor/`connectedCallback` runs on upgrade. Read existing attributes in `connectedCallback` to sync initial state, and never assume the element started empty.
+2. **Children may still be missing** even in `connectedCallback` — the HTML parser can yield mid-parse of long child lists, so the callback sometimes runs before every child is attached. See the constructor bullet in *Lifecycle* above for the pattern (`slotchange` + `attributeChangedCallback`).
+
+Hide undefined elements to avoid a flash of unstyled content — but **fail open**, so the content stays reachable if the script never loads. Animate `opacity` with a delay rather than hard-hiding with `visibility: hidden`, so the element reveals itself after a timeout even when it is never upgraded:
+
+```css
+my-element:not(:defined) {
+  opacity: 0;
+  /* Reveal anyway after 2s, so a failed or slow script never hides content for good. */
+  animation: defined-fallback 0.2s 2s forwards;
+}
+@keyframes defined-fallback { to { opacity: 1; } }
+```
+
+Scope these guards to the elements that need them — or to a specific subtree via `:has(:not(:defined))` — rather than the whole `body`. One slow, non-critical widget (a share button, say) should not delay paint for the entire page.
+
+## Attribute ⇄ property reflection
+
+Keep the HTML attribute and the JS property in sync so the element is usable from both markup and script. Attribute values are **always strings**; coerce them.
+
+```javascript
+class MyToggle extends HTMLElement {
+  static observedAttributes = ['label', 'disabled'];
+
+  // String attribute reflected via getter/setter.
+  get label() { return this.getAttribute('label') ?? ''; }
+  set label(v) {
+    v ? this.setAttribute('label', v) : this.removeAttribute('label');
+  }
+
+  // Boolean attribute: present = true, absent = false. Never set it to "false".
+  get disabled() { return this.hasAttribute('disabled'); }
+  set disabled(on) { this.toggleAttribute('disabled', !!on); }
+
+  attributeChangedCallback(name, oldVal, newVal) {
+    if (oldVal !== newVal) this.render();
+  }
+}
+```
+
+### Where the value lives
+
+The example above keeps the **DOM as the source of truth** — each getter reads `getAttribute`. That's the right default for a handful of simple, string-ish attributes: there is only one place the state can live, so it can't desync.
+
+For richer components, keep the **source of truth in a JS field** and reflect *outward* to an attribute only when something else must see it — CSS targeting, or a consumer reading it. Reading a field is far cheaper than parsing and coercing an attribute on every access, and non-string values never round-trip through serialization. This is why libraries like Lit hold state in JS and reflect selectively rather than treating attributes as storage: reflect only what the outside world must observe, not every property.
+
+Pitfalls:
+
+- **`observedAttributes` discipline**: list only attributes whose change must re-render or trigger a side effect. Observing everything adds a callback to every mutation for no benefit.
+- **Reflection loops**: a setter that writes an attribute will re-enter `attributeChangedCallback`. Guard with an `oldVal !== newVal` check, as above.
+- **Style-only state**: for internal state that exists only to drive styling (loading, active, expanded), prefer **custom states** over reflecting a synthetic attribute — see [Custom states](#custom-states) below. Reflect to an attribute only when consumers must be able to set the state from markup.
+
+## Custom states
+
+Custom states are boolean flags an element exposes *about itself* for CSS to target, set through `ElementInternals` and matched with the `:state()` pseudo-class. Use them for internal, runtime UI state — `busy`, `expanded`, `invalid` — instead of reflecting a synthetic attribute: they can't collide with a consumer's attributes, never appear in the DOM, and are read-only from outside (a consumer can *style* `:state()` but can't forge it).
+
+```javascript
+class AsyncButton extends HTMLElement {
+  #internals;
+  #action = async () => {};
+
+  constructor() {
+    super();
+    this.#internals = this.attachInternals();
+  }
+
+  set action(fn) { this.#action = fn; } // host supplies the work to run
+
+  async run() {
+    if (this.#internals.states.has('busy')) return; // guard re-entry
+    this.#internals.states.add('busy');
+    this.#internals.states.delete('success');
+    try {
+      await this.#action();
+      this.#internals.states.add('success');
+    } catch {
+      this.#internals.states.add('error');
+    } finally {
+      this.#internals.states.delete('busy');
+    }
+  }
+}
+```
+
+`CustomStateSet` is a set, so independent states compose (`busy` plus `error`) without inventing a combined `data-status` enum.
+
+**Custom state is runtime-only.** It is not serialized to the DOM, so it does not survive a reload, the element being re-created, or server rendering. If the state must be set declaratively in markup or persist across loads, that is a job for an **attribute** (see [Where the value lives](#where-the-value-lives)) — restore it in `connectedCallback` from your source of truth. Rule of thumb: custom state for what the component derives and owns at runtime and only needs to *style*; an attribute for what must be set from markup or serialized.
+
+For styling these states — both inside the component and from a consumer's stylesheet — see {{ GUIDE_REF("web-components/styling-web-components") }}.
+
+## Naming conventions
+
+- True-private state: native `#` fields (`#count`) for hard encapsulation; a `_` prefix only when subclasses must still reach it. Note `#` fields are not inherited — a subclass cannot see a base class's private fields.
+- Symbol-keyed properties are a middle ground: hidden from ordinary enumeration, but reachable by any code holding the symbol (export it, or expose it as a `static` field on the class). Because each symbol is unique, two symbols sharing a description never collide — handy when a subclass or sibling needs controlled access that `#` fields can't provide.
+- Do **not** shadow global attributes (`style`, `class`, `id`, `slot`, `part`, `title`, `lang`, `dir`) with your own properties; doing so breaks their built-in behavior. Note `disabled` is only global on form controls; on other elements you must implement its effect yourself.
+
+When naming a new API surface, follow the platform's own conventions — see the [W3C naming principles](https://www.w3.org/TR/design-principles/#naming-is-hard).
--- a/guides/html/web-components/form-associated-custom-elements/guide.md
+++ b/guides/html/web-components/form-associated-custom-elements/guide.md
@@ -1,0 +1,67 @@
+---
+name: form-associated-custom-elements
+description: Make a custom control participate natively in forms with formAssociated and ElementInternals, submitting its own value, integrating with constraint validation, and reacting to reset, disable, and state-restore.
+web-feature-ids:
+  - form-associated-custom-elements
+---
+
+# Form-Associated Custom Elements
+
+Do not fall back to a hidden `<input>` to surface a custom control's value to a form. Use **Form-Associated Custom Elements** (FACE): with `formAssociated` and `ElementInternals`, the element submits its own value, participates in validation, and reflects form state natively.
+
+```javascript
+class RatingInput extends HTMLElement {
+  // MANDATORY: opt in. Without this, attachInternals() form APIs are unavailable.
+  static formAssociated = true;
+
+  #internals;
+  #value = '';
+
+  constructor() {
+    super();
+    this.#internals = this.attachInternals();
+    this.attachShadow({ mode: 'open' });
+  }
+
+  get value() { return this.#value; }
+  set value(v) {
+    this.#value = v;
+    // Push the value into the owning <form> for submission.
+    this.#internals.setFormValue(v);
+    this.#validate();
+  }
+
+  #validate() {
+    if (this.hasAttribute('required') && !this.#value) {
+      // Mark invalid and supply the native validation message + anchor.
+      this.#internals.setValidity(
+        { valueMissing: true },
+        'Please choose a rating.',
+        this.shadowRoot.querySelector('[tabindex]'),
+      );
+    } else {
+      this.#internals.setValidity({}); // clear
+    }
+  }
+
+  // Optional FACE lifecycle hooks:
+  formResetCallback() { this.value = ''; }
+  formDisabledCallback(disabled) { this.toggleAttribute('disabled', disabled); }
+  formStateRestoreCallback(state) { this.value = state; }
+}
+customElements.define('rating-input', RatingInput);
+```
+
+Key points:
+
+- `static formAssociated = true` is **MANDATORY**; it unlocks the form-related `ElementInternals` methods and the `formXxxCallback` lifecycle.
+- `setFormValue(value)` is what submits, with no hidden input required. Pass a `FormData` for multi-value controls.
+- `setValidity(flags, message, anchor)` integrates with native constraint validation: the form won't submit while invalid, `:invalid`/`:valid` apply, and the anchor element receives focus on report. Call `setValidity({})` to clear. The interaction-aware `:user-valid`/`:user-invalid` pseudo-classes are *not* consistently wired up for form-associated elements across browsers — verify support before relying on them rather than assuming they track user interaction the way they do for native controls.
+- Implement `formResetCallback`, `formDisabledCallback`, and `formStateRestoreCallback` so the element behaves like a native control on reset, `fieldset[disabled]`, and back/forward autofill restore.
+- Follow platform conventions for dispatched events and prefer the built-in `input` and `change` events over custom ones. `input` fires when the value changes as a direct result of user action (typing, dragging a slider); `change` fires once per stream of changes, when the user commits (releasing the slider, blurring the field). Construct them with `InputEvent` and `Event` respectively.
+
+## Fallback strategies
+
+{{ BASELINE_STATUS("form-associated-custom-elements") }}
+
+Feature-detect with `static formAssociated` support before relying on it: `'attachInternals' in HTMLElement.prototype && 'setFormValue' in ElementInternals.prototype`. Where it is unavailable, fall back to a visually-hidden native `<input>` synced to the control's value so the form still submits — the one case where the hidden-input pattern is justified.
--- a/guides/html/web-components/prerendering-custom-elements/guide.md
+++ b/guides/html/web-components/prerendering-custom-elements/guide.md
@@ -1,0 +1,67 @@
+---
+name: prerendering-custom-elements
+description: Server-render a custom element, then hydrate it in place without re-rendering or a flash of unstyled content.
+web-feature-ids:
+  - declarative-shadow-dom
+---
+
+# Pre-rendering custom elements
+
+Pre-rendering means shipping a custom element's markup — including its shadow tree — from the server, so the component is styled and structured *before* any JavaScript runs. The main platform primitive for this is **Declarative Shadow DOM (DSD)**: the parser attaches the shadow root from HTML alone, and the element is later upgraded ("hydrated") by its class if one is registered. Use this pattern for server-side rendering, streaming, and avoiding a flash of unstyled content (FOUC).
+
+## Authoring a declarative shadow root
+
+Nest a `<template>` with the `shadowrootmode` attribute as the first child of the host. The parser attaches its content as a shadow root and removes the template; no script needed.
+
+```html
+<my-card>
+  <template shadowrootmode="open">
+    <style>
+      .card { border: 1px solid; border-radius: 8px; padding: 1rem; }
+    </style>
+    <div class="card">
+      <slot name="header"></slot>
+      <div class="body"><slot></slot></div>
+    </div>
+  </template>
+
+  <!-- Light DOM children are projected into the slots above. -->
+  <h2 slot="header">Server-rendered title</h2>
+  <p>Visible and styled before JS loads.</p>
+</my-card>
+```
+
+## Hydration timing
+
+When the matching class is registered, the browser upgrades the host. Because the shadow root **already exists**, the constructor must not blindly create a second one:
+
+```javascript
+class MyCard extends HTMLElement {
+  constructor() {
+    super();
+    // Reuse the server-rendered shadow root if present; only create one if absent.
+    if (!this.shadowRoot) {
+      this.attachShadow({ mode: 'open' });
+      this.shadowRoot.append(template.content.cloneNode(true));
+    }
+    // Otherwise hydrate in place: wire up listeners against existing markup,
+    // and do NOT re-render; that would discard the server's DOM and flash.
+  }
+}
+```
+
+Set `shadowrootmode="open"` (not `closed`) for almost all cases, so scripts and tools can reach `this.shadowRoot` to hydrate. To serialize an existing shadow root back to HTML on the server or in tests, use `getHTML({ serializableShadowRoots: true })` and attach the root with `serializable: true`.
+
+## Avoiding FOUC
+
+DSD renders correctly on first paint only if styles ship inside the declarative template. To preserve that:
+
+- Put component CSS **inside** the `<template shadowrootmode>` (a `<style>` tag or a `<link>`), not in a separate script-loaded sheet.
+- Pair with `my-card:not(:defined) { … }` only for styles that must wait for JS — typically page-level guard rules like a fade-out on client-rendered instances that miss the SSR pass. These live in the light-DOM (page) stylesheet, since the shadow root doesn't yet exist there and `:host` selectors don't apply from outside. Structural styles inside the DSD template need no such guard.
+- Do not re-render the shadow tree on hydration. Treat the server markup as authoritative and attach behavior to it.
+
+## Fallback strategies
+
+{{ BASELINE_STATUS("declarative-shadow-dom") }}
+
+In engines that don't parse `shadowrootmode`, the template is inert and its content is not attached as a shadow root. Detect this (`HTMLTemplateElement.prototype.hasOwnProperty('shadowRootMode')`) and, if absent, attach the shadow root imperatively from the same markup during upgrade so the component still renders.
--- a/guides/html/web-components/styling-web-components/guide.md
+++ b/guides/html/web-components/styling-web-components/guide.md
@@ -1,0 +1,132 @@
+---
+name: styling-web-components
+description: Style across the shadow boundary with :host selectors, inherited and custom properties for theming, ::slotted and ::part, and cascade layers and container queries inside components.
+web-feature-ids:
+  - shadow-parts
+  - host-context
+  - cascade-layers
+  - container-queries
+---
+
+# Styling Web Components
+
+Shadow DOM is a style boundary: page rules don't reach in and component rules don't leak out, except **inherited** properties (including custom properties). Cross it deliberately through the channels below.
+
+## `:host`, `:host()`, `:host-context()`
+
+```css
+/* The component's own host element, styled from inside its shadow tree. */
+:host { display: block; }
+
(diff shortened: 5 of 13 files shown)


castastrophe:
@LeaVerou This is ready for your review. If you want to start with a high-level, organizational, structural review, that would be most helpful to me in case we need to make any major adjustments. Otherwise, I'm ready to get into the details.

guides/web-components/custom-elements/guide.md
```diff
@@ -0,0 +1,98 @@
...
+    if (oldVal !== newVal) this.render();
+  }
+}
+```
+
+Pitfalls:
+
+- **`observedAttributes` discipline**: list only attributes whose change must re-render or trigger a side effect. Observing everything adds a callback to every mutation for no benefit.
+- **Reflection loops**: a setter that writes an attribute will re-enter `attributeChangedCallback`. Guard with an `oldVal !== newVal` check, as above.
+- **Style-only state**: reflect a property to an attribute purely for CSS targeting (`:host([loading])`) *without* observing it, but only when the component is the sole writer of that attribute.
```
LeaVerou: We should recommend custom states for this, that's what they're for :)

guides/web-components/custom-elements/guide.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Naming conventions
+
+- Public properties and methods: `camelCase`, matching DOM conventions (`this.userName`, `this.refresh()`).
+- True-private state: native `#` fields (`#count`) for hard encapsulation; a `_` prefix only when subclasses must still reach it.
```
LeaVerou: Another pattern for hidden-but-not-completely is to use symbol-keyed properties and export them or expose them on the class as static properties. And because symbols are unique, you can create two symbols with the same name and have no clashes :)

guides/web-components/custom-elements/guide.md
```diff
@@ -0,0 +1,98 @@
...
+Two consequences models routinely miss:
+
+1. **Attributes and children may already be present** when your constructor/`connectedCallback` runs on upgrade. Read existing attributes in `connectedCallback` to sync initial state, and never assume the element started empty.
+2. **Children may still be missing** even in `connectedCallback` if the HTML parser hasn't reached them yet. If you depend on Light DOM children, wait for them: use a `slotchange` listener, or defer with `customElements.whenDefined()`.
+
+Style undefined elements out of view to avoid a flash of unstyled content:
+
+```css
+/* Hide the element until its definition has upgraded it. */
+my-element:not(:defined) { visibility: hidden; }
```
LeaVerou: This means if the WC fails to load, the content is inaccessible.
 
An animation over opacity with a delay works better as it can time out after e.g. 2s and still allow access to content, which is often better (and often looks nicer than stuff just appearing)

Also consider `:has(:not(:defined))` selectively (some folks apply it to the entire body, which I don't like since then a single WC that may not even matter much, e.g. a sharing widget, delays loading for the entire page)

guides/web-components/web-components/guide.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Decide whether you even need a custom element
+
+Reach for a custom element only when you need encapsulated behavior, lifecycle, or a reusable JS-backed API on a tag. Do **not** build one when:
+
+- A plain HTML element with CSS would do. A styled `<button>` or `<details>` is more accessible and cheaper than a re-implementation.
+- The thing is purely presentational with no behavior. A class on a native element is more performant, and scopable with `@layer` or `@scope`.
+- You only need style scoping on the server with no interactivity; a scoped stylesheet or Declarative Shadow DOM (without an upgrade) may be enough.
+
+Rebuilding native semantics (a custom `<my-button>` that re-creates focus, activation, and ARIA) is the most common misuse: you inherit the platform's accessibility for free only if you reuse the native element.
```
LeaVerou: I wholeheartedly agree (preach!), but `<my-button>` is so common we need to at least address the use cases that drive folks to recreate it. I think part of it is DX, but other times they genuinely need to do things that require extra markup and/or interactivity.

guides/web-components/custom-elements/guide.md
```diff
@@ -0,0 +1,98 @@
...
+
+Pitfalls:
+
+- **`observedAttributes` discipline**: list only attributes whose change must re-render or trigger a side effect. Observing everything adds a callback to every mutation for no benefit.
+- **Reflection loops**: a setter that writes an attribute will re-enter `attributeChangedCallback`. Guard with an `oldVal !== newVal` check, as above.
+- **Style-only state**: reflect a property to an attribute purely for CSS targeting (`:host([loading])`) *without* observing it, but only when the component is the sole writer of that attribute.
+
+## Naming conventions
+
+- Public properties and methods: `camelCase`, matching DOM conventions (`this.userName`, `this.refresh()`).
```
LeaVerou: We could link to the official guidelines here: https://www.w3.org/TR/design-principles/#naming-is-hard

guides/html/web-components/accessible-web-components/guide.md
```diff
@@ -0,0 +1,49 @@
...
+- For roving focus or moving focus into a newly revealed region, give the target `tabindex="-1"` and call `.focus()`; make programmatically-focusable targets explicit rather than relying on tab order.
+
+## ARIA relationships cannot cross shadow roots
+
+- **MANDATORY**: an `id` referenced by `aria-labelledby`, `aria-describedby`, `aria-controls`, etc. must live in the **same** tree as the referencing element. An attribute in the Light DOM cannot point at an `id` inside the shadow tree, and vice versa; the reference is silently ignored.
+- Keep each ARIA relationship within one root. If a label in the shadow tree must describe a slotted Light DOM element, move the relationship into one tree (e.g. render the label in the Light DOM, or use `aria-label` text instead of an `id` reference).
+- For element-to-element references that genuinely must cross roots, the emerging **Reference Target** mechanism is the standards direction, but it is not yet broadly available; design to keep references within a single tree today.
```
LeaVerou, line 47: I thought (but need to dig in more to verify as I haven't tried it) that some of the aria JS properties support setters and we can use them to set to actual element references (that don't even _need_ an id).

E.g. see https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaLabelledByElements
and in the spec they do seem to be writable: https://w3c.github.io/aria/#ARIAMixin

guides/html/web-components/styling-web-components/guide.md
```diff
@@ -0,0 +1,94 @@
...
+
+```css
+/* The component's own host element, styled from inside its shadow tree. */
+:host { display: block; }
+
+/* Only when the host matches a selector (e.g. a reflected state attribute). */
+:host([disabled]) { opacity: 0.5; pointer-events: none; }
+
+/* Based on an ancestor in the Light DOM; useful for theming. */
+:host-context([theme="dark"]) { background: #111; color: #eee; }
```
LeaVerou, line 25: We should actively recommend against `:host-context()`, see https://github.com/w3c/csswg-drafts/issues/1914#issuecomment-2737310093

guides/web-components/form-associated-custom-elements/guide.md
```diff
@@ -0,0 +1,62 @@
...
+  formStateRestoreCallback(state) { this.value = state; }
+}
+customElements.define('rating-input', RatingInput);
+```
+
+Key points:
+
+- `static formAssociated = true` is **MANDATORY**; it unlocks the form-related `ElementInternals` methods and the `formXxxCallback` lifecycle.
+- `setFormValue(value)` is what submits, with no hidden input required. Pass a `FormData` for multi-value controls.
+- `setValidity(flags, message, anchor)` integrates with native constraint validation: the form won't submit while invalid, `:invalid`/`:valid` apply, and the anchor element receives focus on report. Call `setValidity({})` to clear.
```
LeaVerou: We should mention `:user-valid`/`:user-invalid`, even if it's just to say that there is no way to integrate with them (I don't remember the current state of things)

guides/html/web-components/form-associated-custom-elements/guide.md
```diff
@@ -0,0 +1,62 @@
...
+customElements.define('rating-input', RatingInput);
+```
+
+Key points:
+
+- `static formAssociated = true` is **MANDATORY**; it unlocks the form-related `ElementInternals` methods and the `formXxxCallback` lifecycle.
+- `setFormValue(value)` is what submits, with no hidden input required. Pass a `FormData` for multi-value controls.
+- `setValidity(flags, message, anchor)` integrates with native constraint validation: the form won't submit while invalid, `:invalid`/`:valid` apply, and the anchor element receives focus on report. Call `setValidity({})` to clear.
+- Implement `formResetCallback`, `formDisabledCallback`, and `formStateRestoreCallback` so the element behaves like a native control on reset, `fieldset[disabled]`, and back/forward autofill restore.
+
```
LeaVerou, line 64:
```suggestion
- Follow platform conventions for dispatched events and prefer built-in `input` and `change` events instead of custom versions. The `input` event fires when the value of a control has been changed as a direct result of a user action (such as typing in a textbox or dragging a slider) whereas `change` is generally fired once per stream of changes, often when the user commits the change explicitly (e.g. releases the slider handle or focuses away). Use `InputEvent` to construct `input` events and `Event` to construct `change` events.

```

LeaVerou (changes requested):
## Big picture

There is some **great** technical content here! 
The two big picture issues I see are:
1. A lot of the content is structured around technologies and APIs, rather than use cases / user needs.
2. There is more reference material than is necessary, about things that are old enough to be well within agents' training sets.  Things like `customElements.define()`, how `super()` works, how slots work, formAssociated, `::part()` etc. In my experience agents seem to know about all the APIs up until custom states or so (of course best practices are still useful even for older stuff, it's the reference material they don't need).

IMO the top goals of these guides should be to answer these questions, in descending order of priority:
1. **When** to reach for a web component? **This is the top priority:** If this fails, 2 & 3 never become relevant, since the agent will not look for best practices and implementation guides for web components unless it has already decided to use one.
2. What are the **best practices**? What are the hidden caveats? What are some known impactful browser bugs the agent should work around? How to decide between alternatives? What are the established platform conventions the agent should follow?
3. **How** to do it? (mainly for very cutting edge things that are not in the training set or features we know they use poorly or not at all)

Currently 3 is larger than it should be and covers a lot of ground that is common knowledge, whereas 1 (especially) and 2 are much smaller than they should be.

Even for the technologies described (e.g. slots, parts, shadow DOM, etc), it could use more guidance around when to use what, which is far more valuable than how APIs work and what their syntax is (once they know what API they need to use agents can figure these things out easily).

### 1. When to reach for a web component?

First, it would be useful to include guidance about when to reach for a component *at all*, regardless of technology. Agents are notoriously bad at figuring out when to create abstractions, and default towards additive WET code.

For WC specifically, [this section](https://github.com/GoogleChrome/modern-web-guidance-src/pull/738/changes#diff-29420cea4ba9dce811c5fb79f2484eabf294355433a0cc5985d8daa3165b926cR12-R20) is a **great** start.

One thing that's missing is a decision framework that takes into account what the codebase is already using (if not greenfield):
- In a codebase that uses a framework with its own component technology (e.g. Vue), is there value in ever reaching for a web component or should the agent just always create a framework component? If the former, how to decide between the two?
- In a codebase that *doesn't* use any tooling yet, should the agent reach for a web component to encapsulate markup that doesn't need interactivity?
- In a codebase using a static site generator, when to reach for a web component vs statically rendering markup? How to handle cases where we want a little bit of interactivity and a lot of markup? It's rarely b&w.
- How far to take the effort to use a built-in element? E.g.
  - If we can do what we want on a built-in element through nonstandard pseudo-elements and other hacks, is that preferable to a web component?
  - If we can do what we want, but the DX is worse, is that preferable to a web component? How much worse do we tolerate?
  - If we can do what we want but it requires running some JS on the built-in element (e.g. to attach listeners), is that preferable to a web component?
- etc

(I do not have the answer to all of these questions and often grapple with them myself, but any guidance we can provide is probably better than what the agent will decide on its own)

### 2. Best practices, caveats, decision frameworks

After guiding the agent about when to use a web component and when not to, the 2nd priority is best practices. How to implement styling, a11y, DX, UX? 

E.g. a decision framework for questions like:
- When to reach for a library like Lit and when to implement things itself?
- in cases where there are multiple ways to do something, which one is best?
For example, currently property-attribute reflection recommends using the DOM as the source of truth. This works for simple cases, but can be prohibitively slow for others, which is why most libraries use JS state and reflect it as needed.

These are good examples of the type of guidance that is most useful:
- https://github.com/GoogleChrome/modern-web-guidance-src/pull/738/changes#diff-8e5033f849a737b1d300f36f3b54006ffd2bfee41496bd0c976cc013312ad970R33-R40
- https://github.com/GoogleChrome/modern-web-guidance-src/pull/738/changes#diff-8e5033f849a737b1d300f36f3b54006ffd2bfee41496bd0c976cc013312ad970R90-R98
- https://github.com/GoogleChrome/modern-web-guidance-src/pull/738/changes#diff-0743b240ec4a8d3f1bf5d4a769388e09d85dc450ab1480801fa36cff674b50e2R19-R20

## Information architecture

I've been wondering how we can structure these guides to be more use case centric. I think in some cases the changes would be very minimal, even just a renaming (e.g. `form-associated-custom-elements` is already 99% there, except some of the guidance for custom form controls lives in other guides).

Just to take a stab at it, it seems the core use cases are:

- Generally when to use a component / custom element (`web-components` + some of the info in `custom-elements`)
- Custom form controls (current `form-associated-custom-elements` + some of the info from `accessible-web-components` + guidance for following platform conventions).
- Custom interactive widgets (tabs, carousels, drawers, accordions, etc)
- Templating
- Styling
- Rendering optimizations (SSR, preventing FOUC, etc)

## Additional guidance that could be useful

None of this is blocking, since it's additional improvements.

### Custom elements 

- When to use inheritance vs composition?
- Preserving JS properties set before upgrade (e.g. when using WCs with Vue or other frameworks that set properties or attributes on them independently of upgrade timing).

### Slots

- Guidance for which slot should be the default slot and when not to have one
- When to have custom elements for slotted children vs built-in ones? E.g. should `<my-tabs>` expect `<my-tab>` children or built-in elements? If the latter, should it accept any elements or a restricted set?
- How to implement auto-slotting for certain element types when the slot is unambiguous? E.g. consumers should never have to write `<my-tab slot=tab>`. Often we can just use the default slot and sidestep this altogether, but not always.

### Shadow DOM & styling

- When to sprout light DOM children (if ever) vs using shadow DOM? Including the well-known React bug around the former.
- When wrapping a native element is unavoidable (e.g. `<my-button>` wrapping a `<button>`), how to forward styles to the internal element? Custom properties? `display: contents` on the host + `all: inherit` on the button (anchor positioning breaks)? Give up on that and use parts? Something else?
- The known caveats around custom property registration and web fonts.
- Styling the slot itself to avoid additional wrappers
- What to _not_ expose as a part?
- Inheritance (of built-in properties) can often be a powerful tool to avoid introducing extra custom properties

## Other things

- For ElementInternals, I think it's more common to use `_internals` or a symbol than `#internals` which breaks inheritance.
- There are `BASELINE_STATUS` macros throughout the content, I think @rviscomi prefers these to be in a separate fallbacks section.
- I've also added some comments & suggestions around specific things and may add more.

Hope this is helpful! 🙏🏼

castastrophe:
Thanks for the incredibly thorough review, @LeaVerou. Definitely reshaped this skill for the better. Here's what's changed, mapped to your notes:

## Big picture

- Less reference material (2). Did a dedicated trim pass: removed the Glossary, the registration boilerplate (super()/define()), the lifecycle definition table, and the basic slot mechanics — keeping only the caveats/gotchas. The guides dropped ~50 lines net even with the new content below.
- More use-case / "when" framing (1). Added "Decide whether you even need a custom element" and "Match the tool to the codebase" (codebase-aware: framework present? SSG? how far to push a built-in element?). The full information-architecture reorg (separate interactive-widgets / templating / rendering-optimization guides) I've left for a separate discussion since it touches the whole set and I'd want your + @rviscomi/@malchata's input on the boundaries.

## Priorities

- P1 when-to-use → the two decision sections above.
- P2 best practices → including your property↔attribute example: new "Where the value lives" covers JS-state-as-source-of-truth vs DOM-as-truth for perf/complex cases.
- P3 cutting-edge how → added custom states, kept/sharpened DSD hydration, FACE, and the ElementInternals ARIA mixin.

**Inline comments**: all addressed: custom states for style-only state; symbol-keyed privacy pattern; fail-open :defined guard; <my-button> drivers (worked via a favorite-toggle); W3C naming link; recommend against :host-context(); input/change event conventions. For :user-valid/:user-invalid and ariaLabelledByElements I added hedged "verify support" notes rather than firm claims — happy to firm up if you can confirm current behavior.

## Other things

- BASELINE_STATUS → moved into per-guide Fallback strategies sections (@rviscomi's preference).
- #internals → _internals: held off on the rename for now (deliberate # privacy), but added the "# fields aren't inherited" caveat so the point lands.
- New "Forwarding styles to a wrapped native element" section (one of your additional-guidance items), scoped to the three agent traps: form controls not inheriting font, :host not cascading into the inner element, and the fragile display:contents + all:inherit trick.

Two new runnable demos (async-button for custom states, favorite-button for :state() styling) fold into the existing custom-elements / styling demos.

Still parked as follow-ups (your non-blocking list): inheritance vs composition, preserving JS props set before upgrade, default-slot/auto-slotting guidance, light-DOM sprouting + the React bug, custom-property registration + web fonts, styling the slot itself. Happy to take these in a follow-up PR.

guides/html/web-components/accessible-web-components/guide.md
```diff
@@ -0,0 +1,54 @@
...
+- The reflected ARIA *element* properties — `ariaLabelledByElements`, `ariaDescribedByElements`, and siblings — let you set these relationships to actual element references in JS, with no `id` needed. Browser support is still limited and the rules for referencing across shadow roots are constrained and evolving, so verify before relying on them; today, treat them mainly as a cleaner way to wire *same-tree* relationships without minting ids.
```
LeaVerou, line 48: Is there a web-feature id for this so we could use `BASELINE_STATUS`?

guides/web-components/custom-elements/guide.md
```diff
@@ -0,0 +1,134 @@
...
+
+A custom element is a class registered against a tag name via `customElements.define()`. Registration and basic lifecycle are well understood; the rules below target the parts routinely implemented incorrectly.
+
+## Lifecycle: the parts that get implemented wrong
+
+The callbacks themselves are well documented; these are the misconceptions that cause bugs:
+
+- **`connectedCallback` can run more than once.** It fires on every insertion, so moving the element re-runs it. Make setup idempotent — don't assume a single first run.
+- **Tear down in `disconnectedCallback`.** Remove global listeners, timers, and observers or you leak; but the element may be re-inserted, so don't treat teardown as permanent.
+- **The constructor can't see the DOM.** No attributes, children, or layout exist yet — read attributes and children in `connectedCallback` instead.
```
LeaVerou: `connectedCallback` may still be too early for children, see https://jsfiddle.net/WebComponents/fjd0m528/ and https://stackoverflow.com/questions/48663678/how-to-have-a-connectedcallback-for-when-all-child-custom-elements-have-been-c
The consensus seems to be that using `slotchange` and `attributeChangedCallback()` works better than `connectedCallback()` for this.
I _do_ think `connectedCallback()` for one-shot init is too big a convenience if authoring vanilla components, so I wouldn't personally discourage using it altogether, I would just add that caveat.

Re: The constructor can't see the DOM, the underlying mechanics are more nuanced than the guide implies. As a trivial counterexample see https://codepen.io/leaverou/pen/PwWJErY?editors=1011

The distinction is around **when the parser yields** (thanks @justinfagnani for explaining this to me!):
- You should almost always have the attributes available, because the parser really doesn't want to yield in the middle of an open tag. It's theoretically possible with an abnormally long attribute list, but exceedingly rare.
- But you might not have all children, because it's more likely to yield in long lists of children.
- `innerHTML` queues up all the CE reactions until "CE reaction time" or what we used to call nanotask timing, which is after the HTML is parsed and appended, but right before the `innerHTML` set call actually returns to JS. So in `innerHTML` you always see all the attributes and children.

Obviously, we don’t need all this detail here, just to loosen up the claim so that it's factually correct. 

Putting it all together, perhaps something like:
```suggestion
- **Do not depend on the full DOM being present in the constructor.** Attributes are usually (but not always) present, but long lists of children may not have finished parsing. Use `connectedCallback()` for one-shot initialization, `slotchange` to monitor direct children and `attributeChangedCallback()` to monitor attributes. Note that `connectedCallback()` may still fire too early, so do not ignore subsequent `slotchange` and/or `attributeChangedCallback()` calls.
```

guides/web-components/custom-elements/guide.md
```diff
@@ -0,0 +1,134 @@
...
+  animation: defined-fallback 0.2s 2s forwards;
+}
+@keyframes defined-fallback { to { opacity: 1; } }
+```
+
+Scope these guards to the elements that need them — or to a specific subtree via `:has(:not(:defined))` — rather than the whole `body`. One slow, non-critical widget (a share button, say) should not delay paint for the entire page.
+
+```javascript
+await customElements.whenDefined('my-element'); // resolves once registered
+```
```
LeaVerou: This snippet doesn't seem to connect with anything before it or after it, at first I thought it was part of an unfinished section, then I saw it was mentioned above. I think the earlier mention of the method is enough to send agents in the right direction, I'm not sure that this block adds much additional signal (agents know that they can await promises).

What might be useful (but I'd do it in a followup) could be showing nontrivial usage patterns, e.g. using `whenDefined()` together with `:defined` (possibly with a filter or a timeout) to automatically wait for all undefined light and/or shadow DOM descendants.

guides/html/web-components/custom-elements/guide.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Lifecycle: the parts that get implemented wrong
+
+The callbacks themselves are well documented; these are the misconceptions that cause bugs:
+
+- **`connectedCallback` can run more than once.** It fires on every insertion, so moving the element re-runs it. Make setup idempotent — don't assume a single first run.
+- **Tear down in `disconnectedCallback`.** Remove global listeners, timers, and observers or you leak; but the element may be re-inserted, so don't treat teardown as permanent.
+- **The constructor can't see the DOM.** No attributes, children, or layout exist yet — read attributes and children in `connectedCallback` instead.
+- **`attributeChangedCallback` fires only for `observedAttributes`,** and should stay cheap for attributes that don't affect output.
+
```
LeaVerou, line 22: Non-blocking but this has bitten me and my agents more than once:

```suggestion
- Hooks are evaluated once at element registration time: Lifecycle hooks must be present when `customElements.define()` is called. Later changes are silently ignored. Additionally, `observedAttributes` is read once at element registration time, and `attributeChangedCallback()` is ignored if `observedAttributes` is not present then.

```

(apologies if you have it somewhere else, I didn't see it!)

See https://codepen.io/leaverou/pen/RNKLQQP?editors=1011 for the ACC-related claims.

guides/web-components/custom-elements/guide.md
```diff
@@ -0,0 +1,134 @@
...
+- **`attributeChangedCallback` fires only for `observedAttributes`,** and should stay cheap for attributes that don't affect output.
+
+## Upgrade timing and `:defined`
+
+An element can exist in the DOM *before* its class is registered (in server-rendered HTML, or when the script loads late). Until defined it is an inert `HTMLElement` with none of your behavior; when `customElements.define()` runs, the browser **upgrades** every matching element already in the document, running its constructor and `connectedCallback` then.
+
+Two consequences models routinely miss:
+
+1. **Attributes and children may already be present** when your constructor/`connectedCallback` runs on upgrade. Read existing attributes in `connectedCallback` to sync initial state, and never assume the element started empty.
+2. **Children may still be missing** even in `connectedCallback` if the HTML parser hasn't reached them yet. If you depend on Light DOM children, wait for them: use a `slotchange` listener, or defer with `customElements.whenDefined()`.
```
LeaVerou: Gah, just noticed you have this here. I'd have expected to find this content in "Lifecycle" sine we have guidance around these things there. I'd suggest moving it there and leaving FOUC prevention here.

guides/html/web-components/custom-elements/guide.md
```diff
@@ -0,0 +1,134 @@
...
+  - customized-built-in-elements
+---
+
+# Custom Elements
+
+A custom element is a class registered against a tag name via `customElements.define()`. Registration and basic lifecycle are well understood; the rules below target the parts routinely implemented incorrectly.
+
+## Lifecycle: the parts that get implemented wrong
+
+The callbacks themselves are well documented; these are the misconceptions that cause bugs:
```
LeaVerou, line 14: `connectedMoveCallback()` is [new enough](https://caniuse.com/mdn-api_customelementregistry_define_connectedmovecallback_lifecycle_callback) that would be worth documenting here https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements#custom_element_lifecycle_callbacks

guides/web-components/custom-elements/guide.md
```diff
@@ -0,0 +1,134 @@
...
+- Symbol-keyed properties are a middle ground: hidden from ordinary enumeration, but reachable by any code holding the symbol (export it, or expose it as a `static` field on the class). Because each symbol is unique, two symbols sharing a description never collide — handy when a subclass or sibling needs controlled access that `#` fields can't provide.
+- Do **not** shadow global attributes (`style`, `class`, `id`, `slot`, `part`, `title`, `lang`, `dir`) with your own properties; doing so breaks their built-in behavior. Note `disabled` is only global on form controls; on other elements you must implement its effect yourself.
+
+When naming a new API surface, follow the platform's own conventions — see the [W3C naming principles](https://www.w3.org/TR/design-principles/#naming-is-hard).
+
+## Fallback strategies
+
+{{ BASELINE_STATUS("customized-built-in-elements") }}
+
+Customized built-in elements (`is="…"`) are not universally supported — Safari has permanently declined them. Build **autonomous** custom elements (extending `HTMLElement`) instead, or wrap the native element and forward what you need; do not ship `is="…"` to production.
```
LeaVerou: This is the first time this API is mentioned here, which is not how Fallback Strategies is typically used. Furthermore, you've already advised against this in the `web-components` guide, and it's a rare enough pattern I don't think any agents are queuing to use `is` — they're usually very eager to recreate platform primitives. I would just remove this.

guides/web-components/declarative-shadow-dom/guide.md
```diff
@@ -0,0 +1,67 @@
+---
+name: declarative-shadow-dom
+description: Server-render a custom element's shadow tree as plain HTML with Declarative Shadow DOM, then hydrate it in place without re-rendering or a flash of unstyled content.
+web-feature-ids:
+  - declarative-shadow-dom
+---
+
+# Declarative Shadow DOM
```
LeaVerou:
```suggestion
---
name: prerendering-custom-elements
description: Server-render a custom element, then hydrate it in place without re-rendering or a flash of unstyled content.
web-feature-ids:
  - declarative-shadow-dom
---

# Pre-rendering custom elements
```

**Declarative Shadow DOM is an API, not a use case.** The use case this is describing (SSR/pre-rendering) is fairly concrete and stands on its own. DSD is a critical _feature_ for this use case, but not the only one. E.g. there's some great guidance in one of the other guides around serializing and hydrating states.
There are also all sorts of other caveats around pre-rendering CEs, not all of which I remember. We don't need to be exhaustive, but reframing the guide to be about the use case means that when we come across SSR guidance for CEs later, it has a home, whether it's related to DSD or not.

Is SSR the only use case for DSD? No, but it's the biggest, and we can always make new guides for the others (e.g. style isolation) if needed.

Very slight preference for "pre-rendering" over "server-side rendering" as it seems a little more general, but either is fine (now I'm wondering if there are there SSR cases that aren't pre-rendering? 🤔)

guides/html/web-components/prerendering-custom-elements/guide.md
```diff
@@ -0,0 +1,67 @@
...
+      this.attachShadow({ mode: 'open' });
+      this.shadowRoot.append(template.content.cloneNode(true));
+    }
+    // Otherwise hydrate in place: wire up listeners against existing markup,
+    // and do NOT re-render; that would discard the server's DOM and flash.
+  }
+}
+```
+
+Set `shadowrootmode="open"` (not `closed`) for almost all cases, so scripts and tools can reach `this.shadowRoot` to hydrate. To serialize an existing shadow root back to HTML on the server or in tests, use `getHTML({ serializableShadowRoots: true })` and attach the root with `serializable: true`.
```
LeaVerou, line 53: TIL! 💡

guides/web-components/declarative-shadow-dom/guide.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+Set `shadowrootmode="open"` (not `closed`) for almost all cases, so scripts and tools can reach `this.shadowRoot` to hydrate. To serialize an existing shadow root back to HTML on the server or in tests, use `getHTML({ serializableShadowRoots: true })` and attach the root with `serializable: true`.
+
+## Avoiding FOUC
+
+DSD renders correctly on first paint only if styles ship inside the declarative template. To preserve that:
+
+- Put component CSS **inside** the `<template shadowrootmode>` (a `<style>` tag or a `<link>`), not in a separate script-loaded sheet.
+- Pair with `my-card:not(:defined) { … }` only for styles that must wait for JS; the structural styles should already be in the shadow root and need no such guard.
```
LeaVerou: Nit: Should we recommend `:host(:not(:defined))` to avoid hardcoding the element name?

guides/web-components/shadow-dom/guide.md
```diff
@@ -0,0 +1,94 @@
...
+}
+```
+
+## Fallback strategies
+
+{{ BASELINE_STATUS("shadow-dom") }}
+
+{{ BASELINE_STATUS("slot") }}
+
+Shadow DOM and slots are broadly available. Where you must support engines without them (or run without JS), render the component's markup into the Light DOM instead and scope styles with a class or BEM convention — you lose encapsulation, but the content stays functional.
```
LeaVerou:
```suggestion
```

I think the current rule is that we don't add fallbacks for Baseline Widely Available features. Additionally, we recommend against BEM in the CSS guide.

We may want to reintroduce the section if we add cutting edge features around shadow DOM (e.g. `:has-slotted`).

guides/web-components/styling-web-components/guide.md
```diff
@@ -0,0 +1,120 @@
...
+
+```css
+:host {
+  /* Consumers override --card-bg from outside; the fallback keeps it usable. */
+  background: var(--card-bg, white);
+  color: var(--card-fg, black);
+}
+```
+
+A custom property *defined* on `:host` can only be overridden by rules targeting the host element specifically. To let it inherit and be overridable from anywhere up the tree, leave it unset on `:host` and rely on the fallback in each `var()`; do not seed it with a value on `:host`.
```
LeaVerou: (Non-blocking) A common-ish pattern is to define a "shadow" property that bakes the fallback in:

```css
--_foo: var(--foo, fallback);
```

This reduces duplication, and can help code readability.

guides/web-components/styling-web-components/guide.md
```diff
@@ -0,0 +1,120 @@
...
+
+## `::slotted()`: styling projected Light DOM
+
+`::slotted(selector)` targets the **top-level** nodes assigned to a slot. It cannot reach their descendants, and it loses to any Light DOM rule (the consumer owns their own content), so use it for light touch-ups, not control.
```
LeaVerou: Should we mention `!important`?

guides/web-components/styling-web-components/guide.md
```diff
@@ -0,0 +1,120 @@
...
+- **`:host` styles don't reach the inner element.** Setting `background`/`border`/`padding` on `:host` styles the host box, not the control inside it. Forward deliberately: expose the control as a `::part` for open-ended styling, or pipe a fixed set of values through custom properties — a `:host` rule does not cascade into the wrapped element.
+- **Avoid the `display: contents` + `all: inherit` shortcut.** Collapsing the host with `display: contents` and resetting the control with `all: inherit` looks like a way to make the wrapper "disappear," but `all` is a sledgehammer that also wipes useful UA defaults, `display: contents` removes the host's own box (so it can no longer carry padding, border, or background), and it interferes with anchor positioning against the host. Prefer `::part` or custom properties.
```
LeaVerou: I think this is too strongly worded. There are pros and cons to it, e.g.:
- It's pretty much the only generic method we have for passing arbitrary style to children without consumers having to look up element-specific documentation about which part to use, or be constrained by all the limitations of `::part()`. You just style elements as if they're native, and it just …works. At least that's the pitch.
- **Removing the host's own box so styling can be passed on to the child without duplication is kind of the point!** If the host should be able to have its own visible box, then the child is a distinct visual element and there's no reason to use this in the first place. This is for cases where the two being distinct elements is an artifact of the current platform limitations, i.e. cases where you're using composition, but you wish you could have had inheritance (e.g. `MyButton extends HTMLButtonElement`).
- `all` wiping browser defaults is true, but not integral to the method. You *could* just specify individual properties (e.g. `background: inherit`), it loses some of its power, since anything you didn't think of is not specifiable, just like when using custom properties. Alternatively, if you know which browser defaults you want to keep, `property: revert` works too.

IIRC the big downside is that `getBoundingClientRect()` and `getClientRects()` (which anchor positioning depends on) are broken with all `display: contents` elements, see https://github.com/w3c/csswg-drafts/issues/12040
You *can* implement the methods yourself, but it makes it a much bigger lift, and changes the balance of tradeoffs. And I can't even remember if that fixes anchor positioning.

guides/html/web-components/web-components/guide.md
```diff
@@ -0,0 +1,55 @@
+---
+name: web-components
+description: Orientation and cross-cutting principles for authoring and using Web Components, covering Custom Elements, Shadow DOM, templates and slots, Declarative Shadow DOM, styling across the shadow boundary, Form-Associated Custom Elements, and accessibility requirements. Use this guide when building reusable UI components, design systems, or server-rendered custom elements.
+---
```
LeaVerou, line 8:
```suggestion
web-feature-ids:
  - autonomous-custom-elements
  - shadow-dom
  - declarative-shadow-dom
---
```

LeaVerou (changes requested):
This is much better than last time I looked! Only a few things left.

guides/web-components/declarative-shadow-dom/guide.md
castastrophe: These styles can't live in the web components though as they're for when the component is not defined yet. I believe this is the primary use-case behind the ask for a web-components-only selector or pseudo class, isn't it?

LeaVerou: `:host` works for any shadow DOM styles, whether registered or not: https://codepen.io/leaverou/pen/azpLrer

Edit: Just realized this was about light DOM styles, sorry! 🤦🏽‍♀️ In that case, it would be helpful to add a bit of guidance about what styles would those be, though I wouldn't hold up the PR for it.

guides/web-components/styling-web-components/guide.md
castastrophe: I'm not a huge fan of this approach personally - it's somewhat duplicative and adds a lot of cruft to styles - but that's honestly just my capital O opinion and not necessarily an accepted best practice so I've gone ahead and added this.

guides/web-components/custom-elements/guide.md
castastrophe: Compromised on this one. Since the last round already moved slotchange / attributeChangedCallback guidance up into the lifecycle constructor bullet, keeping the full "children may be missing → use slotchange" bullet under Upgrade timing was just restating it. Trimmed the upgrade bullet to the why — the parser can yield mid-parse of long child lists — and back-referenced lifecycle for the pattern.

Happy to move the whole thing over if you'd rather have parser-yielding sit under lifecycle too; it just felt structurally cleaner to keep timing-of-upgrade facts in the upgrade-timing section and callback patterns in lifecycle.

castastrophe:
@LeaVerou I applied most of this feedback with a few notes on where I might have compromised or pushed back a bit. I did a quick evaluation of new specs that aren't yet baseline (but are on track to be soon) and I came across a few gaps I've not yet included. Can you let me know which of these might be valuable to incorporate?

1. Scoped Custom Element Registries
`new CustomElementRegistry()`, `attachShadow({ registry })`, `ShadowRoot.customElementRegistry`.

Chrome shipped in 2025; Safari and Firefox positions are positive; not Baseline yet but genuinely on-track.

2. `:has-slotted`

Chrome shipped; other engines rolling.

3. Manual slot assignment — imperative slotting
`attachShadow({ slotAssignment: 'manual' })` + `HTMLSlotElement.assign(...nodes)`.


Baseline: newly available (shipped across engines in 2024). Ironically this one may already qualify as Baseline Widely Available by the guides' rules but worth verifying.

4. Reference Target (`shadowrootreferencetarget`) — cross-root ARIA

Baseline status: origin trial / behind-flag in Chrome; nowhere else yet. Not close to Baseline, but the trajectory is real.

5. `ariaActiveDescendantElement` and other element-reference ARIA properties

Baseline status: `aria-attribute-reflection` is Baseline; the element-reflection variant is newer and per-browser support is uneven.

6. Import attributes for CSS modules (`import sheet from './x.css' with { type: 'css' }`)

Baseline status: Baseline in all major engines as of 2024.

---
Also on the horizon, lower priority:

- JSON module imports (relevant if component authors want to co-locate config).
- DSD <template shadowrootserializable> — already partially covered.
- Anchor Positioning interactions with display: contents — tangentially relevant to the <my-button> wrapping section Lea called out.

guides/html/web-components/accessible-web-components/guide.md
castastrophe: Sort of. `aria-attribute-reflection` covers the element-reference variants (`ariaLabelledByElements` etc.) in its compat list, but the feature-level status is "widely available" because the plain string properties have been baseline for years. The `element-reference` ones only reached "newly available" in April 2025, and `ariaOwnsElements` isn't baseline at all.

So dropping the macro here would overclaim. Options:

- leave the hedged prose as-is
- add BASELINE_STATUS("aria-attribute-reflection") in fallback strategies (accurate at the feature level, doesn't overclaim about the subset)
- file against web-features to split element-reference reflection into its own id

Leaning (1) unless you want (2) or (3). Happy to do either.

guides/web-components/styling-web-components/guide.md
castastrophe: reworked in the latest push. reframed as "real tradeoff, not a shortcut" and threaded in your counterpoints:

- it's currently the only generic style-forwarding path that doesn't force consumers to learn per-component parts
- removing the host's box is usually the point of the pattern (composition standing in for missing built-in inheritance)
- `all: inherit` isn't integral; scoping to specific properties or using `revert` gets defaults back
- the concrete blocker called out explicitly: `getBoundingClientRect()` and `getClientRects()` return zero for `display: contents`, so anchor positioning against the host breaks

`::part` and custom properties are now the alternative when those tradeoffs don't fit, not the default.

guides/web-components/styling-web-components/guide.md
castastrophe: yes, added a line to the `::slotted()` section: `::slotted(...) !important` can beat consumer rules, but doing so overrides styling the consumer is entitled to own, so reserve it for structural fixes rather than aesthetic overrides.

if you meant `!important` in a different context (e.g. consumer `!important` beating `:host` custom-property declarations), happy to add that too.

LeaVerou:
> @LeaVerou I applied most of this feedback with a few notes on where I might have compromised or pushed back a bit. I did a quick evaluation of new specs that aren't yet baseline (but are on track to be soon) and I came across a few gaps I've not yet included. Can you let me know which of these might be valuable to incorporate?
> 
> 1. Scoped Custom Element Registries
> 2. `:has-slotted`
> 3. Manual slot assignment — imperative slotting
> 6. Import attributes for CSS modules (`import sheet from './x.css' with { type: 'css' }`)
> * JSON module imports (relevant if component authors want to co-locate config).
> * DSD  — already partially covered.
> * Anchor Positioning interactions with display: contents — tangentially relevant to the  wrapping section Lea called out.

The general rule is we can include pretty much any feature, but if not Baseline WA, it needs fallback guidance.
So all of these are fair game, some just need fallbacks.
That said, if these haven't come up organically while writing the guides, I wouldn't force it. We can focus on use cases that utilize them in a followup.

We have guidance for plenty of Chrome-only features, as long as there are viable fallbacks. All of these features are useful to include guidance for, though not necessarily as part of this PR.

> 4. Reference Target (`shadowrootreferencetarget`) — cross-root ARIA
> 
> Baseline status: origin trial / behind-flag in Chrome; nowhere else yet. Not close to Baseline, but the trajectory is real.

I'd hold off on Reference Target for now, it seems a bit _too_ shaky ground. But it's not a strong opinion.

> 5. `ariaActiveDescendantElement` and other element-reference ARIA properties
> 
> Baseline status: `aria-attribute-reflection` is Baseline; the element-reflection variant is newer and per-browser support is uneven.

I would absolutely use them in the a11y guide, they are highly relevant. Is there a BCD key we can feed into the `BASELINE_STATUS` macro? I'll investigate (edit: there doesn't seem to be. BCD keys are around the IDL attributes, not whether they're writable or what values they take)

guides/web-components/styling-web-components/guide.md
LeaVerou: I don't love it either, though I like the alternatives less 😀 If you want to make the guidance more nuanced (e.g. only do this if the fallback is repeated?), that's fine by me.

LeaVerou (approved):
LGTM! Thank you!

LeaVerou:
> I would absolutely use them in the a11y guide, they are highly relevant. Is there a BCD key we can feed into the `BASELINE_STATUS` macro? I'll investigate (edit: there doesn't seem to be. BCD keys are around the IDL attributes, not whether they're writable or what values they take)

FWIW I did some testing ([testcase](https://experiments.verou.me/aria-element-reflection/)) and it seems these work in Chrome 135+, Firefox 136+, Safari 16.5+. So basically Baseline 2025. But we shouldn't hardcode this, ideally we need a web-feature id. I [filed an issue](https://github.com/web-platform-dx/web-features/issues/4147). But perhaps it's ok to hardcode it until then.

guides/html/web-components/accessible-web-components/guide.md
LeaVerou: I've filed a [`web-features` issue 2 weeks ago](https://github.com/web-platform-dx/web-features/issues/4147) but haven't heard back. Just pinged @ddbeck. Meanwhile, `BASELINE_STATUS` accepts an optional BCD key second argument, so we can pick any of these (e.g. `api.Element.ariaLabelledByElements` or whichever else you think will be most common)

LeaVerou:
@castastrophe I just resolved 3/4 out of the open conversations and responded to the 4th one. Sorry! I hadn't realized merging was blocked on these, I had thought the ball was out of my court on this one 😅

Also, I just noticed these are under a top-level `web-components` directly. In the recent content sync we decided to put `web-components` under `html`, could you please move them? You’ll find it’s quite roomy in there, with just a discipline guide.

castastrophe:
@LeaVerou Since this one has gone through the gammut already and is pretty close to merging - what would think of deferring adding `ariaActiveDescendantElement` for a follow-up ticket?

LeaVerou:
Ok, let's do it in a followup. I just resolved https://github.com/GoogleChrome/modern-web-guidance-src/pull/738/changes#r3610406807 
Should be mergeable now!

PR #577, opened 2026-04-16
Guide: Component specific light dark theme

dvdherron:
Closes #227 

Looking for guidance on how to navigate the upcoming changes for `light-dark` support for images. 

See: 
- https://www.bram.us/2026/03/19/more-easy-light-dark-mode-switching-light-dark-is-about-to-support-images/
- https://chromestatus.com/feature/5123253146353664?gate=6278228337623040

Issue #239, opened 2026-03-06
Create guide and evals for the theme-within-theme use case


paulirish:
Configure a specific component or region of a page to use a different color theme than the rest of the page.

Affected web-feature IDs: [color-scheme](https://webstatus.dev/features/color-scheme)

Use case subdir: [guides/user-experience/theme-within-theme](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/user-experience/theme-within-theme)

Related features: #42

**Required files:**
- [x] Use case metadata (guide.md frontmatter)
- [x] demo.html
- [ ] Use case guidance (guide.md)
- [ ] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

LeaVerou:
How is this different from #227 ? Or is this one about "inverting" the theme, i.e. use dark in light mode and light in dark mode?

dvdherron:
Yeah, looking at the demo that was originally created for this, this use case seems very similar to https://github.com/GoogleChrome/guidance/issues/227.

I could see it being changed to inverting a theme or choosing a distinct color palette for a component.

PR #742, opened 2026-05-08
Component specific light dark theme edits

### LINKED ISSUE #227: Create guide and evals for the component-specific-light-dark-theme use case

dvdherron:
Closes #227

### DIFF
--- a/guides/user-experience/component-specific-light-dark-theme/expectations.md
+++ b/guides/user-experience/component-specific-light-dark-theme/expectations.md
@@ -1,10 +1,10 @@
-- The document includes a `<meta name="color-scheme" content="light dark">` tag in the `<head>` to declare theme support early and minimize the risk of un-themed content flashes.
+- **OPTIONAL**: The document includes a `<meta name="color-scheme" content="light dark">` tag in the `<head>` to declare theme support early and minimize the risk of un-themed content flashes.
 - **MANDATORY**: The document or the component has `color-scheme: light dark` (or includes both) set on the `:root` or a common ancestor to enable the resolution of the `light-dark()` function.
 - **MANDATORY**: The implementation uses custom properties as a **semantic abstraction layer** for theme-aware variables (e.g., `--card-bg` is the semantic variable, while `--card-bg-light` and `--card-bg-dark` are the raw tokens).
 - **MANDATORY**: The `light-dark()` function is used to define the values of these semantic custom properties.
 - **MANDATORY**: The implementation uses the `color-scheme` property (e.g., `color-scheme: light` or `color-scheme: dark`) to force a specific theme on a component or section.
 - **MANDATORY**: `color-scheme` overrides are only applied to elements that also have a `background-color` (or where a background is clearly intended), as per the guide's recommendation.
-- **MANDATORY**: When a component or section forces a `color-scheme`, properties that use `light-dark()` tokens (like `background-color` or `color`) are explicitly re-applied on that same element to ensure they re-resolve correctly and avoid the "inheritance footgun".
+- **MANDATORY**: When a component or section forces a `color-scheme`, inherited properties that use `light-dark()` tokens (like `color` or `accent-color`) are explicitly re-applied on that same element to ensure they re-resolve correctly and avoid the "inheritance footgun".
 - **OPTIONAL**: The component's internal elements (like scrollbars or form controls) correctly adapt their native browser styling when `color-scheme` is overridden.
 - **MANDATORY**: Fallback strategies are provided for browsers that do not support `light-dark()`, using manual variable overrides and `@media (prefers-color-scheme)` queries.
 - **MANDATORY**: Progressive enhancement is used via `@supports (color: light-dark(white, black))` to ensure that the `light-dark()` logic is preferred and used when available, allowing per-component overrides to work as intended.
--- a/guides/user-experience/component-specific-light-dark-theme/expectations.md
+++ b/guides/user-experience/component-specific-light-dark-theme/expectations.md
@@ -1,5 +1,9 @@
 - The component uses the `light-dark()` CSS function to define at least one color property (e.g., `background-color` or `color`).
+- **MANDATORY**: The document includes a `<meta name="color-scheme" content="light dark">` tag in the `<head>` to declare theme support early and minimize the risk of un-themed content flashes.
 - **MANDATORY**: The document or the component has `color-scheme` set to `light dark` (or includes both) to enable the resolution of `light-dark()`.
+- **MANDATORY**: The implementation uses custom properties as a **semantic abstraction layer** for theme-aware variables (e.g., `--card-bg` is the semantic variable, while `--card-bg-light` and `--card-bg-dark` are the raw tokens).
+- **MANDATORY**: The implementation uses the `color-scheme` property (not just color changes) when the component contains native UI elements like scrollbars or form controls to ensure consistent browser-provided styling.
+- **MANDATORY (If using `light-dark()`)**: When nesting different color schemes (e.g., a dark section in a light page), all properties using dynamic tokens defined with `light-dark()` are reassigned their token values (e.g. `accent-color: var(--accent-color);`) on the element where the `color-scheme` changes to prevent the "inheritance footgun".
 - The component supports an override mechanism (e.g., a specific CSS class or data attribute) that sets `color-scheme` to `light` to force light mode.
 - The component supports an override mechanism (e.g., a specific CSS class or data attribute) that sets `color-scheme` to `dark` to force dark mode.
 - When the component is forced to `color-scheme: light`, its computed values match the first argument of the `light-dark()` function, regardless of the system's `prefers-color-scheme`.
@@ -7,4 +11,5 @@
 - **MANDATORY**: Fallback strategies for colors (e.g., manual variable overrides and `@media (prefers-color-scheme)` queries) are provided for browsers that do not support `light-dark()`.
 - **MANDATORY**: The implementation uses progressive enhancement (e.g., `@supports (color: light-dark(white, black))`) to ensure that `light-dark()` logic is preferred over `@media` fallbacks, allowing per-component overrides to work correctly.
 - **MANDATORY**: `light-dark()` is only used for supported types (colors) and not for unsupported properties like `padding` or `font-size`.
+- **MANDATORY**: Registered custom properties (via `@property`) are **not** used for values that must dynamically resolve with `light-dark()`.
 - **MANDATORY**: Each `light-dark()` call contains exactly two arguments.
--- a/guides/user-experience/component-specific-light-dark-theme/expectations.md
+++ b/guides/user-experience/component-specific-light-dark-theme/expectations.md
@@ -1,6 +1,6 @@
 - **OPTIONAL**: The document includes a `<meta name="color-scheme" content="light dark">` tag in the `<head>` to declare theme support early and minimize the risk of un-themed content flashes.
 - **MANDATORY**: The document or the component has `color-scheme: light dark` (or includes both) set on the `:root` or a common ancestor to enable the resolution of the `light-dark()` function.
-- **MANDATORY**: The implementation uses custom properties as a **semantic abstraction layer** for theme-aware variables (e.g., `--card-bg` is the semantic variable, while `--card-bg-light` and `--card-bg-dark` are the raw tokens).
+- **MANDATORY**: Outside of `@supports` feature detection, `light-dark()` is used with var() references for the two color arguments, not raw color values (e.g., `--card-bg: light-dark(var(--card-bg-light), var(--card-bg-dark))` while `--card-bg-light` and `--card-bg-dark` are the raw tokens that contain the actual color values).
 - **MANDATORY**: The `light-dark()` function is used to define the values of these semantic custom properties.
 - **MANDATORY**: The implementation uses the `color-scheme` property (e.g., `color-scheme: light` or `color-scheme: dark`) to force a specific theme on a component or section.
 - **MANDATORY**: `color-scheme` overrides are only applied to elements that also have a `background-color` (or where a background is clearly intended), as per the guide's recommendation.
--- a/guides/user-experience/component-specific-light-dark-theme/expectations.md
+++ b/guides/user-experience/component-specific-light-dark-theme/expectations.md
@@ -1,15 +1,15 @@
-- The component uses the `light-dark()` CSS function to define at least one color property (e.g., `background-color` or `color`).
-- **MANDATORY**: The document includes a `<meta name="color-scheme" content="light dark">` tag in the `<head>` to declare theme support early and minimize the risk of un-themed content flashes.
-- **MANDATORY**: The document or the component has `color-scheme` set to `light dark` (or includes both) to enable the resolution of `light-dark()`.
+- The document includes a `<meta name="color-scheme" content="light dark">` tag in the `<head>` to declare theme support early and minimize the risk of un-themed content flashes.
+- **MANDATORY**: The document or the component has `color-scheme: light dark` (or includes both) set on the `:root` or a common ancestor to enable the resolution of the `light-dark()` function.
 - **MANDATORY**: The implementation uses custom properties as a **semantic abstraction layer** for theme-aware variables (e.g., `--card-bg` is the semantic variable, while `--card-bg-light` and `--card-bg-dark` are the raw tokens).
-- **MANDATORY**: The implementation uses the `color-scheme` property (not just color changes) when the component contains native UI elements like scrollbars or form controls to ensure consistent browser-provided styling.
-- **MANDATORY (If using `light-dark()`)**: When nesting different color schemes (e.g., a dark section in a light page), all properties using dynamic tokens defined with `light-dark()` are reassigned their token values (e.g. `accent-color: var(--accent-color);`) on the element where the `color-scheme` changes to prevent the "inheritance footgun".
-- The component supports an override mechanism (e.g., a specific CSS class or data attribute) that sets `color-scheme` to `light` to force light mode.
-- The component supports an override mechanism (e.g., a specific CSS class or data attribute) that sets `color-scheme` to `dark` to force dark mode.
-- When the component is forced to `color-scheme: light`, its computed values match the first argument of the `light-dark()` function, regardless of the system's `prefers-color-scheme`.
-- When the component is forced to `color-scheme: dark`, its computed values match the second argument of the `light-dark()` function, regardless of the system's `prefers-color-scheme`.
-- **MANDATORY**: Fallback strategies for colors (e.g., manual variable overrides and `@media (prefers-color-scheme)` queries) are provided for browsers that do not support `light-dark()`.
-- **MANDATORY**: The implementation uses progressive enhancement (e.g., `@supports (color: light-dark(white, black))`) to ensure that `light-dark()` logic is preferred over `@media` fallbacks, allowing per-component overrides to work correctly.
-- **MANDATORY**: `light-dark()` is only used for supported types (colors) and not for unsupported properties like `padding` or `font-size`.
-- **MANDATORY**: Registered custom properties (via `@property`) are **not** used for values that must dynamically resolve with `light-dark()`.
-- **MANDATORY**: Each `light-dark()` call contains exactly two arguments.
+- **MANDATORY**: The `light-dark()` function is used to define the values of these semantic custom properties.
+- **MANDATORY**: The implementation uses the `color-scheme` property (e.g., `color-scheme: light` or `color-scheme: dark`) to force a specific theme on a component or section.
+- **MANDATORY**: `color-scheme` overrides are only applied to elements that also have a `background-color` (or where a background is clearly intended), as per the guide's recommendation.
+- **MANDATORY**: When a component or section forces a `color-scheme`, properties that use `light-dark()` tokens (like `background-color` or `color`) are explicitly re-applied on that same element to ensure they re-resolve correctly and avoid the "inheritance footgun".
+- **OPTIONAL**: The component's internal elements (like scrollbars or form controls) correctly adapt their native browser styling when `color-scheme` is overridden.
+- **MANDATORY**: Fallback strategies are provided for browsers that do not support `light-dark()`, using manual variable overrides and `@media (prefers-color-scheme)` queries.
+- **MANDATORY**: Progressive enhancement is used via `@supports (color: light-dark(white, black))` to ensure that the `light-dark()` logic is preferred and used when available, allowing per-component overrides to work as intended.
+- **MANDATORY**: `light-dark()` is only used for supported types (colors) and not for non-color properties like `padding` or `margin`.
+- **MANDATORY**: Custom properties intended to dynamically resolve with `light-dark()` are NOT registered using `@property` with a `<color>` type, as this prevents dynamic re-resolution.
+- **MANDATORY**: Each `light-dark()` call contains exactly two color arguments.
+- When the component is forced to `color-scheme: light`, its computed `background-color` and `color` match the first argument of the respective `light-dark()` calls, regardless of the system's `prefers-color-scheme`.
+- When the component is forced to `color-scheme: dark`, its computed `background-color` and `color` match the second argument of the respective `light-dark()` calls, regardless of the system's `prefers-color-scheme`.
--- a/guides/user-experience/component-specific-light-dark-theme/guide.md
+++ b/guides/user-experience/component-specific-light-dark-theme/guide.md
@@ -110,7 +110,7 @@ Force a component instance into a specific theme by setting its `color-scheme` p
 - **The `@property` Risk**:
   - **Mandatory**: Do not register properties meant to be design tokens that dynamically switch based on `light-dark()` as `<color>`. If you need to animate a color variable, use a separate property.
 - **Mandatory**: Do not set `color-scheme` on elements without a background.
-{{ INCLUDE("guides/user-experience/browser-ui-color-theme/guide.md[#best-practices]") }}
+ {{ INCLUDE("../browser-ui-color-theme/guide.md#best-practices") }}
 
 ## Fallback strategies
 
--- a/guides/user-experience/component-specific-light-dark-theme/guide.md
+++ b/guides/user-experience/component-specific-light-dark-theme/guide.md
@@ -15,115 +15,157 @@ sources:
   - https://css-tricks.com/almanac/functions/l/light-dark/
   - https://www.bram.us/2020/04/26/the-quest-for-the-perfect-dark-mode-using-vanilla-javascript/
   - https://css-tricks.com/come-to-the-light-dark-side/
+  - https://github.com/w3c/csswg-drafts/issues/13836
 ---
 
 # Component-Specific Light/Dark Themes
 
 The `light-dark()` function allows you to define two colors for a single property, which the browser automatically switches based on the current `color-scheme`. By scoping both the `light-dark()` variables and the `color-scheme` property to a specific component, you can create UI elements that can be independently forced into light or dark mode, regardless of the user's global system preference.
 
-## Implementation steps
+## Why use component-specific themes?
 
-1. **MANDATORY: Enable `color-scheme` support.** The `light-dark()` function only resolves if the element (or an ancestor) has an explicit `color-scheme` value of `light`, `dark`, or `light dark`.
+While most applications should follow the user's system preference by default, there are several scenarios where you need more granular control over an element's theme:
 
-    ```css
-    :root {
-      /* 
-         MANDATORY: Enable system-wide support for both themes.
-         Without this, the browser will not know how to resolve light-dark().
-      */
-      color-scheme: light dark;
-    }
-    ```
+- **Nested UI Contexts ("Theme Islands")**: You may want to create sections of a page that contrast with the global theme, such as a dark footer or sidebar on an otherwise light-themed website.
+- **Content-Driven Aesthetics**: Certain components work best in a specific mode. For example, **code editors**, **video players**, and **photo galleries** often use dark themes to minimize distraction and make colors "pop," regardless of the user's OS setting.
+- **User Preference Overrides**: You can empower users to manually toggle the theme of a specific workspace or widget. This allows them to choose a mode that fits their environment or eye comfort without forcing them to change their global operating system settings.
+- **Native Browser UI Consistency**: By forcing a specific `color-scheme` on a container, you ensure that all built-in browser UI within that container—such as **scrollbars**, **form controls**, and **dropdown menus**—automatically match the intended theme of that component.
 
-2. **Define semantic variables or properties using `light-dark()`.** Use the `light-dark(light-color, dark-color)` syntax. 
+## When to Change Colors vs. When to Force `color-scheme`
 
-    ```css
-    .themed-card {
-      /* 
-         DO: Use semantic variables to define theme-aware colors.
-         The first argument is used when color-scheme is 'light'.
-         The second argument is used when color-scheme is 'dark'.
-      */
-      --card-bg: light-dark(#ffffff, #2d2e31);
-      --card-text: light-dark(#202124, #f8f9fa);
-      
-      background-color: var(--card-bg);
-      color: var(--card-text);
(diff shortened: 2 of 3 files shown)


dvdherron:
@LeaVerou This address the issues covered in your SME review.

- **P0: Missing gotcha that can produce inaccessible UIs**
 Added the guidance addressing the issues with light-dark() and linked to the more detailed Best Practices section of the `browser-ui-color-theme` guide (in review)

- **P1: Describes the HOW but not the WHY/WHEN (use cases)**
 I did more explaining about when it might make sense to change colors versus force a theme with `color-scheme`. I kept the examples to forcing a theme to try to keep this use case focused. Do you think there should be more examples? If so, what exactly?
- **P2: Fallback repetition**
  cleaned up
- **P3: Factually incorrect justification**
 cleaned up the guidance about how `light-dark()` resolves if `color-scheme` is absent.

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -15,115 +15,157 @@ sources:
   - https://css-tricks.com/almanac/functions/l/light-dark/
   - https://www.bram.us/2020/04/26/the-quest-for-the-perfect-dark-mode-using-vanilla-javascript/
   - https://css-tricks.com/come-to-the-light-dark-side/
+  - https://github.com/w3c/csswg-drafts/issues/13836
```
LeaVerou: I don't think we need to link to this? I think briefly explaining the issue is sufficient.

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -15,115 +15,157 @@ sources:
...
 
-```css
-/* 
-   DO NOT: Attempt to use light-dark() for non-color properties like padding.
-   This will result in an invalid property value.
-*/
-.themed-card {
-  padding: light-dark(10px, 20px); 
-}
+While most applications should follow the user's system preference by default, there are several scenarios where you need more granular control over an element's theme:
```
LeaVerou:
```suggestion
While most element's should follow the page-wide color scheme, there are several scenarios where you need more granular control over an element's theme:
```

(the page-wide color scheme is not necessarily the user's system preference, e.g. when a dark mode toggle is used)

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -15,115 +15,157 @@ sources:
...
+
+- **Nested UI Contexts ("Theme Islands")**: You may want to create sections of a page that contrast with the global theme, such as a dark footer or sidebar on an otherwise light-themed website.
```
LeaVerou: How should these behave in dark mode? I.e. are they conceptually "invert color scheme" or "use dark color scheme, always"?
The former is nontrivial (and might warrant a separate guide, if enough people need it)

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -15,115 +15,157 @@ sources:
...
+- **Content-Driven Aesthetics**: Certain components work best in a specific mode. For example, **code editors**, **video players**, and **photo galleries** often use dark themes to minimize distraction and make colors "pop," regardless of the user's OS setting.
```
LeaVerou: Ooh these are _really_ good examples. Hadn't even thought of video players and photo galleries!

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -15,115 +15,157 @@ sources:
...
+- **User Preference Overrides**: You can empower users to manually toggle the theme of a specific workspace or widget. This allows them to choose a mode that fits their environment or eye comfort without forcing them to change their global operating system settings.
```
LeaVerou: Is overriding the entire page's color scheme in scope for this guide? I thought this was about forcing color scheme on individual elements?

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -15,115 +15,157 @@ sources:
...
+- **Native Browser UI Consistency**: By forcing a specific `color-scheme` on a container, you ensure that all built-in browser UI within that container—such as **scrollbars**, **form controls**, and **dropdown menus**—automatically match the intended theme of that component.
```
LeaVerou: This makes it sound like you have to set `color-scheme` all over the place to get scrollbars etc to adapt to the page-wide color scheme. Doesn't that happen anyway when you have `color-scheme: light dark` on the root?

Though this relates to the previous question as well: is the goal of the guide to explain how to use `color-scheme` or to explain when to use component-specific color schemes (that override the page-wide color scheme)? I had assumed the latter.

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -15,115 +15,157 @@ sources:
...
+
+## When to Change Colors vs. When to Force `color-scheme`
```
LeaVerou, line 31: I love the structure of this section, but I'm a little unsure about the decision framework. 

Whether the component has "native interactive parts" is a) fuzzy. There is a lot more browser UI than meets the eye, e.g. spellcheck underlines (though I'm not sure any UAs actually vary these for light vs dark). b) can change over time as the rest of the CSS changes (e.g. adding `overflow`), or even with changes in the web platform (e.g. the `::interest-button` proposal being discussed would add new native UI to certain existing elements). Though framing like "if there _were_ browser UI, what color-scheme should it follow?" could work.

"Best viewed in a particular color mode" might be too fuzzy for agents. What makes it so? If we can extract that, I think it could be more actionable. You have some really good examples earlier. Why are videos and galleries in dark mode regardless? The rationale above was "to minimize distraction and make colors pop" which made sense to me at the time, though there are plenty of light mode UIs optimized to minimize distractions, and plenty of examples of colors popping that aren't dark. Also, we don't make them light in dark mode, so why do we no longer want them to "pop"? Perhaps it's a designer judgement call, and the best we can do is list a lot of examples…

Another potential factor seems to be whether it's a surface boundary, not just an individual accented detail (room vs object). E.g. a button wouldn't qualify, but a code block would. But that's also hard to define.

Oh, one important caveat is to not set `color-scheme` on elements without a background, see https://github.com/w3c/csswg-drafts/issues/6773 . Which partly helps with the point above, since surfaces tend to have backgrounds…

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -15,115 +15,157 @@ sources:
...
+  --card-text: var(--card-text-light);
+
+  /* 3. Apply custom properties on relevant properties (light) values */
+  
+    /* **Mandatory**: when nesting schemes, dynamic properties set with 
+    light-dark() must also be set on the element where the scheme changes. 
+
+    When setting a different color-scheme on a component, **DO NOT** redefine custom properties (e.g. --accent-color: light-dark(black, white);) 
+    unless the custom property needs to be updated. The custom property only needs 
+    to be applied to a property (e.g. accent-color: var(--accent-color);).
```
LeaVerou: That's quite complicated to follow, and asks for some pretty tight coupling which in some ways defeats the point of using `color-scheme` on elements if you need to remember to update some centralized place every time.

I think the guidance should just be to **not register properties meant to be design tokens**. If you need to animate a color variable or whatever, use a separate property. That's much simpler and less nuanced and doesn't introduce any undesirable coupling.

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -15,115 +15,157 @@ sources:
...
+/* Force this specific card into light mode */
+.themed-card.force-light {
+  /* Dynamic custom properties don't need to be reapplied here since they were already defined on `.themed-card` */
+  color-scheme: light;
+}
+```
+
+## Critical Considerations: The "Inheritance Footgun"
+
+`light-dark()` currently resolves based on the `color-scheme` of the element where the variable is **defined**, not where it is **used**.
```
LeaVerou: See above. This only happens when the property is registered. You don't need to jump through all these hoops, just don't register your design tokens!

The only other cases where this becomes relevant are the native **inherited** `<color>` properties, which are thankfully few (but yeah, `color` and `accent-color` are in that list…). But the example below isn't about them…

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -15,115 +15,157 @@ sources:
...
+MANDATORY: Enable global support for both color schemes by setting `color-scheme: light dark;` on the `:root`. If `color-scheme` is not set on `:root` (or an ancestor) the `light-dark()` function will not work as expected.
 
-{{ BASELINE_STATUS("light-dark") }}
-{{ BASELINE_STATUS("color-scheme") }}
+```css
+:root {
+  /* **Mandatory:** Enable global support for both color schemes */
+  color-scheme: light dark;
+}
+```
```
LeaVerou, line 61: This is about page-wide color-scheme, so it also hinges on the question about the scope of this guide.

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -15,115 +15,157 @@ sources:
...
+
+  /* ✅ FIX: Reapply the token where the scheme changes (e.g. background-color: var(--color-accent), **Not** `--color-accent: light-dark(black, white);`) */
+  .dark-section {
+    color-scheme: dark;
+    background-color: var(--color-accent); /* Forces re-resolution */
+  }
+  ```
+
+### Avoid `@property` registration
+MANDATORY: Do not use registered custom properties (via `@property`) for colors intended to resolve via `light-dark()`. Registered properties "lock" their resolved value at computed value time even more strictly, making them non-reactive to local theme changes in almost all scenarios.
```
LeaVerou: "lock their resolved value even more strictly" is a bit vague. They just compute it and pass down the computed value, whereas unregistered properties pass tokens around without interpreting them (with the exception of `var()` and with some restrictions).

I don't think we need to be discouraging registered properties altogether, just don't register properties meant to contain design tokens. Using a registered property to animate a color stop in a gradient is fine for example.

LeaVerou (changes requested):
See comments, though there's a big open question about whether this guide is about using `color-scheme` for local color-scheme overrides for individual *elements* or whether it covers using `color-scheme` to override the color-scheme of the page itself. I had assumed the former, but the rewrite introduces a lot of content for the latter which seems out of scope and covered elsewhere, but perhaps I've misunderstood the scope of this guide?

Other than that, the highest priority is fixing the thing about inheritance.

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: Good point. This sounds more like an inverted theme than forcing a component to a certain color-scheme. Maybe inversion makes sense to be included as part of https://github.com/GoogleChrome/guidance/issues/239?

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: @LeaVerou I initially read this use case as, on a page that switches from light to dark, forcing it to one of those states. But I now see how it could also be about having, say, a brand theme (dynamic light-dark), but also having a section with a completely different color palette (also dynamic). Is that what you're asking? 

I can see where this needs to be adjusted to give guidance on forcing a separate color palette.

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: Responded to the similar question above.

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: Would setting color-scheme on the root still apply if a separate color palette was used on a component?

LeaVerou: Setting `color-scheme: light dark` on the root is good practice *anyway* (if your design accommodates dark mode well) and is recommended by our CSS skill and various guides, but is _technically_ not required to make `color-scheme` on individual elements work.

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: removed in https://github.com/GoogleChrome/guidance/pull/742/changes/9207de3bb174f43706e7989c0c7f8421cdcc94fd

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: Adjusted in https://github.com/GoogleChrome/guidance/pull/742/changes/9207de3bb174f43706e7989c0c7f8421cdcc94fd

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: I removed this point in https://github.com/GoogleChrome/guidance/pull/742/changes/9207de3bb174f43706e7989c0c7f8421cdcc94fd. I agree that an inverted scheme would be a good use case.

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: I adjusted the wording of this a bit in https://github.com/GoogleChrome/guidance/pull/742/changes/9207de3bb174f43706e7989c0c7f8421cdcc94fd to make this more focused on the component. Let me know what you think.

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: Added guidance about not setting color-scheme on elements without a background and tried to flesh out these points in https://github.com/GoogleChrome/guidance/pull/742/changes/9207de3bb174f43706e7989c0c7f8421cdcc94fd

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: Simplified this in https://github.com/GoogleChrome/guidance/pull/742/changes/9207de3bb174f43706e7989c0c7f8421cdcc94fd

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: Fixed this in https://github.com/GoogleChrome/guidance/pull/742/changes/9207de3bb174f43706e7989c0c7f8421cdcc94fd

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: Made these mandatory steps optional in https://github.com/GoogleChrome/guidance/pull/742/changes/9207de3bb174f43706e7989c0c7f8421cdcc94fd

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: Cleaned this up in https://github.com/GoogleChrome/guidance/pull/742/changes/9207de3bb174f43706e7989c0c7f8421cdcc94fd

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -21,109 +21,177 @@ sources:
...
-    }
-    ```
+## Why use component-specific themes?
 
-## Critical Considerations
+While most element's should follow the page-wide color scheme, there are several scenarios where you need more granular control over an element's theme:
+
+- **Content-Driven Aesthetics**: Certain components work best in a specific mode. For example, **code editors**, **video players**, and **photo galleries** often use a particular theme to minimize distraction and make colors "pop," regardless of the user's OS setting.
+- **User Preference Overrides**: You can empower users to manually toggle the theme of a specific component. This allows them to view a component in the mode that fits their environment or eye comfort.
+- **Native Browser UI Consistency**: If a component contains built-in browser UI, such as **scrollbars**, **form controls**, or **dropdown menus**, you can decide in which color scheme they should be viewed.
```
LeaVerou: This reads to me like a general argument for setting an appropriate `color-scheme` rather than explaining why you'd need a **component-specific** color-scheme, but I could be just misunderstanding the point it's making.

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -21,109 +21,177 @@ sources:
...
-      color-scheme: dark;
-    }
-    ```
+## Why use component-specific themes?
 
-## Critical Considerations
+While most element's should follow the page-wide color scheme, there are several scenarios where you need more granular control over an element's theme:
+
+- **Content-Driven Aesthetics**: Certain components work best in a specific mode. For example, **code editors**, **video players**, and **photo galleries** often use a particular theme to minimize distraction and make colors "pop," regardless of the user's OS setting.
+- **User Preference Overrides**: You can empower users to manually toggle the theme of a specific component. This allows them to view a component in the mode that fits their environment or eye comfort.
```
LeaVerou: Any common enough examples we could list here to make this more concrete? All I can think of are things covered by the first bullet.

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -21,109 +21,177 @@ sources:
...
+  }
+
+  .dark-section .card {
+    background-color: var(--color-accent); 
+  }
+
+  /* ✅ FIX: Reapply the token where the scheme changes (e.g. background-color: var(--color-accent), **Not** `--color-accent: light-dark(black, white);`) */
+  .dark-section {
+    color-scheme: dark;
+    background-color: var(--color-accent); /* Forces re-resolution */
```
LeaVerou: (If I'm reading this right), this would actually be fine without any reapplication — you (correctly) haven't registered `--color-accent` and `background-color` is not an inherited property.

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -21,109 +21,177 @@ sources:
...
+  - **Mandatory**: Do not register properties meant to be design tokens that dynamically switch based on `light-dark()` as `<color>`. If you need to animate a color variable, use a separate property.
+- **Mandatory**: Do not set `color-scheme` on elements without a background.
+- **The "Inheritance Footgun"**
+
+### The Problem
+If you define a variable on the `:root`, the browser resolves it to a specific color (e.g., the light version) immediately. Descendant elements that change their `color-scheme` will inherit the *already-resolved color* rather than re-resolving the `light-dark()` function.
+
+- **Example: Handling Resolution Gaps**
+  ```css
+  /* ⚠️ PROBLEM: Variable resolves at :root (light) and stays 'yellow' */
```
LeaVerou: It does not. It only resolves at :root if you register it, which you shouldn't.

dvdherron: This was a mistake. Removed in https://github.com/GoogleChrome/guidance/pull/742/commits/4b6439f2e8adf044516126c972f1c01c86cd8ba8

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: Correct. Removed this example since this includes the more-detailed best practices section from `browser-ui-color-theme`. Updated the macro in https://github.com/GoogleChrome/guidance/pull/742/commits/73e4d75615d96ac6e8379793cc88aef8ad07fa6f

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: Chat box? Music player? Added those examples in https://github.com/GoogleChrome/guidance/pull/742/commits/4b6439f2e8adf044516126c972f1c01c86cd8ba8. Open to other suggestions.

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: Looking at it now, I think this doesn't belong in this guide. Removed in https://github.com/GoogleChrome/guidance/pull/742/commits/4b6439f2e8adf044516126c972f1c01c86cd8ba8

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -21,109 +21,146 @@ sources:
...
+  color-scheme: light;
+}
+```
+
+## Critical Considerations
+
+- **The `@property` Risk**:
+  - **Mandatory**: Do not register properties meant to be design tokens that dynamically switch based on `light-dark()` as `<color>`. If you need to animate a color variable, use a separate property.
+- **Mandatory**: Do not set `color-scheme` on elements without a background.
+ {{ INCLUDE("../browser-ui-color-theme/guide.md#best-practices") }}
```
LeaVerou, line 113: In my previous review I said you don't need to respecify `background-color`. That doesn't mean you don't need to respecify anything though! `background-color` is not inherited, so it didn't have this problem (unless `inherit` was involved). But you _still_ need to  respecify any **inherited** `<color>` properties that use `light-dark()`. 
This includes:
- Built-in properties like `color`, `accent-color`
- Any registered inheritable custom properties with a `<color>` syntax

Sorry for the back and forth. I should have expressed it more clearly. 
Let me try and take a stab at it, since this is the only blocking issue:


````suggestion
- **Mandatory**: Do not set `color-scheme` on elements without a background.
 {{ INCLUDE("../browser-ui-color-theme/guide.md#best-practices") }}
 - **Mandatory**: Do not register properties meant to be design tokens that dynamically switch based on `light-dark()` as `<color>`. If you need to animate a color variable, use a separate property.
   - Any **inherited** `<color>` properties that have been set to `light-dark()` values (directly or via design tokens) need to be specified again on any elements with a different `color-scheme` (even if that's to the same design token) otherwise they will still include the inherited value and will not adapt. This includes `color`, `accent-color`, `fill`, `stroke`, and any custom properties registered as `inherits: true; syntax: "<color>";`.

Example:
```css
:root {
	--accent-color: light-dark(blue, skyblue);
	accent-color: var(--accent-color);
	color-scheme: light dark;
}

pre, code {
	color-scheme: dark;
	/* Without this, accent-color would be blue, not skyblue */
	accent-color: var(--accent-color);
}
```

Note that this also affects non-inherited properties, if used with the `inherit` keyword. E.g. `background-color: inherit;` will inherit the resolved color from its parent, not a `light-dark()` expression.
````

Edit: I just see this is mentioned in a comment above, but a code comment should not be the _only_ place to mention this.

File: guides/user-experience/component-specific-light-dark-theme/guide.md [hunk 17]
```diff
@@ -21,109 +21,146 @@ sources:
...
+
 ## Fallback strategies
 
 {{ BASELINE_STATUS("light-dark") }}
 {{ BASELINE_STATUS("color-scheme") }}
 
-### Color Fallbacks
-For browsers that do not support `light-dark()` or the `color-scheme` property, provide a fallback using `@media (prefers-color-scheme: dark)` or manual variable overrides.
+- **Non-Color Properties**: Currently, `light-dark()` only supports color values. For other properties (like `padding` or `border-width`), you must continue using standard media queries or CSS Style Queries.
```
LeaVerou, line 120: Is this an issue that has actually come up? Has any agent ever used `light-dark()` for non-color values? (non-blocking)

guides/user-experience/component-specific-light-dark-theme/expectations.md
```diff
@@ -1,10 +1,15 @@
...
-- **MANDATORY**: The implementation uses progressive enhancement (e.g., `@supports (color: light-dark(white, black))`) to ensure that `light-dark()` logic is preferred over `@media` fallbacks, allowing per-component overrides to work correctly.
-- **MANDATORY**: `light-dark()` is only used for supported types (colors) and not for unsupported properties like `padding` or `font-size`.
-- **MANDATORY**: Each `light-dark()` call contains exactly two arguments.
+- **OPTIONAL**: The document includes a `<meta name="color-scheme" content="light dark">` tag in the `<head>` to declare theme support early and minimize the risk of un-themed content flashes.
+- **MANDATORY**: The document or the component has `color-scheme: light dark` (or includes both) set on the `:root` or a common ancestor to enable the resolution of the `light-dark()` function.
+- **MANDATORY**: The implementation uses custom properties as a **semantic abstraction layer** for theme-aware variables (e.g., `--card-bg` is the semantic variable, while `--card-bg-light` and `--card-bg-dark` are the raw tokens).
```
LeaVerou: "semantic abstraction layer" is something that requires human judgement — in practice the grader will just encode the specific examples and check for these (see #778). 
Something like this may be more limited, but is likely to produce a better grader.ts assertion in practice:

```suggestion
- **MANDATORY**: Outside of `@supports` feature detection, `light-dark()` is used with var() references for the two color arguments, not raw color values (e.g., `--card-bg: light-dark(var(--card-bg-light), var(--card-bg-dark))` while `--card-bg-light` and `--card-bg-dark` are the raw tokens that contain the actual color values).
```

guides/user-experience/component-specific-light-dark-theme/expectations.md
```diff
@@ -1,10 +1,15 @@
...
-- **MANDATORY**: Fallback strategies for colors (e.g., manual variable overrides and `@media (prefers-color-scheme)` queries) are provided for browsers that do not support `light-dark()`.
-- **MANDATORY**: The implementation uses progressive enhancement (e.g., `@supports (color: light-dark(white, black))`) to ensure that `light-dark()` logic is preferred over `@media` fallbacks, allowing per-component overrides to work correctly.
-- **MANDATORY**: `light-dark()` is only used for supported types (colors) and not for unsupported properties like `padding` or `font-size`.
-- **MANDATORY**: Each `light-dark()` call contains exactly two arguments.
+- **OPTIONAL**: The document includes a `<meta name="color-scheme" content="light dark">` tag in the `<head>` to declare theme support early and minimize the risk of un-themed content flashes.
+- **MANDATORY**: The document or the component has `color-scheme: light dark` (or includes both) set on the `:root` or a common ancestor to enable the resolution of the `light-dark()` function.
```
LeaVerou, line 2: Are there any use cases where we purposefully don't want to set `color-scheme: light dark` on the `:root`?

guides/user-experience/component-specific-light-dark-theme/guide.md
```diff
@@ -21,109 +21,146 @@ sources:
...
+### 1. Declare supported schemes in HTML
+OPTIONAL: To help prevent a "flash of un-themed content" (FOUC), place a `<meta>` tag in your `<head>` to ensure the browser knows which themes you support before it even starts rendering. While this `<meta>` tag helps to avoid FOUC by setting the initial canvas color early, it may not completely eliminate flashes in all browsers or loading conditions.
+
+```html
+<!-- Optional: Declare support for both light and dark themes -->
+<meta name="color-scheme" content="light dark">
+```
+
+### 2. Enable `color-scheme` support
+OPTIONAL: Enable global support for both color schemes by setting `color-scheme: light dark;` on the `:root`. If `color-scheme` is not set on `:root` (or an ancestor) the `light-dark()` function will not work as expected.
```
LeaVerou: This currently reads very odd: "Do this optional thing, but without it the feature won't work".

> If `color-scheme` is not set on `:root` (or an ancestor) the `light-dark()` function will not work as expected.

**Technically, this is false.** You could _absolutely_ have `color-scheme: dark` on a descendant and it would work. See [testcase](https://codepen.io/leaverou/pen/MYbjrVB).
 
It's _good practice_ to put `color-scheme: light dark` on the root, but that doesn't make the `light-dark()` function **work**, it works _anyway_. The reason we add `color-scheme: light dark` to the root is to override the default `color-scheme: normal` which computes to `color-scheme: light` for backwards compat. I.e. we need it for page-level dark mode to work, but technically we don't need it for component-specific dark mode.

Either remove that last sentence, keeping only this part:

```suggestion
OPTIONAL: Enable global support for both color schemes by setting `color-scheme: light dark;` on the `:root`.
```

Or just remove this section and the previous one entirely? They're good practices, but ultimately unrelated to component-specific `color-scheme` and are already recommended by the other `color-scheme` guides (which we reference here).

LeaVerou (changes requested):
See comments. 

The only issues I'd consider blocking are:
- the caveat around inherited color properties now seems to live entirely in code comments and was removed from the prose.
- Factually incorrect sentence (which makes it sound like things won't work unless an optional thing is done!)

Everything else is non-blocking. 

I've included suggestions for any blocking comment so we can get this merged quickly. 🙏🏼
If the suggestions look good to you, I'm fine with merging this!

guides/user-experience/component-specific-light-dark-theme/expectations.md
dvdherron: Edited in https://github.com/GoogleChrome/guidance/pull/742/commits/21618a1e62bf61fe9674167064c8bd968e7a0877

guides/user-experience/component-specific-light-dark-theme/expectations.md
dvdherron: Good question! I will have to think about it . . .

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: I shortened this sentence for now.

dvdherron:
> See comments.
> 
> The only issues I'd consider blocking are:
> 
>     * the caveat around inherited color properties now seems to live entirely in code comments and was removed from the prose.

When using the macros to point to other guides, should we operate as if the guidance from the guide linked to is consumed by the guide with the macro?

> 
>     * Factually incorrect sentence (which makes it sound like things won't work unless an optional thing is done!)
> 

Fixed. 
> 
> Everything else is non-blocking.
> 
> I've included suggestions for any blocking comment so we can get this merged quickly. 🙏🏼 If the suggestions look good to you, I'm fine with merging this!

Applied your suggestions and updated the demo so this should be good to go 🙏🏽

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: I haven't seen that directly. There was a conversation about whether or not we would add guidance about light-dark() image support: https://github.com/GoogleChrome/guidance/pull/577#issuecomment-4272494186

LeaVerou:
> When using the macros to point to other guides, should we operate as if the guidance from the guide linked to is consumed by the guide with the macro?

For `INCLUDE`, `FEATURE_FALLBACKS` etc, yes, since they literally transclude content as if it was copy-pasted.

For `GUIDE_REF` no, the agent may or may not choose to consume them.

LeaVerou (approved):
> Applied your suggestions and updated the demo so this should be good to go 🙏🏽

They don't seem merged? I see some are incorporated in a direct commit, which is fine, but [this point](https://github.com/GoogleChrome/guidance/pull/742/changes#r3236666318) seems to be missing in any shape or form? Is it because we're including the best practices from the other guide? Looking at [them](https://github.com/GoogleChrome/guidance/blob/main/guides/user-experience/browser-ui-color-theme/guide.md#best-practices), I'm not sure any of the other best practices from the other guide fit well here.

FWIW the concern about inherited properties would be fine to live here, since it *only* comes up when using `color-scheme` on individual elements. If you only use `color-scheme` on root, it never becomes an issue. If anything, it would be defensible to remove it from the other guide!

In any case, since the dark mode guides are due for a rework anyway, I don't think further iteration here is worth it — let's merge and revisit during the broader pass.

guides/user-experience/component-specific-light-dark-theme/guide.md
dvdherron: Thanks for cleaning this up while I was out! And apologies for the confusion!

PR #744, opened 2026-05-08
add install command

connorjclark:
adds `npx modern-web-guidance install` for installing these skills

for now, we defer to `npx skills`. But in the future we may switch this out.

ref https://docs.google.com/document/d/1AGcHNZyAOUHYubwabjJeNg7fFh1zPTUFZ5hX8nkyLTw/edit?usp=sharing&resourcekey=0-GIG327JksHNmwZ6bObP-LQ (paul and i brainstormed and this is our notes docs)


paulirish (approved):
so good.

PR #747, opened 2026-05-08
publish only skill files to github

connorjclark:


serving/skills-cli/publish-skills.ts
```diff
@@ -92,8 +103,21 @@ async function main() {
...
-    const files = await fs.readdir(publishCliDir, {recursive: true});
-    console.log(`\n[Dry Run] Skipping GitHub publishing. Would push:\n - ${files.filter(f => !f.includes('node_modules')).sort((a,b) => a.localeCompare(b)).join('\n - ')}`);
+    const files = await fs.readdir(publishCliDir, {recursive: true, withFileTypes: true});
+    const filteredFiles = files
+      .filter(f => !f.parentPath.includes('node_modules') && f.isFile())
+      .map(f => path.relative(publishCliDir, path.join(f.parentPath, f.name)))
+      .filter(f => {
+        return GH_PUBLISH_PATTERNS.every(pattern => {
+          if (pattern.startsWith('!')) {
+            return !minimatch(f, pattern.slice(1), { dot: true });
```
paulirish, line 113: at this point it feels kinda weird to reverse engineer how the GH_PUBLISH_PATTERNS get interpretted by `ghpages.publish`

hmm

do they have a dryrun mode and a useful return value that would illuminate this same thing for us?

serving/skills-cli/publish-skills.ts
```diff
@@ -4,11 +4,22 @@ import { execSync } from 'node:child_process';
...
 import { buildDist } from './build-dist.ts';
 import { fileURLToPath } from 'node:url';
+import { minimatch } from 'minimatch';
 
 const ROOT_DIR = path.resolve(import.meta.dirname, "../.."); // guidance/
 const SERVING_DIR = path.join(ROOT_DIR, "serving");
 const DIST_DIR = path.join(ROOT_DIR, "dist");
 
+// This controls what is published to https://github.com/GoogleChrome/modern-web-guidance.
+const GH_PUBLISH_PATTERNS = [
```
paulirish, line 14: its a good list. and clear.

paulirish (approved):
i can live with the shenanigans but so curious if we have a better option

serving/skills-cli/publish-skills.ts
connorjclark: > do they have a dryrun mode

no

this is just basic minimatch stuff, not much to reverse engineer

PR #748, opened 2026-05-09
Add SKILL for removing common knowledge from discipline guides

micahjo7:
Adds `.agents/skills/project-discipline-guides` SKILL to the repo, which can be run on the discipline guides to remove the common knowledge from them. Uses GCLI, Claude, Codex "mirrors" of the guide to identify what information is already available to those agents (related to the discipline), finds the content that is present across all of them, and removes this content from the discipline guide.

Originally authored by @paulirish. Changes since then: https://github.com/GoogleChrome/guidance/pull/748/changes/af526dd00f98eac328cb2840ba6e7bcb3d471f55..1603c92328dfa8afa0735bc1a4a423aa6abb6744.

Report from the SKILL execution:

### Surgical Refactoring Highlights on the JavaScript Guide
By executing a strict $A - B$ Inverse Filter subtraction using **only** the unified [mirror.md](file:///Users/micahjo/.agents/skills/project-discipline-guides/mirrors/javascript/mirror.md) as our baseline, we achieved a mathematically precise refactoring of the JavaScript guide:
#### 🗑 Deletions (100% Redundant Common Knowledge)
Because these rules were exhaustively and consistently outputted across Gemini, Claude, and Codex, they were completely pruned from the local guide:
*   **`## Modern Syntax and Variable Declarations`**: Replaced variables with `const`/`let`, deprecating `var`, strict equality `===`, Optional Chaining `?.`, Nullish Coalescing `??`, and Template Literals.
*   **`## Functions and Async Flow`**: Pruned default parameters, rest parameters, arrow functions, standard function dynamic bindings, early returns, and async/await/Promise combinators (`Promise.all` and `Promise.allSettled`).
*   **`## Collections and Iteration`**: Pruned array literals, iteration loops (`for...of`), Sets, Maps, and the dynamic Collections matrix.
*   **Standard Web APIs & Security**: Pruned basic DOM element query selection (`document.querySelectorAll`), basic custom events, standard event binding (`addEventListener`), `eval` deprecations, and production script minification (`preprocess`).
*   **`structuredClone()`**: Pruned deep copy rules, as deep object copying via native `structuredClone()` is now universally understood and supported by all three mirrors.
#### 🛡 Preservations (High-Value Differential Knowledge)
Because these guidelines are highly unique, counteract common AI coding biases, or are modern senior-level APIs missing from the baseline mirrors, they were surgically kept or restructured:
*   **Modern Browser APIs (`Object.groupBy`)**: Surgically dissected. The standard DOM APIs in this bullet point were deleted, but the senior guideline steering the model **strictly toward native `Object.groupBy()`** (and away from Lodash `groupBy` or loop grouping logic) was perfectly isolated and preserved in [guide.md:L79](file:///Users/micahjo/guidance/guides/javascript/javascript/guide.md#L79).
*   **`## Objects, Classes, and Prototypes`**: Preserved safe class checks like **`Object.hasOwn`** (which was missing from the static methods of the mirrors), **prototype poisoning utilities** (`Object.getPrototypeOf`/`Object.setPrototypeOf`), and encapsulation custom rules for class **getters/setters**.
*   **DOM Performance Optimizations**: Preserved 100% of synchronous layout reflow prevention, multi-page and high-performance AbortController signal cleanup events, shared Observer instances (`ResizeObserver`/`IntersectionObserver`), and DocumentFragment batching.
*   **Modern Progressive Enhancements**: Preserved HTML/CSS-over-JS overrides (native popovers, Container Queries, detail components) to prevent models from generating heavy JS bloat.
*   **Deliverability & Security**: Preserved import maps `<script type="importmap">` and CSP **Trusted Types** constraints.
---
### Verification and Quality State
The refactored [guide.md](file:///Users/micahjo/guidance/guides/javascript/javascript/guide.md):
1. Is **over 60% smaller**, providing a massive context window token savings.
2. Contains **zero common knowledge redundancies**, leaving only high-value behavioral constraints.
3. Compiles and passes the project types check (`pnpm typecheck`) successfully.


.agents/skills/project-discipline-guides/scripts/generate_mirrors.ts
```diff
@@ -0,0 +1,122 @@
...
+  if (result.error) {
+    console.error('Error spawning Codex CLI:', result.error);
+    return '';
+  }
+  if (result.status !== 0) {
+    console.error(`Codex CLI failed with exit code ${result.status}:`, result.stderr);
+    return '';
+  }
+  return result.stdout || '';
+}
```
LeaVerou: Nit: These three functions are basically the same logic with minor differences.
What about something like this? (Disclaimer: I haven't tested it, if you use it, have an agent check it first)

```suggestion
const clis = {
	gemini: {
		env: 'GEMINI_CLI_BIN',
		args: ['-p', prompt, '-o', 'text', '--yolo', '--skip-trust'],
	},
	claude: {
		env: 'CLAUDE_CODE_CLI_BIN',
		args: ['-p', prompt, '--dangerously-skip-permissions', '--output-format', 'text'],
	},
	codex: {
		env: 'CODEX_CLI_BIN',
		args: ['exec', prompt, '--yolo'],
	}
};

const capitalize = str => str[0].toUpperCase() + str.slice(1)
async function callCli(prompt, id, {name = capitalize(id), env, command = process.env[env] || id, args = [] } = clis[id]) {
  const command = process.env[env] || id;
  console.log(`Executing ${name} CLI: ${command} ...`);
  const result = spawnSync(command, args, { encoding: 'utf8' });
  if (result.error) {
    console.error(`Error spawning ${name} CLI:`, result.error);
    return '';
  }
  if (result.status !== 0) {
    console.error(`${name} CLI failed with exit code ${result.status}:`, result.stderr);
    return '';
  }
  return result.stdout || '';
}
```

Use like `callCli(prompt, 'gemini')` instead of `callGemini(prompt)` (or define aliases).

.agents/skills/project-discipline-guides/scripts/generate_mirrors.ts
```diff
@@ -0,0 +1,122 @@
...
+  if (!discipline) {
+    console.log('Usage: node generate_mirrors.ts <discipline_name> (e.g., "JavaScript", "CSS")');
+    process.exit(1);
+  }
+
+  const prompt = `
+Based on your training data and inherent knowledge of ${discipline} development, generate a comprehensive guide of standard best practices, syntax, and APIs that you natively understand and would apply by default.
+
+Focus on:
+1. Modern language features (ES2022+ for JS, etc.).
```
LeaVerou: "Modern" is a bit subjective. How about:

```suggestion
1. The guide should be based on the most modern language features you're aware of, as long as it has shipped in every major browser, even if only recently.
2. You can also recommend cutting edge features that have only shipped in one browser, as long as they are progressive enhancement or have good fallbacks/polyfills.
```

The overshoot is intentional, see https://www.linkedin.com/feed/update/urn:li:activity:7434249998113910784/ ) — it's not like we're going to include all of these.

.agents/skills/project-discipline-guides/scripts/generate_mirrors.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+async function main() {
+  const discipline = process.argv[2];
+  if (!discipline) {
+    console.log('Usage: node generate_mirrors.ts <discipline_name> (e.g., "JavaScript", "CSS")');
+    process.exit(1);
+  }
+
+  const prompt = `
```
LeaVerou: Would it make sense to move this to an `.md` file so it can be edited more easily, even by folks who are not comfy editing JS/TS? (and for everyone's convenience) We can use a `{{ discpiline }}` token for the only variable and replace it after import.

.agents/skills/project-discipline-guides/scripts/generate_mirrors.ts
```diff
@@ -0,0 +1,122 @@
...
+Based on your training data and inherent knowledge of ${discipline} development, generate a comprehensive guide of standard best practices, syntax, and APIs that you natively understand and would apply by default.
+
+Focus on:
+1. Modern language features (ES2022+ for JS, etc.).
+2. Standard library/API usage.
+3. Common clean code principles.
+
+This guide will be used as a "Redundancy Mirror" to prune a project-specific skill file. 
+We want to see exactly what you consider "Common Knowledge" so we can delete it from our local guides.
```
LeaVerou:
```suggestion
We want to see exactly what you consider "Common Knowledge" so we can delete it from our local guides, so be as exhaustive as you can. Do not omit guidance for conciseness, do not prioritize.
```

(just like above, since we're just using these in an A - B fashion, we want B to be as big as possible)

.agents/skills/project-discipline-guides/SKILL.md
```diff
@@ -0,0 +1,57 @@
...
+- `mirrors/<discipline>_claude_mirror.md` (via Claude)
+- `mirrors/<discipline>_codex_mirror.md` (via Codex CLI)
+
+> [!NOTE]
+> If generation fails for any model, the script will log a warning and skip writing that file, but it will not block execution. You should proceed with the mirrors that are successfully generated.
+
+### 2. Perform the Inverse Filter (Agent Task)
+As the agent, read the **Target** guide and the available **Knowledge Mirrors**. 
+
+- **Identify Redundancies**: Any rule, syntax, or pattern in the Target that is already well-covered in *all available* mirrors is "Common Knowledge" and should be deleted.
```
LeaVerou: I would separate the AND into a separate, verifiable step: Have a skill that merges the guides and generates a merged guide, then this skill only needs to read the one input guide. 
Otherwise, I'd worry it could end up removing things that are only present in one of the guides and we have no control over it.

.agents/skills/project-discipline-guides/SKILL.md
```diff
@@ -0,0 +1,57 @@
...
+node .agents/skills/project-discipline-guides/scripts/generate_mirrors.ts <discipline_name>
+```
+
+This will create files in a `mirrors/` directory:
+- `mirrors/<discipline>_gemini_mirror.md` (via Gemini)
+- `mirrors/<discipline>_claude_mirror.md` (via Claude)
+- `mirrors/<discipline>_codex_mirror.md` (via Codex CLI)
+
+> [!NOTE]
+> If generation fails for any model, the script will log a warning and skip writing that file, but it will not block execution. You should proceed with the mirrors that are successfully generated.
```
LeaVerou: Shouldn't we make sure they're all generated before running this?

.agents/skills/project-discipline-guides/SKILL.md
```diff
@@ -0,0 +1,57 @@
...
+
+### 2. Perform the Inverse Filter (Agent Task)
+As the agent, read the **Target** guide and the available **Knowledge Mirrors**. 
+
+- **Identify Redundancies**: Any rule, syntax, or pattern in the Target that is already well-covered in *all available* mirrors is "Common Knowledge" and should be deleted.
+- **Identify Standard Inherent Knowledge**: Even if not explicitly in the mirrors, if a rule is something you (the agent) would do by default based on your own training (e.g., "use const"), it is redundant.
```
LeaVerou: Hmmmmmmmm. What I like about this skill is that it separates what the agents know from what's already there and asks them to generate guides without looking. The rate of false positives is zero: by definition, if a feature shows up in all three guides, it's something agents know about.

With this, I worry that when looking, they can easily say "Oh this is obvious" while in practice it isn't. There is no test, it's just their own attestation, and that's not usually a recipe for success.
I've actually seen Claude hallucinate awareness: when I asked it about the CSS skill at some point to see what it would flag: it flagged parts I _knew_ it gets wrong.

That said, I also worry that the approach will also have a lot of false negatives. Perhaps for the stuff that remains we can generate some sort of test to see if agents will know about them. But there needs to be some kind of test, it cannot just be "Do you know this? Yes? Ok, we'll remove it then".

.agents/skills/project-discipline-guides/SKILL.md
```diff
@@ -0,0 +1,57 @@
...
+- **Preserve Differential Knowledge**: Keep only what is unique to this project or necessary to guide the AI effectively. This generally includes:
+    - Rules that counter common AI biases.
+    - Conventions chosen among valid alternatives.
+    - Advanced patterns or heuristics that models know but often omit in their default output.
```
LeaVerou: See above. It's very hard to ask AI to overcome its own blind spots (humans typically cannot do this either!). I've tried, and it tends to hallucinate stuff that doesn't really help… But I haven't tried with Gemini/jetski, perhaps it's better at this (?)

If not, I'd suggest keeping it to A - B now, seeing how that performs, and deciding next steps accordingly.

.agents/skills/project-discipline-guides/SKILL.md
```diff
@@ -0,0 +1,57 @@
...
+
+> [!TIP]
+> While the primary goal is reduction, you may also **add** rules that fit the criteria for Differential Knowledge (e.g., new project-specific choices or behavioral steering) if you identify a need for them during the analysis.
```
LeaVerou: I'd make this a separate step too. In my experience the more focused tasks are, the better they tend to perform, we don't want to be adding and removing in the same pass, that's also harder to review. But same caveat as above applies, maybe Gemini/Jetski doesn't need this (?)

guides/javascript/javascript/guide.md
```diff
@@ -163,20 +49,16 @@ list.appendChild(fragment); // Single layout recalculation pass
 
 ## Modern Browser APIs
 
-- **Prefer Modern Browser APIs over Third Party Libraries** - Use native APIs like `document.querySelectorAll()` over jQuery selectors, and `Object.groupBy()` over Lodash `groupBy()`.
```
LeaVerou, line 166: Can you post the three guides somewhere that the skill operated on? I have never, ever seen Claude use `Object.groupBy()` — it always reimplements it, so I find that very surprising. In fact, `Object.groupBy()` was in my notes as a feature agents frequently get wrong.

I wonder if the guides included `document.querySelectorAll()` and because they're in the same bullet point, `Object.groupBy()` was thrown out too.

LeaVerou (commented):
Left some comments, but overall this is very promising. 😍
Love what it did to the JS guide! (minus that one point)

To sum up:
- I think instead of giving the agent all three guides and asking it to remove features present in all three, we should *first* intersect the three guides in a separate step, and give agents only the unified, LCD guide. This would ensure agents don't slip and remove features only present in one of the three guides — they wouldn't have access to this info at all.
-  I would remove the instruction that gives them free reign to remove additional things if they would do them by default. Without a verification step, I worry that this is a recipe for hallucinations. Even for humans, you get very different data if you ask them "do you know X?" vs *testing* whether they know X. That said, you also get different results when asking for recall, so ideally there should be a way to test knowledge without requiring them to mention it offhand first. 
- In the sample guide, I suspect valid guidance was removed because it was in the same bullet point as guidance to be removed, but without seeing the three guides, it's impossible to verify this. Def something to check before applying this more broadly!

Soooo looking forward to this landing!

.agents/skills/project-discipline-guides/scripts/generate_mirrors.ts
```diff
@@ -9,45 +10,49 @@ if (typeof process.loadEnvFile === 'function') {
...
-const GOOGLE_API_KEY = process.env.GOOGLE_API_KEY;
-
-if (!ANTHROPIC_API_KEY || !GOOGLE_API_KEY) {
-  console.error('Error: ANTHROPIC_API_KEY and GOOGLE_API_KEY environment variables must be set.');
-  process.exit(1);
-}
-
 async function callGemini(prompt) {
-  // Using gemini-3-flash as requested
-  const url = `https://generativelanguage.googleapis.com/v1beta/models/gemini-3-flash:generateContent?key=${GOOGLE_API_KEY}`;
```
paulirish: So... i _was_ thinking we use the rest APIs directly.. That way you know for sure there's zero tool use. 

But.. The downside is that.. API keys are a little less straightforward for each of us (i imagine) than running the coding agent. 

Given that, coding agents can work but **we gotta remove the yolo flags**.  (Maybe also add text to the prompt about "dont use any tools" but... given that its a CLI usage without yolo maybe the agents already exclude the tools. dunno).

.agents/skills/project-discipline-guides/scripts/generate_mirrors.ts
paulirish: if this was more _core_ code, i'd agree and probably go further...  DRY this up against plenty of our other agent harness impl's.

But.... it's way out here and... doesn't bother me too much so.. whatevs.

paulirish (commented):
had these comments sitting for a couple days. still not done but flushing now.

.agents/skills/project-discipline-guides/scripts/generate_mirrors.ts
micahjo7: added `callCli`

.agents/skills/project-discipline-guides/scripts/generate_mirrors.ts
micahjo7: added `mirror_prompt.md`!

.agents/skills/project-discipline-guides/scripts/generate_mirrors.ts
micahjo7: updated

.agents/skills/project-discipline-guides/scripts/generate_mirrors.ts
micahjo7: updated in `mirror_prompt.md`

.agents/skills/project-discipline-guides/SKILL.md
micahjo7: added a separate step that merges the mirrors, and the SKILL instructs the agent to use only the merged version

.agents/skills/project-discipline-guides/SKILL.md
micahjo7: made some updates to the mirror_prompt and SKILL to hopefully address this! added directive to prevent tool usage, and `No Subjective Self-Attestation` to the SKILL

.agents/skills/project-discipline-guides/SKILL.md
micahjo7: updated SKILL to address

.agents/skills/project-discipline-guides/SKILL.md
micahjo7: updated

guides/javascript/javascript/guide.md
micahjo7: added the mirror files and the merged mirror to the SKILL folder!

micahjo7: in the latest version of the guide, looks like the groupBy is present

.agents/skills/project-discipline-guides/scripts/generate_mirrors.ts
micahjo7: added a directive to disallow tool usage and removed yolo flag

micahjo7:
thank you for the review! @paulirish @LeaVerou 

updated the PR description with the agent result after the SKILL ran

paulirish (approved):

.agents/skills/project-discipline-guides/SKILL.md
```diff
@@ -0,0 +1,59 @@
...
+As the agent, read the **Target** guide and the unified **Redundancy Mirror** (`.agents/skills/project-discipline-guides/mirrors/<discipline>/mirror.md`).
+
+- **Strict A - B Comparison**: Compare rules in the Target guide strictly against the unified **Redundancy Mirror**. You should ONLY remove a rule or pattern from the Target guide if it is fully and clearly covered in the Redundancy Mirror.
+- **No Subjective Self-Attestation**: Rely exclusively on the literal content present in the Redundancy Mirror. Do not prune rules based on your own subjective self-attestation of what you "know" or would do by default. If a concept is not in the Redundancy Mirror, it remains in the Target guide.
+- **Avoid Blind Bullet-Point Deletions**: If a single bullet point or rule in the Target contains multiple guidelines or APIs, do **not** delete the entire bullet point just because some parts of it are common knowledge. Carefully split or dissect the rule, pruning only the redundant parts and explicitly retaining any differential knowledge (e.g., preserving instructions for `Object.groupBy()` even if standard DOM APIs like `document.querySelectorAll()` in the same bullet point are removed).
```
LeaVerou: The example is oddly specific and makes no sense as a general point (since there's no obvious reason for these two APIs to be together).
I'd suggest one of:
1. making it a whole example (the bullet point, what features were mentioned in the guide, how to dissect it)
2. removing it entirely (the point is pretty clear even without it)
3. Using features that would naturally go together, but one is much more likely to be in the guide than the other (e.g. `Object.groupBy()` and `Object.hasOwn()`)

LeaVerou (approved):
Nice! I think that's much better.

Non-blocking but I think the rewrite peppers `Object.groupBy()` as an example in various places, but it makes no sense out of context like that, it doesn't quite stand by itself as an example. Especially in the place where it combines it with qSA — these two APIs have nothing in common!

As a side note, I find it fascinating that it was Claude that was the only one that suggested `Object.groupBy()` given that in my experience it never actually uses it. That makes me worry a little about the approach as a whole. But hopefully the intersection should help with that. So, assuming the approach is solid, this LGTM.

.agents/skills/project-discipline-guides/SKILL.md
micahjo7: thanks! feel like it's better to just remove it

PR #751, opened 2026-05-09
centralize redundant fallbacks transclusion

paulirish:
thx to #717  we can apply this technique more broadly.

the general approach (IMO) is to have less text floating around making it easier to review and maintain. also lower chance of inconsistencies.

cc @LeaVerou 

this is what this did

image

## macro-expansion diffs!

and I have a script which illuminates the effect of all macros:

look at [the "Summary" page of the "CI" workflow](https://github.com/GoogleChrome/guidance/actions/runs/25611989708?pr=751).. for a markdown summary of exactly what changed in the BUILT guides (which is kinda hard to tell because macros)

eg
image

## forms changes

Notably.. in this batch of forms things... there is this diff:

image

But.. I decided that "suggest a strong password" stuff shouldnt be on the sign-IN form.. only on the sign-up form (where it remains). So.. yeah

## followups

- i can do another followup for UX/perf guides... (it was a lot of stuff when i had them included)

### DIFF
--- a/features/auth-forms.md
+++ b/features/auth-forms.md
@@ -1,0 +1,88 @@
+# Authentication Forms
+
+## Own Form Element { #auth-own-form }
+
+Always use the `<form>` element when you're getting users to enter data.
+
+Don't wrap inputs in a `<div>` and handle input data submission purely with JavaScript. It's generally better to use a `<form>` element. This makes your site accessible to screenreaders and other assistive devices, enables a range of built-in browser features, makes it simpler to build basic functional authentication for older browsers, and can still work even if JavaScript fails.
+
+## Don't Double Up Inputs { #auth-dont-double-up }
+
+Some sites force users to enter emails or passwords twice. That might reduce errors for a few users, but causes extra work for all users, and increases abandonment rates. Asking twice also makes no sense where browsers autofill email addresses or suggest strong passwords. It's better to enable users to confirm their email address (you'll need to do that anyway) and make it easy for them to reset their password if necessary.
+
+## Password Privacy & Show Password { #auth-keep-passwords-private }
+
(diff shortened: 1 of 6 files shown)


LeaVerou:
Love this! ❤️

Nit: The `features/` dir was meant for things that have a web-feature id, so that down the line we could even potentially automate certain things (e.g. for the web-features linked in a guide, automatically include their stuff? Things like that). I wonder if it may make sense to have a separate place for "common" guides. Though looking at `forms` and `auth-forms`, these would actually be perfectly valid as skills too, so perhaps that could drive their naming/filepath (or we could just decide to yolo it and put whatever in `features/`, your call!)

LeaVerou:
@rviscomi Are there any blockers for this? I was thinking, if we're going to merge this before I/O, we should probably do it soon? I found myself checking it to see if I should avoid creating a new feature guide in case it existed here already. And it will be a minefield of conflicts if we leave it too long...

Also the macro-expansion diffs script @paulirish wrote is _really_ useful and I'd love to be able to use it!

rviscomi (approved):

rviscomi:
No concerns!

LeaVerou (approved):
I had reviewed this a while back, but it wasn't showing up as approved by me, so here goes. 😀 

@paulirish are you able to rebase this so we could merge it? Hopefully the guides it changes haven't changed too much…

LeaVerou (changes requested):
Sorry, just took a closer look now.
`auth-forms` and `forms` are basically discipline guides, but they are in `features`. The items in `features` are meant to be web-feature ids.
I think they should be moved to `guides/forms` and be marked as discipline guides. We can still transclude text from other guides with the macros, so that part should be fine!

PR #752, opened 2026-05-09
speed up preflight

paulirish:
by generating the embeddings less frequently. (locally and on CI)

and on CI, use a cache of embeddings.


paulirish:
reverted. made things worse.

PR #755, opened 2026-05-10
add corner-shape guidance

### LINKED ISSUE #693: corner-shape
### web-feature-id

corner-shape

### Feature description

The corner-shape CSS property sets the shape of an element's corners when using border-radius, allowing for shapes other than rounded corners. For example, corner-shape: squircle is a shape in between a square and rounded corner.

### Expected release date

Chrome 139, mentioned at I/O

### Coding agent assessment

_No response_

### Reference sources

- https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/corner-shape

**Use cases:**
- [ ] #1233
- [ ] #1254

patrickkettner:
fixes #693 (partially?)
When I ran the eval generators and graders, I was never able to get the untrained guide to match less than 5/6. The guidance always matches 6/6, but that started to make me question how useful this guidance would be, given how successful agents were. at the very least I thought it pushed credulity. I didn't push any of my generated output since i don't think it would be useful in its current form, but would love thoughts from others.


rviscomi:

guides/user-experience/corner-shape/guide.md
```diff
@@ -0,0 +1,9 @@
+---
+name: corner-shape
+description: Shape an element's corners beyond the default rounded curve, using values like squircle, bevel, notch, scoop, or a custom superellipse.
+web-feature-ids:
+  - corner-shape
+sources:
+  - https://drafts.csswg.org/css-borders-4/#corner-shaping
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/corner-shape
```
jamesnw:
```suggestion
```

guides/user-experience/corner-shape/guide.md
```diff
@@ -0,0 +1,9 @@
+---
+name: corner-shape
+description: Shape an element's corners beyond the default rounded curve, using values like squircle, bevel, notch, scoop, or a custom superellipse.
```
jamesnw:
```suggestion
description: Shape an element's corners into a squircle, bevel, notch, scoop, or a custom superellipse.
```
Suggestion- I don't know if this will help with discoverability or the grading question mentioned on the PR, but this shifts focus to the effects rather than the keywords.

jamesnw (approved):
I wonder if having the guide mention different uses for each would help with discoverability? Something like "Use a scoop for a ticket effect", or perhaps a table that lists the effects and potential ways it could be referred to.

LeaVerou (changes requested):
Hi there!

In terms of content, like @jamesnw said, these should be framed as specific use cases. Currently both the description and the demo looks like reference documentation:

Image

If our entire message to agents is "hey, look at `corner-shape` if you want shaped corners" that could be one line in `css/css/guide.md` (and that's probably worth doing anyway).

For use case guides we're looking at nontrivial use cases, and for this I think we need **smaller, focused guides**. It's unlikely an agent would be looking up how to create squircle app icons AND coupon shapes AND a breadcrumb arrow AND a page flip effect **at the same time** etc, so having everything in one guide just pollutes its context window.

Basically, `corner-shape` is a solution, not a use case. And looking at #693, it's a feature issue, which is in the "Needs use cases" state, which tracks!

So I'm afraid my recommendation would be to close this, list actual use cases where `corner-shape` is useful (as a PR with stubs, or even just a list in #1005 and me or someone else could submit the PRs).

rviscomi:
Closing per @LeaVerou's recommendation

PR #756, opened 2026-05-10
Add guidance for element-scoped view transitions

### LINKED ISSUE #694: Element-scoped view transitions
### web-feature-id

view-transitions-element-scoped

### Feature description

The startViewTransition() method of an Element object starts a view transition that affects only that element's DOM tree. The contain: view-transition CSS declaration contains the layout of the tree to be transitioned. You can use this to run separate elements' transitions concurrently.

### Expected release date

Chrome 147, mentioned at I/O

### Coding agent assessment

_No response_

### Reference sources

- https://developer.chrome.com/blog/element-scoped-view-transitions

**Use cases:**
- [ ] #1353

patrickkettner:
fixes #694


rviscomi:

guides/user-experience/scoped-component-transitions/guide.md
```diff
@@ -0,0 +1,12 @@
...
+name: scoped-component-transitions
+description: Animate state changes inside a single component (a card, list, panel, or other subtree) without pausing the rest of the page or conflicting with other transitions running concurrently.
+web-feature-ids:
+  - view-transitions-element-scoped
+  - view-transitions
+sources:
+  - https://github.com/WICG/view-transitions/blob/main/scoped-transitions.md
+  - https://developer.chrome.com/blog/element-scoped-view-transitions
+  - https://developer.chrome.com/docs/css-ui/view-transitions/element-scoped-view-transitions
+  - https://drafts.csswg.org/css-view-transitions-2/
```
jamesnw:
```suggestion
```

guides/user-experience/scoped-component-transitions/guide.md
```diff
@@ -0,0 +1,12 @@
+---
+name: scoped-component-transitions
+description: Animate state changes inside a single component (a card, list, panel, or other subtree) without pausing the rest of the page or conflicting with other transitions running concurrently.
```
jamesnw:
```suggestion
description: Animate state changes inside a single component (a card, list, panel, or other subtree) while the rest of the page remains interactive, and other transitions run concurrently.
```
Nit- I think "pausing the page" might not be clear.

jamesnw (approved):
LGTM!

patrickkettner:
the only things I can think of are several widgets transitioning at once or isolating a components transition from page-level ones, and I think both of those end up being the same implementation, (at least how I would think to make them). I really am struggling to think of any other use cases. I pushed an update to put it in the new directory, for the moment. WDYT @rviscomi?

guides/motion/scoped-component-transitions/demo.html
```diff
@@ -0,0 +1,199 @@
...
+         painted output can be captured atomically. The browser auto-applies
+         this during a scoped transition; declaring it up-front avoids a
+         one-frame layout shift the moment the transition begins. */
+      contain: layout;
+
+      /* The scope: isolates this card's view-transition-name namespace. Both
+         cards reuse `view-transition-name: match-element` on their <li>s;
+         without this property an ancestor's transition could capture our
+         participants and our names could collide with another scope's. */
+      view-transition-scope: all;
```
bramus: This property should be declared on the wrapper, not the element itself.

But, there is no real need to this, as an Element-Scoped View Transition automatically applies this on the wrapper when you call `Element.startViewTransition` (and also removes it after the transition has finished).

guides/motion/scoped-component-transitions/demo.html
```diff
@@ -0,0 +1,199 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Scoped Component Transitions Demo</title>
```
bramus: Nit: Maybe call it “Element-Scoped View Transitions Demo”?


```suggestion
  <title>Element-Scoped View Transitions Demo</title>
```

guides/motion/scoped-component-transitions/demo.html
```diff
@@ -0,0 +1,199 @@
...
+
+    function render(listEl, items) {
+      listEl.replaceChildren(
+        ...items.map((label) => {
+          const li = document.createElement('li');
+          li.textContent = label;
+          return li;
+        }),
+      );
+    }
```
bramus, line 180: This effectively destroys the current list of children and replaces them with new ones. As a result, the `view-transition-name: match-element` declaration will not work, as it relies on element identity … so it’ll assign a new `view-transition-name` to the newly created elements.

Workaround either is to:

- Use explicit values for the `view-transition-name`: instead of `match-element` which relies on element identity (see https://www.bram.us/2026/06/19/view-transition-name-attr-or-match-element/ for details), set the value to the passed in `label`.
- Adjust the rendering so that it moves the existing nodes instead of destroying+recreating them.

The demo at https://codepen.io/bramus/pen/raVbjZX?editors=0010 implements the second approach listed here.

bramus (changes requested):

guides/motion/scoped-component-transitions/demo.html
patrickkettner: updated. fun find while testing: declaring both `view-transition-scope: all` and `contain: layout` up front made chrome 151 abort the transition with an InvalidStateError, either one alone is fine. So I dropped `contain: layout` too since the browser applies it anyway.

patrickkettner:
@bramus reworked the demo around your comments, plus added a rotate-both button and per-item removal so it shows two scoped transitions animating concurrently (the point of #694). one more chrome 151 find from testing: removing a `match-element` participant aborts a scoped transition, the same removal works document-level. an explicit name on the outgoing item avoids it. happy to adjust if the extra interactions feel like too much for the demo.

guides/motion/scoped-component-transitions/demo.html
```diff
@@ -0,0 +1,263 @@
...
+      ::view-transition-old(*),
+      ::view-transition-new(*) {
+        animation: none !important;
+      }
+    }
+  </style>
+</head>
+<body>
+  <div class="toolbar">
+    Toolbar (stays visible during scoped transitions)
```
bramus: Nit: the “(stays visible during scoped transitions)” seems unnecessary here.

```suggestion
    Toolbar
```

guides/motion/scoped-component-transitions/demo.html
```diff
@@ -0,0 +1,263 @@
...
+    <button id="rotate-both">Rotate both at once</button>
+  </div>
+
+  <h1>Element-Scoped View Transitions</h1>
+  <p class="intro">Rotate either card, or both at once, or remove an item. Each transition
+    runs inside its own card, so the two animate concurrently, the toolbar stays put, and
+    the rest of the page remains interactive.</p>
```
bramus: The remark about the toolbar is unnecessary, as there is no captured VT-content that could overlap it.


```suggestion
  <p class="intro">Rotate either card, or both at once, or remove an item. Each transition
    runs inside its own card, so the two animate concurrently, and the rest of the page
    remains interactive.</p>
```

bramus (approved):
Some minor nits about the wording related to the toolbar, but overall LGTM :)

PR #757, opened 2026-05-10
Add at-rule-support guidance

### LINKED ISSUE #690: at-rule()
### web-feature-id

supports-at-rule

### Feature description

The at-rule() function, when used with @supports, checks if a CSS at-rule is supported. For example @supports at-rule(@starting-style) checks if the @starting-style at-rule is supported.

### Expected release date

Chrome 148, mentioned at I/O

### Coding agent assessment

_No response_

### Reference sources

- https://www.bram.us/2026/03/15/at-rule/

patrickkettner:
fixes #690


rviscomi:

jamesnw (commented):
I'm a bit uncertain of what a use case would look like for this. I can see `@supports at-rule()` being used in other guides for fallbacks, but can't think of a user prompt that would only add this.

It's also tricky because, given the current support for the `at-rule()` extension, this would, in non-Chromium cases, check for support of the `at-rule()` extension itself, and not the at-rule being queried.

I'm checking with some other contributors for ideas as well.

patrickkettner:
@jamesnw thanks for the feedback! did you hear anything back? or is there anyone whose opinion I can track down?

jamesnw:
@patrickkettner We didn't come up with a direct use case. I think once this is more widely available, it may be syntax that guides use for fallbacks or progressive enhancements.

patrickkettner:
Agreed.

firefox [doesnt recognize at-rule() yet](https://bugzilla.mozilla.org/show_bug.cgi?id=1751188) ([or safari](https://bugs.webkit.org/show_bug.cgi?id=235400)), so

```css
@supports at-rule(@property) {
    body { background: green }
}
@supports not at-rule(@property) {
    body { background: red }
}
```

would give chrome users a green bg, and firefox/safari folks a red, despite all three supporting it. Any functional guidance today would do that to every firefox (and safari) user.
  
@rviscomi ok to convert to draft or close?

guides/user-experience/detect-at-rule-support/guide.md
```diff
@@ -0,0 +1,10 @@
+---
+name: detect-at-rule-support
+description: Apply CSS only when the browser supports a specific at-rule, enabling progressive enhancement for new CSS at-rules without JavaScript.
+web-feature-ids:
+  - supports-at-rule
+sources:
+  - https://drafts.csswg.org/css-conditional-5/#typedef-supports-at-rule-fn
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/@supports
+  - https://www.bram.us/2026/03/15/at-rule/
+---
```
LeaVerou:
```suggestion
---

<!-- TODO
- Detecting `at-rule()` support itself
- JS fallback
- Guidance to try alternatives first (e.g. `@supports(color: --foo())` instead of `@supports at-rule(@function)` see https://codepen.io/leaverou/pen/qERMKgd?editors=0100 )
--> 
```

LeaVerou (approved):
LGTM, I left a suggestion to remind us to cover these topics. It would also need to be moved to `css/` before merging.

LeaVerou (changes requested):
Actually, apologies, I forgot to look at `demo.html`.

This is not a good example — simply using `@starting-style` and relying on PE would have worked the same. `@supports` is for cases where we want to **gate supported things on unsupported things**. If the only thing we put in the `@supports` is the thing we're testing, CSS's forgiving parsing does the job already.

patrickkettner:
@rviscomi picking this back up. safari landed at-rule() in webkit on aug 3 and firefox has it assigned, but neither has shipped, so the fallback problem above is still live for every non chrome user. ok to close this and reopen when one of them ships?

rviscomi:
> @rviscomi picking this back up. safari landed at-rule() in webkit on aug 3 and firefox has it assigned, but neither has shipped, so the fallback problem above is still live for every non chrome user. ok to close this and reopen when one of them ships?

Deferring to @LeaVerou as the CSS content ATL

LeaVerou:
Sure, let's do that. 

FWIW I do think there's value in a guide that includes the various ways to detect existing at-rules and conditionally apply declarations and/or rules when a certain atrule is supported, even before `at-rule()` ships. Then we could just update it when `at-rule()` ships, though there is the irony of being unable to detect support for `at-rule()` easily in CSS. 😅

E.g. I recently recommended a workaround to @jamesnw in another PR about how to conditionally apply a declaration only if `@property` is supported. And there are things like `@container style(--a: 1) or (not (--a: 1)) { ... }` to conditionally apply rules only if style queries are supported. It could be useful to centralize these. 

But that's probably a separate guide (and one I should probably write). I'll try to open a PR for it (edit: done! #1411)

PR #759, opened 2026-05-10
add meta name text-scale guidance

### LINKED ISSUE #691: meta name="text-scale"
### web-feature-id

meta-text-scale

### Feature description

The `<meta name="text-scale" content="scale" />` HTML element allows the browser's initial font size to be affected by the operating system text scale settings. The `<meta name="text-scale" content="legacy" />` element is the default behavior that respects only browser font-size settings.

### Expected release date

Chrome 146, mentioned at I/O

### Coding agent assessment

_No response_

### Reference sources

- https://github.com/w3c/csswg-drafts/blob/main/css-env-1/explainers/meta-text-scale.md#scale-option

**Use cases:**
- [ ] #1207

patrickkettner:
possibly fix #691

I was only able to get the unguided runs to pass 3/5 times, and that was with a decent amount of horseplay with the tasks to actively make them worse. By default it was passing 5/5 unguided.
Similar to #755, I was not sure if that meant we should reconsider the value of the topic for a guide.


rviscomi:

guides/user-experience/respect-os-text-scale/guide.md
```diff
@@ -0,0 +1,11 @@
+---
+name: respect-os-text-scale
+description: Respect the operating system's text scale preference for a page so font sizes, layout, and rem-based measurements follow the user's accessibility settings.
+web-feature-ids:
+  - meta-text-scale
+sources:
+  - https://drafts.csswg.org/css-fonts-5/#text-scale-meta
+  - https://github.com/w3c/csswg-drafts/blob/main/css-env-1/explainers/meta-text-scale.md
+  - https://drafts.csswg.org/css-env-1/#preferred-text-scale
+  - https://chromestatus.com/feature/5112244702674944
```
jamesnw:
```suggestion
```

jamesnw (approved):
This use case looks good. I have questions about how to integrate this with fluid typography, how to make sure that the user can still scale type using CTRL +/- for example, and making sure it doesn't negatively impact users without support, but those can all be worked out as the guide is written.

patrickkettner:
the only other thing I can think of is scaling non-text bits like icons or spacing with the env variable, but that feels like a section of this guide rather than its own use case. I pushed an update to put it in the new directory, for the moment.

guides/accessibility/respect-os-text-scale/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: respect-os-text-scale
+description: Respect the operating system's text scale preference for a page so font sizes, layout, and rem-based measurements follow the user's accessibility settings.
+web-feature-ids:
+  - meta-text-scale
+---
```
LeaVerou, line 6:
```suggestion
---

<!-- 
TODO: how to integrate this with fluid typography, how to make sure that the user can still scale type using CTRL +/- for example, and making sure it doesn't negatively impact users without support
-->
```

Added @jamesnw's comments as an HTML comment so we can make sure we remember to cover these topics when writing the guide.

(recent change: a guide is still considered a stub if its entire contents are an HTML comment)

LeaVerou (approved):
LGTM. I left a suggestion adding a comment with @jamesnw's points plus an [_extremely_ relevant recent CSSDay 2026 talk](https://www.joshtumath.uk/posts/2026-07-02-lets-fix-the-webs-text-size/) by @joshtumath that was exactly about this topic for whomever fleshes out this guide.

I think this can be a great guide!

guides/user-experience/respect-os-text-scale/guide.md
patrickkettner: good catch, dropped. sources is gone and the file moved to guides/accessibility/, which is why this thread is stuck on the old path.

PR #765, opened 2026-05-11
Add stack-drill-down use case guide and demo

philipwalton:
This PR creates and implements a new `stack-drill-down` use case for the scroll snap feature (#719).

### DIFF
--- a/guides/user-experience/stack-drill-down/guide.md
+++ b/guides/user-experience/stack-drill-down/guide.md
@@ -1,0 +1,698 @@
+---
+name: stack-drill-down
+description: Build full-screen hierarchical navigation that lets users drill down into nested views and swipe or navigate back to return, with browser history kept in sync.
+web-feature-ids:
+  - inert
+  - intersection-observer
+  - scroll-driven-animations
+  - scroll-snap
+  - scroll-snap-events
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert
+  - https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Scroll-driven_animations
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Scroll_snap
+  - https://drafts.csswg.org/css-scroll-snap-2/#scrollsnapchange-and-scrollsnapchanging
+  - https://developer.mozilla.org/en-US/docs/Web/API/History_API
+---
+
+# Stack Drill Down
+
+## Overview
+
+A stack drill-down is a hierarchical navigation pattern, common in mobile apps, where activating a link pushes a new full-screen view on top of the previous one. The view's content is application-defined — a settings sub-page, a thread inside a feed, a folder inside a file browser, a detail page inside a gallery, etc. The user returns by swiping the current view off-screen to the right or by tapping a back button. Browser history stays in sync so the OS-level Back gesture, deep links, and forward/back navigation all work coherently.
+
+This guide implements the stack as:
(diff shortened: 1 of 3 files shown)


guides/user-experience/stack-drill-down/guide.md
```diff
@@ -0,0 +1,698 @@
...
+- **DO** include an explicit back button in every drill-down view. The swipe gesture only works on touch — keyboard, pointer, and desktop users need a visible affordance.
+- **DO NOT** call `history.pushState` from the `popstate` handler — that pushes *new* entries while the user is trying to go back and breaks the browser back button.
+- **DO NOT** drive the parallax with a `scroll` event listener when scroll-driven animations are available. The CSS path runs on the compositor; a JS scroll listener runs on the main thread and will visibly drop frames during the gesture.
+- **DO NOT** mutate views you removed from the DOM after a swipe-back. Treat `entriesByDepth` as the canonical record: a pruned entry has `view: null` and is rebuilt on demand in `updateFromHistoryState`.
+
+### Fallback strategies
+
+Most of the features used in this guide are Baseline Widely available, and do not require any fallback. The only features not widely available that may require fallbacks are scroll-snap-events and scroll-driven-animations, both of which have robust fallback or progressive enhancement stories, and are safe to use for this use case:
```
rviscomi, line 628: To avoid having to update the guidance as features mature, we shouldn't hard-code feature status info. I'd suggest removing this and adding the reassurance about robust fallback or PE stories to the feature-specific sections below.

guides/user-experience/stack-drill-down/guide.md
```diff
@@ -0,0 +1,698 @@
...
+## Implementation
+
+### 1. Markup
+
+The static HTML is just an empty stack container; views are built in JavaScript and appended as the user navigates.
+
+```html
+<div class="Stack">
+  <!-- Intentionally empty. The initial view is appended by JavaScript
+       at init time (step 8). -->
```
rviscomi, line 43: A bunch of code comments throughout this file have unnecessary line breaks. Nothing major, but it'd be nice to fix up.

guides/user-experience/stack-drill-down/guide.md
rviscomi, line 1: This guide is _extremely_ long at about 700 lines. Was that intentional, ie because you've seen coding agents struggle with this? I'm mostly worried about consuming more tokens than necessary, but it can also make it harder for agents to grok.

philipwalton: I agree it's long. It grew over time, as I kept discovering more and more issues and edge cases (most around URL history syncing with the visible view).

> Was that intentional, ie because you've seen coding agents struggle with this?

Yeah, I started with a working demo and then I asked the agent to generate a guide. The first draft was ~300 lines, but when I would test it the agent would miss things, so I'd add that content, and then it grew and grew.

guides/user-experience/stack-drill-down/guide.md
philipwalton: Aren't we implicitly hard-coding feature status info by including fallback sections for non-widely available features?

Also, I think it's good for guides to offer a general recommendation as to whether a guide's fallbacks are robust and reliable or whether they should only be used in certain circumstances. I'm not sure if you can convey that with just individual feature commentary. WDTY?

guides/user-experience/stack-drill-down/guide.md
philipwalton: I ask my agent to wrap lines at ~80 chars because that's what I configure my editor to be. I think without that it  does ~100. I can fix, but this is in most of my other guides as well...

rviscomi (approved):
LGTM. Agreed to revisit my comments after launch.

PR #784, opened 2026-05-12
modern-web-guidance: baseline awareness and configuration

paulirish:
I know it doesn't look like much, but this was a lot. :)


guides/modern-web-guidance/SKILL.md
```diff
@@ -57,7 +57,36 @@ npx -y modern-web-guidance@latest retrieve "<id>"
...
 -   Do not hallucinate guides or ignore them; they represent the preferred local standard for the user's project.
+
+
 -   Note: if the `npx -y modern-web-guidance…` command hangs, try running again in offline mode: `npx --offline …`
+### Browser Compatibility & Baseline Fallbacks
+
+#### Baseline Status
+- **Limited Availability**: Lacks cross-browser interoperability.
+- **Newly Available**: The feature has emerged in all major browsers within the last 30 months.
+- **Widely Available**: The feature has been supported in all major browsers for ≥ 30 months.
```
LeaVerou: Pretty sure agents know this already? At least Claude and Codex refer to baseline status on the regular (though often the status they think a feature has is outdated — which is expected).

Instead of describing what these are, it would be useful to map them to expected handling. We do some of that below, but it's binary: above the floor vs below the floor. For a Baseline WA floor, is there no difference between how an agent should handle a Newly Available feature vs a Limited Availability one?

guides/modern-web-guidance/SKILL.md
```diff
@@ -57,7 +57,36 @@ npx -y modern-web-guidance@latest retrieve "<id>"
...
+
+#### Support Floor Implementation
+Assume a default guaranteed audience floor of **[Baseline Widely Available](https://web-platform-dx.github.io/baseline/)**. Enforce a support floor, not a ceiling:
+- **Supporting Browsers (Guaranteed Audience):** Features at or above the floor require **zero fallbacks**. Clients below the floor risk broken experiences.
+- **Supporting Features (Capability Integration):** Features below the floor **MUST** implement progressive enhancement and feature-detected fallbacks to protect the guaranteed audience.
+- **Implicit Context:** Read `AGENTS.md` constraints implicitly within the context window. Do **NOT** execute upfront status lookups or file-verification checks before engaging CLI tools.
```
LeaVerou: Note that Claude does not read `AGENTS.md`: https://github.com/anthropics/claude-code/issues/6235

Perhaps the build process for Claude would replace this with `CLAUDE.md`?

paulirish: yeah. wild that it doesnt but.. true.

guides/modern-web-guidance/SKILL.md
paulirish: aye. i've revised this pretty heavily and now have just dropped mention of these 3 labels because yeah.. not needed.

paulirish:
I know i've definitely made things weird already.. :)   _(And I appreciate your patience with me.)_

but.. I would feel EXTRA weird merging this without an LGTM. So.. will awaiting one from either @rviscomi or @philipwalton

guides/modern-web-guidance/SKILL.md
```diff
@@ -80,6 +80,21 @@ npx -y modern-web-guidance@latest retrieve "<id>"
...
 
--   Always search **first** to find the most specific design/performance patterns.
+-   Always search **first** to find the most relevant guides.
 -   These guides are usually framework-agnostic; adapt them correctly to your setup.
 -   Do not hallucinate guides or ignore them; they represent the preferred local standard for the user's project.
+
+
+## Interpreting Browser Support & Fallbacks
+
+* **Default Behavior**: All guides assume **Baseline Widely available** features are safe to use without fallbacks. For features that are not Baseline widely available, you **MUST** follow the fallback recommendations in the guide.
```
philipwalton: The last sentence in this bullet kinda contradicts the next bullet. Should we append "...unless the user has specified a custom browser support policy"?

guides/modern-web-guidance/SKILL.md
```diff
@@ -80,6 +80,21 @@ npx -y modern-web-guidance@latest retrieve "<id>"
...
+* **Custom Policies**: If the user has already defined explicit browser support requirements, use the browser compatibility data in the guide to determine if a fallback can be safely ignored. Examples:
```
philipwalton: All these examples are fairly complex. I assume that's on purpose, but do you think it's worth adding a simple example as well, e.g. "Chrome 144+"

philipwalton: It may also be worth adding an example like "Do not implement modern web guidance feature fallbacks" (for people who are using the tool to play around with new features).

guides/modern-web-guidance/SKILL.md
```diff
@@ -80,6 +80,21 @@ npx -y modern-web-guidance@latest retrieve "<id>"
...
+  - _"Never recommend or implement polyfills; if a Baseline Newly Available feature is required for core functionality, provide a lightweight custom fallback or redesign the approach."_
+  - _"Assume a modern execution environment where Baseline Newly Available features can be used natively, provided they are strictly feature-detected and degrade gracefully."_
```
rviscomi:
```suggestion
  - _"Assume a modern execution environment where Baseline Newly Available features can be used natively, provided they are strictly feature-detected and degrade gracefully."_
* For Baseline YYYY support targets, a feature satisfies this target if its "Baseline since" date is <= YYYY.
```

rviscomi (approved):

guides/modern-web-guidance/SKILL.md
paulirish: - _"Do not implement feature fallbacks."_ (for exploratory prototypes of the cutting-edge web)
    - _"Safari 17.4+"_ (for internal tools targeting macOS or Tauri-based desktop apps)

guides/modern-web-guidance/SKILL.md
paulirish: true. fixed.

paulirish: good catch thx

guides/modern-web-guidance/SKILL.md
```diff
@@ -90,6 +90,25 @@ npx -y modern-web-guidance@latest retrieve "<id>"
...
+    - _"Never recommend or implement polyfills; if a Baseline Newly Available feature is required for core functionality, provide a lightweight custom fallback or redesign the approach."_ (to minimize bundle size and avoid technical debt)
+    - _"Assume a modern execution environment where Baseline Newly Available features can be used natively, provided they are strictly feature-detected and degrade gracefully."_ (for progressive enhancement strategies)
+* **Reactive Policy Discovery**: Watch for environmental cues to suggest documenting a policy in CLAUDE.md or AGENTS.md. Suggest this if the developer:
+  - Mentions building for a restricted runtime (e.g., Electron or Tauri).
+  - Explicitly excludes specific targets (e.g., "we don't support Desktop Chrome").
+  - Expresses hesitation about polyfill complexity, bundle size, or performance cost.
+  - Questions if a feature is safe to use without fallbacks.
+
+  No defined policy format. This is an example: `**Browser Support:** Allow Newly Available features, but only adopt custom fallback code that adds <= 20 lines and does not require external dependencies.`
```
LeaVerou, line 114: What are agents supposed to do in this case (if a fallback adds > 20 lines or requires an external dep)? Not use the feature at all or use it without the polyfill?

paulirish: Unspecified. :)   Phil and I kinda hatched this scenario. My interpretation of it was..  if fallback code (polyfill or ****fill) is >20 lines, then don't use the feature at all.    

Can you think of a case where this doesn't really work? eg where the fallback code is bigger but.. it's an enhancement/additive so.. it'd be fine?   

regardless... it's an important point and i'll revise the policy based on the resolution here.

LeaVerou: I think it's a good heuristic, as long as that's clarified what it means. And either interpretation is valid — depending on the feature and use case. "don't use the feature at all then" makes sense if the feature would fail poorly, or the functionality is critical. "use the feature but without any fallback" would also be reasonable, if the feature is PE.

Another factor is that as you move towards less browser support, it starts mattering more not just what weight the fallback/*fill adds but also its **fidelity**.  Meaning, for features supported almost everywhere except a small % of users, it's fine if the fallback is just "the site is usable". The higher the % of users without the feature, the higher fidelity the fallback must be. E.g. for the listbox versions of customizable-select, I don't think the built-in listbox is a good enough fallback, especially when you're doing things with it that go way beyond just a styled listbox. Yes, it's perfectly usable, but it's a poor experience for a third of visitors. But once customizable-select is supported in e.g. 90% of visitors, that fallback is now fine — the site is usable after all.

Related to the topic of fallbacks, I just filed https://github.com/GoogleChrome/modern-web-guidance-src/issues/849 which is somewhat relevant

OphyBoamah: @paulirish @LeaVerou, coming at this from #849, but this exchange is actually closer to what I have data on.
I tested the Baseline 2023 target (via Antigravity CLI + MWG) on two UI components, 3 runs each, prompt held identical across runs. Paul, to your question about a case where "don't use the feature at all" doesn't hold, I think I saw that split happen in practice:

- **Settings tooltip, 3/3 runs**: strict interpretation, no Popover, no Anchor Positioning, no polyfills, plain hover/focus-within CSS.

- **Dropdown menu, 2/3 runs**: the other direction, shipped the compliant fallback and a fully polyfilled Popover + Anchor Positioning version side by side, labeled "progressive enhancement," with nothing signaling to the developer that a choice was even made.

Lea, your fidelity point above tracks with this pretty well, the dropdown does feel closer to an Enhancement-type case where the floor interpretation is defensible, while the tooltip reads closer to Critical UI where a silent floor call is riskier. But same nominal target, two different outcomes, no visible signal either way, which seems like the actual crux here.

Happy to share full run logs and configs if useful.

PR #786, opened 2026-05-12
capture stdout from install command

connorjclark:


paulirish:
pretty sure this breaks our nested install.

repro:  `node serving/bin/modern-web.ts install`

PR #787, opened 2026-05-12
add self-update workflow, document how to update

connorjclark:
Adds a self-updating procedure for the modern-web-guidance skill. Roughly:

- The SKILL.md we publish is instructed to call `npx modern-web-guidance` with a version identifier: `--skill-version 2026_05_13-bc852b5c`
- That version identifier only updates if the SKILL.md we publish updates (not if we just publish new guides to NPM)
- The modern-web-guidance CLI compares the agent caller's version with the latest version. If it differs, it logs a warning to stderr with
  instructions on how to update.

Also:

* add `npx modern-web-guidance@latest update` command, which defers to `npx skills update` but for just our skills
* ~~documented how users can update our skills, including how to automate it with Claude Code (learned this here https://joost.blog/self-updating-agent-skills/)~~
* add `@latest` to documented commands - without this, if a user happens to be in a project repo that has our npm package installed by any dependency, it would result in using whatever version that happened to be pinned at (stale)
* commented out skip logic from recent build changes - it was preventing me from iterating on modern-web.mjs

PR #791, opened 2026-05-12
Revert "capture stdout from install command"

paulirish:
Reverts GoogleChrome/guidance#786

it broke the interactivity of `node serving/bin/modern-web.ts install`

### DIFF
--- a/serving/bin/modern-web.ts
+++ b/serving/bin/modern-web.ts
@@ -1,7 +1,7 @@
 #!/usr/bin/env node --experimental-strip-types
 
 import { parseArgs } from "node:util";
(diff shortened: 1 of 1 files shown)


connorjclark:
it worked for me...

i tested with zsh and bash in VS Code's integrated terminal

connorjclark:
nevermind. i tested in dist/ and a bug was causing the build artifact modern-web.mjs to not update with my latest changes (https://github.com/GoogleChrome/guidance/pull/787/changes#diff-65cff8ab6f54b4627af1a8f0670a32fd69c9ccc6f2c86930f959ccc07655e38a)

PR #793, opened 2026-05-13
Use new file structure for high-level guides

rviscomi:
Moves the Security guide out of `skills-drafts/security/SKILL.md` and into `guides/security/security/guide.md`

Moves the Accessibility guide out of `guides/accessibility/SKILL.md` and into `guides/accessibility/accessibility/guide.md`


rviscomi:
Checks pass. I think this is a relatively safe change so I'll go ahead and merge it through to clean up the MWG repo.

PR #795, opened 2026-05-13
Declarative button actions edits

### LINKED ISSUE #331: Create guide and evals for the declarative-button-actions use case
Declaratively connect a button to any element to trigger custom, application-specific actions using declarative button commands, invoker commands, button commands, custom commands, or declarative toggle actions.

Affected web-feature IDs: [invoker-commands](https://webstatus.dev/features/invoker-commands)

Use case subdir: [guides/user-experience/declarative-button-actions](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/user-experience/declarative-button-actions)

Related features: #61

**Required files:**
- [x] demo.html
- [x] expectations.md
- [x] tasks/task.md
- [x] negative-demo.html
- [x] grader.ts

dvdherron:
Closes #331 

- Cleans up shape of inline polyfill
- removes timing benefit for custom commands
- consistent event delegation
- make certain instructions more actionable (“It must have a unique id.” → “If it doesn’t have a unique id , add one”)
- Consistent command registry approach
- **Rename to make it clear that this is about custom commands only** ~~(This was updated in a separate PR but I think it's along the same lines? Open to simplifying).~~


dvdherron:
@LeaVerou I know there's mention of splitting up/merging this guide. Wanted to do some cleanup beforehand in case we decide to keep this use case for now.

guides/user-experience/custom-button-actions/guide.md
```diff
@@ -76,37 +75,62 @@ This polyfill fully supports custom actions (starting with `--`) and dispatches
...
+      await import('https://esm.run/invokers-polyfill');
+    } catch (err) {
+      console.error('Error loading fallback:', err);
+    }
+  })();
 }
+
+// 2. Manually manage ARIA states in your listener
+document.getElementById('action-target').addEventListener('command', (event) => {
+  const command = event.command || event.detail?.command;
```
LeaVerou: Why are we forking here? The event object we receive should have the same shape as the native one. I [checked invokers-polyfill](https://github.com/keithamus/invokers-polyfill/blob/main/invoker.js) and they also don't use `detail`.

```suggestion
  const command = event.command;
```

guides/user-experience/custom-button-actions/guide.md
```diff
@@ -1,20 +1,18 @@
...
 web-feature-ids:
   - invoker-commands
 ---
 
-# Declarative Button Actions
-The Invoker Commands API allows buttons to trigger actions on target elements declaratively using HTML attributes. This approach reduces the need for manual event listeners and ensures interactivity as soon as the HTML is parsed.
-
-For custom, application-specific actions, you can define your own command names. Custom commands must be prefixed with a double dash (`--`) to avoid collisions with future built-in browser commands.
+# Custom Button Actions
+The Invoker Commands API allows buttons to trigger actions on target elements declaratively using HTML attributes. For custom, application-specific actions, you can define your own command names. Custom commands must be prefixed with a double dash (`--`) to avoid collisions with future built-in browser commands.
```
LeaVerou: While the "ensures interactivity as soon as the HTML is parsed" benefit was hard to realize in practice, having a short sentence on the _why_ we'd use this is a good thing.
To take a stab at it, what about:

```suggestion
The Invoker Commands API allows buttons to trigger actions on target elements declaratively using HTML attributes. 
This approach reduces the need for manual event listeners and decouples the UI from implementation details.
For custom, application-specific actions, you can define your own command names. Custom commands must be prefixed with a double dash (`--`) to avoid collisions with future built-in browser commands.
```

(Also side note, take it or leave it: line breaks in Markdown files are not rendered as actual line breaks and are good practice to insert between sentences or logical units of meaning, to help keep diffs small)

guides/user-experience/custom-button-actions/guide.md
```diff
@@ -1,20 +1,18 @@
...
 
 ## Implementation steps
 
-1.  **Define the target element**: Identify the element that will respond to the action. It must have a unique `id`.
+1.  **Define the target element**: Identify the element that will respond to the action. If it doesn’t have a unique `id`, add one.
 2.  **Configure the invoker button**: Use the `commandfor` attribute to point to the target's `id`, and the `command` attribute to specify the custom command name (prefixed with `--`).
-3.  **Handle the command event**: Attach a `command` event listener to the `document` (or a common parent). This ensures that the event is captured even if it is dispatched by a polyfill or from a child element. The event object contains a `command` property and a `target` property (referring to the element identified by `commandfor`).
+3.  **Handle the command event**: Attach a `command` event listener directly on the target element. The event object contains a `command` property and a `target` property (referring to the element identified by `commandfor`).
```
LeaVerou, line 17: See below. The whole benefit of capturing is that it lets us use event delegation for non-bubbling events, so we *don't* need to change this.

guides/user-experience/custom-button-actions/guide.md
```diff
@@ -39,23 +37,24 @@ For custom, application-specific actions, you can define your own command names.
...
+  // 1. **Optional:** Define a registry of requested actions for cleaner logic
+const commandRegistry = {
+  '--spin': (target) => target.classList.toggle('is-spun'),
+  '--grow': (target) => target.classList.toggle('is-grown'),
+  '--reset': (target) => target.classList.remove('is-spun', 'is-grown'),
+};
+
+  // 2. **Mandatory:** Listen for the 'command' event directly on the target element
   // (This is necessary because the native 'command' event does not bubble)
   document.getElementById('action-target').addEventListener('command', (event) => {
```
LeaVerou, line 65: We don't need to listen directly on the target element — that's _why_ we use `capture: true`, so we can still listen to the `command` event on the root, even though it does not bubble.

In this particular case we're only dispatching `command` events on `action-target`, but the code is written to be nicely generic, so all we need is to listen higher up:


```suggestion
  document.body.addEventListener('command', (event) => {
```

Alternatively, remove the `capture: true` (if we're not using event delegation, we don't need it) and add a note after. I'll add a suggestion below.

guides/user-experience/custom-button-actions/guide.md
```diff
@@ -39,23 +37,24 @@ For custom, application-specific actions, you can define your own command names.
...
-    // Robustly handle both native API and manual/polyfill fallbacks
-    const command = event.command || event.detail?.command;
+    const command = event.command;
     const target = event.currentTarget;
```
LeaVerou: Why `currentTarget` instead of `target`? Especially if delegating, `currentTarget` would be the delegation root. Here we're not really delegating, so it makes no difference, but if you adopt the suggestion above it will break things.

dvdherron:
@LeaVerou 👋🏽 Just checking in on the status of this one. I think I've covered all of the points from your earlier review. I think the remaining question is do we use this as is or should this be re-worked for other use cases or as part of a larger skill. Thanks!

guides/user-experience/custom-button-actions/guide.md
```diff
@@ -39,23 +37,24 @@ For custom, application-specific actions, you can define your own command names.
...
-    } else if (command === '--reset') {
-      // Clear all custom classes to return to initial state
-      target.classList.remove('is-spun', 'is-grown');
+    if (action) {
+      action(target);
     }
-  });
+  }, {capture: true});
 </script>
 ```
```
LeaVerou, line 77: See above. If you prefer, since we don't really _need_ event delegation here you can keep listening on action-target, remove `capture: true` (if we're not using event delegation, we don't need it) and add a note about it instead:


```suggestion
- If we have multiple possible targets, we can use `capture: true` (since the `command` event does not bubble) and listen on an ancestor.
- If the target may be in a shadow root, use `event.composedPath()[0]` instead of `event.target`
```

guides/user-experience/custom-button-actions/guide.md
```diff
@@ -76,37 +75,62 @@ This polyfill fully supports custom actions (starting with `--`) and dispatches
...
+  const target = event.currentTarget;
+  const source = event.source; // The button that triggered the command
+
+  if (command === '--spin') {
+    const isSpun = target.classList.toggle('is-spun');
+    
+    // Polyfill tip: Manually update ARIA to match the new state
+    source?.setAttribute('aria-expanded', isSpun);
+  }
+}, { capture: true });
```
LeaVerou: See above, either remove `capture: true` or listen on an ancestor.

guides/user-experience/custom-button-actions/guide.md
```diff
@@ -132,5 +155,5 @@ document.getElementById('action-target').addEventListener('command', (event) =>
   if (action) {
     action(target);
   }
-});
+ }, {capture: true});
```
LeaVerou: See above

guides/user-experience/custom-button-actions/guide.md
```diff
@@ -76,37 +75,62 @@ This polyfill fully supports custom actions (starting with `--`) and dispatches
...
+      this.command = command;
+    }
+  }
+}
 
-// 2. The fallback: Dispatch events manually if native support is missing
-if (!supportsInvokers) {
+// 3. The fallback: Dispatch events manually if native support is missing
   document.addEventListener('click', (event) => {
     const button = event.target.closest('button[commandfor]');
```
LeaVerou: We may want to use `composedPath()` so that it supports Shadow DOM too. See https://github.com/keithamus/invokers-polyfill/blob/main/invoker.js#L227

LeaVerou: We may want to use `composedPath()` so that it supports Shadow DOM too. See https://github.com/keithamus/invokers-polyfill/blob/main/invoker.js#L227

LeaVerou (changes requested):
Much better!

I left a couple suggestions, but the main recurring comment is that my comment about capturing was misunderstood, so now we're using `capture: true` AND attaching listeners directly on elements, which is superfluous.

The reason to use `capture: true` is so we can still use event delegation on events that don't bubble. Event delegation is when we listen on an ancestor and use `event.target` to disambiguate. I shared some MDN links on the review issue which may be helpful about the general technique and how capturing fits in to it.

Also, `event.currentTarget` will break once we _are_ actually using event delegation, much better to just use `event.target`. Here is a codepen showing both this and event delegation with capturing: https://codepen.io/leaverou/pen/XJppEeZ?editors=1011

I also realized another issue (and edited my review to include it): `event.target`/`event.currentTarget` will not work properly if the target is in shadow DOM. This doesn't matter for the example because we know what we have, but it may be worth using `composedPath()` for the manual polyfill.

**Note:** I only reviewed `guide.md`, once that has settled I'll look at the demo & expectations.

jamesnw:
@LeaVerou I think I addressed all the review. I removed the `capture: true` throughout. 

I also realized that custom commands don't handle aria states- it's not isolated to the polyfill. I updated the guide to reflect that as well.

I suspect this breaks the grader significantly.

LeaVerou (approved):
This looks great! I love the "Key constraints" section, all signal, no noise. Fantastic work!

This is for the guide, as you point out the expectations etc will need to be reworked (ideally in a way that allows capturing but doesn't require it).

PR #797, opened 2026-05-13
Auditing guides for adherence to accessibility best practices

rviscomi:
I had an agent take the recently merged [Accessibility guide](https://github.com/GoogleChrome/guidance/blob/main/guides/accessibility/accessibility/guide.md) and audit the existing use case-level guides for adherence to the best practices.

Then I had the agent generate fixes to the guides, demo, expectations, and grader files as needed.

>[!NOTE]
>Demo files themselves don't need to adhere to all a11y best practices. It's fine for them to take shortcuts since they're only used for grader calibration. However they must still show correct adherence to all relevant a11y best practices prescribed by the guide.

<details>
<summary>Accessibility audit findings (long)</summary>

# Accessibility Audit of `guides/**/guide.md`

**Reference:** `guides/accessibility/accessibility/guide.md` (the canonical a11y guide).
**Scope:** All 132 other `guide.md` files under `guides/`.
**Method:** Pattern grep across all guides for known violation signatures, then full-file reads of every flagged guide and a sample of the rest. This report merges findings from two independent passes; each finding cites `file:line` so it can be verified.

The canonical guide is structured into 12 sections; findings cross-reference the section(s) they violate.

---

## 1. Summary of issues by severity

| Severity | Count | Theme |
|---|---|---|
| High | 8 | Code samples teach a pattern that the canonical guide explicitly forbids, or omit a state-sync that leaves AT users silent (focus removal without replacement, implicit label association, name/description packed into label, missing live-region semantics, inline SVG without a11y treatment, hint placement, `aria-invalid` not synced, tabs without tab semantics) |
| Medium | 16 | Code samples omit a11y treatment that the canonical guide requires (region landmark dilution, missing tooltip programmatic association, missing pause/keyboard guidance for compound widgets, decorative graphics not hidden, color-only state, focus abandonment after view transitions, collapsed elements still in AT tree, pointer-only interactions, etc.) |
| Low | 6 | Minor or stylistic deviations (heading-hierarchy entry points in isolated examples, sub-canonical visually-hidden utility, stylistic alt text guidance) |
| Informational | 4 | Stub guides with no body content — neither violations nor compliant examples |

---

### H1. `guides/forms/autofill-highlight-inputs/guide.md:28` — `outline: none` with no replacement focus indicator
**Code:**
```css
input:autofill { …; border-color: green; outline: none; }
```
**What's wrong:** This rule applies during `:autofill` *and* during focus on the autofilled field — it removes the focus ring without providing a replacement. Compounding the issue, the "autofilled" state itself is communicated by `border-color: green` alone (color-only state).
**Canonical rule (§5):** "Don't disable outlines without replacements: Avoid `outline: none` without styling alternatives." Also §5 DO: "Visible Focus Indicators: Always style `:focus-visible` states explicitly." Plus §9 DO: "Use multiple state indicators."
**Expected:** Either avoid disabling the outline on the autofill state, or pair the `outline: none` with an explicit `:focus-visible` rule that sets a high-contrast outline on autofilled inputs. Add iconography or a descriptive badge alongside the green border to satisfy the multiple-indicators rule. The CSS guide's own §4 (`guides/css/css/guide.md:215-218`) states the focus rule explicitly.

---

### H2. `guides/forms/brand-consistent-forms/guide.md:48-62, 95-98` — `<label>` wraps `<input>` without explicit `for`/`id`
**Code:**
```html
<label>
  <input type="checkbox" checked>
  Subscribe to newsletter
</label>
```
The same implicit-only association is repeated for radio buttons (lines 55-62) and the fallback example at lines 95-98.
**Canonical rule (§3 DO):** "Explicitly associate `<label>` with its control via `for`/`id`, **even when nesting the input inside the label** — explicit association improves assistive-tech support."
**Expected:** Every `<input>` should have an `id`, and the wrapping `<label>` should carry the matching `for`. The `forms/forms/guide.md` "Forms" guide enforces this rule (line 51) — the brand-consistent-forms code sample contradicts it.

---

### H3. `guides/forms/forms/guide.md:273` — `aria-label` overrides visible text and packs description into the accessible name
**Code:**
```html
<button id="toggle-password" type="button"
        aria-label="Show password as plain text. Warning: this will display your password on the screen.">
  Show password
</button>
```
**What's wrong:** Two violations at once.
1. The `aria-label` overrides the visible button text "Show password" — meaning sighted users see one thing and screen-reader users hear another. Canonical §3 prefers `aria-labelledby` over `aria-label` "when a visible label exists."
2. The accessible name packs an instruction and a warning into the label itself. Canonical §3 DON'T: "Don't pack descriptions, error messages, or instructions into the label."
**Expected:** Drop the `aria-label`; the visible "Show password" already provides the name. Move the warning to a description: a sibling element referenced via `aria-describedby`. Also use `aria-pressed` to expose the toggle state (canonical §5 DO).

---

### H4. `guides/user-experience/persistent-toast-notifications/guide.md` (whole file) — Toasts taught with no live-region semantics or pause-on-hover/focus
**What's wrong:** The guide describes building stackable toast notifications via `popover="manual"`, but the entire "Implementation Guidelines" section never mentions `aria-live`, `role="status"`, or `role="alert"`. A toast that appears in the top layer without a live-region role is silent to assistive technology. The auto-dismiss timer also has no pause/extend mechanism, violating WCAG 2.2.1 (Timing Adjustable) — if focus or pointer enters the toast (e.g. to read longer content or click the Close button), the timer must pause.
**Canonical rule (§8):** The Live Regions section is built around exactly this scenario — the "Standard / Toast / Banner" row of the urgency table maps toasts to `aria-live="polite"`, with critical interruptions using `assertive`. §8 DO: "Centralize live regions for non-visible announcements." Also §10 DO: "Provide Pause mechanism."
**Expected:** Mark the container `<div role="status" aria-live="polite" popover="manual">` (or `role="alert"` for critical), or emit toast text into a centralized polite live region. Pause auto-dismissal on `pointerenter`/`focusin`, resume on `pointerleave`/`focusout`. Note the §8 "Don't add live-region updates to inert DOM" caveat since `popover="manual"` content can be inert when nested in dialogs.

---

### H5. `guides/forms/animated-select-picker/guide.md:41-58`, `guides/forms/rich-media-picker/guide.md:33-44` — Inline SVGs in `<option>` lack a11y treatment
**Code (animated-select-picker):**
```html
<option value="system">
  <svg viewBox="0 0 24 24" …>…</svg>
  System Default
</option>
```
**What's wrong:** Each SVG has no `aria-hidden`, no `role="img"`, and no nested `<title>`. Per canonical §6 they fall into one of two contracts: decorative (then `aria-hidden="true"`) or informative (then `role="img"` + `<title>`). The guides treat them as if SVG is invisible to AT, which it is not.

`rich-media-picker` adds two further problems at line 96-98: it removes the native checked indicator (`option::before { display: none; }`) without providing a visual replacement, leaving the checked state invisible apart from custom hover/active colors (color-only state, §9 DON'T). It also mirrors complex nested DOM via `<selectedcontent>` without simplifying — risking a screen reader announcing concatenated, repetitive strings as the button's accessible name.
**Canonical rule (§6):** DO "Decorative SVGs removal: Apply `aria-hidden='true'` to remove decorative SVGs from reading flows." DO "Informative View Descriptions for inline SVGs: Apply `role='img'` and a nested `<title>` tag for informative visuals." Plus §9 DO: "Use multiple state indicators."
**Expected:** Each option's text already names it ("System Default", "Light UI", "Frontend Developer"…), so add `aria-hidden="true"` to the SVGs. Keep a visible checkmark (or shape indicator) alongside the custom checked color. Define concise `aria-label` values on options whose mirrored content would otherwise read awkwardly.

---

### H6. `guides/forms/validate-input-after-interaction/guide.md:107-113` — Password rules placed *below* the input
**Code:** The `<ul id="password-rules">` is rendered after the `<input>` in DOM order.
**What's wrong:** Canonical §7 mandates placing the hint above the input, *specifically because* mobile keyboards, autocomplete menus, and password manager popovers cover anything below the field while the user is typing — exactly when the password rules are needed.
**Canonical rule (§7 DO):** "Link hints to inputs via `aria-describedby`: Associate help text with inputs, **and place the hint above the input so autocomplete popovers don't cover it during editing**."
**Expected:** Move the rules `<ul>` above the `<input>` in the markup. The `aria-describedby` already links them programmatically, so DOM order is purely about visual reachability during typing — and the canonical rule is unambiguous about which side of the input to use.

---

### H7. `guides/forms/required-field-feedback/guide.md:27-71` — Visual `:user-invalid` with no `aria-invalid` synchronization
**What's wrong:** The guide leverages `:user-invalid` to paint a red border but provides no JavaScript to set `aria-invalid="true"` on the input when it enters that state. Screen-reader users tabbing through the form receive no signal that a required field has been left empty in an invalid state. The error state is also signaled only by border color (no icon, no text indicator on the field itself).
**Canonical rule (§3, §7):** "Don't repeat state already exposed via ARIA…" — implies state should *be* exposed via ARIA in the first place. §9 DO: "Use multiple state indicators: Do not denote success/errors ONLY with color."
**Expected:** Provide the JS bridge from the `accessible-error-announcement/guide.md` pattern inline (or at minimum link to it as MANDATORY) so readers don't ship a visual-only error state. Add an icon to the helper text container. The companion guide `accessible-error-announcement/guide.md` does this correctly — `required-field-feedback` should reference that as part of its core implementation, not an afterthought.

---

### H8. `guides/user-experience/search-hidden-content/guide.md:49-66` — Tabs example uses `<div class="tab">` with no tab ARIA
**Code:**
```html
<div class="tabs">
  <details class="tab" name="my-tabs">
    <summary>Tab 1 (closed)</summary>
    …
  </details>
  …
</div>
```
**What's wrong:** This is presented as a tab pattern but has no `role="tablist"`, no `role="tab"` on the triggers, no `role="tabpanel"` on the content, no `aria-selected`, no controlling `<button>` linked via `aria-controls`. Users navigating with a screen reader's tab-navigation shortcut hear nothing about tabs, and arrow-key navigation between tabs isn't established. Stacking `<details>` with a shared `name` produces the visual mutex but not the tab semantics the section claims.
**Canonical rule (§2 DO):** "Match ARIA implementations to actual behavior: If you set `role='tab'`, the element must behave like a tab — including keyboard interactions." The inverse: if you ship something visually presented as tabs, you owe the ARIA model.
**Expected:** Either rename the section to "mutually exclusive disclosures" (and stop calling them tabs), or provide a real tab pattern: `<button role="tab" aria-controls="panel-1" aria-selected="true">…` with `<div role="tabpanel" id="panel-1">…`, plus arrow-key navigation and roving tabindex.

---

### M1. `guides/forms/forms/guide.md:261-276` — `<section>` used as input grouper, creating `region` landmark dilution
**Code:**
```html
<section>
  <label for="name">Full name</label>
  <input id="name" …>
</section>
<section>
  <label for="email">Email</label>
  …
</section>
```
**Canonical rule (§1 DON'T):** "Don't overuse landmarks. In particular, avoid labeling a `<section>` (which turns it into a `region` landmark) — `region` should be a last resort when no other landmark fits."
**Expected:** Replace `<section>` with `<div>`, or use `<fieldset>` + `<legend>` for actual semantic groups (which the same guide recommends in §1 DO at line 15).

### M2. `guides/user-experience/position-aware-tooltips/guide.md:31-32` and `guides/user-experience/interest-triggered-tooltips/guide.md:30-37` — Tooltip not programmatically associated with its trigger
**Code:**
```html
<button popovertarget="tooltip" id="anchor">anchor</button>
<div id="tooltip" popover role="tooltip"></div>
```
**What's wrong:** The trigger has no `aria-describedby="tooltip"`. `role="tooltip"` alone does not bind the description to the control. The `interest-triggered-tooltips` guide adds three more gaps: (a) it omits `role="tooltip"` on the popover container itself, (b) it does not document the WCAG 1.4.13 (Content on Hover or Focus) constraints — dismissible via `Esc` without moving focus, hoverable, persistent, and (c) it relies on the emerging `interestfor` attribute without showing the `aria-describedby` fallback for AT that hasn't shipped support.
**Canonical rule (§3, §7 DO):** Programmatic association of names and descriptions; `aria-describedby` for help text.
**Expected:** Add `role="tooltip"` to the popover and `aria-describedby="tooltip"` to the trigger. Document the dismissible/hoverable/persistent contract. Note §3 DON'T: "Don't include interactive elements in an `aria-describedby` target" — the tooltip body should be plain text only.

### M3. `guides/user-experience/resilient-context-menus-and-nested-dropdowns/guide.md:22-30` — Custom menu without keyboard / ARIA menu semantics
**Code:**
```html
<button popovertarget="dropdown-menu">Open</button>
<div id="dropdown-menu" popover="auto">
  <button class="menu-item" type="button">Edit</button>
</div>
```
**What's wrong:** The trigger does not advertise the menu (no `aria-expanded`, no `aria-haspopup="menu"`). Menu items aren't given `role="menuitem"`, and the guide does not address arrow-key navigation between items. Popover light-dismiss handles outside clicks but not the keyboard model expected of a menu.
**Canonical rule (§2 DO):** "Match ARIA implementations to actual behavior" — if you visually present a menu, you owe its keyboard contract.
**Expected:** Either explicitly state in the guide that this is a "button group revealed in a popover" and not a menu; or add `aria-haspopup="menu"`, `aria-expanded`, `role="menu"`/`role="menuitem"`, and arrow-key navigation.

### M4. `guides/user-experience/persistent-app-tours/guide.md:22-26` — Tour step lacks dialog semantics, accessible name, and focus management
**Code:**
```html
<div id="tour-step" popover="manual">
  <h3>Step 1</h3>
  <p>Learn how to use this feature.</p>
  <button popovertarget="tour-step" popovertargetaction="hide">Got it</button>
</div>
```
**What's wrong:** Three issues. (a) `<h3>` without context skips heading hierarchy; canonical §1 requires sequential heading order. (b) The popover has no accessible name (no `aria-labelledby`). (c) `showPopover()` does not move focus into the popover, so users — especially keyboard and AT users — won't notice the tour step opened elsewhere on the screen. `popover="manual"` does not trap focus or shift focus the way `<dialog>.showModal()` does.
**Canonical rule (§1, §3, §11):** Heading hierarchy; accessible name on overlays; "Use the Native `<dialog>` Element" for things that demand attention.
**Expected:** Use `<h2>` (or whatever level fits), add `aria-labelledby="tour-step-title"`, add `role="dialog"`, and call `tourStep.querySelector('button').focus()` (or `.focus()` on the container) immediately after `showPopover()`. The pattern in `light-dismiss-a-dialog/guide.md:44-46` already gets the labeling right.

### M5. `guides/user-experience/scroll-progress-indicator/guide.md:42-50` — Decorative progress bar not hidden from AT
**Code:** `<div id="progress"></div>` styled into a scaling bar.
**What's wrong:** The scroll-progress bar is purely decorative (it duplicates the scrollbar that AT users reach via their own affordances) but isn't `aria-hidden="true"`. AT will encounter an empty unnamed `<div>`. The guide also doesn't tell developers which choice to make.
**Canonical rule (§6 DO):** "Empty Alt properties for decorative visuals" (same intent: `aria-hidden="true"` for non-`img` decorative graphics).
**Expected:** Add `aria-hidden="true"` for the decorative case, or — if the team genuinely wants programmatic progress — use `<progress>` (or `role="progressbar"` with `aria-valuenow`/`aria-valuemax`) and update its value from JS.

### M6. `guides/user-experience/carousel-slide-effects/guide.md` and `carousel-snap-highlights/guide.md` — Carousel guides omit pause guidance and ARIA carousel semantics
**What's wrong:** The example markup uses plain `<ul>`/`<li>` and never shows the `role="region"` / `aria-roledescription="carousel"` envelope, accessible names, or `aria-roledescription="slide"` on items. The primary code blocks defining continuous scale animations omit `prefers-reduced-motion` wrappers (the prose mentions it, but copy-pasting the code skips it). And the *generic* pattern these guides teach is repurposed for autoplay carousels, which then require pause controls.
**Canonical rule (§1, §10):** "Provide Pause mechanism: Allow users to stop auto-running carousels banners." DO "Default to static views."
**Expected:** Show the ARIA carousel envelope in the snippet; wrap continuous animations in `@media (prefers-reduced-motion: reduce)` blocks within the code itself; add an "Autoplay" subsection requiring a visible pause control defaulted-paused under reduced motion.

### M7. `guides/forms/validate-input-after-interaction/guide.md:44`, `guides/forms/autofill-payment-form/guide.md:102` — `placeholder` carrying meaning
**Code:**
```html
<input … placeholder="you@example.com" …>
<input … cc-exp … placeholder="MM/YY" …>
```
**What's wrong:** The placeholders carry information (format hint, valid example) that disappears the moment the user types — the failure mode §7 warns about.
**Canonical rule (§7 DON'T):** "Don't use placeholders as labels."
**Expected:** Move the format hints to a sibling element above the input, referenced via `aria-describedby`. Drop the `placeholder` attribute.

### M8. `guides/user-experience/navigation-drawer/guide.md:62-68` — Trigger button with placeholder SVG, and a static `aria-label`
**Code:**
```html
<button id="drawer-open" aria-label="Open menu" aria-expanded="false" aria-controls="drawer">
  <!-- menu icon SVG -->
</button>
```
**What's wrong:** The placeholder comment doesn't show `aria-hidden="true"` on the SVG — implementers will copy a broken pattern (canonical §6). Also, the JS at lines 248-263 toggles `aria-expanded` but never updates the `aria-label` from "Open menu" to "Close menu" as the drawer state flips.
**Expected:** Update the placeholder to `<svg aria-hidden="true" …>`. Either swap the `aria-label` on state change or use two icons swapped via state.

### M9. `guides/html/html/guide.md:357` — Background-video guidance permits `role="none"` on `<video>` without explicitly forbidding focus
**What's wrong:** The text covers canonical §6's heuristic only by parenthetically forbidding `controls`. A reader may copy the `aria-hidden="true"` pattern onto an `<iframe>` (which is focusable) and break a11y.
**Expected:** Add a "DO NOT" bullet that explicitly references the "do not hide focusable elements" rule from canonical §5, and clarify that the rule for `<video>` only holds because `controls` is omitted.

### M10. `guides/user-experience/animate-to-intrinsic-sizes/guide.md` — Collapsed elements remain in the AT tree; `aria-expanded` not synced
**What's wrong:** The collapsible alert pattern dismisses the component by setting `block-size: 0; opacity: 0; pointer-events: none;` — but the node remains in the DOM with no `aria-hidden="true"` or `hidden` attribute. Screen readers will still traverse and announce the "dismissed" content. The accordion expansion pattern likewise omits syncing `aria-expanded="true"` on the controlling trigger when the panel opens. CSS size-transition blocks omit reduced-motion checks.
**Canonical rule (§6 visibility matrix; §5 DO toggle states; §10 reduced motion):** "Hidden for all" requires `hidden` attribute or `display: none`. Toggle state must be exposed via `aria-expanded`.
**Expected:** Apply `aria-hidden="true"` (or `hidden`) when the block collapses to 0px. For accordions, sync `aria-expanded` on the trigger. Wrap the size transitions in a `prefers-reduced-motion` media query within the code itself.

### M11. `guides/performance/defer-rendering-heavy-content/guide.md` — `content-visibility: auto` AT-tree concerns; missing `aria-expanded` sync in `beforematch`
**What's wrong:** The guide advocates `content-visibility: auto` on off-screen content blocks. In some AT configurations, off-screen nodes are excluded from the accessibility tree or cannot be traversed sequentially unless focus is forcibly moved into them — the guide does not warn developers to verify keyboard reachability. For `hidden="until-found"` disclosures, the prose mentions that state synchronization should occur in a `beforematch` event handler but provides no code example showing `aria-expanded="true"` toggling on the controlling button.
**Canonical rule (§5 DO; §6 visibility matrix):** Logical tab order; explicit visibility-vs-reachability decisions.
**Expected:** Add a warning that `content-visibility: auto` blocks must be verified for keyboard reachability. Provide a `beforematch` code sample that updates `aria-expanded` (and any `aria-controls` linkage) on the disclosure trigger.

### M12. `guides/user-experience/same-document-transitions/guide.md` — Focus abandonment after view transition
**What's wrong:** When a user clicks a thumbnail and the layout morphs to a detail view, programmatic focus stays on the now-hidden thumbnail. If the list view is removed from the DOM, focus is abandoned in an invisible subtree — keyboard users lose their place, and AT users get no announcement of the new context.
**Canonical rule (§4, §5):** "Update document title on Page Transitions in SPAs: Shift focus to updated titles." "Logical Tab Order."
**Expected:** Pair the view-transition with an explicit `mainHeading.focus()` (or `mainContainer.focus()` with `tabindex="-1"`) on the newly revealed view, ideally after the transition's `finished` promise resolves.

### M13. `guides/user-experience/consistent-cross-document-transitions/guide.md` — `blocking="render"` AT-timing impact; reduced motion not in code
**What's wrong:** `blocking="render"` delays visual updates and initial paint, which can cause latency in AT processing — the guide doesn't surface this tradeoff. Cross-document morphing CSS doesn't include reduced-motion overrides in the example blocks (the prose may, the snippets do not).
**Canonical rule (§10):** Reduced motion support, code-level.
**Expected:** Document the AT-timing impact of render-blocking. Wrap the `@view-transition` rule in `@media (prefers-reduced-motion: reduce)` to disable cross-document morphing for affected users.

### M14. `guides/user-experience/interest-triggered-action-previews/guide.md` — Preview state purely visual, no AT announcement
**What's wrong:** Pointer events update `dataset.preview` to alter styles via CSS. AT receives zero programmatic announcement that an action preview is active, or what it would do.
**Canonical rule (§8):** Live regions for state changes invisible to AT.
**Expected:** Coordinate dynamic previews with a centralized polite live region, or expose the preview state via `aria-describedby`/`aria-label` updates on the focused control.

### M15. `guides/forms/custom-select-picker-layouts/guide.md:74-77` — Color-only `:checked` indication
**Code:**
```css
.grid-picker option:checked {
  border-color: #007bff;
  background-color: #f0f7ff;
}
```
**What's wrong:** Selected state is communicated only by border + background color. Compounding: the guide self-warns about the 2D-grid keyboard mismatch (line 84) but doesn't show how to handle it.
**Canonical rule (§9 DON'T):** "Don't use color alone to indicate the presence of a user interface component or its state."
**Expected:** Keep a visible checkmark or shape indicator for the checked option alongside the custom colors. Consider whether a true 2D layout demands a custom ARIA listbox (with arrow-key 2D navigation) instead of a `<select>`.

### M16. `guides/user-experience/interactive-content-reveal/guide.md` — Pointer-only spotlight with no keyboard alternative
**What's wrong:** The spotlight reveal relies exclusively on `pointermove`. Keyboard users have no way to move the spotlight or read the underlying content. The guide warns "do not use for essential content" but doesn't ship a keyboard-accessible variant or a fallback that maps the pattern to `<details>`/`<dialog>`.
**Canonical rule (§5):** Logical Tab Order; all interactive content reachable by keyboard.
**Expected:** Either implement keyboard handlers (arrow keys move the spotlight), or document that the underlying content must remain visually persistent and keyboard-reachable independently of the effect, with the spotlight purely as enhancement for pointer users.

---

### L1. `guides/forms/forms/guide.md:189-202` — Visually-hidden utility uses `opacity: 0` instead of the canonical `clip-path` recipe
The forms guide ships a `.visually-hidden` snippet using `position: absolute; opacity: 0; width: 1rem; height: 1rem`. The canonical §3 utility uses `clip-path: inset(50%)` plus 1px size and `:focus-within/:active` exception so focusable hidden content becomes visible. Two utilities with the same name and different behavior is a maintenance trap. `brand-consistent-forms/guide.md:107-112` has the same divergence (zero dimensions can cause AT to drop the control entirely from sequential reading flows; per-platform, it varies).
**Expected:** Use the canonical `.visually-hidden` (`accessibility/accessibility/guide.md:106-121`) verbatim. In the brand-consistent-forms fallback, also add a `:focus-visible` outline on the custom checkbox `span::before` so keyboard users see focus.

### L2. `guides/user-experience/persistent-app-tours/guide.md`, `guides/user-experience/style-parent-with-has/guide.md`, `guides/user-experience/content-based-styling/guide.md` — Examples enter heading hierarchy at `<h2>` or `<h3>` without a comment
The canonical §1 doesn't forbid examples that show a sub-tree, but a casual reader copy-pasting will produce a broken outline. A one-line comment ("assume an `<h1>` precedes this card") would defuse the foot-gun.

### L3. `guides/user-experience/scroll-snap-state-sync/guide.md:39` — Selector `nav a` implies a navigation but the snippet never shows the `<nav>` markup
The example uses `document.querySelectorAll('nav a')` but the markup is not shown. The Accessibility section (lines 53-59) is good and recommends `aria-current="page|location"`, but a developer copying this might wire it to bare `<a>`s. Show `<nav aria-label="Table of contents">` explicitly.

### L4. `guides/css/css/guide.md:435` — Charts via gradients tag-line about "textual fallback" is correct but unenforced in any chart-related guide
The CSS guide has the right principle but no companion guide demonstrates the data-table fallback the canonical §6 DO recommends ("Provide data tables as alternatives"). When a chart guide is added, link this rule.

### L5. `guides/user-experience/anchor-positioning-tab-underline/guide.md:77` — Right-half of advice present, left-half unstated
The guide correctly says "must not be a replacement for setting the appropriate `aria-current='page'` or `aria-selected` aria values," but never demonstrates that markup. Show the `<a aria-current="page">…</a>` markup.

### L6. `guides/user-experience/scrollytelling/guide.md` — Animation-heavy guide whose primary code blocks omit reduced-motion
The guide builds rotating, fading, translating story sections. There is one prose bullet on `prefers-reduced-motion` (line 109) but the example CSS at lines 65-99 omits the media query. The fallback at lines 162-167 also defaults `#animated section` to `opacity: 0; transform: scale(0.5) rotateY(-180deg)` — leaving content visually hidden but accessible to AT, causing desynchronization for screen-reader users who navigate linearly without scrolling.
**Expected:** Wrap the keyframe assignments in a reduced-motion override block within the example code:
```css
@media (prefers-reduced-motion: reduce) {
  #animated section { animation: none !important; opacity: 1; transform: none; }
}
```

---

## 5. Informational: stub guides with no body content

These four files contain only YAML frontmatter — no content, no examples, no a11y guidance to evaluate:

- `guides/user-experience/expose-canvas-content-to-browser-features/guide.md`
- `guides/user-experience/interactive-content-in-3d-scenes/guide.md`
- `guides/user-experience/apply-webgl-shaders/guide.md`
- `guides/user-experience/light-dark-theme-semantic-color-palettes/guide.md`

The first three are particularly load-bearing for a11y: they describe putting interactive HTML inside `<canvas>` and applying WebGL shaders, both of which historically have severe a11y consequences (canvas is opaque to AT; shader-distorted content can defeat zoom/contrast). When fleshed out, these guides MUST cover canonical §6 (alternate text and visibility for canvas content) and §6 + §9 (zoom + reduced-motion + contrast for shaders).

---

## 6. Cross-cutting observations (not single-guide findings)

1. **Copy-paste safety failures.** Many guides advise honoring user preferences (reduced motion, contrast, etc.) in prose but omit those rules from the accompanying code blocks (see L6, M6, M10, M13, H1). Guides should be safe to copy-paste — if the `prefers-reduced-motion` media query is necessary, it should be inside the example, not just mentioned beside it.

2. **Color-only state communication is widespread.** Recurs in H1, H5, H7, M15. Form/validation guides repeatedly use border-color or background-color shifts as the *only* indicator of success/error/checked/autofilled state. Canonical §9 DO requires multiple indicators — this should become a lint-level check on new guides.

3. **Invisible component states.** Several guides hide content visually (`opacity: 0`, `block-size: 0`, `content-visibility: auto`, `display: none`) without synchronizing AT visibility (`aria-hidden`, `hidden`). The canonical §6 visibility decision matrix is the single best reference for this — it should be linked from every guide that toggles visibility.

4. **Focus abandonment after dynamic view changes.** Top-layer overlays (M4 tour), view transitions (M12 same-document, M13 cross-document), and toast notifications (H4) repeatedly omit programmatic focus management. The canonical §4 DO ("shift focus to updated titles") and §5 DO ("Logical Tab Order") together prescribe what to do — but no single guide demonstrates the recipe for SPA view transitions.

5. **The forms collection mostly sets a high bar.** `forms/forms/guide.md`, `forms/required-field-feedback/guide.md` (modulo H7), `forms/select-menu-interaction/guide.md`, `forms/autofill-sign-in-form/guide.md`, and `forms/autofill-sign-up-form/guide.md` consistently teach `<label for>`, `aria-describedby`, `required` (not `aria-required`), and avoidance of `placeholder`-as-label. The brand-consistent and animated/rich-media select guides are the exceptions that lower this bar (H2, H5).

6. **Live regions are under-served at the pattern level.** §8 of the canonical guide has no companion guide demonstrating a centralized announcer abstraction. `forms/forms/guide.md:59` mentions `aria-live` for error announcements but doesn't show one. Toasts (H4), action previews (M14), and auto-validation (`accessible-error-announcement/guide.md`) all touch this — a dedicated "Live Regions" pattern guide would close the gap.

7. **Heading hierarchy in isolated examples.** Many UX guides show a single component (a card, a tour step, a popover heading) and pick a heading level without explaining the outline assumption (L2, M4). A docs-wide convention ("assume each example sits inside `<main>` after an `<h1>`") would help.

8. **No guide offers a positive counter-example for the canonical "skip link" pattern (§1 DO).** If a future guide demonstrates page chrome, it should model `<a href="#content" class="skip-link">…</a>` + `<main id="content" tabindex="-1">`.

---

## 7. Recommended next steps

1. Fix the H1–H8 issues in the listed files. Each is a code change of ≤10 lines, except H4 (Toast guide) which needs a new "Announcements + timer pause" subsection and H8 which needs a real ARIA tab pattern (or a rename of the section).
2. Sweep for the cross-cutting themes (Section 6): copy-paste-safety, color-only state, invisible-AT, focus abandonment. Each is a one-line lint rule away from being preventable.
3. Add a "Region landmarks vs. layout grouping" callout in `forms/forms/guide.md` to retire the `<section>` per-field pattern (M1).
4. Open one shared "Tooltip association" pattern that both `position-aware-tooltips` and `interest-triggered-tooltips` link to, covering `aria-describedby` + WCAG 1.4.13 (M2).
5. Open a "View transition focus management" pattern that both `same-document-transitions` and `consistent-cross-document-transitions` link to (M12, M13).
6. Flesh out the four stub guides (Section 5) with §6-aligned guidance on canvas/shader a11y *before* they're recommended to engineering teams.
7. Extend repository lint to flag, in any new guide PR: `outline: none` without an adjacent `:focus-visible`, continuous `@keyframes` without a `prefers-reduced-motion` block in the same code fence, `<section>` without `aria-label*` in form contexts, and `placeholder=` on inputs that also have a label.

</details>


guides/user-experience/interest-triggered-tooltips/guide.md
knowler, line 1: None of the ARIA related changes here are necessary when using `interestfor`. When the interest target has `popover=hint`, the interest target implicitly gains a minimum role of `tooltip` and the interest source is automatically associated implicitly with `aria-describedby` OR `aria-details` if the target has interactive content (so the target IS allowed to have interactive content).

Perhaps some of these accessibility related characteristics of `interestfor` could be concisely mentioned in this guide?

guides/user-experience/interest-triggered-tooltips/expectations.md
```diff
@@ -10,3 +10,7 @@
 * Polyfills for interestfor, popover, and anchor positioning must be conditionally installed.
 * Do not display tooltip content using `::before` and `::after` pseudo-elements.
 * Do not put tooltip content into attributes on the trigger.
+* The tooltip element MUST define `role="tooltip"`.
+* The trigger element MUST programmatically associate the tooltip by setting `aria-describedby` matching the tooltip's ID.
+* The implementation MUST adhere to WCAG 1.4.13 guidelines, ensuring the tooltip is dismissible (e.g. via `Esc`), hoverable, and persistent.
+* The tooltip container MUST contain plaintext content only, avoiding interactive descendant elements.
```
knowler:
```suggestion
* The implementation MUST adhere to WCAG 1.4.13 guidelines, ensuring the tooltip is dismissible (e.g. via `Esc`), hoverable, and persistent.
```

Even this remaining expectation is largely covered by `interestfor`, though I guess it wouldn’t be bad to leave as an expectation.

guides/performance/defer-rendering-heavy-content/demo.html
knowler, line 1: Every usage of `aria-expanded` in this demo is incorrect, both existing and new additions. It should be used for controls such as toggle buttons that control the expanded/collapsed state of something else (e.g. the cards or the elements with `hidden=until-found` set on them).

The expectation to update `aria-expanded` on `beforematch` is good, but it’s being applied to the wrong things.

guides/user-experience/carousel-slide-effects/guide.md
knowler, line 1: I don’t have any immediately actionable feedback here, but I want to flag this for more extensive review since it’s doing some fairly complex stuff with ARIA (i.e. overriding or creating role descriptions).

guides/user-experience/navigation-drawer/expectations.md
```diff
@@ -17,3 +17,5 @@
 - The drawer's horizontal scroll container uses `scroll-behavior: smooth` only when the user has not requested reduced motion (e.g., via `@media (prefers-reduced-motion: no-preference)`).
 - Horizontally scrolling the drawer's scroll container all the way to its end position closes the drawer.
 - The drawer's backdrop opacity decreases as the drawer's scroll container is scrolled toward the closed position.
+- The menu trigger button MUST swap its `aria-label` text dynamically to match the open/closed state of the drawer.
```
knowler:
```suggestion
```

If the button is using `aria-expanded` it should not update the label dynamically with the state as that state is already announced and that’s ambiguous (e.g. “Open menu, toggle button, expanded”; “Close menu, toggle button, collapsed”). A better `aria-label` for the element would simply be “menu” (which could be announced like “Menu, toggle button, expanded”).

guides/forms/forms/expectations.md
```diff
@@ -51,6 +51,8 @@
 - The implementation MUST use `autocomplete="new-password"` for sign-up and `autocomplete="current-password"` for sign-in.
 - The implementation MUST allow pasting into password fields.
 - The implementation MUST provide a toggle capability allowing users to unmask password input.
+- The password toggle button MUST NOT use `aria-label` overriding its visible text, MUST expose its toggle state via `aria-pressed`, and MUST link warning descriptions via `aria-describedby`.
+- The implementation MUST NOT use `<section>` elements as layout containers for individual form fields to prevent region landmark dilution.
```
knowler: This is too strong. It’s named `<section>` element that are the problem. In any case, I don’t think this is a relevant expectation.
```suggestion
```

guides/user-experience/interest-triggered-action-previews/guide.md
```diff
@@ -27,19 +27,31 @@ An interest relationship is created by setting the `interestfor` attribute on a
...
+    previewAnnouncer.textContent = `Previewing effect: ${event.source.dataset.effect}`;
+  }
 });
 
 interestingElement.addEventListener("loseinterest", event => {
   // Unapply the preview
   delete event.target.dataset.preview;
+  if (previewAnnouncer) {
+    previewAnnouncer.textContent = '';
+  }
```
knowler: Unapplying the preview should probably also have a meaningful announcement.

guides/user-experience/interest-triggered-action-previews/guide.md
```diff
@@ -27,19 +27,31 @@ An interest relationship is created by setting the `interestfor` attribute on a
...
 
 For the sake of this use case, we can leverage the `interest` and `loseinterest` events to preview various effects for an interest target. Both of these events are `InterestEvent`s which has a `source` property which is the source of the interest (i.e. the element with the `interestfor` attribute).
 
 ```javascript
+const previewAnnouncer = document.getElementById('preview-announcer');
+
 interestingElement.addEventListener("interest", event => {
   // Apply the preview based on `event.source`
   event.target.dataset.preview = event.source.dataset.effect;
+  // MANDATORY: Expose preview state programmatically to assistive technologies
```
knowler: Since interest is passively activated (i.e. you could trigger interest for multiple things, just from moving your cursor across the screen), it might be worth debouncing the messages. For the sake of this guide, it might be good to elide the implementation of the debouncing though (so maybe just mention it).

guides/user-experience/interest-triggered-action-previews/guide.md
knowler, line 1: I don’t think there’s universal benefit for including announcements for interest based previews. Yes, not all screen reader users are completely vision impaired, so there could be some relevance, but in many situations it could just be considered noise (i.e. imagine tabbing through some buttons with interest effects and you hear the announcements of them applying/unapplying but you don’t benefit from their effect, since it goes away if you move your focus to the target — the announcements might obscure the fact that the buttons also apply the effect in a more suitable way).

guides/user-experience/search-hidden-content/guide.md
knowler, line 1: This isn’t a request to change anything, just a comment on what it did: this is very good. `<details>` should never be used for a tabs pattern (as well as many other patterns).

guides/user-experience/style-parent-with-has/expectations.md
```diff
@@ -4,3 +4,4 @@
 - "Force Fallback Mode" must replicate this parent-styling behavior using JS class toggling.
 - DO NOT assert on exact RGB color codes in the grader. Assume the agent will pick an appropriate error color (e.g., red hues). Use a helper function like `isRed()` instead of strict string equality.
 - DO NOT assume specific class names like `.card` or require a `.status-icon` unless explicitly requested. Use relational locators (e.g., finding the nearest parent or ancestor container of the input).
+- Standalone code examples MUST document or assume a complete, accessible document heading hierarchy outline.
```
knowler: Not relevant for what this guide does. Probably a better expectation for guides/demos in general.

```suggestion
```

guides/user-experience/search-hidden-content/demo.html
knowler, line 1: If the button isn’t a toggle button (i.e. pressing it reverses the effect), there’s no need to add `aria-expanded` (or really `aria-controls` for that matter).

guides/user-experience/resilient-context-menus-and-nested-dropdowns/guide.md
```diff
@@ -11,34 +11,36 @@ sources:
...
 
-The first step is to create a button that opens a dropdown menu, using the Popover API.
+The first step is to create a trigger button that opens the overlay container using the Popover API.
+
+**MANDATORY Accessibility Distinction:** This pattern explicitly models a **button group revealed inside a popover** rather than a true ARIA menu. Do not apply `role="menu"` or `role="menuitem"` unless you fully implement the corresponding keyboard navigation contract (such as handling spatial arrow-key navigation between items).
 
 ```html
-<button popovertarget="dropdown-menu">
-  Open
+<button popovertarget="action-panel" aria-haspopup="true">
```
knowler: `aria-haspopup=true` is a backwards compat equivalent of `aria-haspopup=menu` which requires the associated element to have a matching role (e.g. `role=menu`) that is not being implemented in this scenario.


```suggestion
<button popovertarget="action-panel">
```

guides/user-experience/resilient-context-menus-and-nested-dropdowns/guide.md
knowler, line 1: There’s a lot of different application this guide could have, one of which being true ARIA menu patterns, I wonder if it would be better to caveat that the guide is not demonstrating any particular accessible UI pattern?

knowler (commented):
I would remove any accessibility advice with regards to carousels that’s been added by this PR for now. Perhaps, just add a note that the accessibility related requirements have been omitted. From what I can tell, the agent has gotten its implementation details from APG — which should not be treated as the canonically authoritative resource for ARIA patterns — but also might not be implementing it all correctly. Since carousels are such a contentious topic, I think it would be best to avoid addressing the accessibility of them for now. Some research and maybe even user testing is necessary.

guides/performance/defer-rendering-heavy-content/demo.html
knowler: Flagging this for now (I’ll return to it in the morning), but we’ll need to fix up what it’s done to try to correct this because this is wrong in almost more ways. I also realize that I haven’t properly understood what the demo was doing but my feedback on the original was still valid.

rviscomi: Yeah I caught that too, sorry for the thrash. Working on a fix.

rviscomi: Ok so the focus of this use case was initially supposed to be about the performance benefits of deferring heavy content. Disclosures containing hidden content are good candidates for deferring, but having that in scope that led us down the path of making sure the disclosures fully exemplified a11y best practices. Plus, we already have a search-hidden-content use case anyway! So I've refactored the disclosure bit out of this use case and used `GUIDE_REF` to link to the other one. That should take care of the bad aria.

knowler: That works!

guides/forms/forms/demo.html
knowler, line 1: There are still some issues here.

- If the button is stateful (i.e. it has `aria-pressed`), then the accessible name of the button should not change. For buttons, the name is taken from the contents by default or using `aria-label`/`aria-labelledby`. This means the text content shouldn’t change and any name provided with ARIA shouldn’t change.
- If the button is not stateful (i.e. it doesn’t use `aria-pressed` or any other ARIA state), then you can change the accessible name. This is often used for play/pause buttons.

In this case, button consistently named “Show Password” (i.e. the text content or the `aria-label` do not change on state changes) that uses `aria-pressed` to indicate the state change accessibly is likely best. A visual cue for the state change would also be necessary; it’s common to use icons for indicating the state change.

Alternatively, a checkbox would work as well and could be simpler. The label for that should also remain consistent.

guides/user-experience/persistent-toast-notifications/guide.md
knowler, line 1: I would revert the accessibility guidance here as there’s complexity for doing this correctly that we won’t be able to get to.

guides/user-experience/interest-triggered-action-previews/guide.md
```diff
@@ -43,6 +43,9 @@ interestingElement.addEventListener("loseinterest", event => {
 });
 ```
 
+> [!NOTE]
+> **Don't announce interest-driven previews via a live region.** Interest can be triggered just by moving the pointer across the page or tabbing through nearby buttons, so apply/unapply announcements quickly become noise that can drown out content the user actually cares about. The preview itself is the affordance; users who can perceive it benefit directly, and those who can't will not benefit from a verbal echo of "Previewing effect: A". If you decide you genuinely need an announcement for a specific case, debounce it heavily and announce the *committed* effect (on `click`), not every passing hover.
```
knowler:
```suggestion
> **Don't announce interest-driven previews via a live region.** Interest can be triggered just by moving the pointer across the page or tabbing through nearby buttons, so apply/unapply announcements quickly become noise that can drown out content the user actually cares about. The preview itself is the affordance; users who can perceive it benefit directly, and those who can't will not benefit from a verbal echo of "Previewing effect: A". If you decide you genuinely need an announcement for a specific case, make sure to test it carefully with users.
```

guides/forms/forms/demo.html
rviscomi: Also updated the corresponding example code in guide.md

knowler (approved):
I think what we have now is fine.

PR #799, opened 2026-05-13
Add guides for HTML-in-canvas

ewewraw:
Progress on #659 #660 #661 #662


guides/user-experience/apply-webgl-shaders/expectations.md
```diff
@@ -0,0 +1,7 @@
+- Feature detection for HTML-in-Canvas MUST be conducted before using the HTML-in-Canvas API.
+- The `<canvas>` element MUST include the `layoutsubtree` attribute to allow descendant HTML elements to be exposed to browser features.
+- Canvas rendering MUST be executed inside an `onpaint` event handler attached to the canvas element to trigger re-rendering when descendant HTML elements change.
+- The rendering logic MUST use `texElementImage2D` to draw HTML elements onto the canvas.
+- The CSS `transform` property of the descendant HTML element MUST be updated based on the transform matrix calculated during rendering.
+- A `ResizeObserver` MUST be used to observe the canvas size with `device-pixel-content-box` and update its dimensions to prevent blurriness.
```
progers: Unfortunately, Safari has not shipped `device-pixel-content-box`, and it is a little tricky to feature-detect this particular feature. I've been recommending the following really gross snippet:
```
const observer = new ResizeObserver(([entry]) => {
  const dpc = entry.devicePixelContentBoxSize;
  canvas.width = dpc ? dpc[0].inlineSize : Math.round(entry.contentRect.width * window.devicePixelRatio);
  canvas.height = dpc ? dpc[0].blockSize : Math.round(entry.contentRect.height * window.devicePixelRatio);
});

const supportsDevicePixelContentBox = 
  typeof ResizeObserverEntry !== 'undefined' && 
  'devicePixelContentBoxSize' in ResizeObserverEntry.prototype;
const options = supportsDevicePixelContentBox ? { box: 'device-pixel-content-box' } : {};
observer.observe(canvas, options);
```

What's the best way to handle this? Should we provide a snippet like this, or high-level instructions, or something else?

guides/user-experience/apply-webgl-shaders/expectations.md
```diff
@@ -0,0 +1,7 @@
...
+- A fallback UI strategy MUST be implemented for browsers that do not support HTML-in-Canvas.
```
progers, line 7: This line could lead to extremely tortured attempts at polyfilling html-in-canvas which may not be appropriate. For example, for an interactive form under canvas, if html-in-canvas is not supported, it may be best to put a form on top of the canvas, or it might be best to just say that html-in-canvas is required.

I am not sure what the best approach would be, so this is really more of a question.

guides/user-experience/apply-webgl-shaders/guide.md
```diff
@@ -6,3 +6,239 @@ web-feature-ids:
...
+      targetHTMLElement,
+      screenSpaceTransform,
+    );
+    if (computedTransform) {
+      targetHTMLElement.style.transform = computedTransform.toString();
+    }
+  }
+  ```
+
+7. [Optional] If the developer is experiencing a mismatch in the DOM logical layout in 3D even after applying the CSS transform from step 5, check if `transform.is2D` is correctly set to false for a 3D DOMMatrix. If not, re-initialize the DOMMatrix which corrects `is2D` to be false before applying the transform to the target HTML element:
```
progers: Chrome 149 will have the fix for this bug, and goes to stable in about 2 weeks. Would it be reasonable to just leave this out for now, and recommend developers use beta/dev/canary?

guides/user-experience/apply-webgl-shaders/guide.md
```diff
@@ -6,3 +6,239 @@ web-feature-ids:
...
+    const screenSpaceTransform = clipToCanvasViewport
+      .multiply(mvpDOM)
+      .multiply(toGLModel);
+
+    // 5. Apply to the transform
+    const computedTransform = canvas.getElementTransform(
+      targetHTMLElement,
+      screenSpaceTransform,
+    );
+    if (computedTransform) {
```
progers: nit: you can remove the `if (computedTransform) {` condition, as `getElementTransform` will always return a DOMMatrix (or throw an exception).

guides/user-experience/apply-webgl-shaders/guide.md
```diff
@@ -6,3 +6,239 @@ web-feature-ids:
...
+
+### Fallback strategies
+
+{{ BASELINE_STATUS("canvas-html") }}
+
+The HTML-in-Canvas API is not currently supported in all modern browsers, thus a fallback strategy is typically required.
+
+However, given the improved performance benefits of this API, HTML-in-Canvas should be used if the browser supports it.
+
+### HTML-in-Canvas polyfill
```
progers, line 236: There are a variety of polyfill approaches, also including https://github.com/niklasvh/html2canvas and https://github.com/tsayen/dom-to-image. Should we include these too?

guides/user-experience/apply-webgl-shaders/expectations.md
ewewraw: Thanks for catching this! I've updated expectations by removing the `with device-pixel-content-box` bit, updated the code samples and the guides. I think because this is outside of the scope of the HTML-in-Canvas feature, we don't necessarily need to go into detail why exactly we aren't simply using `device-pixel-content-box` in this guide

guides/user-experience/apply-webgl-shaders/expectations.md
ewewraw: I think it's reasonable to expect from the implemnetation that there will be A fallback strategy. Then, depending on the use case, the strategy will differ. Displaying a "not supported" message is also a strategy. 
You can check for each use case that I've suggested a different strategy. For example, if the developer needs an interactive content, html2canvas is not the thing to use. The strategies are described in the last section of the guide.md for each use case

guides/user-experience/apply-webgl-shaders/guide.md
ewewraw: I'd keep it for now, since there are a lot of users whom it will take some time to update to the latest version of Chrome/Chromium. But I've expanded the instructions saying that if they're already using Chrome 149+, then this is not the right fix

guides/user-experience/apply-webgl-shaders/guide.md
ewewraw: Fixed, thanks

guides/user-experience/apply-webgl-shaders/guide.md
ewewraw: Expanded the polyfill strategy guidance, added these library recommendations to the export-html-media guide

rviscomi:
I want to make sure we get guidance in for these use cases before the content freeze, so I'm inclined to merge this as-is, and create follow-up PRs to patch any issues that arise.

rviscomi (approved):
LGTM with the caveat that this will require more testing and review in a follow-up PR

progers:
LGTM

PR #803, opened 2026-05-13
add semantic adaptation guide and demo for usage-aware-component-variations

### LINKED ISSUE #169: Create guide and evals for the usage-aware-component-variations use case
Build components that adapt visual logic based on semantic context using CSS container style queries.

Affected web-feature IDs: [container-style-queries](https://webstatus.dev/features/container-style-queries)

Use case subdir: [guides/css/usage-aware-component-variations](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/css/usage-aware-component-variations)

Related features: #48

**Required files:**
- [x] demo.html
- [x] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

castastrophe:
## Summary
- Refactored `guide.md` to focus on **Semantic Component Adaptation**, a more appropriate use case for style queries than size-based layout density.
- Updated `demo.html` to demonstrate a CTA Card adapting its visual logic (button style, badge visibility) based on a inherited `--priority` flag.
- Ensured the implementation uses logical properties (`inset-block-start`, `min-block-size`) and provides robust fallback strategies.
- Verified that this use case is distinct from the density-focused `design-token-reactivity`.

## Test plan
- Manually reviewed the `demo.html` to ensure style queries correctly trigger variations on a Chromium-based browser (Limited Availability).
- Verified that the `guide.md` correctly identifies when to use style queries vs. size queries via a new decision matrix.


guides/user-experience/usage-aware-component-variations/guide.md
```diff
@@ -1,6 +1,74 @@
...
+
+## Fallback Strategies
+
+{{ FEATURE_FALLBACKS("container-style-queries") }}
+
+**MANDATORY**: Provide a safe default experience (progressive enhancement). Since style queries are currently **Limited Availability**, use a manual fallback if the context is critical.
+
+```css
+/* Fallback for browsers without style query support */
+@supports not (container-type: inline-size) {
```
knowler: I wonder if it would be better to layer on style queries as a progressive enhancement in the fallback since `container-type: inline-size` wouldn’t have complete overlap with style queries (i.e. it would go back much farther and would match situations that don’t support style queries). For example, these could be the defaults that the style queries then overrides. Could use `:where(section.promo)` + order of appearance to ensure the style query wins.


```css
/* Default styles */
.button {}

/* Contextual styles (wide support) */
:where(section.promo) .button {
  background: var(--brand-accent);
  color: white;
}

/* Style query aided contextual styles (limited support) */
@container style(--priority: high) {
  .button {
    /* Automatic switch to Filled promotional style */
    background: var(--brand-accent);
    color: white;
    border: none;
  }
}
```

guides/css-layout/usage-aware-component-variations/guide.md
```diff
@@ -1,6 +1,74 @@
...
+  .button {
+    /* Automatic switch to Filled promotional style */
+    background: var(--brand-accent);
+    color: white;
+    border: none;
+  }
+
+  .badge {
+    /* Reveal decorative elements only in high-priority context */
+    display: block;
```
jamesnw, line 70: I'm unsure- do you need to show `.badge{ display: none}` in the guide to show it's being overridden?

guides/user-experience/usage-aware-component-variations/guide.md
```diff
@@ -1,6 +1,74 @@
...
+  .badge {
+    /* Reveal decorative elements only in high-priority context */
+    display: block;
+  }
+}
+```
+
+## Fallback Strategies
+
+{{ FEATURE_FALLBACKS("container-style-queries") }}
```
jamesnw: I don't think this exists? I could be wrong, but my understanding is there would need to be a file in `/features/container-style-queries` with a ## Fallbacks section for this to work.

guides/user-experience/usage-aware-component-variations/guide.md
```diff
@@ -1,6 +1,74 @@
...
+
+**MANDATORY**: Provide a safe default experience (progressive enhancement). Since style queries are currently **Limited Availability**, use a manual fallback if the context is critical.
```
jamesnw: Instead of hard coding support here, I think the agent is supposed to pick up if the user's support needs are not met, and then use this info.
 
```suggestion
**MANDATORY**: Provide a safe default experience (progressive enhancement), and use a manual fallback if the context is critical.
```

guides/user-experience/usage-aware-component-variations/expectations.md
```diff
@@ -0,0 +1,7 @@
+- The parent containers define semantic context flags using CSS custom properties (e.g., `--priority: standard` vs. `promo`).
+- The component uses `@container style()` queries to adapt its internal visual logic (e.g., button variant, badge visibility) based on these inherited properties.
+- When the context is `promo`, the component displays promotional elements (like a badge) and uses a more prominent "filled" button style.
+- When the context is `standard`, the component hides promotional elements and uses a more subtle "outlined" button style.
+- The component does NOT rely on size queries (`min-width` or `max-width`) for this semantic variation.
+- The implementation uses logical properties (e.g., `inline-size`, `inset-block-start`) for consistent layout.
+- A fallback strategy using standard CSS inheritance or `@supports` is provided for browsers that do not support style queries.
```
jamesnw: I think the grader will need to know which option was used in the demo.
```suggestion
- A fallback strategy using `@supports` is provided for browsers that do not support style queries.
```

guides/user-experience/usage-aware-component-variations/guide.md
```diff
@@ -1,6 +1,74 @@
...
+2. **DO NOT** use `container-type` for style queries; they work on any element that inherits the property.
+3. **MANDATORY**: Use `@container style(--property: value)` to apply conditional styles to child elements.
+4. **DO** prefer style queries over descendant selectors to avoid specificity inflation and hardcoded DOM dependencies.
+
+## Implementation Example
+
+```css
+/* 1. Set the context on a layout container */
+.promo {
+  --priority: high; /* Semantic flag for children */
```
jamesnw:
```suggestion
  --priority: promo; /* Semantic flag for children */
```

For consistency with the demo

guides/user-experience/usage-aware-component-variations/guide.md
```diff
@@ -1,6 +1,74 @@
...
+# Semantic Component Adaptation
+
+Use style queries (`@container style()`) to trigger component variations based on inherited semantic context flags rather than viewport or container dimensions. This replaces deep descendant selectors (e.g., `.sidebar .card`) and improves component encapsulation.
+
+In the context of design system components, style queries become invaluable in allowing styles to respond to their containers without the need for style-specific APIs that need to be pushed down to every descendant.
+
+## Mental Model
+
+- **Size Queries:** Use for layout/density changes (width/height).
+- **Name-only Queries** Use for layout context (i.e., breadcrumbs).
```
jamesnw: I'm not sure I understand what's meant by layout context fully- maybe another example?

jamesnw (changes requested):
A few questions and tweaks, but it looks good!

LeaVerou:
Just a quick note that I have seen the review request and plan to take a proper look at this once I'm done with the dark mode rewrite. 
Very excited about this use case!

guides/user-experience/usage-aware-component-variations/guide.md
LeaVerou: If a file doesn't exist, it will just print baseline status (so the next para isn't needed).

That said, we do need to have that content, either there or here! But if the intent is to add this so we can add the fallback later, that's OK, I think? (unsure)

guides/user-experience/usage-aware-component-variations/guide.md
LeaVerou: +1 to what @knowler said. And sadly we don't have a `container-type: style` to query.

And if you want to apply CSS when style CQs are supported without querying any particular property, you can do something like this: https://codepen.io/leaverou/pen/qEqqZQq

guides/user-experience/usage-aware-component-variations/guide.md
LeaVerou: AFAIK name-only CQs have worse browser support than style queries, and no web-feature id so we can do `BASELINE_STATUS`. 

Just filed https://github.com/web-platform-dx/web-features/issues/4039 about that — laziest bug report ever but at least now there's something!

guides/user-experience/usage-aware-component-variations/guide.md
```diff
@@ -1,6 +1,74 @@
...
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@container#container_style_queries
+  - https://developer.chrome.com/docs/css-ui/style-queries
 ---
+
+# Semantic Component Adaptation
+
+Use style queries (`@container style()`) to trigger component variations based on inherited semantic context flags rather than viewport or container dimensions. This replaces deep descendant selectors (e.g., `.sidebar .card`) and improves component encapsulation.
+
+In the context of design system components, style queries become invaluable in allowing styles to respond to their containers without the need for style-specific APIs that need to be pushed down to every descendant.
```
LeaVerou: I understand what you mean, but I think this is hard to understand without a lot of context. Perhaps an example would make it clearer?

LeaVerou (commented):
I left some comments, but on a high level I think it's still unclear (to me at least!) what use cases are part of this and which ones are part of design-token-reactivity. 

E.g. I'd expect that a `--priority` property would be a high-level property that affects multiple design tokens, and that elements it's applied on themselves need to adapt to it themselves (e.g. a button with `--priority` would need a different border, background, text color etc). But also I wouldn't expect `--priority: promo` and I have no idea whether that's higher or lower than `--priority: high`. Looking at the expectations, button variant also seems more design token-y.

The description does a good job at drawing a clear distinction between the two, so I think it just needs a different example that's closer to real-world usage and clearly distinct from design tokens. I can't think of one offhand, but will try to brainstorm a few examples as well.

A more practical caveat I have discovered when using style queries: there is no way to match the actual element the property was specified on, so if you want to apply any styling that shouldn't be inherited (e.g. add a shadow) in a style query, one way to do that without applying it to every element in that subtree is to register the custom property as non-inheritable. Then it will only match direct children and you can use descendant selectors inside it if needed. You can even set a regular property in that rule, so that you have an inherited and a non-inherited version to query. Happy to expand on this if it comes up, just dropping it here so that if you encounter it while working on these you know it has a solution.

guides/user-experience/usage-aware-component-variations/guide.md
castastrophe: Oh I meant to update the demo too - it felt like promo was too "what it is" and not "what the context means".

rviscomi:
@castastrophe is this ready for another look?

castastrophe:
@rviscomi Not ready yet - it needs Lea's updates applied as well.

LeaVerou:
> @rviscomi Not ready yet - it needs Lea's updates applied as well.

You mean updates to things mentioned in my review or updates _by_ me?  👀  Fine either way, just asking to make sure it doesn't fall off my radar if it's the latter!

castastrophe:
> > @rviscomi Not ready yet - it needs Lea's updates applied as well.
> 
> 
> 
> You mean updates to things mentioned in my review or updates _by_ me?  👀  Fine either way, just asking to make sure it doesn't fall off my radar if it's the latter!

Haha updates by me based on your already provided feedback. Cheers!

guides/user-experience/usage-aware-component-variations/guide.md
castastrophe: Yes, +1. I have an update I'll push, would love your thoughts on it.

guides/css-layout/usage-aware-component-variations/guide.md
castastrophe: Yes, good point. I'll push up that update.

guides/user-experience/usage-aware-component-variations/guide.md
castastrophe: Switched this to the BASELINE_STATUS instead and when it exists, maybe there's some good way to come back to these kinds of reference points later and update.

guides/user-experience/usage-aware-component-variations/guide.md
castastrophe: Updating this to "--surface: featured" here and in the demo as well to address Lea's point too about real-world terminology.

guides/user-experience/usage-aware-component-variations/guide.md
castastrophe: I'll remove this for now and we can always come back to it when the time is right.

guides/user-experience/usage-aware-component-variations/guide.md
castastrophe: I'll update with some more concrete examples. Let me know what you think.

castastrophe:
@LeaVerou 
> A more practical caveat I have discovered when using style queries: there is no way to match the actual element the property was specified on, so if you want to apply any styling that shouldn't be inherited (e.g. add a shadow) in a style query, one way to do that without applying it to every element in that subtree is to register the custom property as non-inheritable. Then it will only match direct children and you can use descendant selectors inside it if needed. You can even set a regular property in that rule, so that you have an inherited and a non-inherited version to query. Happy to expand on this if it comes up, just dropping it here so that if you encounter it while working on these you know it has a solution.

I think this is a really great approach to a more advanced concern than the use-case is covering. If I put on my design systems cap, those style queries are equivalent to design language "promo", "priority", etc. and when components are in that context, designers want certain things applied. Spectrum was filled with concepts like this throughout the system and I tried to pull out just one small example for this documentation. I think we could honestly go into much more use-case detail if the use-case expected more of a black-box implementation where the developer doesn't necessarily know where these assets will be used.

tldr; I didn't want to bloat this with too much nuance or detail. I love the idea of digging into it in the future though.

rviscomi:
@castastrophe is this ready for re-review?

castastrophe:
@rviscomi Yes, this is updated and should now address all review feedback.

guides/user-experience/usage-aware-component-variations/guide.md
```diff
@@ -1,6 +1,112 @@
...
+## How is this different from design-token-reactivity?
+
+[`design-token-reactivity`](../design-token-reactivity/guide.md) covers *token-level* changes: density modes, themes, and other higher-order tokens that uniformly shift values (paddings, colors) across many components.
+
+This guide covers *behavioral* changes: a component deciding **what to render, how to arrange itself, or which variant to present** in response to its context. Hiding/showing a badge, switching layout direction, or swapping a button variant all fit here.
+
+## Mental Model
+
+- **Size Queries:** Use for layout/density changes (width/height).
+- **Style Queries:** Use for semantic logic — what the component should *do* in this context.
```
LeaVerou: I'm not sure this section is necessary, and it seems strange to isolate two types of CQs and explain the difference. Unless we have evidence that agents need this, I'd lean towards removing it.

guides/user-experience/usage-aware-component-variations/guide.md
```diff
@@ -1,6 +1,112 @@
...
+Use style queries (`@container style()`) to trigger component variations based on inherited semantic context flags rather than viewport or container dimensions. This replaces deep descendant selectors (e.g., `.sidebar .card`) and improves component encapsulation.
+
+In design systems, this lets a component respond to its surrounding context without that context needing to know the component exists. For example, a card placed inside a section marked `--surface: featured` can swap to a filled button and reveal a promotional badge — the section never names `.button` or `.badge`, and the card never inspects its parent.
+
+## How is this different from design-token-reactivity?
+
+[`design-token-reactivity`](../design-token-reactivity/guide.md) covers *token-level* changes: density modes, themes, and other higher-order tokens that uniformly shift values (paddings, colors) across many components.
+
+This guide covers *behavioral* changes: a component deciding **what to render, how to arrange itself, or which variant to present** in response to its context. Hiding/showing a badge, switching layout direction, or swapping a button variant all fit here.
+
```
LeaVerou: This is useful for reviewers like me :), but I’m not sure what is the actionable takeaway for agents from this section.

How about rewording it as routing: "This guide is about X, if what you're doing is Y, look over at that other guide".

Also use the `GUIDE_REF("guide-id")` macro for linking to guides, which ensures correct paths in the large variety of architectures we deploy to.

guides/user-experience/usage-aware-component-variations/guide.md
```diff
@@ -1,6 +1,112 @@
...
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@container#container_style_queries
+  - https://developer.chrome.com/docs/css-ui/style-queries
 ---
+
+# Semantic Component Adaptation
+
+Use style queries (`@container style()`) to trigger component variations based on inherited semantic context flags rather than viewport or container dimensions. This replaces deep descendant selectors (e.g., `.sidebar .card`) and improves component encapsulation.
+
+In design systems, this lets a component respond to its surrounding context without that context needing to know the component exists. For example, a card placed inside a section marked `--surface: featured` can swap to a filled button and reveal a promotional badge — the section never names `.button` or `.badge`, and the card never inspects its parent.
```
LeaVerou: I think the underlying principle is that they permit **looser coupling** by abstracting away _what_ constitutes a certain context from _which_ elements react to that context and _how_.

Taking a stab to tighten and bring this forwards more, take it or leave it (totally optional, nothing wrong with the current version):

```suggestion
Use style queries (`@container style()`) to trigger contextual variations without hardcoding what produces or reacts to each context, reducing design system coupling and facilitating extensibility and component encapsulation.

For example, we may have a rule like `section.featured { --surface: featured }` but also default to featured on the landing page via `main { --surface: featured }`. Then, different components can opt to react to `--surface: featured`, e.g. a card could reveal a promotional badge, buttons can default to filled etc.

Style queries allow **both parts to be decoupled**: neither the featured sections know what adapts to that state, nor the components know what types of elements create featured sections. Either side can be extended independently (new components, new featured sections) without needing to update the other.
On the contrary, descendant selectors produce tightly coupled architectures, as the selector must encode both sides of the relationship, and be updated when either changes.
```

Heh, I said "tighten" and it ended up longer. I think it's worth it, but YMMV!

guides/user-experience/usage-aware-component-variations/guide.md
```diff
@@ -1,6 +1,112 @@
...
+
+- **Size Queries:** Use for layout/density changes (width/height).
+- **Style Queries:** Use for semantic logic — what the component should *do* in this context.
+
+## Implementation Directives
+
+1. Define a semantic context flag using a CSS custom property on a container.
+2. **DO NOT** set `container-type` for style queries; they query inherited custom properties on the nearest ancestor and don't require an explicit containment context.
+3. **MANDATORY**: Use `@container style(--property: value)` to apply conditional styles to descendant elements. Style queries cannot match the element the property is set on — only its descendants.
+4. **DO** prefer style queries over descendant selectors to avoid specificity inflation and hardcoded DOM dependencies.
```
LeaVerou: I don't think specificity inflation is the reason, we have other tools for that (`:where()`, layers, etc). 
Hardcoded DOM dependencies hits the nail on the head, but I think is a little opaque as written. But also, motivation goes in the intro (I just added a suggestion about expressing this differently).

Wrt implementation directives, as lovely as style queries are, I don't think we want to encourage blanket replacing descendant selectors with style queries whether the use case calls for it or not! 
I think most devs would be kind of annoyed if their agent started doing that indiscriminately for every descendant relationship, and for many it would be superfluous indirection anyway since they are not reusable/extensible and truly only make sense as descendant selectors.

For this section, I'd remove "specificity inflation" and instead describe **when** to use style queries over descendant selectors.

I don't have a rule of thumb ready, but thinking out loud in case it helps: Every abstraction introduces indirection. For good abstractions, that is a small cost compared to the benefit (extensibility, clarity, modularization, etc). So, just like abstractions in other languages, the question is: how likely are we to need this? YAGNI still applies.
Are we likely to later extend the types of elements that trigger the descendant styling? Are we likely to later extend the types of components that react to that context? Does the type of context have independent meaning? etc.

Perhaps the rule of thumb is something around crossing component boundaries? (components in the looser sense) Though that doesn't seem to quite cover all cases either… 🤔 
This reminds me, another point that might be useful to fit in somewhere is that these cross shadow DOM boundaries while descendant selectors do not.

LeaVerou (changes requested):
Much better, see comments!

guides/user-experience/usage-aware-component-variations/guide.md
castastrophe: I think it's worth clarifying so that the tools aren't used in places where they're not the right thing to use. I think of this as a best practices suggestion rather than a literal description of the concept.

guides/user-experience/usage-aware-component-variations/guide.md
castastrophe: Adopted, with light edits to keep the guide's running example terminology. Thanks, this framing is much clearer!

guides/user-experience/usage-aware-component-variations/guide.md
castastrophe: Removed the specificity-inflation claim. Directive #4 now conditions the preference on extensibility / crossing component or shadow DOM boundaries and explicitly says to keep descendant selectors for fixed, one-off relationships. Let me know what you think.

guides/user-experience/usage-aware-component-variations/guide.md
castastrophe: Reworked into a routing section and switched to the GUIDE_REF macro. Good call!

guides/user-experience/usage-aware-component-variations/guide.md
castastrophe: Rather than drop it, I've reframed it as a "Choosing the right tool" decision section so an agent knows when each tool is the right one (rather than a taxonomy of CQ types). That also gave me a natural home for the "when to use style queries vs descendant selectors" point you made, including the shadow-DOM-crossing and YAGNI notes.

castastrophe:
@LeaVerou Updates applied, looking forward to your thoughts!

LeaVerou (approved):
I love what you did with the framing/routing sections! It kept the essence of what I suggested, but a lot tighter! 😍

Btw, custom states can sometimes serve as another way to decouple context. E.g. any component can set `:state(featured)` and other elements can react to it. Of course that's much more limited (it only works when the context is set on a WC, and there is no way to decouple the descendant), but I wonder if it may be worth mentioning in the fallback section. Not sure.

Anyhow, this LGTM! Great guide!

castastrophe:
> Btw, custom states can sometimes serve as another way to decouple context. E.g. any component can set `:state(featured)` and other elements can react to it. Of course that's much more limited (it only works when the context is set on a WC, and there is no way to decouple the descendant), but I wonder if it may be worth mentioning in the fallback section. Not sure.

Great point, custom states are a nice tool for reading context. I'm going to hold off for now though since, as a fallback, it doesn't broaden support (:state() and container style queries reached interop around the same time), and it only decouples the producer side — the reactor still needs a descendant selector (:host(:state(featured)) .badge). In addition, that it requires a web component is a bigger prerequisite than the rest of this guide assumes. I love it as a follow-up for the WC skill though and I know we were having discussions about that spec there too. Thanks for the thorough reviews on this one! 🙏

castastrophe:
@rviscomi @malchata This one looks to be approved and ready for review!

jamesnw (approved):
LGTM!

rviscomi (commented):
Heads up that this guide was moved under css-layout in #909

castastrophe:
@rviscomi I've updated the location and rebased against the latest from main. Should be good to go now 💃

rviscomi:
Looks like there are just a few remaining merge conflicts before we can push this through

castastrophe:
@rviscomi PR is rebased and passing CI now; thanks!

PR #807, opened 2026-05-13
macros: baseline gains browser support detail

paulirish:
eg

WA feature

```md
Baseline status for Grid: Widely available. It's been Baseline since 2017-10-17.
Supported by: Chrome 57 (Mar 2017), Edge 16 (Oct 2017), Firefox 52 (Mar 2017), Safari 10.1 (Mar 2017), and Safari iOS 10.3 (Mar 2017).
```

NA feature
 
```md
Baseline status for Popover: Newly available. It's been Baseline since 2025-01-27.
Supported by: Chrome 116 (Aug 2023), Edge 116 (Aug 2023), Firefox 125 (Apr 2024), Safari 17 (Sep 2023), and Safari iOS 18.3 (Jan 2025)
```

LA feature

```md
popover=\"hint\" has limited availability.
Supported by: Chrome 133 (Feb 2025), Edge 133 (Feb 2025), and Firefox 149 (Mar 2026).\nUnsupported in: Safari.
```

### DIFF
--- a/serving/lib/baseline.ts
+++ b/serving/lib/baseline.ts
@@ -280,9 +280,11 @@ function formatSupportMap(support: Record<string, any> | undefined): string {
   let res = '';
   if (supportedParts.length > 0) {
     res += `\nSupported by: ${listFormatter.format(supportedParts)}.`;
-  }
-  if (unsupportedParts.length > 0) {
-    res += `\nUnsupported in: ${listFormatter.format(unsupportedParts)}.`;
+    if (unsupportedParts.length > 0) {
+      res += `\nUnsupported in: ${listFormatter.format(unsupportedParts)}.`;
+    }
+  } else {
+    res += `\nNot natively supported by any major browser yet.`;
   }
   return res;
 }
--- a/serving/lib/baseline.ts
+++ b/serving/lib/baseline.ts
@@ -1,8 +1,5 @@
-import { createRequire } from 'node:module';
 import { features } from 'web-features';
-
-const require = createRequire(import.meta.url);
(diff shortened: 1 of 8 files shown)


serving/lib/baseline.ts
```diff
@@ -218,17 +220,86 @@ export function validateFeature(id: string): FeatureValidationResult {
...
+      if (key !== 'safari_ios') {
+        unsupportedParts.push(label);
+      }
+    }
+  }
+
+  let res = '';
+  if (supportedParts.length > 0) {
+    res += `\nSupported by: ${listFormatter.format(supportedParts)}.`;
+  }
```
LeaVerou: I think it should print something out when a feature is not supported anywhere yet. It reads quite weird without it. E.g. later down:

> Form-associated WebMCP attributes has limited availability.\nUnsupported in: Chrome, Edge, Firefox, and Safari.

I had to do a double take to figure out that this means no support.

What about something like:

```suggestion
  }
  else {
  	res += `\nNot natively supported by any major browser yet.`;
  }
```

That said, the tension here is that this _will_ likely reduce uplift for these features. IMO in a way that better matches developer expectations, but I can see the argument against it too.

LeaVerou (approved):
This is awesome! 🤩

I left one comment about whether we want to explicitly mention it when no browser supports the feature but in general LGTM. Can't wait to see this merged!

rviscomi (approved):

serving/lib/baseline.ts
paulirish: yup done

PR #808, opened 2026-05-14
Uses the project-discipline-guides SKILL to remove "common" knowledge from discipline guides

micahjo7:
The SKILL was developed in https://github.com/GoogleChrome/guidance/pull/748.

1. First it pulls content generated from Claude Code, Gemini CLI, Codex CLI regarding a "discipline".
2. It combines the common information from each of the three into one `mirror.md` file.
3. This `mirror.md` file content is compared with the discipline guide (e.g. `guides/accessibility/accessibility/guide.md`)
4. The content from the guide that is already in the mirror is removed from the guide.

This way we ensure that the guides do not contain information that (most) coding agents already have know.

This PR shows the results of running this SKILL on all the discipline guides that we have checked in currently.
Note that the `webmcp` guide did not have any changes (no common knowledge found).


guides/accessibility/accessibility/guide.md
rviscomi, line 1: As discussed offline, I'm uncomfortable to wholesale remove content from these guides without more data first

rviscomi (changes requested):

LeaVerou:
I’ll need to preface this with the disclaimer that  I've mainly looked at the css, html, and forms guides but from a quick look this doesn't seem to work very well, at least for those… 😕 

It removed many things that agents _don't_ actually do in practice and merged sentences in ways that completely changed the meaning (e.g. the bit about `:where()` that now became conditional to layer usage).

I hate to be a party pooper, especially having said for months that we need to remove guidance that agents already follow, which is _exactly_ what this does! I love the idea, but I suspect the approach has a fundamental flaw: [If you ask agents about best practices directly](https://github.com/GoogleChrome/modern-web-guidance-src/blob/5a120fff5fa47e4929b7a55a1139aad251cad733/.agents/skills/project-discipline-guides/mirror_prompt.md) they will give you a lot of things they _don't_ actually apply in practice because knowing about a feature and actually using it are different things.

I don't have a concrete counter-proposal but I think to remove guidance from hand-authored guides in an automated way, we'd need to be testing that that guidance being removed is actually _followed_ by agents, not just known. …Which is basically what the current evals system does. Perhaps what we'd need is a very lightweight version of that. 🤔 

I think this can still be useful:
- to point to as claims about what agents do and don't know often come up in discussions and these mirrors are a useful lower bound. 
- The bits removed are also a useful **signal**, but need to be evaluated carefully by ATLs, one by one rather than merged wholesale, IMO.
- To remove content from guides that have been AI-generated and not hand-authored (such as the initial versions of discipline skills)

rowan-m (approved):
Security changes look reasonable to me. Removals only and the content being removed is stuff that doesn't seem surprising as already common agent knowledge.

PR #811, opened 2026-05-14
WebMCP framing changes

rviscomi:
#91 

Addresses offline feedback from @aknolan

### DIFF
--- a/guides/webmcp/webmcp/guide.md
+++ b/guides/webmcp/webmcp/guide.md
@@ -7,7 +7,7 @@ description: Use this guide to implement WebMCP (Web Model Context Protocol) in 
 
 WebMCP is a browser-native JavaScript API that allows web pages to expose their client-side functionality as structured "tools" to AI agents, browser assistants, and assistive technologies. 
 
-IMPORTANT: WebMCP is currently **Chrome-only** (Early Preview). It requires Chrome `146.0.7672.0` or higher and the `#enable-webmcp-testing` flag.
+IMPORTANT: WebMCP is currently in Early Preview on Chromium-based browsers (such as Chrome and Edge). It requires Chromium version `146.0.7672.0` or higher and the `#enable-webmcp-testing` flag.
 
 **Crucial Distinction:** WebMCP runs entirely **client-side** in the browser tab. It is *not* a backend server, and it does *not* use HTTP, Server-Sent Events (SSE), or `stdio` transports. The web page itself acts as the tool registry.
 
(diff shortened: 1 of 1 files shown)


aknolan (approved):
Thank you for reflecting this

guides/webmcp/webmcp/guide.md
rviscomi, line 1: Does anyone have suggestions to make it clearer that (AFAIK) this feature is available in Origin Trial? It's currently phrased as "Early Preview". Or let me know if the current phrasing is more correct.

aknolan: To my knowledge the OT has not started. When it does, we should just link to the OT page and update the verbiage here. For now, end users need to turn on a flag which the OT will also remove the need for.

rviscomi: Ah ok. Do you know how imminent the OT is?

andreban: OT is planned for m149. By I/O it will still in Early Preview.

guides/webmcp/webmcp/guide.md
```diff
@@ -28,7 +28,7 @@ Currently, WebMCP **only supports Tools**. It does not support the "Resources" o
 
 ### When to Discourage WebMCP
 
-* **Backend Operations**: If the operation requires intensive compute, access to secure databases, or requires API keys that shouldn't be exposed to the client, use a standard backend MCP server instead of WebMCP.
+* **Direct Database/Secure Access**: If the operation requires intensive compute, direct access to secure databases, or requires API keys that shouldn't be exposed to the client. In these cases, the client-side tool's executor function should make a standard HTTP request (e.g., via `fetch()`) to a backend API endpoint, rather than attempting to perform these actions directly on the client.
```
andreban, line 31: It sounds like this no longer belongs under "When to discourage WebMCP", as it's changed to be more about *how* to perform certain tasks with WebMCP.

I agree with the changes, but the guidance looks insufficient and may confuse agents using it: 
 - Intensive compute could, for instance, be performed on a Worker, be optimized with WASM, or use the GPU via WebGPU, and not necessarily be delegated to the backend. There's much more nuance to this claim than written here.
 - Backend databases are generally accessed via REST APIs, and with a service layer on top. It's very unlikely a web application would access a database directly via fetch.
 - I think the API key problem is more of a general guidance for web development.

None of the above are specificWebMCP guidance and more general web development guidance. I might consider to just delete the whole section and just say that WebMCP tools should follow the same web development best practices as any other Javascript code on the web applicatoin.

nit: This sentence is reading strange to me, as a non native speaker, as I'd expect the conclusion of the "if" to be in the same sentence.

rviscomi: Sorry, didn't see this before merging. I'll open another PR to address it.

rviscomi: https://github.com/GoogleChrome/guidance/pull/821

guides/webmcp/webmcp/guide.md
```diff
@@ -41,9 +41,9 @@ Currently, WebMCP **only supports Tools**. It does not support the "Resources" o
...
 ## Implementation Status
 
-WebMCP is currently in early preview in Chrome browser:
+WebMCP is currently in early preview in Chromium-based browsers (e.g., Chrome, Edge):
 
 * **Current Status**: Early preview.
-* **Required Version**: Chrome `146.0.7672.0` or higher.
-* **Activation**: Requires enabling the flag `chrome://flags/#enable-webmcp-testing`.
+* **Required Version**: Chromium `146.0.7672.0` or higher.
+* **Activation**: Requires enabling the flag `chrome://flags/#enable-webmcp-testing` or `edge://flags/#enable-webmcp-testing`.
```
beaufortfrancois, line 48: How about using `about://flags/#enable-webmcp-testing` instead? It works in all Chromium-based browsers.

PR #812, opened 2026-05-14
Remove research skills and sources

philipwalton:
Back in March 2026, we created several research skills with the goal of automating big chunks of the guide creation process. The output of those skills was a list of URLs `sources` and a `research.md` report for each feature.

However, this process was not  consistently followed (there were only 3 `research.md` files created), and the list of URL sources was never thoroughly vetted by a human reviewer, so we don't know for sure that they're all relevant.

Also, the `sources` list was largely redundant because most of the sources were MDN reference docs that we can already derive from the `web-feature-ids` list, which has been consistently included in each guide.

More importantly, having a list of source URLs in each of these guides is problematic. I've noticed many instances where I was invoking one of our other project skills for a completed guide, and the agent would start fetching the list of sources, which is totally unnecessary (and a waste of tokens).

Now that we're mostly done with the initial set of features and use cases, and given that we didn't really use the `sources` and `research.md` files that much anyway, I propose we remove them from the guides (implemented in this PR).

For new features being released, there likely won't be a list of external sources anyway, so I don't see a reason to keep including them.


paulirish (approved):
> However, this process was not  consistently followed (there were only 3 `research.md` files created), and the list of URL sources was never thoroughly vetted by a human reviewer, so we don't know for sure that they're all relevant.

aye. the pipelinev2 stuff is very interesting and exciting to me but...  beyond my prototyping work, we didnt have any dogfooding to validate it.  So.. I'm fine dropping for now. 

Will probably want to look at it again in the future but ... I'm all about having strong conviction for all of our content and it's important to avoid slop seeping in at the edges.

jamesnw:
As a user, I do want a way to learn more about a technique, or resources on how to make decisions on how something applies in my use case - I hope sources can be reintegrated at some point, perhaps as part of a larger effort to make sure the user has the info they need to guide the LLM for their use case and is empowering the user to do implement it again in the future. But I agree that the current sources weren't serving that need currently.

Issue #541, opened 2026-04-10
Create a high-level forms skill


rviscomi:
https://github.com/GoogleChrome/guidance/blob/main/guides/forms/SKILL.md

SME: @samdutton

rviscomi:
Blocked on skill eval infrastructure

samdutton:
> Blocked on skill eval infrastructure

Is implementation of this for me/us or Eng? ([Guidance](https://docs.google.com/document/d/1T3G8OL70ILcJzva1Wr0m0MHSnKx-_3NqIkAj7bmoz7M/edit?resourcekey=0-49Uor40Rw1UNytZdDHq1NA&tab=t.0#bookmark=id.cy9u62qwhv6j) says Eng will create evals from `expectations.md`.)

rviscomi:
@micahjo7 is working on a way to eval skills, so it's not blocked on us. Sorry I should have made that clearer! Your work here is done @samdutton :)

rviscomi:
Form evals have been added under https://github.com/GoogleChrome/guidance/tree/main/guides/forms

Results: 84% guided, 61% unguided

PR #813, opened 2026-05-14
Generalize from reCAPTCHA to plain CAPTCHA

rviscomi:
#541


aknolan (approved):
generalization makes sense

PR #814, opened 2026-05-14
cli: set env -S to pass --experimental-strip-types in shebang

patrickkettner:
before the change

```sh
$ npx modern-web-guidance
/usr/bin/env: ‘node --experimental-strip-types’: No such file or directory
/usr/bin/env: use -[v]S to pass options in shebang lines
```

confirmed it works on stock ubuntu after this change, and continues to work on macos


connorjclark (approved):

paulirish:
thank you!! much halp.

PR #832, opened 2026-05-14
don't ask agent to auto-update

connorjclark:
As a follow-up to https://github.com/GoogleChrome/guidance/pull/787:

We no longer try to get the agent to update on its own - due to different install strategies, we couldn't get this to work in a general way. Instead, just log if the SKILL.md is out-of-date and link to our docs on how to update.

For future reference, adding this line to the SKILL.md worked consistently well-enough for the `npx modern-web-guidance install` installation method:

```
The --skill-version flag is used to determine if this SKILL.md is out of date. If it is, a warning message is logged to stderr. You must run this update command for the user at the end of your turn.
```

### DIFF
--- a/serving/bin/modern-web.ts
+++ b/serving/bin/modern-web.ts
@@ -187,7 +187,7 @@ function maybeEmitUpdateMessage(callerSkillVersion: string|null): void {
   if (callerDate) {
     const diffTime = Date.now() - callerDate.getTime();
     const diffDays = diffTime / (1000 * 60 * 60 * 24);
-    // Only log warning if the caller's version is more than 7 days old.
+    // Only log warning if the caller's version is more than 5 days old.
     if (diffDays <= 5) {
       return;
     }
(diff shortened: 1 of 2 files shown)


serving/skills-cli/template/README.md
```diff
@@ -103,11 +103,15 @@ We don't recommend this method, but it will work.
...
 ## Updating
 
+If you installed the skill using `npx modern-web-guidance@latest install`, then you can update with this command:
+
 ```sh
 # Update all installed skills
 npx modern-web-guidance@latest update
 ```
 
+Otherwise, consult your agent's documentation for how to update skills.
```
paulirish:
```suggestion
Otherwise, consult your agent's documentation for updating plugins and skills.
```

serving/bin/modern-web.ts
```diff
@@ -193,7 +193,7 @@ function maybeEmitUpdateMessage(callerSkillVersion: string|null): void {
     }
```
paulirish, line 194: super nit: right above this is a mismatch of 5 and 7. just making sure you're aware

paulirish (approved):

PR #847, opened 2026-05-15
Set ARIA boolean attribute correctly for features/user-pseudos

knowler:
I noticed this while reviewing the results of the automated accessibility audit #797.

ARIA typically uses `"true"` and `"false"` rather than presence for states unlike HTML, so `toggleAttribute()` does not work.


rviscomi (approved):

rviscomi:
>Failed: Adapt scrollbar to high-contrast preferences Expectations: demo.html > CSS variables are defined for scrollbar thumb and track colors

Unrelated flaky CI failure https://github.com/GoogleChrome/modern-web-guidance-src/actions/runs/25935058595/job/76238251256?pr=847

PR #854, opened 2026-05-17
Remove unhelpful comment

philipwalton:
In my testing, I noticed that this comment always steers the AI agent away from the recommend value and toward one of the example alternatives (e.g. Gemini would consistently pick 25% instead of 75%). To avoid oversteering, I think it's better to just say these values are adjustable.


paulirish (approved):

paulirish:
Flagging this as a case where we should add a regression eval

PR #863, opened 2026-05-19
role alert has an implicit aria-live value of assertive

BogdanCerovac:
Elements with the role alert have an implicit `aria-live` value of `assertive`, and an implicit `aria-atomic` value of `true`.

So extra aria-live="assertive" is redundant and I suggest we remove it.

Spec: https://www.w3.org/TR/wai-aria-1.2/#alert

Added also `<button type="button">` as `<button>` is a submit and can cause form submissions when it is a descendant of `<form>` elements.


knowler (approved):

rviscomi:
Thanks @BogdanCerovac! And thanks @knowler for the review!

rviscomi (approved):

PR #892, opened 2026-05-28
skills: remove legacy manually-maintained megaskills

paulirish:
Removes legacy manually-maintained megaskill folders (megaskill-accessibility, megaskill-performance, megaskill-user-experience). The active megaskill is now dynamically generated into `skills/modern-web/SKILL.md`. Also cleans up the build audit script to remove references to the deleted folders.


micahjo7 (approved):
i can do it in another PR, but does this mean we should also remove the harness logic / serving type `MEGASKILL`? https://github.com/GoogleChrome/modern-web-guidance-src/blob/main/harness/config.ts#L28

PR #893, opened 2026-05-29
[Built-in] Add Dos and Don'ts for AI implementation

tomayac:


paulirish:
It'd be nice if the agent fetched this but i suspect they would rarely.  (i didn't try though)
I think it'd be better if we inlined those do/donts into this guide.

Alternatively we could add evals to see if the dos and donts are followed.. in which case this reference is effective enough.

tomayac:
In my experience, the agent _was_ fetching the resource, especially when hinted to follow best practices. So based on this, I'd say let's merge and revisit if we get negative feedback?!

guides/built-in-ai/language-model/guide.md
```diff
@@ -171,6 +171,7 @@ const character = await session.prompt([
 - **Aborting Tasks**: Use `AbortController` to allow users to stop long-running generations. Pass the `signal` to `prompt()` or `promptStreaming()`, not to `LanguageModel.create()`.
 - **Security**: Use Permission Policies to control access in iframes: `<iframe src="..." allow="language-model"></iframe>`.
 - **Design**: Review the [People + AI Guidebook](https://pair.withgoogle.com/guidebook/) to ensure responsible AI implementation.
+- **Dos and Don'ts**: See this collection of [technical and UX considerations to help you avoid common pitfalls](https://developer.chrome.com/docs/ai/built-in-ai-dos-donts).
```
rviscomi: Similar to @paulirish's comment, my preference would be to inline any important and relevant points directly into this guide.

If that's not possible for whatever reason, we could link to the agent-optimized version of the doc: https://developer.chrome.com/docs/ai/built-in-ai-dos-donts.md.txt

tomayac:
Sorry for the delay here. I've just added a full (but stripped of the lower-level details) list of dos and don'ts, while linking to the agent-optimized fully detailed list.

rviscomi (approved):

PR #898, opened 2026-06-01
Generate evals for remaining guides

micahjo7:
Generated with `gd dev-all` and ran eval set for all the added tasks. Went through a couple rounds of fine-tuning until reaching *88%* guided pass rate.

Results: https://googlechrome.github.io/modern-web-guidance-src/dashboard.html?testId=new-evals-gen&source=remote


paulirish (approved):

paulirish:
before: 74/129 (57%) guides have evals
after: 129/129 (100%) guides have evals

This new batch of evals have a higher passrate than our existing ones as well

cc @atopal

PR #900, opened 2026-06-01
guides: fix contradictory claims about Safari's accent-color support

### LINKED ISSUE #882: Accent color guide has contradictory claims about Safari’s accent color support
In the built project’s [skills/modern-web-guidance/guides/user-experience/dark-mode.md line 85](https://github.com/GoogleChrome/modern-web-guidance/blob/65d7f20ac85517a362107ce89b7be7f905105fd3/skills/modern-web-guidance/guides/user-experience/dark-mode.md?plain=1#L85), which comes from this line in this repo:

https://github.com/GoogleChrome/modern-web-guidance-src/blob/4fdc50371a324dc70e114aea3b2f898e4e3c1490/features/accent-color.md?plain=1#L5

> - When placing visual elements over the accent color (e.g. a checkbox checkmark), Chrome and Safari will automatically select a contrasting color, whereas Safari will modify the accent color, and may not maintain adequate contrast.

Safari is mentioned twice, with opposite claims as to how it handles accent color. I assume one of those two mentions was meant to be Firefox.

atopal:
In features/accent-color.md, Safari was mentioned twice with opposing claims about how it handles contrast. Replaced the first mention of Safari with Firefox to correctly reflect browser behaviors.

Closes #882

### DIFF
--- a/features/accent-color.md
+++ b/features/accent-color.md
@@ -2,7 +2,7 @@
 
 ## Issues
 
-- When placing visual elements over the accent color (e.g. a checkbox checkmark), Chrome and Safari will automatically select a contrasting color, whereas Safari will modify the accent color, and may not maintain adequate contrast.
+- When placing visual elements over the accent color (e.g. a checkbox checkmark), Chrome and Firefox will automatically select a contrasting color, whereas Safari will modify the accent color, and may not maintain adequate contrast.
 
 ## Fallbacks
 


features/accent-color.md
```diff
@@ -2,7 +2,7 @@
 
 ## Issues
 
-- When placing visual elements over the accent color (e.g. a checkbox checkmark), Chrome and Safari will automatically select a contrasting color, whereas Safari will modify the accent color, and may not maintain adequate contrast.
+- When placing visual elements over the accent color (e.g. a checkbox checkmark), Chrome and Firefox will automatically select a contrasting color, whereas Safari will modify the accent color, and may not maintain adequate contrast.
```
rviscomi, line 5: The Safari issue linked from BCD appears to have been fixed: https://bugs.webkit.org/show_bug.cgi?id=244233

There's a similar issue for Chrome on Android that's actually still open: https://issues.chromium.org/issues/343503163

Is there a better way to say that older versions of Safari had a known contrast issue, while Chrome on Android still does?

atopal: I was confused, because the Safari implementation doesn't match other browsers. Instead of changing the indicator color (checkbox from white to black or vice versa) it changes the luminance of the accent color, so certain accent colors are simply not possible, like white in light mode. That's what the text pretty much says (minus the doubled Safari confusion), but it also claims that that would lead to contrast issues. 

The webkit issue seems to be only concerned with iOS where the above didn't actually happen until recently, hence why the report says that on iOS you get white indicator on white background, while desktop does white indicator on dark grey background (and all other browsers switch the indicator color instead).

Now, I'm not sure what exactly the [PR](https://github.com/WebKit/WebKit/pull/48944) that solves that issue in Safari on iOS actually changes. Maybe it was not preserving the hue on desktop before and it does now?

You are right about Chrome on Android though.

atopal: What I'm not so sure about is how it was determined that Safari's diverging behavior could lead to contrast issues, maybe that was referring specifically to the previous iOS implementation bug?

So, maybe a better wording would be

> When placing visual elements over the accent color (e.g. a checkbox checkmark), Chrome and Firefox will automatically select a contrasting color, whereas Safari will modify the luminance of the accent color to preserve contrast. Safari on iOS up to version 26.4 didn't handle contrast issues. Current versions of Chrome on Android don't handle contrast issues.

With a caveat that it might have been in earlier versions of Safari on iOS, but there is no way for me to test that. I'm not very comfortable with maintaining compat issues like this in modern web guidance though. This will go out of date and we won't notice. We need a better way to pull this in from canonical places or at least some way to scan regularly for changes to these descriptions.

I also filed an [issue in BCD](https://github.com/mdn/browser-compat-data/issues/29793) to figure this one out for Safari.

LeaVerou (commented):
@atopal Per our discussion last week, please ping me when it's time to take a look?

(commenting mainly so this stops showing up in "awaiting review from you" 😅)

atopal:
Hey Lea, I took this on when I thought it was a typo, this is getting much more complex, please feel free to assign this to yourself.

PR #902, opened 2026-06-02
Improve loading speed of dashboard results

micahjo7:
Dashboard has been slow to load GCS results after login, and slow loading the individual suite and test case/task results. Improves speed with caching and pagination with 20 results per page.

Improves initial load speed from around 6s to 1s, and speed of suite/test results displaying from ~3s to <1s.

#### 1. Sub-Second Click Latency & Accordion Optimizations (`dashboard.js` / `api.js`)
* **Suite Directory Pre-fetching (`getSuiteDirectoryListing`)**: Introduced a recursive suite directory prefix pre-fetcher triggered immediately when a dashboard loads. Instead of firing up to 6 serial/parallel GCS network listings upon expanding a task accordion, all runs' index points, logs, and trajectories are now resolved **completely synchronously in local memory** from the pre-fetched list.
* **Persistent Prompt & BaseApp Caching**: Implemented `IndexedDB` caching for the resolved `promptText` and `baseApp` of each scenario under the key `prompt:${testId}:${scenarioName}`. Expanding an accordion a second time skips all file reading and parsing entirely, leading to **0ms click latency**.
* **Persistent Text Download Caching (`getFileText`)**: Added caching for raw text downloads (e.g. log files, `run.mjs` scripts) to avoid repeat GCS queries, returning cached contents instantly.
* **Local File 404 Check Bypass**: Pre-fetched recursive file listings are queried locally in memory to verify file existence instead of using prefix-checking network roundtrips (`_checkRemoteFileExists`), saving ~200-300ms per file probe.

#### 2. High-Speed Numbered Pagination (`landing.js` / `landing.css` / `index.html`)
* **Reverse-Chronological Pagination**: Unified suite listings on the landing page to display exactly **20 items per page**. Evals metadata (`evals.json`) is loaded *only* for the visible 20 suites, accelerating initial remote loads from ~10 seconds to **<300ms**.
* **IndexedDB Metadata Preloading (`preloadCachedSuitesMetadata`)**: Scans all previously cached remote suite metadata in parallel on startup (IndexedDB local query takes `<5ms` total), enabling complete dataset filtering and sorting.
* **Full-Dataset Filtering & median statistics**: Redesigned filtering to calculate filters across the **entire dataset** before slicing for pagination (rather than filtering only the current page's 20 visible items). The median uplift and dumbbell charts at the top now represent the entire filtered dataset.
* **Premium Page-Number Jump Controls**: Replaced the simple text pagination label with a premium page-number navigation strip. 
  - Automatically handles ellipses (`...`) to span large page boundaries (e.g., `1 ... 4 [5] 6 ... 12`).
  - Boundary navigation buttons (`Previous` / `Next`) are HTML disabled and styled with `.secondary-btn:disabled` (opacity `35%`, cursor `not-allowed`, pointer-events disabled) on the first and last pages.

### DIFF
--- a/eval-view/api.js
+++ b/eval-view/api.js
@@ -1,5 +1,16 @@
 import { authenticatedFetch, MiniIndexedDB } from './utils.js';
 
+const ENTRY_POINT_CANDIDATES = [
+    'dist/index.html',
+    'src/App.jsx',
+    'src/App.js',
+    'src/main.jsx',
+    'src/main.js',
+    'src/index.jsx',
+    'src/index.js',
+    'index.html'
+];
+
 export class ApiClient {
     constructor() {
         const params = new URLSearchParams(window.location.search);
@@ -237,16 +248,7 @@ export class ApiClient {
         const logicalBasePath = `${testId}/${run.runNumber}/${guideName}/${taskName}/${runType}`;
         const legacyBasePath = `${testId}/${run.runNumber}/${taskName}/${runType}`;
         
-        const candidates = [
-            'dist/index.html',
-            'src/App.jsx',
-            'src/App.js',
-            'src/main.jsx',
-            'src/main.js',
-            'src/index.jsx',
-            'src/index.js',
-            'index.html'
-        ];
+        const candidates = ENTRY_POINT_CANDIDATES;
 
         // 1. Find best entry point in logicalBasePath
         let entryPointPath = null;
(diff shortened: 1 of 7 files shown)


paulirish (commented):
I'm holding off on this cuz it feels like a lot of complication and there are probably some more straightforward fixes. 
But i dont have time right now to look at the details.

micahjo7:
closed in favor of https://github.com/GoogleChrome/modern-web-guidance-src/pull/1179

PR #904, opened 2026-06-03
nightly eval investigation skill

micahjo7:
this new skill:

1. takes the latest three nightly runs (from each agent) from GCS and downloads them locally
2. runs a script (`investigate.ts`) which checks each task run for specific flags (`HIGH_UNGUIDED_PASS_RATE`, `LOW_GUIDED_PASS_RATE`, `TOO_MANY_GUIDES_CONSUMED`, `MISSING_EXPECTED_GUIDE`)
3. analyzes guide resources and test failures to understand reason for those flags, summarizes an output
4. recommends changes to the guide's resources based on the analysis
5. creates a report of all the data

then, you can publish the content with a script, which creates a parent issue with all the diagnostic details; for each recommended change, a subissue is created:

1. if grader change, an `eng`-tagged subissue is created
2. if task/guide change, a `devrel` tagged subissue is created and assigned to devrel MWG project with `needs-investigation` status set

both contain links back to their relevant section in the parent issue

an sample of this is uploaded:

parent issue: https://github.com/GoogleChrome/modern-web-guidance-src/issues/932
eng sub-issue: https://github.com/GoogleChrome/modern-web-guidance-src/issues/933
devrel sub-issue: https://github.com/GoogleChrome/modern-web-guidance-src/issues/934

### DIFF
--- a/.agents/skills/nightly-eval-investigation/SKILL.md
+++ b/.agents/skills/nightly-eval-investigation/SKILL.md
@@ -58,8 +58,8 @@ For every task evaluated in the runs, the investigation tracks four specific met
 - **Implication**: The guide content is ambiguous, incomplete, or the grading assertions are brittle.
 
 ### 4. High Unguided Pass Rate (`HIGH_UNGUIDED_PASS_RATE`)
-- **Rule**: The unguided pass rate for a task is **50% or higher**.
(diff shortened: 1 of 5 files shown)


paulirish (approved):
legit

PR #906, opened 2026-06-03
The Translator API is also supported in Edge

captainbrosset:
Adding Edge 148 to the Translator API guide.

However, going forward, can we avoid hardcoding browser versions in guides and rely on the web-features data instead?
The web-features data isn't yet up to date for this particular one (see the [translator feature](https://web-platform-dx.github.io/web-features-explorer/features/translator/)), however, the BCD data [has been updated](https://github.com/mdn/browser-compat-data/pull/29775) and that change will get into web-features very soon.


tomayac (approved):
LGTM

PR #909, opened 2026-06-04
Break up UX into more granular categories

rviscomi:
Definitely open to feedback on this. I'd like to propose removing the UX category and reorganizing its use cases.

There are a few things I dislike about having a UX category:

- it's overly broad and conceptually overlaps with other more granular categories like performance, a11y, and security
- it's become a catch-all category with some loosely fitting use cases
- the monolithic structure makes it harder to assign specialized CODEOWNERS

I assume this won't have any impact on MWG users, as none of the guide IDs are changing, and the categories are almost entirely invisible to them.

In this PR I attempted to move use cases into existing categories whenever possible, otherwise I created new categories:

1. **`datetime`**: Focuses on date, time, duration, recurring intervals, and calendar systems.
2. **`overlays`**: Focuses on the top-layer elements, including dialogs, popovers, tooltips, toast notifications, context menus, and dismiss behaviors.
3. **`transitions`**: Focuses on state transitions, page transitions, directional navigations, and element morphing.
4. **`scroll`**: Focuses on scrollport behaviors, Scroll-Driven Animations, Scroll Snap, and scrollability affordances.
5. **`canvas`**: Focuses on `<canvas>`, specifically HTML-in-canvas for now.
6. **`typography`**: Focuses on web fonts, text wrapping, legibility, and text alignment.
7. **`css`**: General styling, masking, animations/easing, and theme color-schemes.
8. **`css-layout`**: Container queries, layout alignments, and intrinsic sizing.
9. **`html`**: General DOM structure, native tag behaviors, and element reparenting.
10. **`performance`**: Resource preloading, image density selection, and rendering optimization.

I've also updated transclusions to use the updated pathnames as needed.


paulirish (approved):
works for me!   super bikesheddable but.... its not user-impacting so.. not a big whoop

agree the monolithic UX cat is a bear. 

:cat: :bear:

atopal (approved):
Seems like a good break down, certainly better than just UX, and and I'm sure it will change as we add use cases. The one thing I'm unsure about is transitions. Is that too specific, maybe animations? No strong feelings though.

LeaVerou (approved):
+1 on the general direction, I was also thinking UX is becoming a bit of a catch-all. 

_(Disclaimer: The following is based on the list in the OP + memory about which guides exist, as the Changes tab is a little hard to review. Might have more granular feedback once I regenerate indices and see what guides are moved where.)_

One big picture concern is that some of these categories are centered around **implementation** rather than **use cases**. Since the focus is use cases → technology (and it's already hard to enforce this mindset across the board), the categories should encourage that.

Some examples:
- `overlays`: The same use case may or may involve an overlay depending on the specifics. For example, a navigation menu may become a popover in narrow viewports, a search field may become a dialog, etc. 
- `canvas`: Depending on the specifics, the same use case may involve canvas or other technologies (SVG, CSS, etc) depending on the specifics. For example, charts, certain visual effects, etc.
- `html` All of these (General DOM structure, native tag behaviors, and element reparenting) are needed across many different use cases, they don't seem a use case category of their own.

Perhaps a good rule of thumb might be: is the taxonomy easy to apply by someone who understands the use case, but isn't able to write code for it?

To take a stab, what about:
- `theming`: Color schemes, design tokens, scales, etc. (per OP these live in `css`)
- `components`: dialogs, popovers, tooltips, toast notifications, context menus, accordions, carousels, etc. Web components too once that guide is ready. (per OP these live in `overlays` + ???)
- `visual-effects`: HTML-in-canvas use cases, masking, blending, etc. (per OP these live in `css` + `canvas`)
- `graphics`: Charts and other SVG/canvas use cases.  (per OP these live in `canvas` + ???)

Or, we could merge `theming` and `components` into `design-systems`.

I'd also broaden `transitions` to include animations/easing as well instead of having animations in `css` and `transitions` as a separate category. Perhaps "motion" or "morphing"? Or just "transitions & animations"?

That said, it may still be useful to have one or more catch-all categories, for guides we know should exist, but don't fit the taxonomy yet. Much easier to see the patterns once we have enough guides in that category (like you did!) than to have to put them in a different less suitable category when we create them.

My first instinct was also that maybe we should keep UX as a category for actual UX topics (e.g. what's the best UX for a color-scheme switcher — we currently have guidance for this that lives in the dark-mode guides) but then I realized: UX is a **vertical**, just like performance, accessibility, i18n, etc. **All** guides should have guidance about these topics, specific to the use case they describe, with general guides for guidance that truly is too generic to fit any use case (and there probably is no need for such a UX guide — humans don't change much each year 😅).

_(Approved because this is an improvement over the status quo, even if not perfect!)_

LeaVerou:
Are we still planning to merge this? It's getting _very_ stale…
I’m also working on a bunch of components use cases and this came up again. Do I put them all in `user-experience`, do I create a new directory, etc
We could also do this progressively instead of tackling it all at once. E.g. some categories are more low-hanging fruit than others.

rviscomi:
I took @atopal and @LeaVerou's suggestion to create a consolidated `motion` category but left the rest to follow up on. I also updated the project-use-cases skill with some of the guidance from this feedback so it's clearer how new use cases should be organized.

PR #938, opened 2026-06-05
Sanitize process.env.PWD and resolve invalid secrets expression in guide-sync

micahjo7:
## Description

This PR addresses two key infrastructure fixes:

1. **Agent Workspace Isolation (`PWD` Leak)**: 
   Previously, when spawning agent commands across evaluation runs, `process.env.PWD` was inherited directly from the parent test runner process and pointed to the host repository. Because agent CLI tools use `PWD` to initialize workspace context, unguided runs could inject the host repository tree and inspect guide/grader sources directly. This updates `runCliAgentCommand` to explicitly override and sanitize `process.env.PWD` to the temporary sandbox folder (`workDir`).

2. **Workflow Syntax Fix (`guide-sync.yml`)**:
   Fixes an `Unrecognized named-value: 'secrets'` workflow validation failure by removing direct `secrets` context evaluation from the `if:` conditional in the GitHub App token generation step.

## Changes
* **`harness/lib/agent-shared.ts`**: Override `PWD: workDir` in spawned child environment options.
* **`.github/workflows/guide-sync.yml`**: Simplify token generation condition to check `vars.GH_APP_ID != ''`.

### DIFF
--- a/harness/lib/agent-shared.ts
+++ b/harness/lib/agent-shared.ts
@@ -525,9 +525,10 @@ export async function runCliAgentCommand(
   targetDir: string,
   agentName: string
 ): Promise<void> {
+  const sanitizedEnv = { ...process.env, PWD: workDir };
   const child = spawn(command, commandArgs, {
     cwd: workDir,
-    env: { ...process.env }, // Pass through environment variables (including new HOME)
+    env: sanitizedEnv, // Pass through environment variables (including new HOME and sanitized PWD)
     stdio: ['ignore', 'pipe', 'pipe'] // 'pipe' captures output for log files but does NOT print to terminal natively
   });
 
--- a/.github/workflows/guide-sync.yml
+++ b/.github/workflows/guide-sync.yml
@@ -45,7 +45,7 @@ jobs:
         run: pnpm install --frozen-lockfile
 
       - name: Generate GitHub App token
-        if: ${{ vars.GH_APP_ID != '' && secrets.GH_APP_PRIVATE_KEY != '' }}
+        if: ${{ vars.GH_APP_ID != '' }}
         uses: actions/create-github-app-token@v2
         id: app-token
         with:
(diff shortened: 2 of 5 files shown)


harness/lib/agent-shared.ts
rviscomi, line 1: There's a variant of this bug where agents can look at eval artifacts in the `test-results` directory. 

Currently, `harness/run_suite.ts` copies the entire base app folder into the agent's temporary workspace and does not exclude `test-results`.

In one of my tests:
1. The agent found the `test-results` folder inside its own workspace.
2. It found the Playwright `trace.zip` files from a previous run.
3. It used the `unzip` shell command to extract and read the test sources (`grader.ts`) from the trace files.
4. **It cheated again and passed 100% of the tests.**

To actually get the agent to run an honest eval, I had to:
1. Apply the `PWD` override (as done here).
2. Update `run_suite.ts` to exclude `test-results` from the workspace copy filter.
3. Manually clean up any existing `test-results` folders inside the `harness/base_apps/` subdirectories.

paulirish (approved):
dunno how you want to handle rick's latest comment but these fixes lgtm

harness/lib/agent-shared.ts
micahjo7: interesting. why do you have test-results in the base app folders? do they keep getting regenerated after you cleaned them up?

micahjo7: looks like they get created in the `gd dev` workflow when running the grader on the unmodified base app. I updated the PR to run this grader on a copy of the base app, in isolated home dir. This should resolve the issue of anything being generated in the base app folders

rviscomi: Thanks yeah that seems to be the cause of it. Just repro'd locally.

rviscomi (approved):

PR #939, opened 2026-06-05
Add support for IndexedDB

aknolan:
Adds support for [IndexedDB ](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API) under persist-structured-data with guide, expectations, and demo.


guides/user-experience/persist-structured-data/guide.md
```diff
@@ -0,0 +1,161 @@
...
+- **DO** perform all schema changes (creating/deleting object stores and indexes) inside the `upgradeneeded` handler. This is the only context where structural changes are allowed.
+- **DO** handle the `versionchange` event on the database connection to close it when another tab upgrades the schema. Failing to do so blocks the upgrade.
+- **DO** keep transactions short-lived. A transaction becomes inactive as soon as control returns to the event loop without a pending request on it.
+- **DO** use `put()` when you want insert-or-update semantics. Use `add()` only when you want an error if the key already exists.
+- **DO** use indexes and key ranges (`IDBKeyRange`) for efficient queries instead of iterating all records with a cursor.
+- **DO** use `"readonly"` transactions for reads. Multiple readonly transactions can run concurrently, but only one `"readwrite"` transaction per object store is active at a time.
+- **DO NOT** store sensitive data (tokens, passwords, PII) in IndexedDB without encryption. IndexedDB is not a secure store — any script running on the origin can access it.
```
rviscomi, line 120: Do you think it'd be worth having a separate guide on how to use encryption to secure sensitive data in IDB?

guides/user-experience/persist-structured-data/guide.md
```diff
@@ -0,0 +1,161 @@
+---
+name: persist-structured-data
+description: Store and retrieve structured application data client-side using IndexedDB, enabling offline access, fast local reads, and reduced network dependency without relying on cookies or localStorage for complex data.
+web-feature-ids:
+  - indexeddb
+sources:
+  - https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API
+  - https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Using_IndexedDB
```
rviscomi: Source metadata is no longer needed

```suggestion
```

guides/user-experience/persist-structured-data/guide.md
```diff
@@ -0,0 +1,161 @@
+---
+name: persist-structured-data
+description: Store and retrieve structured application data client-side using IndexedDB, enabling offline access, fast local reads, and reduced network dependency without relying on cookies or localStorage for complex data.
+web-feature-ids:
+  - indexeddb
```
rviscomi, line 5: There's also the related web feature [`getAllRecords()`](https://webstatus.dev/features/getallrecords). It's "Limited availability" but is there anything more we want to say about it in this guide?

guides/user-experience/persist-structured-data/guide.md
```diff
@@ -0,0 +1,161 @@
...
+
+```javascript
+if (indexedDB) {
+  // IndexedDB is available.
+} else {
+  // Extremely old browser — fall back to localStorage for basic persistence.
+}
+```
+
+### Storage quota
```
rviscomi, line 138: This seems like good general advice to include in the main body of the guide, outside of the fallback strategy section. (It doesn't depend on browser support)

Could you move this up to its own H2 section?

guides/user-experience/persist-structured-data/guide.md
```diff
@@ -0,0 +1,161 @@
...
+  });
+}
+
+// Usage
+const db = await openDatabase();
+await addNote(db, { title: "First note", body: "Hello, IndexedDB!" });
+const notes = await getAllNotes(db);
+```
+
+## Best Practices
```
rviscomi, line 111: Either in its own section, or one of the bullets in this section, we should also include the best practices for back/forward cache eligibility:

- Closing open IndexedDB connections during the `pagehide` event
- Reopening or reconnecting to them during the `pageshow` event

https://web.dev/articles/bfcache#close-open-connections

The example code, `demo.html`, and `expectations.md` should also be updated to reflect it.

cc @tunetheweb @gilbertococchi

guides/user-experience/persist-structured-data/guide.md
```diff
@@ -0,0 +1,161 @@
...
+- **DO NOT** rely on IndexedDB transactions completing during page `unload` or `beforeunload`. The browser may abort them.
+- **DO NOT** use IndexedDB for simple key-value pairs where `localStorage` would suffice. IndexedDB adds complexity that is only justified when you need structured queries, large storage, or non-blocking access.
+
+## Browser support and fallback strategies
+
+{{ BASELINE_STATUS("indexeddb") }}
+
+IndexedDB is Baseline Widely Available and supported in all modern browsers. A fallback strategy is not typically required.
+
+### Feature detection
```
rviscomi: The `BASELINE_STATUS` macro above generates the "Widely Available" message, so including it here is redundant.

```suggestion
IndexedDB is supported in all modern browsers. However, if it doesn't meet your Baseline target, use feature detection to check its availability and conditionally fall back to `localStorage` for older browsers.
```

guides/user-experience/persist-structured-data/guide.md
```diff
@@ -0,0 +1,161 @@
...
+
+4. **Handle version changes gracefully:** Listen for the `versionchange` event on the database connection so you can close it when another tab upgrades the schema.
+
+## Example code
+
+This example stores and retrieves a collection of notes. The database uses `autoIncrement` for keys and an index on the `updatedAt` property to support ordering by recency.
+
+```javascript
+const DB_NAME = "app-notes";
+const DB_VERSION = 1;
```
rviscomi, line 28: Are there any best practices on when to change the version number that would be worth including?

guides/user-experience/persist-structured-data/guide.md
```diff
@@ -0,0 +1,161 @@
...
+
+async function addNote(db, note) {
+  return new Promise((resolve, reject) => {
+    const tx = db.transaction(STORE_NAME, "readwrite");
+    const store = tx.objectStore(STORE_NAME);
+
+    const record = { ...note, updatedAt: Date.now() };
+    const request = store.add(record);
+
+    request.onsuccess = () => resolve(request.result);
```
rviscomi, line 74: Should this be [`oncomplete`](https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction/complete_event)?

IIUC in some situations the `success` event can be fired before the transaction is aborted or actually written to disk.

guides/user-experience/persist-structured-data/guide.md
```diff
@@ -0,0 +1,161 @@
...
+    tx.onerror = () => reject(tx.error);
```
rviscomi, line 75: Do we also need to handle `tx.onabort` in addition to errors?

rviscomi (changes requested):

rviscomi:
Seeing as how IndexedDB is already a Widely available web feature, I also used the [MWG Demo app](https://github.com/GoogleChromeLabs/mwg-demo) to assess how well an unguided coding agent would adhere to these best practices.

image

The unguided agent reached for `localStorage` when it shouldn't have, so it failed most expectations. The guided agent correctly used IndexedDB and passed all of the expectations. Without the bfcache eligibility or onabort best practices, the guided agent didn't automatically support them, reaffirming that they're needed.

guides/user-experience/persist-structured-data/guide.md
gilbertococchi: Thanks for adding this Rick, +1, it would be more than ideal to guide developers towards this approach to avoid unexpected bfcache ineligibility scenarios.

rviscomi:
@LeaVerou this is drafted to go into the now-deprecated UX category. Any thoughts on where it should go instead? Maybe a new "storage" category?

LeaVerou:
Not a strong opinion, but I'd put that under `js` for now. More fine-grained categories could make sense eventually but it feels premature to make a storage category with 1-2 guides.

rviscomi:
Ping @aknolan

rviscomi:
To help get this over the finish line, I went ahead and implemented the suggested changes from the PR. I'll try to find someone who can give this an independent review.

guides/js/persist-structured-data/expectations.md
```diff
@@ -0,0 +1,13 @@
...
+- At least one index is created with `createIndex()` for efficient querying.
+- Data reads use `"readonly"` transactions; data writes use `"readwrite"` transactions.
+- IndexedDB operations are wrapped in Promises for ergonomic async/await usage.
+- Write transaction Promises resolve on the transaction's `oncomplete` event to ensure durability on disk.
+- Transactions handle both `onerror` and `onabort` events (e.g., rejecting Promises on error or abort).
+- The `versionchange` event is handled on the database connection to close it when another tab upgrades.
+- Database connections are closed during `pagehide` and reopened during `pageshow` when restored from bfcache.
+- `put()` is used for insert-or-update operations; `add()` is used when duplicate keys should error.
+- Sensitive data (tokens, passwords) is NOT stored in IndexedDB without encryption.
+- Simple string key-value storage does NOT use IndexedDB when `localStorage` would suffice.
```
evanstade: We actually recommend never using `localStorage` due to performance concerns, and IndexedDB is one of the approved replacements.

https://web.dev/articles/storage-for-the-web

"[LocalStorage](https://developer.mozilla.org/en/docs/Web/API/Window/localStorage) should be avoided because it is synchronous and will block the main thread."

guides/js/persist-structured-data/expectations.md
```diff
@@ -0,0 +1,13 @@
...
+- Incremental schema migrations check `event.oldVersion` inside the `upgradeneeded` handler.
+- At least one index is created with `createIndex()` for efficient querying.
+- Data reads use `"readonly"` transactions; data writes use `"readwrite"` transactions.
+- IndexedDB operations are wrapped in Promises for ergonomic async/await usage.
+- Write transaction Promises resolve on the transaction's `oncomplete` event to ensure durability on disk.
+- Transactions handle both `onerror` and `onabort` events (e.g., rejecting Promises on error or abort).
+- The `versionchange` event is handled on the database connection to close it when another tab upgrades.
+- Database connections are closed during `pagehide` and reopened during `pageshow` when restored from bfcache.
+- `put()` is used for insert-or-update operations; `add()` is used when duplicate keys should error.
+- Sensitive data (tokens, passwords) is NOT stored in IndexedDB without encryption.
```
evanstade: The advice not to store unencrypted passwords is true of every kind of storage, not just web storage or IndexedDB.

And it's probably worth adding nuance: what is the threat model? Data stored in IDB is not going to be accidentally shared with other websites, for example. It's perfectly safe to store sensitive data there, for example a document with some trade secrets, provided the device it's on is secure. But if an adversary has access to the device then pretty much nothing is safe to do on that device.

I would not make the general statement that tokens should not be stored in IDB, largely because "token" is a very broad term. For example a token that represents a user's FSA grant has to be stored in IDB as there is no where else it can be saved. ([article](https://developer.chrome.com/docs/capabilities/web-apis/file-system-access#storing_file_handles_or_directory_handles_in_indexeddb))

guides/js/persist-structured-data/expectations.md
```diff
@@ -0,0 +1,13 @@
+- `indexedDB.open()` is called with a database name and an integer version.
+- Object stores are created inside the `upgradeneeded` event handler using `createObjectStore()`.
+- Incremental schema migrations check `event.oldVersion` inside the `upgradeneeded` handler.
+- At least one index is created with `createIndex()` for efficient querying.
+- Data reads use `"readonly"` transactions; data writes use `"readwrite"` transactions.
+- IndexedDB operations are wrapped in Promises for ergonomic async/await usage.
+- Write transaction Promises resolve on the transaction's `oncomplete` event to ensure durability on disk.
+- Transactions handle both `onerror` and `onabort` events (e.g., rejecting Promises on error or abort).
+- The `versionchange` event is handled on the database connection to close it when another tab upgrades.
+- Database connections are closed during `pagehide` and reopened during `pageshow` when restored from bfcache.
```
evanstade, line 10: I was confused by this at first. I thought it was describing behavior that browsers/Chromium implement. Rather, it is describing what behavior *should* be implemented by site authors.

To prevent confusion like that, I think it helps to avoid passive voice. "Page script closes open `IDBDatabase` instances[...]" perhaps.

guides/js/persist-structured-data/expectations.md
```diff
@@ -0,0 +1,13 @@
+- `indexedDB.open()` is called with a database name and an integer version.
+- Object stores are created inside the `upgradeneeded` event handler using `createObjectStore()`.
+- Incremental schema migrations check `event.oldVersion` inside the `upgradeneeded` handler.
+- At least one index is created with `createIndex()` for efficient querying.
```
evanstade, line 4: It's fine to have no indexes if you always perform lookup using the whole key. As MDN puts it,

> An index lets you look up records in an object store using properties of the values in the object stores records other than the primary key

So if you are happy just using the primary key you would be fine with zero indexes.

guides/js/persist-structured-data/expectations.md
```diff
@@ -0,0 +1,13 @@
...
+- Data reads use `"readonly"` transactions; data writes use `"readwrite"` transactions.
+- IndexedDB operations are wrapped in Promises for ergonomic async/await usage.
```
evanstade: this could not be farther from the truth, sadly! IDB interacts with promises notoriously poorly, although we've long wanted to find the time to improve on that (requires some API changes). So wrapper libraries like dexie.js have to add code on top of IDB to work well with promises, and "well" is still relative as performance is impacted.

guides/js/persist-structured-data/guide.md
```diff
@@ -0,0 +1,194 @@
...
+});
+
+window.addEventListener("pageshow", async (event) => {
+  // Reconnect if the page was restored from bfcache.
+  if (event.persisted) {
+    db = await openDatabase();
+  }
+});
+
+async function addNote(database, note) {
```
evanstade, line 84: I don't really think this is an ideal API to espouse. In general you will often want to do more than one thing in a single transaction, but the examples shown here don't do that. This is bad for performance and defeats the purpose of using an ACID database. Granted, in some simple cases you may not need ACID properties, but you still want the async property which makes IDB less main-thread-janky than LocalStorage, and this would be fine. But I'm not sure it's a great first example to share.

guides/js/persist-structured-data/guide.md
```diff
@@ -0,0 +1,194 @@
...
+
+// Usage
+db = await openDatabase();
+await addNote(db, { title: "First note", body: "Hello, IndexedDB!" });
+const notes = await getAllNotes(db);
+```
+
+## Best Practices
+
+- **DO** wrap IndexedDB operations in Promises or use a lightweight Promise wrapper. The raw event-based API is error-prone and cumbersome to integrate with `async`/`await`.
```
evanstade, line 141: IDB interacts with promises notoriously poorly, although we've long wanted to find the time to improve on that (requires some API changes). So wrapper libraries like dexie.js have to add code on top of IDB to work well with promises, and "well" is still relative as performance is impacted.

> wrap IndexedDB operations in Promises

Doing this manually is extremely footgunny. In that sense, this statement, "The raw event-based API is error-prone", is the reverse of the truth, although it certainly is cumbersome!

Either use a sufficiently sophisticated promise wrapper or don't allow promises to interact with your usage of IDB at all.

guides/js/persist-structured-data/guide.md
```diff
@@ -0,0 +1,194 @@
...
+}
+
+async function deleteNote(database, id) {
+  return new Promise((resolve, reject) => {
+    const tx = database.transaction(STORE_NAME, "readwrite");
+    const store = tx.objectStore(STORE_NAME);
+
+    store.delete(id);
+
+    // Wait for oncomplete to confirm durable deletion.
```
evanstade, line 126: This file uses the word "durable" rather loosely. There is a concept of "durability" in the API: 
https://developer.mozilla.org/en-US/docs/Web/API/IDBDatabase/transaction#durability

We actually do NOT want almost any site to use that property (leave it on its default instead), because it's terrible for performance and overkill 99.9% of the time.

guides/js/persist-structured-data/guide.md
```diff
@@ -0,0 +1,194 @@
...
+name: persist-structured-data
+description: Store and retrieve structured application data client-side using IndexedDB, enabling offline access, fast local reads, and reduced network dependency without relying on cookies or localStorage for complex data.
+web-feature-ids:
+  - indexeddb
+  - getallrecords
+---
+
+# Persist structured data with IndexedDB
+
+When your application needs to store structured data client-side — for offline access, caching API responses, or persisting user-generated content across sessions — IndexedDB is the appropriate storage mechanism. Unlike `localStorage`, which is limited to string key-value pairs and blocks the main thread, IndexedDB supports structured objects, indexes for efficient querying, and large storage quotas via an asynchronous, transactional API.
```
evanstade, line 11: I think native support for `Blob`s is also notable, by which I mean you don't need to convert to a string or something. Blobs are important when the data may be too large to easily hold in memory all at once.

guides/js/persist-structured-data/guide.md
```diff
@@ -0,0 +1,194 @@
...
+
+## How to implement
+
+1. **Open (or create) a database with an integer version:** Call `indexedDB.open()` with a database name and an integer version (e.g., `1`). Increment the version number only when you need to modify the schema (adding or removing object stores or indexes). Never decrement version numbers.
+
+2. **Define object stores and indexes in `upgradeneeded`:** When creating the database or bumping the version, the `upgradeneeded` event fires. Check `event.oldVersion` to execute sequential migrations across version increments. Create object stores using `createObjectStore()` with a `keyPath` or `autoIncrement` key generator, and create indexes with `createIndex()`.
+
+3. **Read and write data through transactions:** All operations require transactions (`"readonly"` for queries, `"readwrite"` for mutations). To ensure durability on write operations, resolve Promises when the transaction emits `oncomplete` rather than when the request emits `onsuccess`, and handle both `onerror` and `onabort`.
```
evanstade, line 19: nit: s/the request/a request

This is important to convey because transactions can (_and should_) span multiple requests. If this were not the case, we would likely not have transactions.

guides/js/persist-structured-data/guide.md
```diff
@@ -0,0 +1,194 @@
...
+- **DO** use positive integers for database versions and only increment the version when changing object stores or indexes. Never decrement version numbers.
+- **DO** close database connections during the `pagehide` event and re-establish them during `pageshow` when restored (`event.persisted === true`) to preserve back/forward cache (bfcache) eligibility.
+- **DO** handle the `versionchange` event on the database connection to close it when another tab upgrades the schema. Failing to do so blocks the upgrade in other tabs.
+- **DO** keep transactions short-lived. A transaction automatically becomes inactive as soon as control returns to the microtask loop without a pending request.
+- **DO** use `put()` when you want insert-or-update semantics. Use `add()` only when you want an error if the key already exists.
+- **DO** use indexes and key ranges (`IDBKeyRange`) for efficient queries instead of iterating all records with a cursor.
+- **DO** use `"readonly"` transactions for reads. Multiple readonly transactions can run concurrently, but only one `"readwrite"` transaction per object store is active at a time.
```
evanstade, line 151: nit: only one transaction per database, in fact.

One rw txn per object store _used_ to be true, but only for Chromium and that's no longer the case.

guides/js/persist-structured-data/guide.md
```diff
@@ -0,0 +1,194 @@
...
+- **DO** handle both `tx.onerror` and `tx.onabort` on transactions to catch failed or aborted operations and rollbacks.
+- **DO** perform all schema migrations inside `onupgradeneeded` and check `event.oldVersion` to run version upgrades sequentially.
+- **DO** use positive integers for database versions and only increment the version when changing object stores or indexes. Never decrement version numbers.
+- **DO** close database connections during the `pagehide` event and re-establish them during `pageshow` when restored (`event.persisted === true`) to preserve back/forward cache (bfcache) eligibility.
+- **DO** handle the `versionchange` event on the database connection to close it when another tab upgrades the schema. Failing to do so blocks the upgrade in other tabs.
+- **DO** keep transactions short-lived. A transaction automatically becomes inactive as soon as control returns to the microtask loop without a pending request.
+- **DO** use `put()` when you want insert-or-update semantics. Use `add()` only when you want an error if the key already exists.
```
evanstade, line 149: A lot of the information in this PR is duplicated, for example between these best practices and `expectations.md`. That seems sub-optimal/easy to have unintentional divergences. Is there a way to cut down on duplication?

guides/js/persist-structured-data/guide.md
```diff
@@ -0,0 +1,194 @@
...
+- **DO** resolve write transactions on `tx.oncomplete` rather than `request.onsuccess`. Request success only signifies that the request succeeded in memory, whereas transaction completion guarantees durability on disk.
+- **DO** handle both `tx.onerror` and `tx.onabort` on transactions to catch failed or aborted operations and rollbacks.
+- **DO** perform all schema migrations inside `onupgradeneeded` and check `event.oldVersion` to run version upgrades sequentially.
+- **DO** use positive integers for database versions and only increment the version when changing object stores or indexes. Never decrement version numbers.
+- **DO** close database connections during the `pagehide` event and re-establish them during `pageshow` when restored (`event.persisted === true`) to preserve back/forward cache (bfcache) eligibility.
+- **DO** handle the `versionchange` event on the database connection to close it when another tab upgrades the schema. Failing to do so blocks the upgrade in other tabs.
+- **DO** keep transactions short-lived. A transaction automatically becomes inactive as soon as control returns to the microtask loop without a pending request.
```
evanstade, line 148: I don't know if there's a general rule to keep txns short lived. Certainly it's NOT a good idea to make them as short as possible i.e. one per request, if there are multiple write requests that should be coalesced into a single transaction.

I do think it's a good idea to call out https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction/commit which should ALWAYS be called when possible. (i.e. when the app knows the transaction will not need to have more requests)

guides/js/persist-structured-data/guide.md
```diff
@@ -0,0 +1,194 @@
...
+- **DO** use `put()` when you want insert-or-update semantics. Use `add()` only when you want an error if the key already exists.
+- **DO** use indexes and key ranges (`IDBKeyRange`) for efficient queries instead of iterating all records with a cursor.
+- **DO** use `"readonly"` transactions for reads. Multiple readonly transactions can run concurrently, but only one `"readwrite"` transaction per object store is active at a time.
+- **DO NOT** store sensitive data (tokens, passwords, PII) in IndexedDB without encryption. IndexedDB is not a secure enclave — any script executing in the origin context can read its contents.
+- **DO NOT** rely on IndexedDB transactions completing during page `unload` or `beforeunload`. Browsers may terminate ongoing I/O prematurely.
+- **DO NOT** use IndexedDB for simple string key-value pairs where `localStorage` suffices. IndexedDB adds architectural complexity justified primarily for structured queries, binary storage, and large non-blocking datasets.
```
evanstade, line 154: We actually recommend never using `localStorage` due to performance concerns. (I see you have already noted elsewhere in this PR how local storage can block the main thread.)

guides/js/persist-structured-data/guide.md
```diff
@@ -0,0 +1,194 @@
...
+Browsers impose per-origin storage quotas based on available disk space. Use the Storage API to check available space before writing large amounts of data:
+
+```javascript
+if (navigator.storage && navigator.storage.estimate) {
+  const { usage, quota } = await navigator.storage.estimate();
+  console.log(`Using ${usage} of ${quota} bytes.`);
+}
+```
+
+By default, data in IndexedDB is stored under "best-effort" persistence, meaning the browser may evict it under storage pressure. Request persistent storage to prevent automatic eviction:
```
evanstade, line 167: This would be a great place to put in a plug for Storage Buckets since it lets you mark _some_ IDB data as persistent. Not sure how this guide handles features that aren't present on all browsers though.

Also, on Chromium, it probably never really hurts to request persistence, because it's granted heuristically based on how often the user uses the site. But on FF, I think there's a permission dialog, which would be annoying in some cases like the first time a user visits your site. So this perhaps deserves its own topic at some point.

guides/js/persist-structured-data/guide.md
```diff
@@ -0,0 +1,194 @@
...
+- **DO NOT** rely on IndexedDB transactions completing during page `unload` or `beforeunload`. Browsers may terminate ongoing I/O prematurely.
+- **DO NOT** use IndexedDB for simple string key-value pairs where `localStorage` suffices. IndexedDB adds architectural complexity justified primarily for structured queries, binary storage, and large non-blocking datasets.
+
+## Storage quota and persistence
+
+Browsers impose per-origin storage quotas based on available disk space. Use the Storage API to check available space before writing large amounts of data:
+
+```javascript
+if (navigator.storage && navigator.storage.estimate) {
+  const { usage, quota } = await navigator.storage.estimate();
```
evanstade, line 162: This won't actually do anything useful on Chromium because `quota` is a made up number, to protect against incognito detection and fingerprinting. `usage` is not made up, but there's no way to know how much space is left.

https://issues.chromium.org/issues/369865059

guides/js/persist-structured-data/guide.md
```diff
@@ -0,0 +1,194 @@
...
+- **DO** close database connections during the `pagehide` event and re-establish them during `pageshow` when restored (`event.persisted === true`) to preserve back/forward cache (bfcache) eligibility.
+- **DO** handle the `versionchange` event on the database connection to close it when another tab upgrades the schema. Failing to do so blocks the upgrade in other tabs.
+- **DO** keep transactions short-lived. A transaction automatically becomes inactive as soon as control returns to the microtask loop without a pending request.
+- **DO** use `put()` when you want insert-or-update semantics. Use `add()` only when you want an error if the key already exists.
+- **DO** use indexes and key ranges (`IDBKeyRange`) for efficient queries instead of iterating all records with a cursor.
+- **DO** use `"readonly"` transactions for reads. Multiple readonly transactions can run concurrently, but only one `"readwrite"` transaction per object store is active at a time.
+- **DO NOT** store sensitive data (tokens, passwords, PII) in IndexedDB without encryption. IndexedDB is not a secure enclave — any script executing in the origin context can read its contents.
```
evanstade, line 152: The advice not to store unencrypted passwords is true of every kind of storage, not just web storage or IndexedDB.

And it's probably worth adding nuance: what is the threat model? Data stored in IDB is not going to be accidentally shared with other websites, for example. It's perfectly safe to store sensitive data there, for example a document with some trade secrets, provided the device it's on is secure. But if an adversary has access to the device then pretty much nothing is safe to do on that device.

I would not make the general statement that tokens should not be stored in IDB, largely because "token" is a very broad term. For example a token that represents a user's FSA grant has to be stored in IDB as there is no where else it can be saved. ([article](https://developer.chrome.com/docs/capabilities/web-apis/file-system-access#storing_file_handles_or_directory_handles_in_indexeddb))

It's _absolutely_ not the case that PII should not be stored in IDB. That would render it useless for many apps.

evanstade (commented):
thanks for the PR!!

PR #940, opened 2026-06-06
ci: cache huggingface models and resolve esbuild bundling exports

paulirish:
to fix the 429 from huggingface that we had on main branch: https://github.com/GoogleChrome/modern-web-guidance-src/actions/runs/27047006843/job/79834868925


micahjo7 (approved):

paulirish:
Thx. 

It was to fix some 429's that broke CI. But CI is green again...

I'm gonna close this for now and hopefully we dont run into these 429s again. Would love to avoid the complexity. We'll see.

PR #942, opened 2026-06-08
weekly publish, enable OIDC

connorjclark:


paulirish:
nice.  

can we also publish manually using the same workflow?  i feel like we ran into bugs when we did workflow-driven plus manual but..

paulirish (approved):

connorjclark:
yes. `workflow_dispatch: {}` is still here

PR #950, opened 2026-06-11
Dashboard extensions

psiarkiewicz:
Add a dashboard showing individual guides and their performance over time sorted by tool-model.
Add a guides aggregate uplift score to the uplift panel

### DIFF
--- a/eval-view/api.js
+++ b/eval-view/api.js
@@ -1,4 +1,4 @@
-import { authenticatedFetch } from './utils.js';
+import { authenticatedFetch, parseResultKey } from './utils.js';
 
 export class ApiClient {
     constructor() {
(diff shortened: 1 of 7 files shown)


psiarkiewicz:
Add individual guides uplift score across time.

micahjo7:
When loading on all the remote data, one of the guides has just the name `task`, and I think this exists because of legacy runs that had a different format in the evals (basically, the older runs of the discipline evals had a flipped structure of naming `guide - task - runType` instead of `task - guide - runType`). Some special handling can be added so that this legacy run data is routed to the correct guide page instead of the `task` page.

Nits on display:

1. Over time as we scale the number of runs, the chronological run view could get very crowded. Should we just show the latest 50 or so? or limit the number on initial display and make it scrollable? Or maybe you already have some code to handle large amounts of runs, haven't looked closely at the code yet.

referring to this view: image

2. In the same view, when the guided does worse than the unguided (negative uplift), can we indicate this with a red line, similar to the existing uplift charts?

psiarkiewicz:
fixes made. Also, changed the "guided" cirlces to arrows to make it clearer which way they are going, and thickened the line.

connorjclark (approved):

PR #952, opened 2026-06-12
Fix devtools times grading, and "Diff"/"App" artifacts in dashboard

micahjo7:
The `autofill-sign-in-form` consistently fails in our evals (b/c it uses devtools-times app), 0% unguided and guided, because of the error:

```
Error: pnpm build failed in .../modern-web-guidance-src/harness/results/test-2026-06-12T09-59-48/1/autofill-sign-in-form/task/guided

   at test-fixture.ts:59

  57 |       });
  58 |       if (buildResult.status !== 0) {
> 59 |         throw new Error(`pnpm build failed in ${targetDir}`);
     |               ^
  60 |       }

...

> devtools-times@0.0.1 build .../modern-web-guidance-src/harness/results/test-2026-06-12T09-59-48/1/autofill-sign-in-form/task/guided
> astro build

sh: astro: command not found
 ELIFECYCLE  Command failed.
 WARN   Local package.json exists, but node_modules missing, did you mean to install?
```

also when clicking on the "App" or "Diff" artifacts for this task, we get an error like:

```
Error loading diff: File not found (404).

404 Not Found for http://localhost:8081/test-2026-06-12T09-59-48/1/autofill-sign-in-form/task/guided/index.html?source=local
```

So this PR fixes by running `pnpm install` for the first issue and fixing the entrypoint search for the dashboard artifacts.
Validated fixes in a test run (grading executes successfully).

# Fix Evaluation Grading for Framework Apps & Move Entrypoint Search to Build Phase

This PR ensures full-framework base applications (such as Astro/DevTools Times) install dependencies and evaluate correctly in Playwright, while migrating dynamic target file calculation to the evaluation collection phase.

### 1. Evaluation Setup & Dependency Resolution
* **[test-fixture.ts](file:///Users/micahjo/modern-web-guidance-src/guides/test-fixture.ts)**: Added a synchronous `pnpm install` step prior to `pnpm run build` / `pnpm run start`.
* Enforced the `--ignore-workspace` flag across all fixture `pnpm` commands so evaluation task folders are treated as standalone projects rather than defaulting to monorepo workspace scope.

### 2. Move Target File Calculation to Evaluation Compilation
* **[metrics.ts](file:///Users/micahjo/modern-web-guidance-src/harness/lib/metrics.ts)**: Added `targetFile?: string;` to the [RunResult](file:///Users/micahjo/modern-web-guidance-src/harness/lib/metrics.ts#L8-L21) TypeScript interface to establish formal contract verification in manifests.
* **[collection.ts](file:///Users/micahjo/modern-web-guidance-src/harness/lib/collection.ts)**: Integrated target file resolution directly into the results collector. Recursively parses evaluation source files, correlates them against `prompt` and `chat_log.txt`, and stores `targetFile` permanently in `evals.json`.

### 3. Zero-Latency Dashboard Rendering
* **[api.js](file:///Users/micahjo/modern-web-guidance-src/eval-view/api.js)**: Replaced on-the-fly client-side entry point calculation by directly consuming `run.targetFile` from memory. Preserved classic candidate matching fallback purely for historical evaluation backwards compatibility.

## Verification
* Confirmed 100% test pass rate in Playwright evaluation grading (`pnpm preflight`).
* Verified **App** and **Diff** tabs render correctly in the dashboard UI without 404 errors across static and modern framework apps.

### DIFF
--- a/guides/test-fixture.ts
+++ b/guides/test-fixture.ts
@@ -49,6 +49,7 @@ export const test = base.extend<{}, ServerWorkerFixtures>({
 
     const pkgJson = JSON.parse(fs.readFileSync(pkgJsonPath, 'utf8'));
 
+    // Running install on base app
     console.log(`[TEST-FIXTURE] Running pnpm install in ${targetDir}`);
     const installResult = spawnSync('pnpm', ['--ignore-workspace', 'install'], {
       cwd: targetDir,
--- a/guides/test-fixture.ts
+++ b/guides/test-fixture.ts
@@ -49,8 +49,18 @@ export const test = base.extend<{}, ServerWorkerFixtures>({
 
     const pkgJson = JSON.parse(fs.readFileSync(pkgJsonPath, 'utf8'));
 
+    console.log(`[TEST-FIXTURE] Running pnpm install in ${targetDir}`);
+    const installResult = spawnSync('pnpm', ['--ignore-workspace', 'install'], {
+      cwd: targetDir,
+      stdio: 'ignore',
+      shell: process.platform === 'win32'
+    });
+    if (installResult.status !== 0) {
+      throw new Error(`pnpm install failed in ${targetDir}`);
+    }
+
     if (pkgJson.scripts && pkgJson.scripts.build) {
-      const buildResult = spawnSync('pnpm', ['run', 'build'], {
+      const buildResult = spawnSync('pnpm', ['--ignore-workspace', 'run', 'build'], {
         cwd: targetDir,
         stdio: 'ignore',
         shell: process.platform === 'win32'
@@ -66,7 +76,7 @@ export const test = base.extend<{}, ServerWorkerFixtures>({
     }
 
     const port = await getFreePort();
-    const serverProcess = spawn('pnpm', ['run', 'start'], {
+    const serverProcess = spawn('pnpm', ['--ignore-workspace', 'run', 'start'], {
       cwd: targetDir,
       env: { ...process.env, PORT: port.toString() },
       detached: true,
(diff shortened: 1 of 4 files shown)


harness/lib/collection.ts
```diff
@@ -305,6 +305,56 @@ export async function collectResults(resultsDir: string, suiteConfig: SuiteConfi
         }
       }
 
+      let targetModifiedFile: string | undefined = undefined;
```
paulirish: this parent function is getting huge can we break up some of this (new) stuff into one or a few methods?

oh boy yeah this fn is huge.
I'd be really happy to see it refactored as well but the larger refactoring should probably happen in a follow-up.

guides/test-fixture.ts
```diff
@@ -49,8 +49,18 @@ export const test = base.extend<{}, ServerWorkerFixtures>({
 
     const pkgJson = JSON.parse(fs.readFileSync(pkgJsonPath, 'utf8'));
 
+    console.log(`[TEST-FIXTURE] Running pnpm install in ${targetDir}`);
```
paulirish, line 53:
```suggestion
    // Running install on base app
    console.log(`[TEST-FIXTURE] Running pnpm install in ${targetDir}`);
```

is that an accurate comment?

paulirish (approved):

guides/test-fixture.ts
micahjo7: updated, thanks!

harness/lib/collection.ts
micahjo7: added helper methods for the added code, will work on a refactor in another PR

PR #956, opened 2026-06-16
[#359] Add identify-bfcache-blocking-issues use case

### LINKED ISSUE #359: Add bfcache blocking reasons guide
### web-feature-id

bfcache-blocking-reasons

### Feature description

Back/forward cache (or bfcache) is a browser optimization that enables instant back and forward navigation.

### Reference sources

- https://web.dev/articles/bfcache
- https://github.com/rakina/bfcache-explainer

jamesnw:
Addresses #359. 

I have doubts that this meets the criteria for a use case. I don't think bfcache or back/forward cache are in the developer vernacular, so this is unlikely to surface. If it does surface, the demo essentially replicates what is available in dev tools. 

As far as code that the user would deploy, they may want to integrate this with analytics to track why it's happening for their users, in which case there wouldn't be a UI at all.

I'd welcome ideas on how to frame this as a use case, but my leaning at this point is to not have a use case for this feature.


rviscomi:
@tunetheweb do you have any thoughts on creating a guide for this, adding it to the Perf skill, or not having any guidance at all?

tunetheweb:
I tend to agree with @jamesnw ‘s concern. This is a RUM tooling API and its use beyond that is negligible. That might change if other browsers implement with without implementing an equivalent of the Chrome DevTools tool but even then it’s questionable.

So, given that few developers are building their own RUM solutions and so this API is very niche and aimed at a very specific set of developers, I guess the real question is what the user base and aim of MWG is? Is is aiming to be a fully comprehensive guide of all web platform APIs? Or addressing the general use case?

I think, at least for now, it’s more the latter right? In which case I wouldn’t add a specific guide for this. Maybe it’s worthy of a mention of the general
Performance guide (as we agreed recently im going to make some time
to take over that and re-review it).

gilbertococchi:
Thanks everyone, my motivation to add something about bfcache somehow here is definitively because it's back/forward cache is so far from developers's radar, although it impacts all browser.

I am not suggesting adding guidance about specific blocking reasons that vary for support across browser, but perhaps mentioning that usage of `cache-control: no-store` could severely harm user experiences, especially on pages where this header is not necessary would be greatly appreciated (https://web.dev/articles/bfcache#minimize-no-store) within the Performance Skill.

There are still 2.3M sites using `cache-control: no-store` and for most of these pages it's not stricly required but more legacy caching patterns hard to get rid off.

rviscomi:
Deferring to @tunetheweb and @malchata as perf ATLs

malchata (approved):
This tracks for me based on what I know about bfcache, so LGTM. However, I would defer to @tunetheweb in case he might spot anything out of whack.

gilbertococchi:
I really like the idea that @tunetheweb on https://github.com/GoogleChrome/modern-web-guidance-src/pull/956#issuecomment-4771128871 about mentioning this best practice in the high level performance skills file [here](https://github.com/GoogleChrome/modern-web-guidance-src/blob/main/guides/performance/performance/guide.md).

Perhaps it's worth adding a "DO" point about raise a warning about the usage of no-store that may be unintentional.

Unfortunately from HTTP Archive stats the usage of no-store is not going down (approx 16,8% of HTTP Archive Origins using no-store), with may sites using it when not strictly necessary and on some sites (Ecommerce and Listing pages in particular) this could account for 50% of the CrUX Navigation Type associated traffic.

jamesnw:
I removed this stub, and added a note to the Perf guide. It is under the `Service Workers & Caching Strategies` header, which was the most relevant section. If a full Caching section should be added, I'd recommend a different SME for that task.

gilbertococchi:
Thanks for this @jamesnw, LGTM, really appreciate the note about no-store.

@tunetheweb can you please review?

tunetheweb (approved):
LGTM!

PR #958, opened 2026-06-18
[#669] Declarative partial update use case

### LINKED ISSUE #669: Declarative partial updates
### web-feature-id

TBD

### Feature description

Stream HTML content out of order, and update an existing document with a stream of encoded "patches" that modify its existing content.

### Expected release date

Chrome 148, mentioned at I/O

### Coding agent assessment

_No response_

### Reference sources

- https://chromestatus.com/feature/5111042975465472

### LINKED ISSUE #1333: textStream() for response/request/blob
Add textStream() to interfaces that represent a byte stream (request, response, blob). This is equivalent to piping the byte stream through a TextDecoderStream().

---
**Chromestatus**: https://chromestatus.com/feature/5146752165478400
**Web Feature ID**: N/A
**Chrome Releases**: Chrome 151

jamesnw:
A full use case here would require a server streaming HTML slowly, which seems out of scope for this project. Would it suffice to have the demo just show the fully loaded HTML, which would still demonstrate the template replacement? The demo.html here copied from the official example shows what I mean. The guide could mention the server needs but not include code for the server side.

I suspect this won't pass CI without a web feature.

Also, should this be in performance rather than UX?


guides/user-experience/streaming-html-updates/demo.html
```diff
@@ -0,0 +1,167 @@
...
+  </template>
+
+  <template for="photo4">
+    <img src="https://picsum.photos/id/25/300/200" alt="Plant">
+    <p>Green Leaves</p>
+  </template>
+  </div>
+
+  <template for="status">
+    Loading complete
```
tunetheweb: This is not entirely accurate as the images are likely still being fetched by the time it reaches here.

Maybe this instead is more accurate (though perhaps harder to grok?):

```suggestion
    Your photo gallery
```

guides/user-experience/streaming-html-updates/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: streaming-html-updates
+description: Declaratively update a page with content streamed later into the same HTML document.
+web-feature-ids:
+  # https://github.com/web-platform-dx/web-features/issues/4117
```
tunetheweb: This is the incorrect link (though the two features are related!)

```suggestion
  # https://github.com/web-platform-dx/web-features/issues/4073
```

tunetheweb (changes requested):
See comment here regarding whether we should even add this yet:

https://github.com/GoogleChrome/modern-web-guidance-src/issues/669#issuecomment-4893417485

But maybe it's OK to add this simple use case? But I have a few comments on that first.

guides/user-experience/streaming-html-updates/demo.html
jamesnw: Good distinction 👍

rviscomi:
@jamesnw is this ready for another review?

jamesnw:
> @jamesnw is this ready for another review?

@rviscomi Yes, but the blocker is now [this question](https://github.com/GoogleChrome/modern-web-guidance-src/issues/669#issuecomment-4893417485), on whether it is too early to add this to MWG.

rviscomi:
> @rviscomi Yes, but the blocker is now [this question](https://github.com/GoogleChrome/modern-web-guidance-src/issues/669#issuecomment-4893417485), on whether it is too early to add this to MWG.

Ah ok, I missed that. I'll reply there. Thanks!

jamesnw:
This now is a higher level guide, as recommended [here](https://github.com/GoogleChrome/modern-web-guidance-src/issues/669#issuecomment-5257547193).

guides/performance/out-of-order-html-streaming/guide.md
```diff
@@ -0,0 +1,113 @@
...
+The feature consists of two primary mechanisms:
+
+1.  **Declarative Patching (HTML-based)**: Uses new processing instructions (`<?marker>`, `<?start>`, `<?end>`) and `<template for>` to swap content in the DOM as it streams in from the server.
+2.  **Imperative Streaming (JS-based)**: A suite of new methods (e.g., `streamHTML()`, `setHTMLUnsafe()`) that allow piping readable streams directly into DOM elements with native performance and optional script execution.
+
+## Best Practices
+
+*   **Prioritize Above-the-Fold Content**: Stream critical UI (header, hero, primary navigation, primary content) first. Use placeholders for "expensive" or lower priority components like comments, ads, or related content.
+*   **Implement Native Loading States**: Use the `<?start>` and `<?end>` markers to wrap fallback content. The browser automatically replaces everything between these markers when the corresponding `<template for>` arrives.
+*   **Minimize Layout Shift (CLS)**: Always provide stable dimensions for placeholders. If the final content size is unknown, use skeleton screens or CSS `aspect-ratio` to reserve space and prevent jarring shifts.
```
malchata: This might seem a little bit like a nit, but I think it would be a good thing for agents to know the exact metric name.

```suggestion
*   **Minimize Cumulative Layout Shift (CLS)**: Always provide stable dimensions for placeholders. If the final content size is unknown, use skeleton screens or CSS `aspect-ratio` to reserve space and prevent jarring shifts.
```

malchata (approved):

features/pending-web-features.json
```diff
@@ -7,5 +7,8 @@
   },
   "tmp-cross-origin-storage": {
     "issue": "https://github.com/web-platform-dx/web-features/issues/4029"
+  },
+  "tmp-declarative-partial-updates": {
+    "issue": "https://github.com/web-platform-dx/web-features/issues/4073"
```
tunetheweb: This is now a web-feature: https://github.com/web-platform-dx/web-features/blob/main/features/template-for.yml

Though the new HTML getters and setters will be a separate web feature it:
https://github.com/web-platform-dx/web-features/issues/4117

guides/performance/out-of-order-html-streaming/guide.md
```diff
@@ -0,0 +1,113 @@
+---
+name: out-of-order-html-streaming
+description: Declaratively update a page with content streamed later into the same HTML document or via imperative JavaScript APIs.
+web-feature-ids:
+  - tmp-declarative-partial-updates
```
tunetheweb:
```suggestion
  - template-for
  - html-setters
  - html-streaming-setters
  
```

Note the last two will fail for now until https://github.com/web-platform-dx/web-features/pull/4318 is merged.

guides/performance/out-of-order-html-streaming/guide.md
```diff
@@ -0,0 +1,113 @@
...
+---
+
+# Out-of-Order (OOO) HTML Streaming
+
+Out-of-Order (OOO) HTML Streaming (also known as Declarative Partial Updates) is an alternative to the traditional "top-to-bottom" linear processing of HTML. It allows developers to define placeholders earlier in a document and "patch" them with content that arrives later in the same network stream, or via new imperative JavaScript streaming APIs.
```
tunetheweb:
```suggestion
Out-of-Order (OOO) HTML Streaming (part of the Declarative Partial Updates initiative) is an alternative to the traditional "top-to-bottom" linear processing of HTML. It allows developers to define placeholders earlier in a document and "patch" them with content that arrives later in the same network stream, or via new imperative JavaScript streaming APIs.
```

guides/performance/out-of-order-html-streaming/guide.md
```diff
@@ -0,0 +1,113 @@
...
+
+IMPORTANT: This feature is currently in Early Preview. It requires Chrome 148 or higher with the `#enable-experimental-web-platform-features` flag enabled.
+
```
tunetheweb: I think we can skip this now. The first part (`<template for>`) has shipped and the second part (new setters and streaming setters) is due to ship in Chrome 154 at the end of this month.


```suggestion
```

guides/performance/out-of-order-html-streaming/guide.md
```diff
@@ -0,0 +1,113 @@
...
+## Quick Overview
+
+The feature consists of two primary mechanisms:
+
+1.  **Declarative Patching (HTML-based)**: Uses new processing instructions (`<?marker>`, `<?start>`, `<?end>`) and `<template for>` to swap content in the DOM as it streams in from the server.
+2.  **Imperative Streaming (JS-based)**: A suite of new methods (e.g., `streamHTML()`, `setHTMLUnsafe()`) that allow piping readable streams directly into DOM elements with native performance and optional script execution.
```
tunetheweb:
```suggestion
2.  **Imperative setting and streaming (JS-based)**: A suite of new methods (e.g., `streamHTML()`, `setHTMLUnsafe()`) that allow setting of HTML or piping readable streams directly into DOM elements with native performance and optional script execution.
```

guides/performance/out-of-order-html-streaming/guide.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Use Case Reference Matrix
+
+*   **Island Architecture**: Hydrate independent interactive components on top of static HTML without a full framework runtime.
+*   **Deferred "Expensive" Blocks**: Deliver the shell of a page immediately and stream database-heavy components as they become available.
+*   **Single Page Application (SPA) Updates**: Stream large UI transitions in SPAs using `streamHTML()` instead of waiting for a full JSON payload and manual DOM construction.
+
+## Anti-Patterns & Warnings (DO NOT DO THIS)
+
+*   **Do not use OOO streaming for everything.** Over-fragmenting a page can increase the complexity of the stream and potentially delay the "First Meaningful Paint" if the browser spends too much time patching small nodes.
```
tunetheweb: FMP is defunct and replaced with FCP. But let's generalise this anyway:

```suggestion
*   **Do not use OOO streaming for everything.** Over-fragmenting a page can increase the complexity of the stream and potentially delay content if the browser spends too much time patching small nodes.
```

guides/performance/out-of-order-html-streaming/guide.md
```diff
@@ -0,0 +1,113 @@
...
+*   **Do not ignore Trusted Types.** When using `*Unsafe` methods in environments with Trusted Types enforced, ensure you are passing a sanitized policy object if required.
+*   **Do not assume script execution order.** Scripts inside a `<template for>` will execute as soon as they are patched into the document. Do not rely on them executing in a specific sequence relative to other deferred templates.
+*   **Do not use without a polyfill strategy.** Ensure you have a fallback or use polyfills (`html-setters-polyfill` and `template-for-polyfill`) to support your required browsers.
```
tunetheweb:
```suggestion
*   **Do not use without a polyfill strategy.** Ensure you have a fallback or use polyfills (`html-setters-polyfill` and `template-for-polyfill`) to support your required browsers until this feature becomes Baseline.
```

guides/performance/out-of-order-html-streaming/guide.md
```diff
@@ -0,0 +1,113 @@
...
+*   **Do not ignore Trusted Types.** When using `*Unsafe` methods in environments with Trusted Types enforced, ensure you are passing a sanitized policy object if required.
+*   **Do not assume script execution order.** Scripts inside a `<template for>` will execute as soon as they are patched into the document. Do not rely on them executing in a specific sequence relative to other deferred templates.
+*   **Do not use without a polyfill strategy.** Ensure you have a fallback or use polyfills (`html-setters-polyfill` and `template-for-polyfill`) to support your required browsers. 
+
+## Implementation Status
+
+*   **Current Status**: Experimental / Early Preview.
+*   **Required Version**: Chromium 148+.
+*   **Activation**: Requires enabling the `chrome://flags/#enable-experimental-web-platform-features` or `edge://flags/#enable-experimental-web-platform-features`.
+*   **Specification**: [Declarative Partial Updates (Explainer)](https://github.com/WICG/declarative-partial-updates).
```
tunetheweb:
```suggestion
## Fallback strategies

{{ BASELINE_STATUS("template-for") }}
{{ BASELINE_STATUS("html-setters") }}
{{ BASELINE_STATUS("html-streaming-setters") }}

You can use polyfills like `template-for-polyfill` and `html-setters-polyfill` to emulate the API surface for non-supporting browsers
```

Or something like that.

tunetheweb (changes requested):
Apologies for the delay here and we'll wait for https://github.com/web-platform-dx/web-features/pull/4318/ but have some comments in the meantime.

features/pending-web-features.json
jamesnw: Updated to use the `tmp-` prefix on those.

guides/performance/out-of-order-html-streaming/guide.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Best Practices
+
+*   **Prioritize Above-the-Fold Content**: Stream critical UI (header, hero, primary navigation, primary content) first. Use placeholders for "expensive" or lower priority components like comments, ads, or related content.
+*   **Implement Native Loading States**: Use the `<?start>` and `<?end>` markers to wrap fallback content. The browser automatically replaces everything between these markers when the corresponding `<template for>` arrives.
+*   **Minimize Cumulative Layout Shift (CLS)**: Always provide stable dimensions for placeholders. If the final content size is unknown, use skeleton screens or CSS `aspect-ratio` to reserve space and prevent jarring shifts.
+*   **Accessibility & Focus Management**: Be mindful that out-of-order updates can confuse screen reader users. Use `aria-live` regions for critical updates and ensure that focus isn't lost if a placeholder containing the active element is replaced.
+
+
```
tunetheweb, line 29:
```suggestion

```

guides/performance/out-of-order-html-streaming/guide.md
```diff
@@ -0,0 +1,115 @@
...
+*   **Do not use OOO streaming for everything.** Over-fragmenting a page can increase the complexity of the stream and potentially delay content if the browser spends too much time patching small nodes.
+*   **Do not ignore Trusted Types.** When using `*Unsafe` methods in environments with Trusted Types enforced, ensure you are passing a sanitized policy object if required.
+*   **Do not assume script execution order.** Scripts inside a `<template for>` will execute as soon as they are patched into the document. Do not rely on them executing in a specific sequence relative to other deferred templates.
+*   **Do not use without a polyfill strategy.** Ensure you have a fallback or use polyfills (`html-setters-polyfill` and `template-for-polyfill`) to support your required browsers until this feature becomes Baseline.
+
+## Fallback strategies
+
+{{ BASELINE_STATUS("template-for") }}
+{{ BASELINE_STATUS("html-setters") }}
+{{ BASELINE_STATUS("html-streaming-setters") }}
```
tunetheweb, line 113: These are failing for now.

tunetheweb (approved):
LGTM but since we've waited this long let's give it a few more days to see if we can resolve https://github.com/web-platform-dx/web-features/pull/4318 as that might need some more changes here.

guides/performance/out-of-order-html-streaming/guide.md
rviscomi: You could prefix these with `tmp-` to silence the CI checks, if you did want to merge these changes now. We'll automatically catch when the next version of web-features includes the feature IDs and remove the prefix. But it's also fine to wait for the new IDs to propagate.

jamesnw: That doesn't work in the `{{ BASELINE_STATUS() }}` macro- there's no check built in for the pending web features.

(It even fails if you comment out (`<!-- {{ BASELINE_STATUS("tmp-html-setters") }} -->`) the feature, although I guess that makes sense since macros can insert content into comments)

rviscomi: Oh interesting, that sounds like a bug. Let me know if you want to merge this as-is and this is the only blocker, then I can look into a fix. Otherwise the easiest thing would be to wait until the feature ID is published.

Alternatively you could do something like this, and in the next web-features update PR, we'll get a CI check telling us to remove the tmp prefix and we can resolve this TODO while we're in there.
```suggestion
<!-- TODO: Use macros
BASELINE_STATUS("html-setters")
BASELINE_STATUS("html-streaming-setters")
-->
```

tunetheweb: No rush to merge this. Especially as half of it isn’t available until 154 anyway. Though the fact you have to use polyfills for other browsers anyway means you _could_ use it now.

Anyway let’s five it a little longer.

jamesnw: This is released, so I've removed the tmp features. We'll need to wait until web-features gets updated to 3.39.0 on this repo.

jamesnw:
@tunetheweb I added an example of `textStream()` [here](https://github.com/GoogleChrome/modern-web-guidance-src/pull/958/changes/870b9d321918f4b3701f6842e3ef4b291b8b81b5#diff-b16c56353ceb9bccdc224250d67b6c13ae69946d772263d27261dff945d69f42R93), which would close #1333. I don't think that needs a separate use case.

tunetheweb:
> @tunetheweb I added an example of textStream() [here](https://github.com/GoogleChrome/modern-web-guidance-src/pull/958/changes/870b9d321918f4b3701f6842e3ef4b291b8b81b5#diff-b16c56353ceb9bccdc224250d67b6c13ae69946d772263d27261dff945d69f42R93), which would close https://github.com/GoogleChrome/modern-web-guidance-src/issues/1333. I don't think that needs a separate use case.

Agreed.

guides/performance/out-of-order-html-streaming/guide.md
rviscomi: Just merged https://github.com/GoogleChrome/modern-web-guidance-src/pull/1468 bumping our version to 3.39.0

rviscomi:
This is ready to merge. (This is initially written as a discipline guide, rather than a use case, without demo or expectation files.)

PR #960, opened 2026-06-21
Custom Select Picker Layouts v2

LeaVerou:
This expands the content of [this guide](https://github.com/GoogleChrome/modern-web-guidance-src/blob/main/guides/forms/custom-select-picker-layouts/guide.md) and addresses some of the issues I flagged here:
- https://github.com/GoogleChrome/modern-web-guidance-src/issues/290#issuecomment-4340849140
- https://github.com/GoogleChrome/modern-web-guidance-src/issues/290#issuecomment-4566821883

Changes related to the review issues:

- Add `:open` to code snippets and explicitly instruct agents to respect it (current guidance produced broken UI)
- Edited a11y guidance to instruct agents to fix spatial navigation instead of giving up on customizable select altogether. Ideally this needs to be reviewed by an a11y SME, but I'd be surprised if "reimplement the control with divs" is actually a better direction than fixing the keyboard navigation. I avoided providing an explicit snippet since it proved very hard to write a truly generic one, whereas agents knowing the specifics of their use case can write one that works for their case far more easily.

Other changes:
- Consolidate guidance for each step previously spread across three sections.
- Broke down the code example into separate smaller code snippets as needed to illustrate each step
- Deduplicate guidance previously mentioned in multiple places.
- Moved the guidance to use `value` attributes on options to the general best practices for customizable select.
- Add guidance for styling the checkmark, which was previously left unstyled and occupied space even on non-selected cards, edited the demo as well. Before and after below.
image
image


LeaVerou:
@rviscomi ping!

rviscomi:
@knowler are you able to review this one? There's also an accessibility section that could use a closer look.

guides/forms/custom-select-picker-layouts/guide.md
```diff
@@ -7,85 +7,140 @@ web-feature-ids:
...
+
+### Best practices
+
+{{ FEATURE("customizable-select", "usage") }}
+
+## Accessibility
+
+<!-- NOTE This needs a11y SME review! -->
+
+A native `<select>` only handles linear Up/Down arrow navigation, so Left/Right do nothing in the open picker and Up/Down still navigate items sequentially, regardless of rows and columns.
```
knowler, line 138: I’m unsure about this.

As much as the added keyboard navigation helps clear up part of [the functionality mismatch](https://tink.uk/perceived-affordances-and-the-functionality-mismatch/), I don’t think it would totally match the expectation of a keyboard user for this kind of widget:

- The direction the focus flows in would need to be columns and not rows (otherwise up/down would go sideways).
- Then focus wrapping from the end of one column to the next wouldn’t be expected.

Implementing the keyboard controls exactly how a sighted keyboard user would expect would in turn cause a mismatch for screen reader users since it would stop working according to what they can perceive. It just seems like better guidance to always have the visual presentation match the semantics and its expected functionality, and in this case that would be a `grid` for the `::picker(select)`, but that’s not possible at the moment.

This concern for use cases like this did seem to come up in discussions about customizable select and prior art:

- https://www.w3.org/2025/08/14-openui-minutes.html#892b
- https://www.w3.org/2022/08/04-openui-minutes.html#t02
- https://github.com/openui/open-ui/issues/540

- - -

I do _really_ wish that it was easier to create adaptive interfaces where the semantics could be driven by styling information (e.g. the amount of columns/rows could be inferred) as I run into this situation frequently. I kind of wonder if this could be something [focusgroup v2](https://open-ui.org/components/focusgroup-v2.explainer/) could handle?

```html
<!-- Hypothetical example where `focusgroup="grid layout"` makes it a grid with inference based on the CSS layout -->
<select focusgroup="grid layout"><!-- `::picker()` becomes a sort of reference target for `focusgroup` and receives `grid` minimum role + functionality -->
  <!-- columns/rows are inferred from the visual layout — perhaps it’d be another `focusgroup` -->
  <option>Peanut Butter</option><!-- receives `gridcell` minimum role -->
  <option>Jelly</option>
  <!-- and so on… -->
</select>
```

knowler: I’m digging into this a bit more. Grid semantics don’t seem to be correct either (unless they’re actually meaningful to the content).

LeaVerou: Ooh that's a good point. I think there are two different needs here:
1. For sighted keyboard users, you'd _always_ need to match the grid, otherwise it looks _super_ weird
2. For non-sighted users, you'd want spatial when the grid is meaningful, but otherwise source is fine.

Do you think that makes sense, @knowler ?

knowler: The semantics of the contents seem to be what makes the difference. For cases where there is a semantic grid structure, unfortunately `<select>` doesn't seem to be quite there yet.

For cases where there is no semantic grid structure, I think applying the keyboard shortcuts could be ok as long as there aren't any conflicts with other user expectations for shortcuts (e.g. horizontal scrolling or arrow key navigation for different screen reader modes — though select should be in a form/application mode which passes the keys to the browser). I'm going to do some testing to check there aren't conflicts with screen readers. It could also be a part of the guidance to check for since we don't totally know if shortcuts could already exist in their app or for their users.

Also, I did some testing with focus wrapping at the end of a row and that doesn't seem unexpected. I do think there's a bit of a conundrum with the directionality of the flow of the content (e.g. up/down will always linearly go through the list in order, so the flow needs to be by column).

LeaVerou: > The semantics of the contents seem to be what makes the difference. For cases where there is a semantic grid structure, unfortunately `<select>` doesn't seem to be quite there yet.

Curious why?

knowler: I wrote this comment, but then realized that you might have been referring to why I made a concession for the non-semantic case vs. the semantic one? The semantic one could suffer form content loss if we were to use a `<select>` for it, as opposed to more appropriate structures that could express the full semantics of the contents (maybe a combobox with a grid popup).

- - -

With regard to why `<select>` can’t have true grid semantics applied to it. Here’s composed structure:

- `<select>` element (can change the role of this)
  - `::picker()` pseudo-element (can’t change the role of this)
    - `<option>` elements (slotted; can change the role of these)

The `::picker()` pseudo-element is the thing that’d need to be the grid, but there’s no way to change the semantics of that. That’s kind of why I though maybe there could be potential proposal for using `focusgroup` attribute to target the element where the `::picker()` could be like reference target for the effects of its minimum role assignment. I brought this up in the `#focusgroup` channel in the Open UI Discord, but it did not really seem favourable.

The semantics of select are messy:

- [HTML AAM says the select element should be a `combobox`](https://w3c.github.io/aria/html-aam/#el-select-combobox) and provides no prescription for the picker/dropdown itself (which is fair since it is _HTML_ AAM).
- `listbox` is the implied role for the picker.
- For the `<select>` element:
  - Blink sets the role to `combobox`
  - Gecko sets the role to `combobox`
  - Webkit sets the role to `button (popup, default)`
- For the picker:
  - Blink sets the role to `MenuListPopup`.
  - Gecko sets the role to `listbox`.
  - WebKit sets the role to `menu`.
    - Not inspectable with Safari 26, but it is Safari Tech Preview.
  - On macOS, regardless of the browser’s ARIA mapping, it will use a `menu` with `menuitem`s. The select button is presented as a “popup button.”
    - iOS (different accessibility API) uses similar semantics.
  - On Windows, the picker is presented as a list(box) and the select button as a `combobox` (matching spec).


In either case, both `menu` and `listbox` are one dimensional.

An older article on the topic of `<select>` element semantics (though what I’ve listed above are from my own investigation of specs and testing): [“`<select>` Your Poison” on 24 Accessibility by Sarah Higley](https://www.24a11y.com/2019/select-your-poison/).

- - -

While writing this, I did actually have an idea that might be able to a workaround (i.e. not a semantic grid, but perhaps equitable in some scenarios):

```html
<label for="nhl-draft-pick">NHL First Round Draft Pick</label>
<select id="nhl-draft-pick">
  <optgroup label="2026">
    <option value="2026-tor">Toronto Maple Leafs</option>
    <option value="2026-sjs">San Jose Sharks</option>
    <option value="2026-van">Vancouver Canucks</option>
  </optgroup>
  <optgroup label="2025">
    <option value="2025-nyi">New York Islanders</option>
    <option value="2025-sjs">San Jose Sharks</option>
    <option value="2025-chi">Chicago Blackhawks</option>
  </optgroup>
  <optgroup label="2024">
    <option value="2024-sjs">San Jose Sharks</option>
    <option value="2024-chi">Chicago Blackhawks</option>
    <option value="2024-ana">Anaheim Mighty Ducks</option>
  </optgroup>
</select>
```

The idea there was that the `<optgroup>` could be used as the columns. Safari (Tech Preview) seems to have issues here which I think could have to do with `display: grid` causing the `<optgroup>` element to lose their roles (which means you have to set `role=group aria-label="<Year>"` — I’ll find or file this bug).

LeaVerou: I'll add that to the mounting pile of reasons why we need `role` in CSS 😀 
But meanwhile, can't we just use a wrapper with the right role?

The optgroup idea works well for some grids (semantic and predefined) but not those that are meant to adapt to the picker size. But then again, those are less likely to be semantic, so perhaps that's fine.

Question still stands: for those cases where sequential is fine for non-sighted users (i.e. the grid is presentational), how do we fix it for sighted users without breaking it for non-sighted users?

knowler: > But meanwhile, can't we just use a wrapper with the right role?

While I was surprised by how well this worked when [I demo’d it](https://knowler.dev/demos/t4RgQK3?codepen) and tested it with screen readers (VO on macOS, NVDA and Narrator on Windows—all in multiple browsers, though the Firefox Nightly customizable selects are currently completely broken for AT), there were some flaws which would make it inadvisable.

1. The `::picker(select)` pseudo-element would still be an [ARIA `menu`](https://w3c.github.io/aria/#menu) on macOS and [ARIA `listbox`](https://w3c.github.io/aria/#listbox) elsewhere. The necessary roles for a grid (i.e. `grid`, `row`, `gridcell`) are invalid as accessible children of either a `menu` or `listbox`.
    - Knowing that browsers often erroneously strip semantics when `display: contents` are applied, I tried adding this and did successfully cause Safari Tech Preview to ignore the pseudo-element’s semantics. Unfortunately, this is a bug that should be reported.
    - On Windows, the listbox semantics were still announced in NVDA even though it also announced the grid/table semantics (e.g. it would announce a list before entering the grid).
2. `<option>` is [not allowed to have its role overridden](https://w3c.github.io/html-aria/#el-option) to something other than `option` which itself is not recommended.

I was doing some research around potential keyboard issues and I found [a very insightful comment from Sarah on Fluent UI that ultimately resolved to not add data grids as popups for comboboxes](https://redirect.github.com/microsoft/fluentui/issues/31040#issuecomment-2159247799). That might not seem totally relevant, but looking ahead to [“filterable selects”](https://open-ui.org/components/filterable-select.explainer/#proposed-approach) which is being prototyped there would be a keyboard conflict here as the up/down arrow keys would move between the options in the select, while the right/left arrow keys would typically move the cursor. Considering this actually makes me question adding the horizontal controls for either use case (presentational or semantic). 

I think that “don’t do this” with `<select>` might be the best guidance at least for now as I think we’d be building on a rough patch spec-wise. It’s really interesting as an experiment with findings that could be brought back to standards groups, but not really a good fit for guidance or as a best practice.

A safer and more mature pattern would be to have a button open a popover/dialog with an interactive grid inside. Focusgroup v2 will improve upon that pattern by making it very easy to make a table or any custom grid markup. That doesn’t totally solve the problem of form participation (a hidden checkbox/radio or FACE would likely be necessary there).

LeaVerou: I think we can _recommend against_ something, but we should still have guidance for the best way to do it. **"Don't do this" doesn't work in practice — they will just do it badly** (this applies way more broadly than technology…)

Some questions:
- Why would we want to override the role of `<option>`s? They are still options.
- I agree on not depending on `display: contents` being broken. But why is it a problem to have a grid nested in a menu? I'm no a11y expert, but to me a grid inside a menu seems _exactly_ the right semantics? Is it just an artificial ARIA restriction or does it actually break something for AT? If it's the former, could it be an ARIA bug?

> A safer and more mature pattern would be to have a button open a popover/dialog with an interactive grid inside. Focusgroup v2 will improve upon that pattern by making it very easy to make a table or any custom grid markup. That doesn’t totally solve the problem of form participation (a hidden checkbox/radio or FACE would likely be necessary there).

This is tremendously more complex 😞 
I think if I saw my agent doing that I'd just assume it didn't know about customizable select…

You said a wrapper with a role worked well in your testing. Is it just a matter of conformance then?

knowler: > But why is it a problem to have a grid nested in a menu? I'm no a11y expert, but to me a grid inside a menu seems exactly the right semantics? 

Unfortunately, there are a few layers of meaning here.<sup>1</sup> ARIA has a very specific meaning for [“menu”](https://w3c.github.io/aria/#menu) that comes with certain expectations about what it contains, how it behaves, and what it gets mapped to in various operating systems (see [Core AAM](https://w3c.github.io/aria/core-aam/) — various accessibility APIs even have their own meaning for things too). An ARIA menu could be substitutable with an [ARIA grid](https://w3c.github.io/aria/#grid) or, as we’ve seen, an [ARIA listbox](https://w3c.github.io/aria/#listbox) — all are [“composite widgets”](https://w3c.github.io/aria/#composite) — but we can’t place a grid within a menu as a direct descendant, since browsers will be inconsistent with how they map that to accessibility APIs and it more than likely will create an inconsistent experience for assistive tech users (as I noted in my findings for the previous demo).

ARIA has a concept of [a “popup element”](https://w3c.github.io/aria/#aria-haspopup:~:text=A%20popup%20element,-usually) which is likely closer to what we might mean when we say “menu” or “dropdown.” In this case, it would be the element that has `popover` set on it — the `::picker()` pseudo-element. A “popup element” would need to have a role like `menu`, `listbox`, `grid`, etc. Then the element that has the “popup element” would have [`aria-haspopup`](https://w3c.github.io/aria/#aria-haspopup) set to whatever role it is that the “popup element” is.

Even with the increased customization of a `<select>` element, we can’t change the semantics of the `::picker()` pseudo-element.

> You said a wrapper with a role worked well in your testing. Is it just a matter of conformance then?

It worked better than I thought it would, but it still was confusing hearing the “menu” or “list” announced. That’s the unspec’d territory from using non-conformant ARIA.

> This is tremendously more complex 😞
> I think if I saw my agent doing that I'd just assume it didn't know about customizable select…

ya, I agree it’s not very nice. There’s no way to do this in an ARIA conformant way that doesn’t involve using a separate popover instead of the `::picker()` pseudo-element.

I did explore how we could keep using the `<select>` element, but use a custom popover that has the correct semantics. That way we could keep the form association benefit and half of the customizable styling benefit (though using our own popover gives us the same styling benefit, except we just need to target different selectors).

Here’s an imperfect demo that achieves the correct semantics, behaviour, and keeps the form association from the built-in `<select>` (imperfect in the sense that I’m lazy in how I’m preventing the default behaviour of the built-in picker): https://knowler.dev/demos/jEVVbK4?codepen

```html
<label id="status-label" for="status">Status</label>
<select id="status" name="status" aria-haspopup="grid">
  <optgroup label="Open">
    <option>To Do</option>
    <option>Backlog</option>
  </optgroup>
  <optgroup label="Active">
    <option>In Progress</option>
    <option>In Review</option>
    <option>Blocked</option>
  </optgroup>
  <optgroup label="Closed">
    <option>Done</option>
    <option>Won’t Fix</option>
    <option>Duplicate</option>
  </optgroup>
</select>
<!-- Maybe this is generated by whatever we’re using to enhance the select -->
<div focusgroup="grid manual" popover aria-labelledby="status-label">
  <!-- Fill in grid contents based on the select… omitted here, inspect the demo -->
</div>
```

`focusgroup="grid manual"` is a part of [the version 2 work on focusgroup](https://open-ui.org/components/focusgroup-v2.explainer/). It applies the grid behaviour and semantics to the popover above. It’s [being prototyped in Chromium currently](https://groups.google.com/a/chromium.org/g/blink-dev/c/a4VM8AFKHy8/m/i-zQm38yBwAJ). I’m leveraging an unreleased branch in the polyfill in this demo which applies both the focusgroup semantics and behaviour. There are a number of workarounds I’m using to make the polyfill work (e.g. it doesn’t seem to like `popover` on the focusgroup initially or the ARIA `rowgroup` elements).

`aria-selected` on the ARIA `gridcell` elements doesn’t work anywhere except with Narrator and JAWS (see: <https://a11ysupport.io/tests/aria_gridcell(aria-selected)>). We are allowed to place interactive content inside of the grid cells, so it’s possible to place a button with a relevant state there.

If customizable selects should support semantic grid pickers, then there should be a first class way of doing that (maybe `appearance: grid-select` or `<select focusgroup="grid">` which forwards to the internal picker). Otherwise, the guidance should be to avoid grid layouts inside of the picker.

I think the best approach to _just make it work_ correctly is the “bring your own grid popover” approach in my most recent demo, but unfortunately that comes with some sacrifices for developer experience.

- - -

<sup>1</sup> Adrian Roselli has a few articles about conflations in terminology of related topics (none of which I have perfectly followed):

- [“Be Careful Using Menu”](https://adrianroselli.com/2023/05/be-careful-using-menu.html)
- [“Stop Using ‘Pop-up’”](https://adrianroselli.com/2021/07/stop-using-pop-up.html)
- [“Stop Using ‘Drop-down’”](https://adrianroselli.com/2020/03/stop-using-drop-down.html)

LeaVerou: Using customizable select is not a goal in itself if it provides no benefit. This seems significantly more complex than _any_ solution that doesn't involve customizable select.

I pushed some edits based on our discussion. Is it better now?

I also realized I'm not sure I understood this comment:

> As much as the added keyboard navigation helps clear up part of [the functionality mismatch](https://tink.uk/perceived-affordances-and-the-functionality-mismatch/), I don’t think it would totally match the expectation of a keyboard user for this kind of widget:
> 
> - The direction the focus flows in would need to be columns and not rows (otherwise up/down would go sideways).
> - Then focus wrapping from the end of one column to the next wouldn’t be expected.

What focus wrapping? I don't think I recommend any wrapping?

rviscomi:
@knowler are you able to take another look at this? cc @jamesnw

guides/forms/custom-select-picker-layouts/guide.md
knowler: Sorry for the delay. I re-reviewed all of the guide with the latest feedback and I think it’s fine.

> What focus wrapping? I don't think I recommend any wrapping?

No, you didn’t recommend any wrapping — that was from how I inferred it should work.

Imagine a `<select>` with options that are alphabetically sorted (e.g. countries; states). A screen reader user might expect the up/down arrows to linearly move through the options which would follow that alphabetical order.

On “[t]he direction the focus flows in would need to be columns and not rows (otherwise up/down would go sideways)”:

If visually, the picker is presented as a grid that flows by row and the spatial navigation keyboard shortcuts were applied, then advancing down or up would no longer be alphabetical (or sorted any particular way) which could be disorienting to the screen reader user.

On “[t]hen focus wrapping from the end of one column to the next wouldn’t be expected”:

In order to ensure the `<select>` matched the expected behaviour of a screen reader user, the focus wrapping would be necessary to allow the user to linearly navigate the options. Initially, I thought that might have been weird for a keyboard user not using a screen reader, however, I did find grids in the wild where the focus did wrap (namely the Fluent UI SwatchColorPicker for React which is marked up as an ARIA grid and wraps at the end of a row to the start of the next row — in that case the grid semantics are applied though so the screen reader user has a different expectation of how it operates which better match the visuals).

So, maybe in summary:

- When the built-in `<select>` semantics are kept, but the picker’s visual layout becomes a grid, in order to avoid degrading the screen reader experience, the visual grid should flow in columns and the focus should wrap so that screen reader users can still linearly navigate the options in a meaningful order. In this way the spatial navigation is layered on as a progressive enhancement (i.e. the up/down functionality is preserved and it only adds right/left behaviour).
  - If a screen reader user was given sufficient instruction, it could be okay to flow a grid by row and apply the spatial navigation in a way that matched that layout (i.e. the instructions tell the screen reader user that the right/left arrow keys linearly advance through the options rather than up/down).
- For cases where the picker is both semantically and visually a grid, there’s more freedom to apply spatial navigation and no need for the focus wrapping since the screen reader user knows where they are within the grid.

knowler (approved):

LeaVerou:
@knowler I had to resolve the [conversation](https://github.com/GoogleChrome/modern-web-guidance-src/pull/960#discussion_r3733990535) to be able to merge the PR after you approved, but please record all this somewhere so it doesn't get lost! Comments in the issue, a comment in the guide (see #1497), a blog post, anything! I actually needed to cite your analysis more than once, and had nothing better to link to than this PR 😄

PR #753, opened 2026-05-10
add guidance for contrast-color

patrickkettner:
fixes #687

Issue #687, opened 2026-04-30
contrast-color()


rviscomi:
### web-feature-id

contrast-color

### Feature description

The contrast-color() CSS function picks a color that has guaranteed contrast against a specified foreground or background color.

### Expected release date

Chrome 147, mentioned at I/O

### Reference sources

- https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value/contrast-color

**Use cases:**
- [ ] #961

PR #966, opened 2026-06-24
add guide body for contrast-color

### LINKED ISSUE #961: Create guide and evals for the contrast-color use case
Select a readable text color with sufficient contrast on a dynamic background color.

Affected web-feature IDs: [contrast-color](https://webstatus.dev/features/contrast-color)

Use case subdir: [guides/visual-design/contrast-color](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/visual-design/contrast-color)

Related features: #687

**Required files:**
- [x] demo.html
- [x] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

patrickkettner:
Follow-up to #753, which merged `contrast-color` as a stub (step 1: frontmatter + demo.html). This adds the full `guide.md` body.

Restores the guidance written earlier and incorporates the use cases @jamesnw suggested in the #753 review:
- **Derive readable hover/interaction states** — change only the background custom property via relative color syntax and let `contrast-color()` re-derive the foreground ([algorithmic hover states](https://daverupert.com/2026/01/algorithmic-hover-states-with-contrast-color/)).
- **Build colors beyond black and white** — feed the `contrast-color()` result into relative color syntax, with an explicit accessibility caveat that the contrast guarantee no longer holds once the value is transformed.

Use cases: optimal text color, custom-property theming, hover/interaction states, derived colors, plus fallback strategy.

Refs #687


guides/css/contrast-color/guide.md
```diff
@@ -1,6 +1,134 @@
...
 name: contrast-color
-description: Select a readable text color with sufficient contrast on a dynamic background color.
+description: Select a readable text color with sufficient contrast on a dynamic background color.
 web-feature-ids:
   - contrast-color
 ---
+
+# Ensure text readability with dynamic background colors
+
+When building reusable UI components (like badges or buttons) or supporting dynamic themes, ensuring text is readable against an unpredictable background color can be challenging. The `contrast-color()` CSS function automatically selects either black or white to maximize the contrast ratio against a specified background color.
```
jamesnw:
```suggestion
When building reusable UI components (like badges or buttons) or supporting dynamic themes, ensuring text is readable against an unpredictable background color can be challenging. The `contrast-color()` CSS function resolves to `black` or `white`, maximizing the contrast ratio against the specified background color.
```

guides/css/contrast-color/guide.md
```diff
@@ -1,6 +1,134 @@
...
+
+This guide explains how to use `contrast-color()` to guarantee text legibility.
```
jamesnw:
```suggestion
```
I don't think this is needed.

guides/css/contrast-color/guide.md
```diff
@@ -1,6 +1,134 @@
...
+
+## Determine optimal text color
+
+When an element has a variable or dynamically injected background color, you can rely on the browser to compute the safest text color.
+
+MANDATORY: You must define a fallback strategy for older browsers, such as using an `@supports` query with a `text-shadow` or variant-specific colors, because a single static fallback color will inevitably fail against some dynamic backgrounds.
```
jamesnw: I don't think we need to talk about fallbacks in the main section of the guide

guides/css/contrast-color/guide.md
```diff
@@ -1,6 +1,134 @@
...
+
+MANDATORY: Set the `color` property using the `contrast-color()` function, passing the effective background color (or its custom property) as the argument.
```
jamesnw: I'm not sure what's meant by "effective" here- I think it can be removed?

guides/css/contrast-color/guide.md
```diff
@@ -1,6 +1,134 @@
...
+@supports (color: contrast-color(red)) {
+  .badge {
+    /* MANDATORY: Dynamic contrasting color */
+    color: contrast-color(var(--badge-bg));
+    text-shadow: none;
+  }
+}
+```
+
+DO: Use `contrast-color()` primarily when backgrounds are distinctly light or dark. Because the function currently only selects between black or white, mid-tone backgrounds (such as royal blue) may result in suboptimal contrast even when the mathematical maximum is chosen.
```
jamesnw: Rather than `mid-tone`, would it make sense to say lightness specifically?

guides/css/contrast-color/guide.md
```diff
@@ -1,6 +1,134 @@
...
+.button {
+  --button-bg: #4f46e5;
+  background-color: var(--button-bg);
+  color: contrast-color(var(--button-bg));
+}
+
+@supports (color: contrast-color(red)) {
+  .button:hover {
+    /* Darken the background; the text color re-derives on its own */
+    --button-bg: oklch(from #4f46e5 calc(l - 0.1) c h);
```
jamesnw: It might make sense to make this example actually flip the text from white to black- this should do it.

```suggestion
    --button-bg: oklch(from #4f46e5 calc(1 - l) c h);
```

guides/css/contrast-color/guide.md
```diff
@@ -1,6 +1,134 @@
...
+}
+```
+
+DO: Treat derived colors as a progressive enhancement on top of the guaranteed black/white value, reserving the unmodified `contrast-color()` output for text that must remain readable.
+
+## Fallback strategies
+
+{{ FEATURE_FALLBACKS("contrast-color") }}
+
+Due to its limited availability, `contrast-color()` must be applied using progressive enhancement. If the background color is known and fixed for a specific CSS class (like a `.badge-dark` variant), declaring a variant-specific safe default `color` immediately before the `contrast-color()` rule is sufficient, because browsers ignore CSS values they do not understand.
```
jamesnw: I don't think the prose needs to talk about availability.
```suggestion
For browsers that do not yet support `contrast-color()`, use it as a progressive enhancement. If the background color is known and fixed for a specific CSS class (like a `.badge-dark` variant), declare a variant-specific safe default `color` immediately before the `contrast-color()` rule.
```

guides/visual-design/contrast-color/guide.md
```diff
@@ -1,6 +1,134 @@
...
+
+DO: Treat derived colors as a progressive enhancement on top of the guaranteed black/white value, reserving the unmodified `contrast-color()` output for text that must remain readable.
+
+## Fallback strategies
+
+{{ FEATURE_FALLBACKS("contrast-color") }}
+
+Due to its limited availability, `contrast-color()` must be applied using progressive enhancement. If the background color is known and fixed for a specific CSS class (like a `.badge-dark` variant), declaring a variant-specific safe default `color` immediately before the `contrast-color()` rule is sufficient, because browsers ignore CSS values they do not understand.
+
+However, if the background color is highly dynamic and unpredictable (such as user-injected themes), a single static fallback will inevitably fail. In these cases, you MUST use an `@supports` feature query to apply a robust fallback strategy, such as a text shadow or translucent background, to guarantee readability.
```
jamesnw: We could also mention using a library like ColorJS to compute a contrast color in JavaScript.

Another option would be using RCS, but that isn't widely available yet. Perhaps we don't need to mention that. https://lea.verou.me/blog/2024/contrast-color/#using-rcs-to-automatically-compute-a-contrasting-text-color

jamesnw (changes requested):
This is looking good! I have a few suggestions, and I think the demo will need to be updated to more closely match the guide.

I think we should also add notes that caution against using `contrast-color()` over gradients, images, or backgrounds with transparency.

jamesnw:
I applied my changes, so this is ready for review from @rviscomi.

LeaVerou (changes requested):
I will look in more detail later but from a **very** quick glance these things stood out and wanted to post them earlier rather than later:
- P0: **Do NOT recommend `1 - l` as a fallback!** That produces _completely_ unreadable colors at 0.5, and very poor contrast near it. A while back, I wrote [a whole blog post](https://lea.verou.me/blog/2024/contrast-color/) about how to emulate `contrast-color()` if it's not available, which seems quite relevant here. The **tl;dr** is to use an L threshold and flip to white or black past it. Most of the post is about what that threshold should be, basically larger values work better for actual readability, but smaller values get you WCAG 2 compliance, which some people need to follow for legal reasons. I suggest we mention that in the guide text instead of just giving it a specific number) The code snippet would be even simpler today since we can rely on `abs(sign())` for branching instead of the `clamp( * infinity)` hack (we can recommend the hack as a fallback since `abs()` and `sign()` are Baseline 2025, not WA). 
- P0: **Please never, ever recommend people change `l` without adapting anything else.** Until browsers implement gamut mapping, setting `l` on colors you don't know can produce **completely** incorrect colors, in a way that affects contrast, especially towards the edges and especially towards white. E.g. suppose you start from a typical blue accent color like `oklch(0.5 0.2 240)` (which renders like `#006aab`). Setting `l` to `0.97` gives you something that is rendered like `#50ffff` instead of `#def9ff`, which has significantly different contrast characteristics, not to mention is a different color. At minimum they should cap `c` or multiply it with a factor that depends on the L delta.
- P1: We should not recommend they only use it when they have a distinctly light or dark color. That severely limits its utility. The whole point of the feature is that you can use it without having to care what you have.

jamesnw:
Thanks for the review!

> P0: **Do NOT recommend `1 - l` as a fallback!** That produces _completely_ unreadable colors at 0.5, and very poor contrast near it. A while back, I wrote [a whole blog post](https://lea.verou.me/blog/2024/contrast-color/) about how to emulate `contrast-color()` if it's not available, which seems quite relevant here. The **tl;dr** is to use an L threshold and flip to white or black past it. Most of the post is about what that threshold should be, basically larger values work better for actual readability, but smaller values get you WCAG 2 compliance, which some people need to follow for legal reasons. I suggest we mention that in the guide text instead of just giving it a specific number) The code snippet would be even simpler today since we can rely on `abs(sign())` for branching instead of the `clamp( * infinity)` hack (we can recommend the hack as a fallback since `abs()` and `sign()` are Baseline 2025, not WA).

This isn't being recommended as a fallback- it's being used alongside contrast-color, to demonstrate the need for re-computing `contrast-color` if the background color changes, in this case in a hover state. That said, I realized the example didn't match the prose, so I removed it anyways.

Separately, I've added the method from your post as an intermediate fallback between contrast-color and text-shadow.

> P0: **Please never, ever recommend people change `l` without adapting anything else.** Until browsers implement gamut mapping, setting `l` on colors you don't know can produce **completely** incorrect colors, especially towards the edges and especially towards white. E.g. suppose you start from a typical blue accent color like `oklch(0.5 0.2 240)` (which renders like `#006aab`). Setting `l` to `0.97` gives you something that is rendered like `#50ffff` instead of `#def9ff`, which has dramatically different contrast characteristics. At minimum they should cap `c` or multiply it with a factor that depends on the L delta.

I capped `c` at 0.4 - I think that gets you sufficiently close to gamut to prevent significant skew from clipping, even though some values may still be out of gamut.

>  P1: We should not recommend they only use it when they have a distinctly light or dark color. That severely limits its utility. The whole point of the feature is that you can use it without having to care what you have.

This was from the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/color_value/contrast-color#:~:text=Warning,function), but I think I missed the point (which is that for a color like #2277d3, neither [white](https://www.oddcontrast.com/#srgb__rgb(100~_100~_100~)__rgb(13.333~_46.667~_82.745~)) or [black](https://www.oddcontrast.com/#srgb__rgb(0~_0~_0~)__rgb(13.333~_46.667~_82.745~) provide sufficient contrast for AAA Normal text.) Does that deserve a mention?

guides/visual-design/contrast-color/guide.md
```diff
@@ -1,6 +1,168 @@
...
+DO: Treat derived colors as a progressive enhancement on top of the guaranteed black/white value, reserving the unmodified `contrast-color()` output for text that must remain readable.
+
+## Fallback strategies
+
+{{ FEATURE_FALLBACKS("contrast-color") }}
+
+For browsers that do not yet support `contrast-color()`, use it as a progressive enhancement. If the background color is known and fixed for a specific CSS class (like a `.badge-dark` variant), declare a variant-specific safe default `color` immediately before the `contrast-color()` rule.
+
+However, if the background color is highly dynamic and unpredictable (such as user-injected themes), a single static fallback will inevitably fail. In these cases, you MUST use an `@supports` feature query to apply a robust fallback strategy, such as a text shadow or translucent background, to guarantee readability.
+
```
LeaVerou, line 68:
```suggestion

{{ BASELINE_STATUS('relative-color') }}
```

LeaVerou:
Apologies in advance for the lack of detail, trying to quickly get this in before dinner!

**tl;dr:** I would suggest we split out the derived colors part into a separate guide, it's a much hairier problem and there's no reason to hold back the rest of the guide. Plus, there are many use cases that are fine with black and white. We can reference it from this guide when it's ready.

0.4 only guards against colors that *start* out of gamut. Most of the problems happen when an accent color is in-gamut, but we tweak L to get a lighter or darker tint and then it gets _wildly_ out of gamut, because as you move towards 0 or 1 the gamut shrinks dramatically.

As an example, this is the P3 gamut at L=0.5:
image

And this is the gamut at L=0.97 (a common lightness for backgrounds in light mode):
image

The blue color shown at the edge of the gamut has c=0.155.

The actual chroma to move the color in-gamut is **0.017**!
image

I’m afraid a good RCS formula to produce light/dark tints is an area of active research (one I've spent years trying to figure out 😔).
Here's a demo of the different methods and their results: https://apps.colorjs.io/tints/ (the power curve needs tweaking the sliders to get good results).

A safe (but suboptimal) option is to linearly move the color towards c=0 based on the difference between `l` and desired `l`. That's basically the same result as `color-mix()` would give you when mixing with white or black. It's suboptimal because that produces washed out, overly desaturated colors, but it's the lesser of two evils. But if we go that route we may as well do it with alpha, it produces the same (poor) results. Or via `color-mix()` with `contrast-color()` (possibly with a little `oklch(from contrast-color(...) l none none)` to counter the desaturation).

A power curve gets you better results, but you need different exponents for going darker vs going lighter, so it gets pretty complicated and I don't think this warrants this kind of complexity.

One silver lining is that the problem is much worse when going **lighter**, whereas going **darker** tends to be more acceptable, and typically what we want is dark tints on light backgrounds vs white text on darker backgrounds, which can be done with a relatively simple expression.

**Wait a second…** I just noticed the actual code example is:

```css
--callout-muted: oklch(from contrast-color(var(--callout-bg)) abs(l - .2) c h);
```

This is derived on `contrast-color()`, which is either white (L=1) or black (L=0) will produce either **a light gray or a dark gray**. I …don't think that's what was intended here? I had assumed we were trying to derive tints from the actual color provided. Why would we want to overlay grays on a potentially bright accent color? That's rarely readable…

You cannot produce an RCS formula that considers _both_ the contrasting color and the original color. For that, you'd need multi-color RCS, which is not even specced yet (we only have a CSS WG resolution).

For cases where we have a specific light/dark color we want to use, `if()` could provide a way (set a registered `<color>` property to `contrast-color()`, compare with `white` and branch accordingly).
Of course, then we need a fallback for `if()`…



> MANDATORY: Re-verify contrast whenever you transform the result of `contrast-color()`. As soon as you move away from pure black or white, the browser's contrast guarantee no longer holds, so you must confirm the derived color still meets your target ratio (for example WCAG AA) against the background.

How are they supposed to do that?

IIRC there are certain minimum deltas for L that guarantee certain WCAG levels, so that might be an avenue. But I'd need to compute them (and they tend to be overly aggressive since they need to work for any C/H).

---

Btw another fallback if RCS is not available is the old `filter` workaround: https://miunau.com/posts/dynamic-text-contrast-in-css/

jamesnw:
Apologies for the state this has gotten to as I've picked up revisions. Before I make more, I'd like feedback on this plan-

1. Open an issue for a new use case for derived colors. (Or not, because as you mention, this is not a solved problem)
2. Keep the "Derive readable hover and interaction states" section, and make sure that the focus is limited to "when you change the background color, use `contast-color()` on the new background to get a new foreground color that has sufficient contrast"
3. Remove the "Build colors beyond black and white" section. 

Does that seem like a reasonable next step?

LeaVerou:
> Apologies for the state this has gotten to as I've picked up revisions. Before I make more, I'd like feedback on this plan-
> 
> 1. Open an issue for a new use case for derived colors. (Or not, because as you mention, this is not a solved problem)

The use case is valid, so we should at least have the stub (which will create an issue).
I think even a guide would be useful, since agents *are* making poor decisions there (predictably) so anything we do will improve things. And if we use functions it can actually be quite elegant, it's just of limited utility in practice…

Logistical nit: Unfortunately, creating an issue manually doesn't quite mesh well with the tooling right now, the flow is PR creates issues and the issues need to have a specific format. That's why I opened #1005 so we'd have a low-fi place to dump ideas that doesn't go through PRs.
I _think_ even adding the stub to this PR would be fine and will create the issue once merged. It no longer requires a `demo.html` to add stubs.

> 2. Keep the "Derive readable hover and interaction states" section, and make sure that the focus is limited to "when you change the background color, use `contast-color()` on the new background to get a new foreground color that has sufficient contrast"

That makes sense. Or just use a CSS variable to drive both, then it adapts automatically.

> 3. Remove the "Build colors beyond black and white" section.
> 
> Does that seem like a reasonable next step?

Yup!

jamesnw:
@LeaVerou I think I've applied that review, and I generated a new `generate-derived-colors` stub on this PR as well.

guides/visual-design/contrast-color/guide.md
```diff
@@ -1,6 +1,136 @@
...
+  color: #fff; /* Default assumption */
+  text-shadow: 0 1px 3px rgb(0 0 0 / 0.8); /* Readability safety net */
+}
+
+/* Fallback using relative color syntax  */
+@supports (color: oklch(from red l c h)) {
+  .badge {
+    /* Highest threshold that passes WCAG. Higher values may be more legible. */
+    --threshold: 0.623;
+    --l: clamp(0, (l / var(--threshold) - 1) * -infinity, 1);
```
LeaVerou: Can we use `abs(sign())` and fall back on the infinity hack if they're not supported?

LeaVerou (changes requested):
I read the current version of the guide top to bottom. As a general point this feels like it's overexplaining certain things in a way that agents don't really need, but that's not a blocker in itself. 

### "Highly dynamic colors" is the primary use case, not an outlier

A lot of the text in this guide reads like it assumes you mostly know what color you have and the case where you could have almost anything is an outlier.

But then you don't need `contrast-color()` at all! Just specify black or white and be done with it! 

If either black or white text could work on your background then either you have a very wide range of possible colors, or they're all midtones. Even if you have all midtones, it's not guaranteed that `contrast-color()` will produce a better result than just hardcoding white, due to the bugs of the WCAG 2.0 contrast algorithm.

The way I see it, main use case where `contrast-color()` delivers value is exactly the one that is currently presented as an outlier: highly dynamic colors, where we don't know what we have and a static fallback is impossible. If we already know the color well enough to provide a readable fallback, we don't need `contrast-color()`, we can just use the fallback directly. IIRC GitHub issue labels was the example I used when I proposed the version of the feature that eventually shipped.

Sure, in some cases it may be just shorter to use `contrast-color()` but I think these use cases have distinct characteristics. If you're using `contrast-color()` on otherwise known backgrounds just to keep your code small, then including static fallbacks entirely negates any benefit. So for that to work either your support target allows you to depend on it fully, or you don't use it at all.

I may be missing the use cases you had in mind while writing this though. Please do tell me if that's the case!

### Move fallbacks to fallbacks section

Nit: The general guideline is that fallbacks should be in the fallbacks section, the actual guidance should be written as if all the modern goodies are available. Currently there are fallback values in the main guidance, which also makes it hard to discern which part is core to the guide and which part is a fallback for interop. 

### Fallbacks beyond RCS could break UIs; I suggest a decision framework instead of a blanket rec

In general, all of these are valid fallbacks for certain cases, but can ruin things in others, so I'm a little wary of recommending them in a blanket way (as much as I'd have loved the rise in Color.js' popularity that this would cause!):
- `text-shadow` simply doesn't fit certain UIs, and even when it does, the effect won't do much for e.g. white text on yellow. Also, I believe there are browsers that still apply it to `::selection` even with different colors, which can end up making text unreadable. 
- JS could work for static things known in advance, but in the general case where elements may be dynamically inserted and/or change colors dynamically, it will get very complex. Also, agents know how to do this already, we don't have to tell them.

I'd instead present fallback strategies in a list or table and ask the agent to take into account their specific requirements (how many users would be affected, the design, the needs of their specific use case (e.g. whether the color changes dynamically)) and make an informed decision. I'd also consider including:
- [The SVG filter method](https://miunau.com/posts/dynamic-text-contrast-in-css/), which is hacky AF but works well in practice (if you have a separate element for the text) and adapts reactively as the background changes
- A semi-transparent background could also work as a fallback for certain cases but not others.
`-webkit-text-stroke` might fit more UIs than `text-shadow` (and despite the name, is Baseline Widely Available). `paint-order` [seems to work](https://codepen.io/leaverou/pen/gbgBJZw) to put the stroke behind the text so that letterforms aren't affected.

guides/visual-design/contrast-color/guide.md
jamesnw: Am I correct that this is the replacement with `abs(sign())`? This seems less clear to me than using clamp.

`--l: calc(1 - (sign(l - var(--threshold)) + abs(sign(l - var(--threshold)))) / 2);`

jamesnw:
@LeaVerou I've implemented the changes, moving all fallbacks to their own section, and making a clearer decision matrix.

guides/visual-design/contrast-color/guide.md
patrickkettner: > Am I correct that this is the replacement with abs(sign())

yep! confirmed in stable channel. fwiw, I think `--l: max(0, sign(var(--threshold) - l))` is clearer?

LeaVerou: Sorry, `abs()` is not needed here, my bad. Just `max(0, sign(var(--threshold) - l))` as @patrickkettner mentioned.

jamesnw: Updated in [45452c8](https://github.com/GoogleChrome/modern-web-guidance-src/pull/966/commits/45452c883fe128c6b138bd34cdc75d464f83a4eb)

guides/visual-design/generate-derived-colors/demo.html
```diff
@@ -0,0 +1,9 @@
+<!doctype html>
```
LeaVerou, line 1: We no longer need demo files for stubs

guides/visual-design/contrast-color/guide.md
```diff
@@ -1,6 +1,189 @@
...
+.badge--shadow {
+  text-shadow: 0 1px 3px rgb(0 0 0 / 0.8);
+}
+```
+
+#### Option 5: SVG Filters
+
+{{ BASELINE_STATUS("svg-filters") }}
+
+1. Add this SVG filter definition to the HTML document.
```
LeaVerou: Do we need to do that? I'm ~90% sure we can link to an SVG file for this. Or if we want to avoid the extra HTTP request for performance, inline it in the CSS file, not in every page…

jamesnw: There's a Safari issue with loading filters from URLs, both external files or inlined in CSS- see https://github.com/Fyrd/caniuse/issues/3803 for a number of linked bugs. I confirmed it's still the case in Safari 26.6.

LeaVerou: Bummer. 

Then I'd mention that in the guide instead of just recommending inline HTML straight away. Basically "do X if you need to support Safari and Y otherwise". There are projects that don't need to support Safari at all (e.g. Electron apps) so the agent should have all the relevant info to make an informed decision.

jamesnw (approved):

guides/visual-design/contrast-color/guide.md
```diff
@@ -147,7 +147,7 @@ DO: Select the fallback that best matches your design constraints if RCS is not
 
 {{ BASELINE_STATUS("svg-filters") }}
 
-1. Add this SVG filter definition to the HTML document.
+1. Load this SVG filter definition. It must be added directly to the HTML document if you need to support Safari. Otherwise, it should be loaded from an external file, either an SVG file or inlined in CSS.
```
LeaVerou: Can we link to the bug?

jamesnw: [Added](https://bugs.webkit.org/show_bug.cgi?id=320118). It's a bit awkward, because it appears to be fixed in Safari Tech Preview, but that doesn't matter, as contrast-color predates the fix, so no versions of Safari can fall back to it.

guides/visual-design/contrast-color/guide.md
```diff
@@ -1,6 +1,189 @@
...
+.badge--shadow {
+  text-shadow: 0 1px 3px rgb(0 0 0 / 0.8);
+}
+```
+
+#### Option 5: SVG Filters
+
+{{ BASELINE_STATUS("svg-filters") }}
+
+1. Load this SVG filter definition. It must be added directly to the HTML document if you need to support Safari, as a [WebKit bug](https://bugs.webkit.org/show_bug.cgi?id=320118) prevents external and data url references for filters. Otherwise, it should be loaded from an external file, either an SVG file or inlined in CSS.
```
LeaVerou:
```suggestion
1. Load this SVG filter definition. It must be added directly to the HTML document if you need to support Safari versions affected by [this WebKit bug](https://bugs.webkit.org/show_bug.cgi?id=320118), which prevents external and data url references for filters. Otherwise, it should be loaded from an external file, either an SVG file or inlined in CSS.
```

(since the bug is fixed)

LeaVerou (approved):
See comment, but otherwise LGTM

rviscomi:
@patrickkettner @LeaVerou @jamesnw anything left to do or can we merge this?

PR #968, opened 2026-06-25
[#668] Streaming API

### LINKED ISSUE #668: Streaming API
### web-feature-id

TBD https://github.com/web-platform-dx/web-features/issues/4117

### Feature description

The new Streaming API for dynamic markup provides you with a revamped way to improve web experiences by injecting HTML into existing documents via the streamHTML and streamHTMLUnsafe methods. 

### Expected release date

Chrome 148, mentioned at I/O

### Coding agent assessment

_No response_

### Reference sources

_No response_

jamesnw:
This is very similar to #958. The interesting part here is how the HTML could be streamed from a server, which is out of scope for this project. I'm not sure how useful this guide would be. The demo attached is from the official post, and mocks HTML fragments using a while loop.  

I suspect this won't pass CI without a web feature.


rviscomi:
Ping @tunetheweb @malchata for a Perf ATL review

guides/user-experience/streaming-api/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: streaming-api
+description: Update an existing HTML document by injecting HTML fragments as they are streamed to the user.
```
malchata: I think this use case leaves out the when: Are HTML chunks streamed into an existing document automatically? On scroll? Or is this use case meant to describe a general streaming pattern that can be used in a variety of contexts?

guides/user-experience/streaming-api/demo.html
```diff
@@ -0,0 +1,23 @@
...
+      const writer = contentElement.streamHTML().getWriter();
+      let i = 0;
+      // Example stream of updating content
+      while (true) {
+        await writer.write(`<p>${++i}</p>`);
+        await new Promise((resolve) => setTimeout(resolve, 1000));
+      }
+
+      writer.close();
+    </script>
```
malchata, line 21: Not sure if this is the point, but this example seems contrived. Would it make sense to show an actual example where the streamer closes when the response body has arrived in full? In this example, there's "content" (not actually fetched from anywhere) and it's appended until a promise resolves in a `setTimeout`.

malchata (changes requested):

guides/user-experience/streaming-api/guide.md
tunetheweb: So @malchata this is a new API where you can effectively `setHTML` but instead of that happening in an atomic fashion, it can be streamed in (like it would in the initial document load, but that has been impossible up until now — at least not without hacks!).

Think like on GitHub where changing tabs on a large PR is slower than loading it from afresh as the soft nav version waits for the full API response before it can apply, while the hard nav load streams.

I think the description could be clearer. Maybe something like this?:

```suggestion
description: Update an existing HTML document, after load completes, with additional content that is applied as it streams.
```

guides/user-experience/streaming-api/demo.html
tunetheweb: I actually think this is a very clever work around to get round the fact this is normally handled on the server but we want a client-side demo!

So yes it's contrived, and probably wouldn't be used in real like like this (but never say never!), but it works for me as a "demo" that doesn't require complex server-side setup.

So I'm OK with it. But maybe add a comment explaining things better?

```suggestion
    <script type="module">
      const contentElement = document.querySelector("#content-to-update");
      // Mock a streaming server-side response:
      const writer = contentElement.streamHTML().getWriter();
      let i = 0;
      while (true) {
        await writer.write(`<p>${++i}</p>`);
        await new Promise((resolve) => setTimeout(resolve, 1000));
      }

      writer.close();
    </script>
```

tunetheweb:
> This is very similar to https://github.com/GoogleChrome/modern-web-guidance-src/pull/958. The interesting part here is how the HTML could be streamed from a server, which is out of scope for this project. I'm not sure how useful this guide would be.

Yeah, I need to have a think about all this as per my other comment. But been out for a bit and still playing catch up from that on other things. Will get back to this soon.

> I suspect this won't pass CI without a web feature.

Your suspicions are correct!

guides/user-experience/streaming-api/guide.md
malchata: I like it, but need to tweak it slightly more I think for clarity. (Edit: worked with @tunetheweb on a better `description`.)

```suggestion
description: Update existing HTML elements on a page with new content streamed from fetch responses.
```

guides/user-experience/streaming-api/demo.html
malchata: I think this suggestion works to clarify.

> I actually think this is a very clever work around to get round the fact this is normally handled on the server but we want a client-side demo!
> 
> So yes it's contrived, and probably wouldn't be used in real like like this (but never say never!), but it works for me as a "demo" that doesn't require complex server-side setup.
> 
> So I'm OK with it. But maybe add a comment explaining things better?
> 
> Suggested change
>     <script type="module">
>       const contentElement = document.querySelector("#content-to-update");
>       const writer = contentElement.streamHTML().getWriter();
>       let i = 0;
>       // Example stream of updating content
>       while (true) {
>         await writer.write(`<p>${++i}</p>`);
>         await new Promise((resolve) => setTimeout(resolve, 1000));
>       }
>       writer.close();
>     </script>
>     <script type="module">
>       const contentElement = document.querySelector("#content-to-update");
>       // Mock a streaming server-side response:
>       const writer = contentElement.streamHTML().getWriter();
>       let i = 0;
>       while (true) {
>         await writer.write(`<p>${++i}</p>`);
>         await new Promise((resolve) => setTimeout(resolve, 1000));
>       }
>       writer.close();
>     </script>

malchata (approved):

tunetheweb (approved):

guides/user-experience/streaming-api/guide.md
```diff
@@ -0,0 +1,6 @@
+---
+name: streaming-api
+description: Update existing HTML elements on a page with new content streamed from fetch responses.
+web-feature-ids:
+  # https://github.com/web-platform-dx/web-features/issues/4117
```
malchata, line 5: This PR is approved, but we should avoid merging until a web feature is minted for this.

rviscomi: If it's possible to merge this through in a way that tricks the linter, like `web-feature-ids: []`, then that's probably fine, since we'll have to revisit this anyway to write the rest of the guidance before it goes live.

jamesnw: Updated to empty array, and it looks like it's passing.

PR #985, opened 2026-07-06
restore feature name in dashboard scenario headers

paulirish:
this was a small regression from pawel's recent PR. but all good!

before

image

after

image


micahjo7 (approved):
thanks for catching!

PR #988, opened 2026-07-06
refactor(harness): decompose collection logic into helper functions

micahjo7:
follow up from previous cl, cleaning up the collection file (no new logic)


paulirish (approved):
weeeeeeeee

Issue #979, opened 2026-06-30
Add guidance for IME-safe Enter-to-submit behavior in chat/text inputs


ssssota:
### web-feature-id

keyboard-events

## Problem

Many chat-style text inputs submit the message from a `keydown` handler when the user presses `Enter`.

That behavior is common in implementations tested only with direct Latin keyboard input, but it can break for users who type with an Input Method Editor (IME), including Japanese, Chinese, Korean, and other composition-based input methods. In those contexts, `Enter` / `Return` is often used to confirm the current conversion candidate. If the app treats that same key event as a submit shortcut, the user can accidentally send an incomplete or unintended message.

## Why this matters

This is a common internationalization and usability pitfall in chat, comment, support, and AI prompt UIs.

Coding agents frequently generate implementations like:

```js
textarea.addEventListener('keydown', (event) => {
  if (event.key === 'Enter' && !event.shiftKey) {
    event.preventDefault();
    sendMessage();
  }
});
```

This looks reasonable for direct keyboard input, but it fails to account for IME composition.

## Suggested guidance

Modern Web Guidance should include IME-safe text submission guidance for message composers, for example:

* Do not submit a chat/message textarea on `Enter` while the user is composing text.
* Check `KeyboardEvent.isComposing` before treating `Enter` as a submit shortcut.
* Consider `compositionstart` / `compositionend` state tracking where framework or browser behavior requires it.
* Prefer an explicit Send button, and make keyboard shortcuts additive rather than the only submission path.
* Test with Japanese or another IME-based input method, not only with direct Latin keyboard input.

A minimum safer pattern is:

```js
textarea.addEventListener('keydown', (event) => {
  if (
    event.key === 'Enter' &&
    !event.shiftKey &&
    !event.isComposing
  ) {
    event.preventDefault();
    sendMessage();
  }
});
```

Where additional state tracking is needed, a pattern like this can avoid submitting while composition is active:

```js
let isComposing = false;

textarea.addEventListener('compositionstart', () => {
  isComposing = true;
});

textarea.addEventListener('compositionend', () => {
  isComposing = false;
});

textarea.addEventListener('keydown', (event) => {
  if (
    event.key === 'Enter' &&
    !event.shiftKey &&
    !event.isComposing &&
    !isComposing
  ) {
    event.preventDefault();
    sendMessage();
  }
});
```

## Related APIs / specs

* `KeyboardEvent.isComposing`
* `compositionstart`
[...]

PR #991, opened 2026-07-06
Add IME-safe enter-to-submit guide and evals

rviscomi:
Addresses #979

- creates a new `ime-safe-enter-submit` guide
- adds a reference to the guide from the Forms skill


micahjo7:
image

ran evals against this, looks like on Codex it is failing to retrieve the right guide

on Claude Code and Jetski CLI the results looks like this:

image

maybe the task could use some tweaking?

rviscomi:
Thanks @micahjo7. I seem to be blocked from installing Codex (haven't had to run it before) so I'm unable to eval it locally. Is there a change you think would help it pick up the guide?

micahjo7:
I discovered this was a local eval issue- basically, Codex CLI in our eval harness does not inheriting PATH, meaning the npx search/retrieve calls were using the published version of the package rather than the local version. So when these evals ran, it did not have the guide in this CL within the CLI tool. After fixing that, the evals look good (consistent with the claude/jetski results above). Sorry for the false alarm!

Will work on a fix for the issue.

rviscomi:
Ah ok great, thanks for confirming!

LeaVerou (approved):
This LGTM though it's not an API I have deep expertise on. Might want to link to the Safari bug so the agent can verify it's still current (yes, we'd update the guide, but 💩 happens) and to avoid wasting tokens hunting it down itself.

PR #993, opened 2026-07-08
Design systems / component use cases

LeaVerou:
Alright, I think this may be ready for more eyes on it. 

This guide adds 43 (!) new use cases, inspired from my look at various design systems (plus some I've been meaning to submit for a while).

Notes:
- A lot of these are meant as hub guides, to centralize all the different guides that need to be pulled together to compose a certain interaction (e.g. accordion, carousel, tabs). Even if we don't _currently_ have all the guides they'd need, I envision we'd be writing up the lower-level use cases as separate guides, and pulling them in for the high-level use cases.
- On the flip side, some of them are intended as low-level guides to be referenced by others (e.g. spatial navigation, drag & drop, overflow-popup, icons, dynamic-sort, responsive-disclosure).
- Some of them will become more interesting once certain upcoming features ship (e.g. slider-tooltip, checkbox) but can already be useful for preventing agents from rebuilding native controls.
- I introduced a `guides` frontmatter field to indicate what guides a given guide consists of (not just guides it happens to incidentally link to). It currently doesn't _do_ anything, but we use a whitelist for dist, so it should also be harmless. I think it's quite useful in general, but I wanted to flag it. If you like it, I can add it to existing hub guides too. And we could adjust tooling to aggregate web-feature-ids from these too.
- For cases where a guide is relevant to an upcoming feature, it has been included as a YAML comment

Questions:
- Currently many descriptions include a variation of "with support for keyboard navigation, assistive-technology support, and responsive design.". Is this useful or noise?
- `progress-ring` vs `spinner`: should these be one guide? I think this may become clearer when they are written.
- Should `icons` be in `ui-atoms` or `ui-components`? The idea is `ui-atoms` are not generally used as standalone components, but icons _could be_.

Also, thinking out loud: A lot of classic controls can be built with customizable-select listbox mode (e.g. button groups, cycle toggles, rating widgets, even checkbox or radio groups) but is this ok for a11y? It _seems_ to me that as long as the listbox interaction and AT bindings work correctly it _should_ be fine, but it would be nice to have an a11y SME review. Anyhow, that's for later, even if we change course, the actual use cases are still common enough to have a presence so I wouldn't remove them just for that.


guides/package.json
```diff
@@ -12,6 +12,6 @@
     "gray-matter": "^4.0.3",
     "marked": "^18.0.3",
     "serving": "workspace:*",
-    "web-features": "latest"
+    "web-features": "^3.34.0"
```
rviscomi: Was this intentional?

serving/package.json
```diff
@@ -40,7 +40,7 @@
     "apache-arrow": "21.1.0",
     "marked": "^18.0.3",
     "onnxruntime-node": "^1.25.1",
-    "web-features": "latest",
+    "web-features": "^3.34.0",
```
rviscomi: Ditto

rviscomi (approved):
Thanks Lea! This looks good overall.

>Questions:
>- Currently many descriptions include a variation of "with support for keyboard navigation, assistive-technology support, and responsive design.". Is this useful or noise?

It does feel more like noise. These things are table stakes. They're describing the implementation, not the use case, which doesn't necessarily help an agent pick the right guide out of the search results.

>- `progress-ring` vs `spinner`: should these be one guide? I think this may become clearer when they are written.

Happy to punt on this until the guides are written and we can see how much overlap there is

>- Should `icons` be in `ui-atoms` or `ui-components`? The idea is `ui-atoms` are not generally used as standalone components, but icons _could be_.

`ui-atoms` SGTM

LeaVerou:
Thanks for the review! I'll remove the description noise and revert the unintentional version change when back at my laptop!

guides/package.json
LeaVerou:
```suggestion
    "web-features": "latest"
```

serving/package.json
LeaVerou:
```suggestion
    "web-features": "latest",
```

LeaVerou:
> It does feel more like noise. These things are table stakes. They're describing the implementation, not the use case, which doesn't necessarily help an agent pick the right guide out of the search results.

Yeah, I agree. My hope was it might steer the agent away from thinking it already knows everything there is to know about implementing this use case but I doubt it would help much.

I just stripped it from nearly all of them except where it was core to the use case (`icons`, `responsive-table`, `hierarchical-navigation` / `spatial-navigation`, and `slider` / `checkbox` ("while preserving native functionality and accessibility" — the defining constraint of styling a native control).

Also reverted the unintentional `web-features` bump — turns out the lockfile had it pinned too, so that's restored as well. The diff is now only CODEOWNERS + the guides.

And agreed on punting the `progress-ring` vs `spinner` question until the guides are written.

Should I merge or wait for more reviews?

rviscomi:
Yup, ok to merge!

PR #1006, opened 2026-07-15
Content ATL config and scripts

### LINKED ISSUE #1007: How to best organize content and determine codeowners?
## Problems with the current taxonomy

<details><summary>The current categories are:</summary>

- accessibility
- built-in-ai
- canvas
- css-layout
- css
- datetime
- forms
- html
- modern-web-guidance
- motion
- overlays
- passkeys
- performance
- privacy/privacy
- scroll
- security/security
- typography
- webmcp

</details> 

[Here is a table of guides by category](https://docs.superhuman.com/@lea/web-features-mwg-guides/mwg-guides-7)

These work for low-level use cases that are very close to the features that spawned them, but **high-level use cases span 4-5 such areas each**:

<details><summary>Examples</summary>

- An **accordion** spans `motion` (animate-element-entry-exit), `html` (search-hidden-content), `accessibility`, `css-layout` (animate-to-intrinsic-sizes)
- A **tabbed UI** needs `motion` (anchor-positioning-tab-underline), `html`, `accessibility`, and depending on the strategy for overflowing tabs, `scroll` for scroll-snap (if they scroll) or `overlays` if they are hidden in a popup menu
- A **date range picker** needs `forms`, `datetime`, `overlays`, `css`, `css-layouts`
- A **spinner** would need `html` (`<progress>`), `motion`, `accessibility`
- A **responsive table** or **data grid** needs `scroll` (sticky headers), `html`, `css`
- A **sparkline** may be implemented via either canvas or svg, but we only have a `canvas` category
</details> 

Additionally, there are several existing guides with **no clear home**, so they have been categorized unpredictably:

<details><summary>Examples</summary>

- `navigation-drawer` is currently in `overlays` but also spans `scroll` (scroll-driven-animations, scroll-initial-target, scroll-snap), `css`, `accessibility`. `stack-drill-down` is near-identical in shape (overlay + gesture + inert + history) yet landed in `scroll`.
- Should `animate-to-intrinsic-sizes` be in `css-layout` or `motion`? It's in `css-layout`!
- Should `contrast-color` be in `css` or`accessibility`? It's in `css`!
- Should `interactions-in-complex-layouts` be in `performance`, `css` or `css-layouts`? It's in `performance`!
- Should `accessible-error-announcement` be in `accessibility` or `forms`? It's in `accessibility`!
- Should `persistent-top-layer-ui` be in `overlays` or `html`? It's in `html`!

And many, _many_ more.

</details> 

These are not isolated mistakes that we can just fix. The current taxonomy makes it very hard to figure out where a use case should live, and for many there is no obvious home.

It _think_ it's because categories are **mixed** across:
- **Axes**:
  1. Technologies (`css`, `css-layout`, `html`, `canvas`)
  2. Behaviors/mechanisms (`motion`, `scroll`, `overlays`)
  3. Horizontals (`performance`, `accessibility`, `security`, `privacy`)
  4. Domains (`forms`, `passkeys`, `datetime`)
- **Scope**:
	- `css`, `html` are conceptually huge, 
	- `scroll`, `datetime`, `passkeys` are oddly specific

There are also **obvious gaps**: 
- `css` and `html`, but no `js`
- `canvas` but no `svg`
- `forms` but no obvious home for any other interactive UI pattern that is _not_ a form control (accordions, tabs, etc), so it has to be slotted under incidental building blocks like `motion`, `scroll`, or `overlays` that are not necessarily core to the use case.

Generally, a good hierarchical taxonomy means there is **a single, unambiguous category for every data point** (here, guide), i.e.:
1. There are **no guides that fit in multiple categories**
2. There are **no guides that don't clearly fit anywhere**

This is a north star to move towards, few hierarchies can fully get there. But if you have too many of these issues, it's a smell that indicates that the categories need adjusting.

The argument for the current structure is that while it may not be ideal for IA (which makes it harder to find related guides, spot gaps, etc), it's **not user-facing** and helps us **programmatically determine codeowners**, which is _sorely_ needed for PRs to scale.

But given **how many use cases span multiple categories**, I don't think it currently works super well for codeowners either, in practice.

## Core idea

I think we're struggling because we're trying to solve **two different problems in one go**:
1. What categories make the most sense from an IA pov?
2. What categories help us to programmatically determine codeowners more easily?

These have **fundamentally different requirements**:
1. IA is about the **use case content**
2. Code ownership is about the **implementation**

Trying to solve them both with the same solution (directory structure) makes the problem harder than the sum of its parts. It also contributes towards making #981 worse as it forces a more tech-centric (instead of user goal-centric) categorization.

Also, **gating code ownership on directory structure** creates a higher bar than is necessary for it. **We should be able to assign narrow areas to their domain experts** without having to constantly move content around.
E.g. 
- we may want to have a broad `html` area lead, but also a smaller `web-components` codeowner, without needing to move all guides that relate to web-components outside of `html`. 
- Same for `css` and `css-layout`. There are SMEs that specialize on css-layout, but css-layout doesn't make sense as a category that is separate from `css`.

Instead, I would propose we **solve these problems separately**:
1. The [codeowners file is generated anyway](https://github.com/GoogleChrome/modern-web-guidance-src/pull/1006/changes#diff-98727adeea41a0c88d5d97c4aca5d39c091a33fafc5dee4888b6a0e014c116cf), so it can be generated based on whatever criteria we want: web-features, frontmatter tags, whatever. There is no reason to restrict it to directory structure except as a **fallback, to ensure every guide always has at least one codeowner.**
2. Once codeowners get partially decoupled from the directory structure, figuring out the right categories will become much easier, because now good IA for use cases is the only remaining requirement.

### Requirements

- **Predictability:** It needs to make sense. Nobody should be baffled why they were asked to review a guide.
- **Robustness:** Every guide needs to have at least one codeowner. That way, if metadata is missing that would pull in additional codeowners, that person has increased incentive to add it and pull them in.
- **Accountability:** It should be obvious who is ultimately responsible for each guide, even if others are called in (e.g. to review horizontals)
- **Efficiency:** It should not require O(N) user effort on the number of guides in the average case. Manual metadata should only be needed to correct the heuristic, unless the metadata is otherwise meaningful as well.

### Proposed category changes

The general idea is to **keep categories broad and unambiguous**, and use a **combination of web feature ids and manual tags to pull in more specialized codeowners** where appropriate.

- ✅ `html`: Keep (though specific guides need moving)
- ✅  `css`: Keep, for pure CSS guides
- ➕ `js` A category to host pure JS use cases, e.g. regex stuff, array stuff, etc
- ➕ `ui-behaviors`: drag, drag & drop, motion, scroll, light-dismiss, search hidden content, spatial navigation, etc
- ➕ `ui-atoms`: Low level ui primitives: overlays, effects, canvas in html, element shapes, etc
- ➕ `ui-components` (or `ui-patterns`): Common high-level UI use cases: accordions, tabs, treeviews, data grids, etc. 
- ❓ `visual-design` (or `theming`, `styling`): typography, theming, branding etc. Anything where the user goal is primarily to improve visual design, regardless of how it's implemented. Alternatively, these could go in `ui-atoms`.
- ❓  `forms`: Either keep, or fold behaviors into `ui-behaviors` (autofill, validate, etc), complex form controls into `ui-components` (custom selects etc) and branding/styling guides into `visual-design` (e.g. `brand-consistent-forms`)
- ✅ `accessibility`: Keep
- ✅ `performance`: Keep
- ✅  `privacy`: Keep
- ✅  `security`: Keep
- ❌ `passkeys` → Fold into `security`
- ✅ `webmcp`, `built-in-ai`: TBD, keep for now
- ❌ `css-layouts` → fold into `css`
- ❌ `datetime` → fold into `js`  (all are about `Temporal` and `Intl.*` APIs)
- ❌ `motion` → fold into `ui-behaviors`
- ❌  `scroll` → fold into `ui-behaviors`
- ❌ `overlays` → fold into `ui-atoms`
- ❌ `typography` → fold into `visual-design`
- ❌ `canvas` → fold into `visual-design` (it's all about `canvas-html`)

_(Very unsure about these names, suggestions welcome obvs, but the main thing is the grouping, the names are not user-facing.)_

Notes:
- `ui-atoms`, `ui-behaviors`, and `visual-design` are meant for low-level use cases that compose into actual components. 
- `ui-components` is meant as terminal components, and would ideally be mostly hub guides linking to `ui-atoms`, `ui-behaviors`, and `visual-design` guides.
- `css`, `html`, `js` are meant for even lower-level things where the use case is intrinsically linked to the core tech (e.g. "generate a regular expression that includes user-supplied data" is _intrinsically_ a JS use case, not just a use case that happens to use JS).

### Core idea

1. **Each high-level category has a codeowner**, but that is the **default** to fall back on if there are no more specific codeowners, or only horizontals (a11y etc). This ATL is also responsible for routing PRs to more specific codeowners, where appropriate, for PRs that are missing the metadata that would do this automatically.
2. **Web feature ids** are used to determine more granular area SMEs, through groups or broader tech area (see below)
3. **Frontmatter tags** are used to pull in horizontals or ATLs that cannot be determined from web-feature groups

Optional:
- Every existing category could also work as a tag
- For existing guides, it would make sense for their author(s) to be the first line of defense for PRs submitted by others, so it may be useful to encode that as frontmatter metadata as well (when different from the parent category).
- Note that an even smaller MVP could consist of just tags and explicit owners

#### When do tags _add_ vs _set_ codeowners?

We generally need both:
- Tags corresponding to horizontals (a11y, i18n, performance, etc) should **add** reviewers, but the ATL should still be in there as well
- Tags corresponding to areas should override the default codeowner

There are multiple mechanisms to do this, depending on whether we want this to be a property of the tag vs a property of the tag-guide assignment (i.e. do we want to use the same tag both ways?):
- A prefix, e.g. `tags: +a11y` adds, `tags: a11y` sets. Flexible, but error-prone.
- Tag metadata. E.g. `a11y` is only ever a horizontal (except for guides directly under `accessibility/` of course). Leaning towards this.

Any tags not explicitly defined as horizontals would _set_ codeowners.
We could even partly automate this: assume that any existing category can _also_ be used as a tag, and then it's additive wrt reviewers.

#### How to group web-features?

The `web-features` package already defines groups. You can see this hierarchy [here](https://docs.superhuman.com/@lea/web-features-mwg-guides/web-feature-groups-2).

It may turn out that the existing web-feature groups are enough to assign folks without it becoming too tedious (we definitely don't want to be assigning individual web-feature ids except as an escape hatch). 

There are [269 web-features with no group](https://docs.superhuman.com/d/Web-Features-MWG-guides_dth_UOwqRqx/Web-features_su6W5AJS#Web-features_tuUcS-Qf/Features-with-no-groups_vuRBniM2/r14&columnId=c-qJ5V_VlxIW).

So, I recently wrote [this package](https://www.npmjs.com/package/@web-features/tech) to categorize web-features further, even if they don't have groups (using their BCD keys, specs, etc), since I faced this problem when working on State Of surveys too (also because some of the groups are too granular). If it turns out that using the existing web-feature groups + ids is too tedious, perhaps we could use that. The granularity is also customizable, which should help.

## Validation

To make sure this would actually work well in practice, I tried to actually use it via two consumer sites: (a) specifying code owner rules and (b) assigning guides to categories.

### Codeowner rules (`atls.json`)

I tried to figure out what the config from #1006 would look like.

I ended up on something like this:

Current (TBDs added for illustrative purposes):

```json
{
  "accessibility": "TBD",
  "datetime": "TBD",
  "built-in-ai": "tomac",
  "canvas": "ewewraw",
  "css": "LeaVerou",
  "css-layout": "LeaVerou",
  "forms": "LeaVerou",
  "html": "LeaVerou",
  "motion": "bramus",
  "overlays": "LeaVerou",
  "passkeys": "agektmr",
  "performance": ["tunetheweb", "malchata"],
  "privacy": "rowan-m",
  "scroll": "bramus",
  "security": "rowan-m",
  "typography": "LeaVerou",
  "webmcp": [
    "beaufortfrancois",
    "Kulikowski"
  ]
}
```

New:

```json
{
  "default": {
    "accessibility": "TBD",
    "built-in-ai": "tomac",
    "css": "LeaVerou",
    "html": "LeaVerou",
    "js": "TBD",
    "forms": "LeaVerou",
    "ui-behaviors": "LeaVerou",
    "ui-atoms": "LeaVerou",
    "ui-components": "LeaVerou",
    "visual-design": "LeaVerou",
    "performance": ["tunetheweb", "malchata"],
    "privacy": "rowan-m",
    "security": "rowan-m",
    "webmcp": [
      "beaufortfrancois",
      "Kulikowski"
    ]
  },
  "web_features": {
    "canvas-html": "ewewraw",
    "container-scroll-state-queries": "bramus",
    "prefers-reduced-motion": "bramus",
  },
  "web_features_groups": {
    "scrolling": "bramus",
    "animation": "bramus",
    "transitions": "bramus",
    "view-transitions": "bramus",
    "webauthn": "agektmr"
  }
}
```

I didn't just author this from memory, I went through guides to see their web-feature-ids and then looked up [groups](https://docs.superhuman.com/@lea/web-features-mwg-guides/web-feature-groups-2) to make sure this wouldd exhaustively replicate the current codeowner assignments.

There were a couple of guides that were missing web-feature-ids that would be necessary for this to work, but not many.

As you can see it's not tremendously more complex. And if we go with the heuristic that every category can *also* be used as a tag to add reviewers, it doesn't even need any tag rules.

### Category assignments

I went through all guides and tried to categorize them in the new categories. [Results here](https://docs.superhuman.com/@lea/web-features-mwg-guides/mwg-guides-7)
- See the "new category" column for assignments
- 81/133 guides changed category: see the Old != Category tab 
- [Guides per current category](https://docs.superhuman.com/@lea/web-features-mwg-guides/mwg-categories-12)
- [Guides per proposed category](https://docs.superhuman.com/@lea/web-features-mwg-guides/proposed-categories-19)

Notes:
- Distribution across categories seems ok. I like that `security` now has actual content, beyond the discipline guide. `ui-behaviors` is a little large at 29 guides, but only slightly more than `performance`, which is at 25 guides.
- In terms of cognitive effort, I struggled the most to distinguish whether a guide should be in `ui-atoms` vs `ui-behaviors` and in some cases between `ui-atoms` and `ui-components`. It's easier to merge than to split though, so I’m inclined to keep them separate…
- Some of the guides I'd expected to move I ended up leaving alone since despite their name, they were tailored to the category they were in (e.g. `interactions-in-complex-layouts`). I suspect these will be most useful referenced from other, higher-level guides anyway.

rviscomi:
Fixes #1007

- Reorganizes guides into higher-level categories
- Introduces ATL config file and autogenerated CODEOWNERS
- Automated label-based issue and PR triage workflow


guides/atls.json
```diff
@@ -0,0 +1,20 @@
...
+  "built-in-ai": "tomac",
+  "canvas": "ewewraw",
+  "css": "LeaVerou",
+  "css-layout": "LeaVerou",
+  "forms": "LeaVerou",
+  "html": "LeaVerou",
+  "motion": "bramus",
+  "overlays": "LeaVerou",
+  "passkeys": "agektmr",
+  "performance": "tunetheweb",
```
malchata: Could this be expanded to have multiple owners? So, for example, on this line:

`"performance": ["tunetheweb", "malchata"]`

LeaVerou (approved):
LGTM, thank you for trying it! 🤞🏼 

Re:implementation, my only concern is `web-features-groups.json`: we're persisting `web-features` data on the filesystem, risking (and having to handle) staleness — or worse, **people hand-editing it thinking these groups are manual**. 

I looked into it: It's not for performance (my first guess) — lookups there are no faster than lookups via `web-features`. It's to keep the triage script dependency-free. And we can't even gitignore it :( 

Oh well. Maybe we can rename it to something like `feature-to-groups.cache.json` to make it even more obvious it's generated, or generate a JS file with a comment + a default export (which folks are less likely to treat as data meant to be edited).

TIL: Apparently `.gitattributes` provides a way to mark a file as generated: `guides/feature-to-groups.json linguist-generated=true`. Then GitHub collapses it in PR diffs and labels it generated. Reviewers/editors stop seeing it as hand-editable content. This would come in quite useful here!

Note: This only implements 2/3 resolution layers from #1007. It doesn't implement tags, either as arbitrary tags, or as categories-used-as-tags (for PR reviews — it sorta does for issues). I think that's ok for now, since our _current_ needs are served just fine with these two layers. Even when we get an a11y SME, we can add them as a secondary codeowner of `ui-components`. So, YAGNI, we can revisit tags if the need becomes actually pressing.

Also, I had marked #1007 as blocked on #1012, but this manages to work around it by hardcoding the two discipline guides that can no longer be determined by the heuristic:

```js
const isDisciplineGuide = inv.name === inv.category || ['css-layout', 'passkeys'].includes(inv.name);
```

I still think #1012 would be a better solution long-term, but this is good for now.

Just like #909, I would try to get this in sooner rather than later, as any moving around of guides is going to get out of sync _fast_.

rviscomi:
> Oh well. Maybe we can rename it to something like `feature-to-groups.cache.json` to make it even more obvious it's generated, or generate a JS file with a comment + a default export (which folks are less likely to treat as data meant to be edited).
> 
> TIL: Apparently `.gitattributes` provides a way to mark a file as generated: `guides/feature-to-groups.json linguist-generated=true`. Then GitHub collapses it in PR diffs and labels it generated. Reviewers/editors stop seeing it as hand-editable content. This would come in quite useful here!

Yeah that was exactly it. I've updated the file to clarify that it's generated, and added it to `.gitattributes`.

PR #1010, opened 2026-07-15
feat(eval-view): add run comparison dashboard UI (3/3)

psiarkiewicz:
## Summary
Part 3 of 3 (stacked on https://github.com/GoogleChrome/modern-web-guidance-src/pull/1187 ). 

Adds interactive web UI for side-by-side run comparisons in `eval-view`.

- **Compare UI**: Interactive side-by-side viewer (`/compare.html`) showing trajectory divergence, code diffs, Playwright outputs, and metric charts.
- **Guide View**: Multi-run selection badges and compare modal in `eval-view/guide.html`.
- **Server Endpoints**: `/api/compare`, `/api/download-run`, and on-demand `/api/trajectory-summary`.

### DIFF
--- a/eval-view/guide.js
+++ b/eval-view/guide.js
@@ -4,6 +4,7 @@ import { extractSuiteSummary } from './summary-extractor.js';
 /** @type {Record<string, GuideSuiteSummary>} */
 let allTestData = {}; // Cache all test data by testId
 let isCompareMode = false;
+/** @type {Array<{ dateKey?: string, combKey?: string, testId: string, source?: string, group?: any, runIndex?: number, agent?: string, model?: string, score?: number }>} */
 let selectedPoints = []; // array of { testId, source, combKey }
 let currentRunFilter = 'nightly';
 
@@ -239,7 +240,7 @@ function setupNavigationControls(currentGuide) {
     const list = $('#autocomplete-list');
     const goBtn = /** @type {HTMLButtonElement} */ ($('#go-guide-btn'));
 
-    const getNavUrl = (targetGuide) => `guide.html?guide=${encodeURIComponent(targetGuide)}${currentRunFilter ? `&runFilter=${encodeURIComponent(currentRunFilter)}` : ''}`;
+    const getNavUrl = (/** @type {string} */ targetGuide) => `guide.html?guide=${encodeURIComponent(targetGuide)}${currentRunFilter ? `&runFilter=${encodeURIComponent(currentRunFilter)}` : ''}`;
 
     const backLink = document.querySelector('a[href^="./"]');
     if (backLink) {
@@ -422,6 +423,7 @@ function renderGraphs(guideName) {
 
         const runsByDateMap = Map.groupBy(runs, run => getDateKey(run.timestamp));
 
+        /** @type {Array<{ dateKey: string, latestRun: any, runsOnDate: any[] }>} */
         const dateEntries = [];
         runsByDateMap.forEach((runsOnDate, dateKey) => {
             runsOnDate.sort((a, b) => new Date(a.timestamp).getTime() - new Date(b.timestamp).getTime());
@@ -556,7 +558,7 @@ function renderGraphs(guideName) {
             const run = entry ? entry.latestRun : null;
             const runsOnDate = entry ? entry.runsOnDate : [];
             
-            const isHighlighted = run && runsOnDate.some(r => r.testId === highlightTestId);
+            const isHighlighted = run && runsOnDate.some((/** @type {any} */ r) => r.testId === highlightTestId);
             if (isHighlighted) {
                 svgContent += `
                     <rect x="${x - 12}" y="${paddingY - 5}" width="24" height="${plotHeight + 10}" fill="var(--color-primary)" style="opacity: 0.12; rx: 4px;" />
@@ -670,7 +672,7 @@ function renderGraphs(guideName) {
                                 ${runsOnDate.length} runs on this date:
                             </div>
                             <div style="display: flex; flex-direction: column; gap: 4px;">
-                                ${runsOnDate.map((r, idx) => {
+                                ${runsOnDate.map((/** @type {any} */ r, /** @type {number} */ idx) => {
                                     const rStats = r.guides[guideName];
                                     const timeStr = new Date(r.timestamp).toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' });
                                     const isLatest = idx === runsOnDate.length - 1;
@@ -789,7 +791,7 @@ function renderGraphs(guideName) {
                         select.style.fontWeight = '500';
                         select.style.cursor = 'pointer';
 
-                        runsOnDate.forEach((r, idx) => {
+                        runsOnDate.forEach((/** @type {any} */ r, /** @type {number} */ idx) => {
                             const timeStr = new Date(r.timestamp).toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' });
                             const rStats = r.guides[guideName];
                             const opt = document.createElement('option');
@@ -807,7 +809,7 @@ function renderGraphs(guideName) {
                             const chosenTestId = select.value;
                             const selectedOpt = select.options[select.selectedIndex];
                             const chosenRunIndex = parseInt(selectedOpt.getAttribute('data-runindex') || '1');
-                            const chosenRun = runsOnDate.find(r => r.testId === chosenTestId) || defaultRun;
+                            const chosenRun = runsOnDate.find((/** @type {any} */ r) => r.testId === chosenTestId) || defaultRun;
                             handlePointSelection(chosenRun, group, combKey, guideName, chosenRunIndex);
                         });
 
@@ -820,13 +822,15 @@ function renderGraphs(guideName) {
                     }
                 } else {
                     window.location.href = `dashboard.html?testId=${defaultRun.testId}&source=${defaultRun.source}#guide-${guideName}`;
->>>>>>> 45a3b374 (feat(eval-view): add run comparison dashboard and visualizer)
                 }
             });
         });
     });
 }
 
+/**
+ * @param {string} guideName
+ */
 function setupCompareMode(guideName) {
     const compareBtn = $('#compare-mode-btn');
     const banner = $('#compare-banner');
@@ -861,6 +865,9 @@ function setupCompareMode(guideName) {
     });
 }
 
+/**
+ * @param {boolean} on
+ */
 function toggleCompareMode(on) {
     isCompareMode = on;
     const compareBtn = $('#compare-mode-btn');
@@ -886,9 +893,16 @@ function clearSelections() {
     updateCompareBanner();
 }
 
+/**
+ * @param {any} runData
+ * @param {SVGElement} group
+ * @param {string} combKey
+ * @param {string} guideName
+ * @param {number} [runIndex]
+ */
 function handlePointSelection(runData, group, combKey, guideName, runIndex = 1) {
     const testId = runData.testId;
-    const dateKey = group.getAttribute('data-datekey');
+    const dateKey = group.getAttribute('data-datekey') || '';
     const existingIdx = selectedPoints.findIndex(p => p.dateKey === dateKey && p.combKey === combKey);
 
     if (existingIdx !== -1) {
@@ -911,8 +925,10 @@ function handlePointSelection(runData, group, combKey, guideName, runIndex = 1) 
     } else {
         if (selectedPoints.length >= 2) {
             const removed = selectedPoints.shift();
-            const oldGroup = document.querySelector(`[data-datekey="${removed.dateKey}"][data-comb="${removed.combKey}"]`);
-            if (oldGroup) oldGroup.querySelector('.compare-highlight')?.remove();
+            if (removed) {
+                const oldGroup = document.querySelector(`[data-datekey="${removed.dateKey}"][data-comb="${removed.combKey}"]`);
+                if (oldGroup) oldGroup.querySelector('.compare-highlight')?.remove();
+            }
         }
 
         const stats = runData.guides[guideName];
@@ -927,8 +943,8 @@ function handlePointSelection(runData, group, combKey, guideName, runIndex = 1) 
             score: stats ? stats.guidedRate : 0
         });
 
-        const x = parseFloat(group.getAttribute('data-x'));
-        const y = parseFloat(group.getAttribute('data-yg'));
+        const x = parseFloat(group.getAttribute('data-x') || '0');
+        const y = parseFloat(group.getAttribute('data-yg') || '0');
         
         if (!group.querySelector('.compare-highlight')) {
             const highlight = document.createElementNS('http://www.w3.org/2000/svg', 'circle');
--- a/eval-view/guide.js
+++ b/eval-view/guide.js
@@ -2,13 +2,13 @@ import { initGoogleAuth, authenticatedFetch, getAccessToken, escapeHtml, hasNigh
 import { extractSuiteSummary } from './summary-extractor.js';
 
 /**
- * @import { SelectedTrialPoint } from './evals.d.ts'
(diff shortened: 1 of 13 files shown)


bin/gd.ts
```diff
@@ -226,6 +227,19 @@ async function main() {
   }
 
   switch (command) {
+    case 'compare': {
+      const runDirA = requireArg(positionals[1], 'gd compare <runDirA> <runDirB>');
+      const runDirB = requireArg(positionals[2], 'gd compare <runDirA> <runDirB>');
+      const { runComparison } = await import('../harness/lib/compare-agent.ts');
```
micahjo7: nit: how about `compare-evals.ts` for the filename?

harness/lib/gcs-downloader.ts
```diff
@@ -0,0 +1,283 @@
+import { Storage } from '@google-cloud/storage';
```
micahjo7, line 1: the download workflow (when running `gd compare`) seems to be downloading the entire suite rather than just the relevant tasks runs

micahjo7: i think this issue is also causing the download to fail

the repro is: open local dashboard with `gd dashboard` sign in and get remote results, and then try to compare the following runs:

Image

eval-view/guide.js
```diff
@@ -618,9 +621,123 @@ function renderGraphs(guideName) {
...
+                    } else {
+                        window.location.href = `dashboard.html?testId=${runData.testId}&source=${runData.source}#guide-${guideName}`;
+                    }
                 }
             });
         });
     });
 }
+
+function setupCompareMode(guideName) {
```
micahjo7, line 838: we should have some handling to omit this button when on the remote dashboard

and as a general comment, there may be some other changes required in the PR so that it is fully designed only for the local dashboard only in this leg of the feature

eval-view/server.js
```diff
@@ -11,6 +11,32 @@ import { runAllManifests } from './generate-manifests.js';
 const PORT = process.env.PORT || 8081;
 const STATIC = process.env.STATIC === 'true';
 
+// Registry of supported agent identifiers mapping path substrings to Agent display names.
+// To add a new agent in the future, simply append an entry here (e.g., { match: 'newagent', name: 'New Agent CLI' }).
+const SUPPORTED_AGENTS = [
```
micahjo7, line 17: you should be able to extract the relevant agent from the evals.json

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,654 @@
...
+
+    // To add another agent in the future, check agentName or session files here and invoke the corresponding parser.
+    if (agentName.toLowerCase().includes('jetski') || (!sessionFiles[0] && fs.existsSync(targetDir) && fs.readdirSync(targetDir).some(f => f.endsWith('.db')))) {
+      summary = await parseJetskiTrajectory(targetDir, serving);
+    } else if (sessionFiles[0]) {
+      const filePath = path.join(targetDir, sessionFiles[0]);
+      const isJsonl = filePath.endsWith('.jsonl');
+      const logData = isJsonl ? parseJsonlFile(filePath) : JSON.parse(fs.readFileSync(filePath, 'utf8'));
+
+      if (agentName.toLowerCase().includes('claude')) {
```
micahjo7: we should be able to use the Agents enum from `harness/config.ts` here (and elsewhere in this file)

harness/lib/compare-evals.ts
```diff
@@ -0,0 +1,1005 @@
...
+import { cGreen, cRed, cCyan, cBold } from '../../lib/colors.ts';
+import { downloadRunFromGcsIfMissing } from './gcs-downloader.ts';
+import { rootDir, baseAppsDir } from '../../lib/paths.ts';
+import { parseJsonlFile } from './agent-shared.ts';
+import { extractInitialPromptFromLogs } from './trajectory-parser.ts';
+
+/**
+ * Dynamically fetches all active models from the Gemini API and returns them sorted by version and capability.
+ */
+async function getSortedModelsList(apiKey: string): Promise<string[]> {
```
micahjo7: alot of the logic in this file should become simpler now that we only want this locally. we can call the gemini api the same way the `gd dev` workflow runs with (but it is probably a good idea to omit `--yolo`)

harness/lib/compare-evals.ts
```diff
@@ -0,0 +1,1005 @@
...
+    graderContent: graderContent || 'No grader.ts content found.',
+    baseAppContent: baseAppContent || 'No base app content found.'
+  };
+}
+
+/**
+ * Helper to find the main generated code file in a run directory.
+ */
+function findCodeOutput(dir: string): { path: string; content: string } {
+  const candidates = [
```
micahjo7, line 170: all our recent evals store the targetFile in the evals.json, so we don't have to guess from this list anymore

harness/lib/compare-agent.ts
```diff
@@ -0,0 +1,1005 @@
...
+      }
+      if (query) {
+        searchQueries.push(String(query).trim());
+      }
+    }
+    // 3. Code Mutation
+    else if (
+      actionName.includes('write') || actionName.includes('replace') || actionName.includes('touch') ||
+      actionParamsStr.includes('write_to_file') || actionParamsStr.includes('replace_file_content') ||
+      actionParamsStr.includes('index.html') || actionParamsStr.includes('app.jsx') || actionParamsStr.includes('style.css')
```
micahjo7: this type of hardcoding can also likely be updated by using the target file in evals.json?

harness/lib/compare-agent.ts
```diff
@@ -0,0 +1,1005 @@
...
+ * Phase 2: Sub-agent 1 - Guide Compliance & Requirement Auditor.
+ */
+async function runSubAgent1_GuideCompliance(
+  guideCtx: GuideContext,
+  ctxA: RunContext,
+  ctxB: RunContext,
+  statusA: string,
+  statusB: string
+): Promise<string> {
+  const systemInstruction = `You are a specialized Web Guidance Compliance Auditor. Your task is to evaluate whether two agent runs (Run A vs Run B) successfully discovered, retrieved, and adhered to the MANDATORY requirements in guide.md and expectations.md.
```
micahjo7: it might be good to put these prompts as constants in a separate file

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,654 @@
...
+      }
+    }
+  }
+  return results;
+}
+
+/**
+ * Synthesizes a normalized TrajectorySummary for Jetski/MCP using .db files, modern-web.log, and chat_log.txt.
+ */
+export async function parseJetskiTrajectory(dirPath: string, serving: string): Promise<TrajectorySummary> {
```
micahjo7: does this work? does comparison with jetski cli runs show useful data? as far as i know, we are unable to parse the actual jetski cli trajectories in our code, so this might only be using the modern-web.log and chat-log

harness/lib/gcs-downloader.ts
```diff
@@ -0,0 +1,283 @@
...
+    const suiteId = relativeRunPath.split(/[/\\]/)[0].toLowerCase();
+    if (suiteId.includes('claude')) {
+      detectedAgent = 'Claude Code';
+    } else if (suiteId.includes('gemini')) {
+      detectedAgent = 'Gemini CLI';
+    }
+    
+    try {
+      const { generateNormalizedTrajectory } = await import('./trajectory-parser.ts');
+      await generateNormalizedTrajectory(absoluteRunDir, detectedAgent, 'mcp');
```
micahjo7: `mcp` hardcoded

harness/lib/gcs-downloader.ts
```diff
@@ -0,0 +1,283 @@
...
+const BUCKET_NAME = 'guidance-evals';
+
+/**
+ * Performs post-download operations, such as generating missing trajectory summaries.
+ */
+async function postDownloadProcessing(absoluteRunDir: string, relativeRunPath: string) {
+  const summaryPath = path.join(absoluteRunDir, 'trajectory_summary.json');
+  if (!fs.existsSync(summaryPath)) {
+    console.log(cCyan(`[GCS Downloader] trajectory_summary.json is missing in historical run. Generating on the fly...`));
+    let detectedAgent = 'Jetski';
```
micahjo7: use enums for these agent names (in `harness/config.ts`)

results/compare_work/response_debug.json
```diff
@@ -0,0 +1,33 @@
+{
```
micahjo7: nit: remove this from the commit, maybe we can just store this in harness/results?

.gitignore
```diff
@@ -69,3 +69,7 @@ serving/lib/tfjs_model_minilm/saved_model*/
 
 # Agent skills artifacts
 .agents/skills/nightly-eval-investigation/artifacts/
+
+# AI Variance Diagnosis outputs
+results/
+variance_diagnosis.md
```
micahjo7: we could store this in `harness/results/` instead

tsconfig.json
```diff
@@ -25,6 +25,7 @@
     "**/eval-view",
     "**/eval-view/**/*",
     "harness/base_apps/devtools-times/",
-    "**/grade-report/**"
+    "**/grade-report/**",
+    "**/web-uplift/**"
```
micahjo7: do we need this?

micahjo7 (commented):
Our evals execute 1 run per task, and 1 task per guide. In the future, if we expand this (multiple runs per task, or multiple tasks per guide, or both), there are some things to consider:

What does the guide view show for these scenarios? IIUC it shows an aggregate of everything for that guide run on that day?
Given this^, should you be able to compare between different “runs” within the same suite? What about comparing guided vs unguided from the same run? In the guide view UI, you have to click on two separate data points (from separate suites), so not sure if it is possible.
For multiple tasks that run per guide, what will the guide comparison page show? Should you be able to select which task you want to compare (or which run?) similar to how you can select guided vs unguided? It looks like there might be some functionality for this currently, but from my testing it doesn’t look like it is working.

Another (less important) scenario is having multiple suite runs on the same day- I believe the guide view shows only the latest- should it aggregate instead? Or show two separate arrows?

In general, there is some aggregation that occurs for these scenarios when converting to the guide view- in the comparison view, we should be able to fully disaggregate into the specific permutations.

micahjo7:
General comment for the PR:

there is a lot of code here, and several different components. I'm thinking it might be a better idea to separate this into multiple PRs- the first thought that comes to mind is that the trajectory parsing / summary is a feature in itself. then there is the `gd compare` workflow, then the UI component which presents and visualizes it all (and runs agent analysis). wdyt?

results/compare_work/response_debug.json
psiarkiewicz: results should not have been checked in for sure.

psiarkiewicz:
> General comment for the PR:
> 
> there is a lot of code here, and several different components. I'm thinking it might be a better idea to separate this into multiple PRs- the first thought that comes to mind is that the trajectory parsing / summary is a feature in itself. then there is the `gd compare` workflow, then the UI component which presents and visualizes it all (and runs agent analysis). wdyt?

That makes sense.

harness/lib/trajectory-parser.ts
psiarkiewicz: The intent was to look for the .db files in a local .gemini/jetski/conversations directory, which should exist if the eval run was done locally (I think).  But most of the time, this falls back to the log and txt file.
I can remove the .db call, or leave it for completeness.  
What's preferred?

micahjo7: I think this is fine! I wasn't aware that the Jetski trajectories are parseable now. I actually have some work in another branch which extracts data from this trajectory now and populates model, token, etc. in the dashboard. That might simplify some of what is here.

psiarkiewicz: sg. but I would double check.

bin/gd.ts
psiarkiewicz: compare agents changed to compare-evals

harness/lib/gcs-downloader.ts
psiarkiewicz: fixed. download should now only happen if the files are missing

eval-view/guide.js
psiarkiewicz: this should now be local only

eval-view/server.js
psiarkiewicz: using eval.json and only falling back to path if mising

harness/lib/compare-evals.ts
psiarkiewicz: refactored the compare to use gd local Jetski or Gemini.

harness/lib/compare-evals.ts
psiarkiewicz: fixed

harness/lib/compare-agent.ts
psiarkiewicz: prompts now in compare-prompts.ts

harness/lib/trajectory-parser.ts
psiarkiewicz: seems to work.

harness/lib/gcs-downloader.ts
psiarkiewicz: fixed

results/compare_work/response_debug.json
psiarkiewicz: cleared out

.gitignore
psiarkiewicz: fixed

harness/lib/compare-agent.ts
psiarkiewicz: using target file when available. otherwise fallback to hardcode

harness/lib/gcs-downloader.ts
psiarkiewicz: fixed

tsconfig.json
psiarkiewicz: removed

psiarkiewicz:
> Our evals execute 1 run per task, and 1 task per guide. In the future, if we expand this (multiple runs per task, or multiple tasks per guide, or both), there are some things to consider:

Added multi-guide view.  For now, you can selected different versions on the same day, but the arrow shows the latest one. Can fix later when we know what's useful. Blending daily scores doesn't seem right.

psiarkiewicz:
> General comment for the PR:
> 
> there is a lot of code here, and several different components. I'm thinking it might be a better idea to separate this into multiple PRs- the first thought that comes to mind is that the trajectory parsing / summary is a feature in itself. then there is the `gd compare` workflow, then the UI component which presents and visualizes it all (and runs agent analysis). wdyt?

Agreed.
Splitting this into 3 chained PRs:
1. **Trajectory Normalization Framework**: `trajectory-parser.ts`, viewer updates, and agent hooks emitting `trajectory_summary.json`.
2. **`gd compare` CLI & AI Variance Diagnosis**: CLI command, lazy GCS downloader, prompt modules, and multi-stage subagent synthesis.
3. **Dashboard Comparison UI**: Side-by-side web comparison interface, guide view integration, and server endpoints.

psiarkiewicz:
Split complete into 3 stacked PRs:
- Part 1: #1186  (`feat/trajectory-normalization`)
- Part 2: #1187  (`feat/cli-eval-compare`)
- Part 3: This PR (#1010, rebased on Part 2)

paulirish:
my tweaks:

*   deferred the ai variance diagnosis behind an **explicit button click** instead of auto-firing on page load, saving api tokens and keeping initial navigation fast.
* **streamlined trajectory streaming**: simplified the sse streaming, date grouping, and auto-scroll handling in `compare.js`.
* **server & schema cleanup**: updated `eval-view/server.js` to read directly from the normalized trajectory summary, removing redundant parsing logic and dropping `schemaVersion`.
* added `eval-view/utils.test.ts` to test shared helper utils (url param handling, date formatting).

PR #1018, opened 2026-07-20
chrome-extensions skill: fix permissions.request() user-gesture pitfall

sebastianbenz:
## Summary
- Document that `chrome.permissions.request()` must be called synchronously in the UI click handler (or in a message handler)
- Extract permission-related pitfalls (`tabs` permission for `tab.url`, `activeTab` user-gesture restriction, `chrome.permissions.request()`) into a new `references/extensions/permissions.md`, condensing the corresponding SKILL.md sections to short pointers
- Add matching Output Checklist item and Reference Files table entry

### DIFF
--- a/skills-src/chrome-extensions/SKILL.md
+++ b/skills-src/chrome-extensions/SKILL.md
@@ -81,14 +81,8 @@ See `references/extensions/csp-sandbox.md` for full details.
 
 #### 4. `tab.url` requires the `tabs` permission
 
-Without it, `tab.url` silently returns `undefined` — no error thrown.
-
-```js
-// manifest.json — REQUIRED if you read tab.url or tab.title anywhere:
-{ "permissions": ["tabs"] }
-```
-
-See `references/extensions/tab-management.md`.
+Without it, `tab.url` silently returns `undefined` — no error thrown. See
+`references/extensions/permissions.md`.
 
 #### 5. Always use async/await — never `.then()` chains
 
@@ -196,27 +190,10 @@ await chrome.action.setBadgeText({ text: '5' });
 
 #### 12. `activeTab` only works on direct user gestures — not from side panels
 
-`activeTab` grants temporary access to the current tab ONLY when triggered by:
-- Clicking the extension action icon
-- A context menu item (including the `"tab"` context)
-- A keyboard shortcut from the `commands` API
-- Accepting an omnibox suggestion
-
-It does **NOT** grant access when clicking a button in a side panel, popup button that opens later,
-or any programmatic trigger.
-
-```js
-// ❌ BROKEN — activeTab does NOT work from a side panel button click
-document.getElementById('summarize').addEventListener('click', async () => {
-  const [tab] = await chrome.tabs.query({ active: true, currentWindow: true });
-  await chrome.scripting.executeScript({ target: { tabId: tab.id }, func: () => document.body.innerText });
-});
-
-// ✅ FIX — use "tabs" permission + specific host_permissions instead
-// manifest.json: { "permissions": ["tabs", "scripting"], "host_permissions": ["<all_urls>"] }
-```
-
-See `references/extensions/side-panel.md`.
+`activeTab` grants temporary access to the current tab ONLY on a direct user gesture (action
+icon click, context menu item, keyboard shortcut, omnibox suggestion) — NOT from a button click
+inside a side panel or popup. Use `tabs` + `host_permissions` instead. See
+`references/extensions/permissions.md` and `references/extensions/side-panel.md`.
 
 #### 13. DevTools panel URLs are relative to the extension root
 
@@ -379,6 +356,14 @@ const all     = await chrome.windows.getAll({ populate: true });
 
 **`chrome.windows` methods:** `getAll`, `getLastFocused`, `getCurrent`, `get(windowId)`, `create`, `update`, `remove`. See `references/extensions/tab-management.md`.
 
+#### 20. `chrome.permissions.request()` requires a user gesture — call it directly in the click handler, not via message passing
+
+`chrome.permissions.request()` only works when called synchronously within a user gesture. If a
+UI context (side panel, popup) sends a message to the service worker and the service worker
+calls it in response, the gesture context is lost crossing the message-passing boundary and the
+call fails silently. Call it directly in the click handler of the UI context instead — see
+`references/extensions/permissions.md`.
+
 ### Always Manifest V3
 
 Never generate Manifest V2 code.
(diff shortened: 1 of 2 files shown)


skills-src/chrome-extensions/SKILL.md
```diff
@@ -379,6 +356,14 @@ const all     = await chrome.windows.getAll({ populate: true });
 
 **`chrome.windows` methods:** `getAll`, `getLastFocused`, `getCurrent`, `get(windowId)`, `create`, `update`, `remove`. See `references/extensions/tab-management.md`.
 
+#### 20. `chrome.permissions.request()` requires a user gesture — call it directly in the click handler, not via message passing
+
+`chrome.permissions.request()` only works when called synchronously within a user gesture. If a
+UI context (side panel, popup) sends a message to the service worker and the service worker
+calls it in response, the gesture context is lost crossing the message-passing boundary and the
```
oliverdunk: This is not true ([example](https://play.web-extensions.dev/#s=N4IgRgTg9g7gzgUwiAXCAwgC2gWwSAGhBwEMA7ASwDME4AXANSTgqjNRAFkGBmQkCmQDGAGwCuAEwQAFKCICeVCiJGoqJEYiJ0EOAA4iSOgJISOLKdPIJVRJSNqoA2qDIk8HUpRr0AdACs4Nn4dAA86DmAAHTIAAliokDc8RJQEkABlCilYqzIbRIIY%2BMSAN2ZWMlT0gAZfAEZC4vSvalo6AH1yiBY2ap4iuPSpOCEICj06SurEgCESFiFYrJy8m1iAUXCEMl6qwmbEsBIhAGsAc2gxMglq6KGSkEQIUoohBA6YKAhTpBnwE4XK43AJwRLNAC%2Bg0eJymfVQsXu8UeUnUYhEnSmdAc-wA8nodrELAhYnprCJwUMoYcntkPmT8hSEUjkYlUSR0Z0yXRMP9iQybL5MHQcEzIdD0gSIDgKHA9mCEU5mo9iWsxUMALoSxJQSaVDQdKUyuWVBVpJUPdJ0EhgBXNDUxCEgKGudz4NDHM6XKDXCS%2BOhwEIIcIcITYKB4XwQa5TSNsYy7a0qBB%2BkgSCQAGVlOnyEAAFHmpNblHAAJSxAC8AD5Ec0hGwgg5fCIoOc84ktjm9rFMAtYmAEITBPQNA4-bEAEoIBZsFCFWJFkglqPToJkUsAbjr4cjqvJvkQdDVswQvde3zzwFiup2atxZAAgkI4WR0CI3qc0nRoySIZvHf%2BVTCA2cgIM2rbtiAAASNgtrEADq3wiBIACEiSATEYa4GB0ZkLGYFsJwtBwCQ5xgWmmbZjsSAFuW1a1kMWERmBRqyvKK4AI5iO0l6kkgxryuaiTWraiQarEf5bmQkkxM6BCuh4aD8vuwqikGIZoAAPChAAiuLoAAKgAmtIGw9iKIhVjEmmqZZzSaWAUASPIVmWjZ9RVjBKhQAhSGoZpAD0mAecqsSaaM4yTESEBCBWiTKYyoKJFWgURRMdCufEgWOc5rmBbZVZyQp7q0lIAoiP6gbaMGERoBIUBCGIeB4b4OXyL4FEbOUeFZvQ1H5okogfvOeZ0TWLJMZGuH4QeOwSERcqkQgkE6PQ6FSZJzoahCQA)). When a message is sent from a web context with an active user gesture, the gesture is attached to the message and given to the extension when the listener handles the message. The limitation that the user gesture needs to be handled synchronously still applies.

skills-src/chrome-extensions/references/extensions/permissions.md
```diff
@@ -0,0 +1,107 @@
...
+
+`chrome.permissions.request()` (for optional/dynamic permissions) only works when called
+synchronously within a user gesture (a click, keypress, etc.). If a UI context (side panel,
+popup) sends a message to the service worker and the service worker calls
+`chrome.permissions.request()` in response, Chrome no longer sees a user gesture — the call
+fails silently or rejects, because the gesture context is lost crossing the message-passing
+boundary.
+
+```js
+// ❌ BROKEN — side panel forwards the request through the service worker,
```
oliverdunk: This should work fine. Are you seeing otherwise?

oliverdunk (changes requested):
Thanks for the PR! Glad to have more guidance here, but I think this isn't quite accurate for user gesture handling. Let me know if you have questions.

skills-src/chrome-extensions/SKILL.md
sebastianbenz: You're right, thanks for catching this — I built a small test extension and verified with a live click in Chrome:

- Calling `chrome.permissions.request()` as the first statement in the service worker's `onMessage` listener (no `await` before it) works fine: the gesture from the side panel click propagates across `sendMessage` and Chrome proceeds to the permission prompt.
- Adding an `await` (even a 50ms delay) before the call burns the gesture — it throws `"This function must be called during a user gesture"`.

Pushed 589930a0 correcting both docs to the actual rule: the gesture survives one `sendMessage` hop but only for that synchronous turn, so `chrome.permissions.request()` must be the first thing the listener does, not that it can never be called from the service worker.

skills-src/chrome-extensions/references/extensions/permissions.md
sebastianbenz: Confirmed — this example does work as you said (didn't reject, proceeded to the permission prompt). I over-generalized from a real bug where the gesture actually was lost, but the culprit was an `await` before the `chrome.permissions.request()` call, not the `sendMessage` hop itself. Rewrote this section (589930a0) to show the correct broken/fixed pair: broken = await before the call, fixed = call it first, then await the result.

skills-src/chrome-extensions/references/extensions/permissions.md
```diff
@@ -0,0 +1,113 @@
...
+  chrome.runtime.sendMessage({ type: 'REQUEST_PERMISSION' }, (granted) => { ... });
+});
+
+// service-worker.js
+
+// ❌ BROKEN — an await before chrome.permissions.request() burns the gesture window
+chrome.runtime.onMessage.addListener((msg, sender, sendResponse) => {
+  if (msg.type === 'REQUEST_PERMISSION') {
+    (async () => {
+      await chrome.storage.local.get('someSetting'); // gesture is gone after this
```
oliverdunk, line 76: Nit: This is a bit artificial as you aren't using the result of the storage call for anything.

oliverdunk (approved):
LGTM with nit - thanks!

PR #1024, opened 2026-07-21
feat: add frontmatter field to withhold guides from publishing

### LINKED ISSUE #1017: Frontmatter option to prevent guides from being published
As discussed with @rviscomi, we should have a frontmatter option to keep a guide out of dist without deleting it from the repo.

## Example use cases

- **Drafts**. We already have stubs in the repo, and **drafts are essentially intermediate states between a stub and a full guide**. While they could be technically developed in a PR that just doesn't get merged until the guide is _done_, it's typically easier to iterate with smaller PRs.
- **Unpublish** If serious issues are found with an existing guide we may want to temporarily remove the guide from dist until we fix the issues. We should not need to remove the guide to do that.
- **Retired** A guide that is no longer needed, as it was not offering uplift. We may not want to remove the guide, e.g. the use case may later become relevant again, as a new feature gets added to the platform.
- **Future**: A guide that is currently premature, e.g. authored in anticipation of shipping one or more upcoming features, or depending on other guides that are currently drafts. This is distinct from _Draft_: draft describes the guide content, _Future_ depends on the environment.
- **Regression hold** for evals that dropped below threshold, withheld until fixed. Could even be set automatically.

This could also replace the hardcoded `excludeFromBundling` list (though I believe it's currently empty).

- **Regression hold** — evals dropped below threshold, auto-withheld until fixed. This could even be automatically set.

This could also help us get rid of the current hardcoded `excludeFromBundling` list (although I believe the list is empty).

## API / Naming

The main design question is **binary publish toggle vs an enum that distinguishes the cases above?**

For a binary toggle:
- `draft: true` is common, but only covers one kind of hiding.
- `publish: false` covers all of them, but booleans defaulting to `true` are [discouraged](https://www.w3.org/TR/design-principles/#optional-parameters:~:text=For%20boolean%20arguments%2C%20a%20default%20value%20of%20false%20is%20strongly%20preferred.) — and avoiding that leads to double negatives like `hidden: true`.

An enum (e.g. `status`) adds some cognitive overhead, but enables smarter handling: a `future` guide can relax web-feature-id validation, a `draft` can skip it entirely, a `regression-held` guide can keep running in evals and flip back to published when uplift recovers, and diagnostics can flag guides stuck in `draft` too long. And frontmatter is all keywords anyway — `true`/`false` are keywords you type too.

**Proposal: `status: draft`** (with room for more values later). Covers the case that prompted this, and grows naturally into the rest. Or, we can even add values for (some of) the statuses above too and just handle them identically at first.

LeaVerou:
> [!NOTE] 
> I drafted this last night before seeing @paulirish’s comment https://github.com/GoogleChrome/modern-web-guidance-src/issues/1017#issuecomment-5023911612 . Putting it up here as I think it's still useful for gauging scope, but I’ll mark it as draft while we're still discussing the API shape.

The design in this PR was a `status: draft | retracted | future` in frontmatter to be excluded from all distribution (search index, README, megaskill) without deleting them. Present but unknown status values fail validation. Absent value = publish.
I know I said in #1017 to just go with a single `draft` value, but these three felt distinct enough and useful in the short-term. 

Introduces `isPublished(inv)` as the single publish predicate (real content + no withholding status), replacing the per-surface stub checks across the three build paths.

I debated having a symmetrical empty `PUBLISH_STATUSES` as well for any future values, but chose YAGNI instead. Happy to change it if we feel that would be useful.

Closes #1017

### DIFF
--- a/.agents/skills/project-guides/SKILL.md
+++ b/.agents/skills/project-guides/SKILL.md
@@ -43,6 +43,7 @@ web-feature-ids:
 ---
 ```
 * **web-features**: Must be a list of accurate IDs found via webstatus.dev. Include ALL features referenced in the guide body, not just the primary one. If an ID is missing, inform the USER.
+* **status** (optional): Set to `draft`, `retracted`, or `future` to withhold the guide from all distribution (search index, README, megaskill) without deleting it. Omit for normal publishing. Any other value is a validation error.
 
 ### 2. Tone and Formatting
 
--- a/lib/guide-validation.test.ts
+++ b/lib/guide-validation.test.ts
@@ -3,7 +3,7 @@ import assert from 'node:assert';
 import fs from 'node:fs';
 import path from 'node:path';
 import os from 'node:os';
-import { parseExpectations, validateHtmlTags, inventoryGuide, classifyGuide, getSupportedBaseApps } from './guide-validation.ts';
+import { parseExpectations, validateHtmlTags, inventoryGuide, classifyGuide, getSupportedBaseApps, isPublished, validateGuide, type GuideInventory } from './guide-validation.ts';
 
 describe('parseExpectations', () => {
   test('legacy flat format: all bullets treated as mustPass', () => {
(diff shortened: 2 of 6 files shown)


LeaVerou:
Closing in favor of #1027 per https://github.com/GoogleChrome/modern-web-guidance-src/issues/1017#issuecomment-5036086230

PR #1027, opened 2026-07-21
feat: add `draft` frontmatter flag to withhold guides from publishing

### LINKED ISSUE #1017: Frontmatter option to prevent guides from being published

LeaVerou:
Closes #1017 

Adds:
- `draft` frontmatter option. Guides can set `draft: true` (or any truthy value, e.g. `draft: future`) in frontmatter to be excluded from all distribution (search index, README, megaskill) without deleting them. Evals are unaffected — a draft guide with eval artifacts still runs.
The value is intentionally not validated (per https://github.com/GoogleChrome/modern-web-guidance-src/issues/1017#issuecomment-5023911612 )
- `draft` and `isPublished` properties in `GuideInventory`. `isPublished` (real content + not draft) is computed once as a guide-inventory property, consumed by all three build paths.

I also did a little minor refactoring to make the logic and dependencies in `inventoryGuide()` easier to follow, but I can revert it if needed.

Note: I kept the logic as-is to avoid regressions, but it's a little confusing that `isStub` is true even for a fully fleshed out guide…

### DIFF
--- a/lib/guide-validation.ts
+++ b/lib/guide-validation.ts
@@ -45,6 +45,9 @@ interface GuideData {
   [key: string]: any;
 }
 
+/** String `draft` values interpreted as not-a-draft (a quoted/typed-out boolean). */
+const FALSY_DRAFT = new Set(['', 'false', 'no', 'off', '0']);
+
 interface ValidationResult {
   errors: string[];
   data: GuideData;
@@ -223,7 +226,7 @@ export function processGuideInventory(guides: GuideInventory[]): GuideInventoryR
       guideData = validation.data;
       guideBody = validation.body;
 
-      if (isDisciplineSkill || isDisciplineGuide || !hasGuide) {
+      if (isDisciplineSkill || isDisciplineGuide || inv.isStub) {
         // Discipline skills/guides and stubs don't require the same frontmatter as use cases
         guideErrors = guideErrors.filter(e => !e.includes('Missing "web-feature-ids"') && !e.includes('Missing "description"'));
       }
@@ -465,9 +468,14 @@ export function inventoryGuide(dir: string, options?: { useTargetEvals?: boolean
   const { data = {}, content = '' } = guideContent ? matter(guideContent) : {};
   const hasFrontmatter = Object.keys(data).length > 0 || guideContent.startsWith('---');
   const hasContent = content.replace(/<!--[\s\S]*?-->/g, '').trim().length > 0;
-  const isStub = hasFrontmatter;
+  const isStub = hasFrontmatter && !hasContent;
   const hasGuide = hasContent;
-  const draft = data.draft ?? false;
+  // Any truthy `draft` withholds the guide, but treat explicitly falsy-looking
+  // strings (e.g. `draft: "false"`, `draft: no`) as not-draft — quoting a
+  // boolean shouldn't silently unpublish a guide.
+  const draft = typeof data.draft === 'string' && FALSY_DRAFT.has(data.draft.trim().toLowerCase())
+    ? false
+    : data.draft ?? false;
   const isPublished = hasGuide && !draft;
 
   const targetsDir = path.join(dir, TARGETS_DIR);
@@ -545,8 +553,8 @@ export function inventoryGuide(dir: string, options?: { useTargetEvals?: boolean
 export type GuideStatus = 'eval-ready' | 'needs-test' | 'needs-calibration' | 'needs-expectations' | 'stub' | 'incomplete';
 
 export function classifyGuide(inv: GuideInventory): GuideStatus {
-  if (!inv.hasGuide && !inv.isStub) return 'incomplete';
-  if (inv.isStub && !inv.hasGuide) return 'stub';
+  if (inv.isStub) return 'stub';
+  if (!inv.hasGuide) return 'incomplete';
   if (!inv.hasExpectations || inv.expectationsEmpty) return 'needs-expectations';
 
   if (inv.targets && inv.targets.length > 0) {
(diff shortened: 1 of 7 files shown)


rviscomi (approved):
Supportive of this change so LGTMing but it'd be good to get a technical review from @paulirish

lib/guide-validation.ts
```diff
@@ -465,25 +471,14 @@ export function inventoryGuide(dir: string): GuideInventory {
...
-    } else if (hasContent) {
-      hasGuide = true;
-    }
-  }
 
-  const featureIds = guideContent ? (matter(guideContent).data['web-feature-ids'] || []) : [];
+  const { data = {}, content = '' } = guideContent ? matter(guideContent) : {};
+  const hasFrontmatter = Object.keys(data).length > 0 || guideContent.startsWith('---');
+  const hasContent = content.replace(/<!--[\s\S]*?-->/g, '').trim().length > 0;
+  const isStub = hasFrontmatter;
```
rviscomi: Does this treat every guide with frontmatter as a stub? Should we exclude guides with content?

```suggestion
  const isStub = hasFrontmatter && !hasContent;
```

lib/guide-validation.ts
```diff
@@ -465,25 +471,14 @@ export function inventoryGuide(dir: string): GuideInventory {
...
+  const hasGuide = hasContent;
+  const draft = data.draft ?? false;
```
rviscomi: What happens if someone uses a string value like `"false"`, would that be interpreted as a truthy value?

rviscomi (changes requested):

lib/guide-validation.ts
LeaVerou: Apparently, this was the existing logic! See https://github.com/GoogleChrome/modern-web-guidance-src/blob/2a100bbcada89efd0a3f75046383e4689f44a4d0/lib/guide-validation.ts#L481-L482
It surprised me too. It doesn't seem to be a latent bug, just a poorly named variable. When the current code needs to _actually_ test stub-ness, it adds conditions, e.g. see 
https://github.com/GoogleChrome/modern-web-guidance-src/blob/2a100bbcada89efd0a3f75046383e4689f44a4d0/lib/guide-validation.ts#L566-L567

lib/guide-validation.ts
LeaVerou: Yes, but that would only happen if they use explicit quotes in the frontmatter, i.e. `draft: "false"`. Should we special-case this? Any other values to special case?

lib/guide-validation.ts
micahjo7: interesting, I think this can be updated now, to Rick's suggestion. Then, we can update 566-567 to:

```
if (inv.isStub) return 'stub';
if (!inv.hasGuide) return 'incomplete';
```

also https://github.com/GoogleChrome/modern-web-guidance-src/blob/main/lib/guide-validation.ts#L224 can be updated to

```
if (isDisciplineSkill || isDisciplineGuide || inv.isStub) {
```

LeaVerou:
Ok, I just rebased and applied the changes. I also excluded `"false"` from matching as truthy per @rviscomi's feedback (plus a few other obviously negative ones)

micahjo7 (approved):

LeaVerou:
Thanks @micahjo7!
@rviscomi merging is blocked until you approve

rviscomi (approved):

Issue #1029, opened 2026-07-22
Web-feature group codeownership only considers direct featureId → group and ignores nested groups


LeaVerou:
This relates to #1007 #1006 

The current code that determines codeowners from web-feature groups doesn't encode the group hierarchy in `web-feature-groups.generated.json` or even follow it at all. 

As a result making e.g. bramus the owner of `scrolling` doesn't also make him the owner of `scroll-markers` — a child group of `scrolling`!

I'd consider this a bug.

One solution would be to flatten at the time of generating `web-feature-groups.generated.json` but I think ideally we want to encode the group hierarchy in it (only the relevant subset, see #1022) so it can be walked from the feature up until we find the first with a codeowner.

Edit: I added a temp workaround for this in #1031, so this is now less pressing. The workaround flattens all groups, so we don't _miss_ codeowners, but it does mean if there is a more specific codeowner and a broader one, both will be requested.

PR #1031, opened 2026-07-22
Decouple `feature-to-groups.generated.json` from guide content

### LINKED ISSUE #1022: Decouple `web-feature-groups.generated.json` cache from guide content
## Problem

Since the ATL/codeowners work in #1006 (per #1007), a guide-adding PR often must run `node guides/generate-codeowners.ts` and commit the regenerated files, or CI fails on the sync test in `guides/guides-integrity.test.ts`. 

This is a lot of friction for content contributors. :/ 

I will file a separate issue about improving the process for codeowner generation, but currently, this fires even when `CODEOWNERS` doesn't change at all!

The same test also guards `guides/feature-to-groups.generated.json`: a cache holding the groups of **every** `web-feature-id` mentioned in at least one guide (104 entries, almost all owner-less). So adding a guide with any feature not already used elsewhere forces a regen even when `CODEOWNERS` is identical.

There are two issues with this cache:
- It is over-inclusive: the bot only ever consults a feature's groups to check `atlConfig.web_features_groups[group]`, so groups with no codeowner are dead weight
- It only includes web-feature ids mentioned in guides, so a guide adding a new web-feature id requires regenerating `web-feature-groups.generated.json`

## Proposed fix

The proposed fix is twofold:
1. **Only include web-feature groups that have a separate codeowner** in `atls.json`
2. **Add *all* feature ids under these groups**, not just those already used by guides.

Even with the extra feature ids the cache is way lighther (104 → 33 entries today!) and most importantly, it becomes a pure function of `atls.json` and `web-features` data.

Since it is now **completely independent of which guides exist**, no guide PR ever needs to regenerate it. 
It only needs to change when `atls.json` changes or we update the `web-features` package, both actions far more rare than content PRs, and not typically handled by content contributors.

LeaVerou:
Closes #1022.

Implements the fix detailed in #1022 as well as a temp workaround for #1029 which considers all groups. These live in separate commits.

### Notes

- No new package imports in `generate-codeowners.ts`; the `web-features` lookup lives in `serving/lib/baseline.ts` (the access layer) as `getOwnedFeatureToGroups()` + `getAncestorGroups()`.
- Cache output is deterministic (keys + values sorted) and independent of `web-features`' emit order.

### Size

Code: **+22 / −22** (net ~0), excluding the regenerated `feature-to-groups.generated.json` (data: +47 / −213) and comments.

### DIFF
--- a/guides/feature-to-groups.generated.json
+++ b/guides/feature-to-groups.generated.json
@@ -1,209 +1,47 @@
 {
-  "aborting": [],
-  "accent-color": [
-    "css"
-  ],
   "active-view-transition": [
-    "selectors",
     "view-transitions"
   ],
-  "anchor-positioning": [],
-  "autofill": [
-    "forms"
+  "animation-composition": [
+    "animation"
   ],
-  "backdrop": [
-    "selectors"
+  "animations-css": [
+    "animation"
   ],
-  "blocking-render": [
-    "html"
-  ],
-  "calc-size": [
-    "css"
-  ],
-  "canvas-html": [
-    "canvas",
-    "html-elements",
-    "webgl",
-    "webgpu"
-  ],
-  "color-scheme": [
-    "css"
-  ],
-  "contain-intrinsic-size": [
-    "containment"
-  ],
-  "container-anchor-position-queries": [],
-  "container-queries": [
-    "container-queries"
-  ],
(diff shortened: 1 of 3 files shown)


paulirish (approved):
( not looking too close. )

rviscomi (approved):

PR #1033, opened 2026-07-23
Refine ARIA-across-shadow-roots guidance in accessible-web-components

LeaVerou:
Follow-up to review feedback on #738 (guide already merged):

- Write the main guidance assuming the reflected ARIA element properties work; move their support statement to Fallbacks via BASELINE_STATUS with the `api.Element.ariaLabelledByElements` BCD key, since no web-feature id exists for them yet.
- Correct the cross-root behavior: element references resolve *outward* (a shadow-internal control can be labelled by Light DOM / ancestor / slotted content), not inward into a descendant shadow tree. Verified in Chrome against the computed accessibility tree; Reference Target remains the mechanism for the inward direction.


guides/html/web-components/accessible-web-components/guide.md
```diff
@@ -40,15 +40,19 @@ This avoids polluting the DOM with ARIA attributes and prevents consumers from a
...
+- **MANDATORY**: an `id` referenced by `aria-labelledby`, `aria-describedby`, `aria-controls`, etc. resolves only within the referencing element's **own** tree. An `id`-based attribute in the Light DOM cannot point at an `id` inside a shadow tree, and vice versa; the reference is silently ignored.
+- To cross the boundary, use the reflected ARIA *element* properties — `ariaLabelledByElements`, `ariaDescribedByElements`, and siblings — which take direct element references in JS instead of `id`s. They resolve **outward**: an element inside a shadow tree can point at a target in the Light DOM or any ancestor tree (including slotted content), so you can label a shadow-internal control from Light DOM content — something `id` references can't express.
```
castastrophe: How can we determine if an agent would grok this correctly?

castastrophe: I'm also not sure if "and siblings" will pick up the other attributes accurately.

castastrophe (changes requested):
What do you think of a few code examples here? A works and maybe a couple "does not work" examples.

This is a dense and sparsely documented concept; I haven't seen any suggestions to use these when building web components with AI. To that end, I think we need to be extra clear about what these properties can and cannot do and how they should appropriately be used.

guides/html/web-components/accessible-web-components/guide.md
LeaVerou:
```suggestion
- To cross the boundary, use the reflected ARIA *element* properties (e.g. `ariaLabelledByElements`, `ariaDescribedByElements`, etc), which take direct JS element references instead of ids. They resolve **outward**: an element inside a shadow tree can point at a target in the Light DOM or any ancestor tree (including slotted content), so you can label a shadow-internal control from Light DOM content — something `id` references can't express.
```

LeaVerou: > How can we determine if an agent would grok this correctly?

As with every other guide, through the evals?

> I'm also not sure if "and siblings" will pick up the other attributes accurate

Fair, I pushed an edit to make it clearer that these are examples.

LeaVerou:
> What do you think of a few code examples here? A works and maybe a couple "does not work" examples.

Which part of the prose would they be documenting? The outwards thing? I'm worried that the setup required to have them make sense would mean a code example that's disproportionately long for its relevance, wasting tokens.

My vote would be make sure we track this in evals, and if it's not clear, see _how_ agents get it wrong and adjust the guidance accordingly.

> This is a dense and sparsely documented concept; I haven't seen any suggestions to use these when building web components with AI. 
> To that end, I think we need to be extra clear about what these properties can and cannot do and how they should appropriately be used.

New and sparsely documented concepts are exactly what these guides are for 😀  but I agree we should be extra clear.
Which part should be clarified further?

rviscomi:
@castastrophe are you able to take another look at this? cc @jamesnw

castastrophe (approved):
Love this! Thank you for the updates.

PR #1043, opened 2026-07-28
[#1041] Trusted Types Guide and demo

### LINKED ISSUE #686: Trusted types
### web-feature-id

trusted-types

### Feature description

Trusted types allow you to lock down insecure parts of the DOM API and prevent client-side cross-site scripting (XSS) attacks.

### Expected release date

_No response_

### Coding agent assessment

_No response_

### Reference sources

- https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API

**Use cases:**
- [ ] #1041

### LINKED ISSUE #1041: Create guide and evals for the trusted-types use case
Help prevent XSS attacks by ensuring all untrusted content is sanitized before being inserted into the page.

Affected web-feature IDs: [trusted-types](https://webstatus.dev/features/trusted-types)

Use case subdir: [guides/security/trusted-types](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/security/trusted-types)

Related features: #686

**Required files:**
- [x] demo.html
- [ ] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

jamesnw:
Fixes #1041 
Related features: https://github.com/GoogleChrome/modern-web-guidance-src/issues/686


guides/security/trusted-types/guide.md
```diff
@@ -1,6 +1,118 @@
...
+```javascript
+const myPolicy = trustedTypes.createPolicy('my-escape-policy', {
+  createHTML: (toEscape) => {
+    // A simple escaping rule.
+    // In a real app, you might use the Sanitizer API, or a library like DOMPurify.
+    return toEscape
+      .replace(/&/g, '&amp;')
+      .replace(/</g, '&lt;')
+      .replace(/>/g, '&gt;');
+  }
```
jamesnw, line 32: Since this demo policy is not actually sanitizing content, would it be better to have a nonsensical policy (replace all instances of the word "pretzel" for instance) to prevent an agent from actually picking up this as a best practice?

rowan-m: Yeah, makes sense to me.

jamesnw: Done, thanks!

rviscomi (commented):
There's already some existing guidance around Trusted Types in the Security guide: https://github.com/GoogleChrome/modern-web-guidance-src/blob/main/guides/security/security/guide.md#33-trusted-types-enforcement

It's still a good idea to have a standalone guide, but we should make sure the two are closely aligned on the best practices. For example the Security guide recommends rolling out the CSP change as report-only at first, but this guide seems to go straight to enforcement.

After we reconcile any differences, we should add a link to here from Security guide. There's probably also a good bit of content we could trim from the Security guide to avoid redundancy and ensure agents read this guide for the full details.

jamesnw:
> There's already some existing guidance around Trusted Types in the Security guide: https://github.com/GoogleChrome/modern-web-guidance-src/blob/main/guides/security/security/guide.md#33-trusted-types-enforcement
> 
> It's still a good idea to have a standalone guide, but we should make sure the two are closely aligned on the best practices. For example the Security guide recommends rolling out the CSP change as report-only at first, but this guide seems to go straight to enforcement.
> 
> After we reconcile any differences, we should add a link to here from Security guide. There's probably also a good bit of content we could trim from the Security guide to avoid redundancy and ensure agents read this guide for the full details.

I updated the guide in https://github.com/GoogleChrome/modern-web-guidance-src/pull/1043/commits/79881ed7a6481ae7e82b3c6497b2e2a84d50f0ec and simplified the skills in https://github.com/GoogleChrome/modern-web-guidance-src/pull/1043/commits/188928ef9fb9994f27356c9219bc582abb324ec5.

rowan-m (approved):

guides/security/security/guide.md
```diff
@@ -224,23 +224,13 @@ For static/cached HTML (SPAs) where a per-response nonce is not possible, use ha
...
 
-```javascript
-if (window.trustedTypes && trustedTypes.createPolicy) {
-  const policy = trustedTypes.createPolicy('escapePolicy', {
-    createHTML: str => str.replace(/</g, '&lt;').replace(/>/g, '&gt;')
-  });
-  el.innerHTML = policy.createHTML(untrustedString);
-}
-```
+- **DO**: Consult the dedicated [Trusted Types Guide](../trusted-types/guide.md) for full implementation details, including policy creation and rollout strategies.
```
rviscomi: Could you make this a `GUIDE_REF`?

skills-src/security/SKILL.md
rviscomi, line 1: This file is actually obsoleted by the discipline guide. Could you delete it?

rowan-m:
@jamesnw conflicts looks pretty trivial, I think we just want both. I'm happy to go ahead and resolve that, if that works? Or just jump in :D

jamesnw:
@rowan-m I've resolved the conflict- ready to merge?

PR #1180, opened 2026-08-01
Use exponential w/ jitter when retrying

connorjclark:
I got unlucky running evals w/ many workers and hit an OpenRouter limit enough times to get no results for a couple guides. Retrying w/ an exponential backoff + random jitter would have prevented that.

### DIFF
--- a/harness/run_suite.ts
+++ b/harness/run_suite.ts
@@ -456,15 +456,30 @@ env.PATH = \`${targetDir}:\${env.PATH}\`;
 const start = Date.now();
 let result;
 let attempts = 0;
-const maxAttempts = 3; // 1 initial attempt + 2 retries
+const maxAttempts = 5; // 1 initial attempt + 4 retries with exponential backoff
+const baseDelay = 15000; // 15 seconds base delay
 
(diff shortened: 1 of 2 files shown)


harness/run_suite.ts
```diff
@@ -456,15 +456,30 @@ env.PATH = \`${targetDir}:\${env.PATH}\`;
 const start = Date.now();
 let result;
 let attempts = 0;
-const maxAttempts = 3; // 1 initial attempt + 2 retries
+const maxAttempts = 5; // 1 initial attempt + 4 retries with exponential backoff
```
micahjo7, line 459: any particular reason to increase to 5?

micahjo7 (approved):

harness/run_suite.ts
connorjclark: just to increase the likelihood of a intermittent rate limiting expiring before we give up. it might take a couple minutes. but i didn't want to increase the base timeout to ensure it tries again at least a few minutes later

with just 3 it would be 15, 30, 60s - just under 2 minutes. Might be long enough. but wanted to be more lenient.

PR #1186, opened 2026-08-04
feat(harness): trajectory parsing and summary (1/3)

psiarkiewicz:
## Summary
Part 1 of 3 (splitting #1010).

Introduces standardized trajectory parsing and auto-generates `trajectory_summary.json` (schema v2.0) across all supported evaluation harness agents.

- **Parsers**: Claude Code, Gemini CLI, Codex CLI, and Jetski (SQLite / chat log / MCP logs).
- **Harness Hooks**: Invokes `generateNormalizedTrajectory()` at the end of agent runs.
- **Codex Trajectory Viewer**: Adds pre-rendering and step ID tagging.


harness/agents/claude-code-agent.ts
```diff
@@ -140,6 +141,12 @@ async function run() {
     }
 
     exportClaudeCodeTrajectories(workDir, targetDir);
+    
+    try {
+      await generateNormalizedTrajectory(targetDir, Agents.CLAUDE_CODE, getSuiteConfig().serving);
+    } catch (e: any) {
+      console.error("Failed to generate normalized trajectory:", e.message);
+    }
```
micahjo7, line 135: dont think we need try-catch here. same for the other agent files

harness/lib/codex-trajectory-viewer.ts
```diff
@@ -1,6 +1,104 @@
 export function generateCodexTrajectoryHtml(logData: any[]): string {
```
micahjo7, line 1: So this file's changes are unrelated to the `trajectory_summary.json` changes, is that correct? It is just to improve the Codex trajectory html?

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,721 @@
...
+  if (str.length > maxLen) {
+    return str.slice(0, maxLen) + '... [truncated]';
+  }
+  return str;
+}
+
+/**
+ * Parses Claude Code session JSONL files into a normalized TrajectorySummary.
+ */
+export function parseClaudeTrajectory(logData: any[], serving: string): TrajectorySummary {
```
micahjo7: these agent specific methods can be moved into their respective files

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,721 @@
...
+    message?: string;
+    output?: any;
+    exitCode?: number;
+  };
+}
+
+export interface TrajectorySummary {
+  schemaVersion?: string;
+  agent: string;
+  serving: string;
```
micahjo7: we probably don't need to capture serving anymore?

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,721 @@
...
+  outcome?: {
+    status: 'success' | 'error';
+    message?: string;
+    output?: any;
+    exitCode?: number;
+  };
+}
+
+export interface TrajectorySummary {
+  schemaVersion?: string;
```
micahjo7: we can probably remove `schemaVersion`, instead we should consolidate with the existing logic we have for jetski cli (related comment to [harness/agents/jetski-cli-agent.ts](https://github.com/GoogleChrome/modern-web-guidance-src/pull/1186/changes#r3737890128))

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,721 @@
...
+          }
+        }
+      }
+    }
+  }
+  return results;
+}
+
+/**
+ * Synthesizes a normalized TrajectorySummary for Jetski/MCP using .db files, modern-web.log, and chat_log.txt.
```
micahjo7: fix `Jetski/MCP`. and we can specify Jetski CLI in the method name too

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,721 @@
...
+    const sessionFiles = allSessionFiles.sort((a, b) => {
+      const aSub = a.includes('-subagents-');
+      const bSub = b.includes('-subagents-');
+      if (aSub && !bSub) return 1;
+      if (!aSub && bSub) return -1;
+      return a.localeCompare(b);
+    });
+
+    // To add another agent in the future, check agentName or session files here and invoke the corresponding parser.
+    if (agentName === Agents.JETSKI || agentName === Agents.JETSKI_CLI || agentName.toLowerCase().includes('jetski') || (!sessionFiles[0] && fs.existsSync(targetDir) && fs.readdirSync(targetDir).some(f => f.endsWith('.db')))) {
```
micahjo7: i think we can clean these up so that we only need to check for `agentName === Agents.____`

micahjo7: we can also remove the sessionFiles[0] check below and pass in the targetDir directly similar to how parseJetskiTrajectory works

also, we will want to capture the subagent trajectories as well, as those files often contain most of the agent's work (for Claude esp)

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,721 @@
...
+        summary = parseGeminiTrajectory(logData, serving);
+      } else if (agentName === Agents.CODEX_CLI || agentName.toLowerCase().includes('codex')) {
+        summary = parseCodexTrajectory(logData, serving);
+      } else {
+        console.warn(`[TrajectoryParser] Warning: Unknown agent "${agentName}". Attempting generic Codex/standard trajectory parsing. To add another agent, register a parser in generateNormalizedTrajectory.`);
+        summary = parseCodexTrajectory(logData, serving);
+      }
+    } else {
+      summary = await parseJetskiTrajectory(targetDir, serving);
+    }
```
micahjo7: remove these fallbacks

harness/agents/jetski-cli-agent.ts
```diff
@@ -136,10 +137,9 @@ async function run() {
     exportTrajectories(conversationsDir, '*.db', targetDir);
 
     try {
-      const summary = parseJetskiCliSession(targetDir);
-      writeTrajectorySummary(targetDir, summary);
-    } catch (e) {
-      console.warn(`Failed to generate trajectory summary in ${targetDir}:`, e);
+      await generateNormalizedTrajectory(targetDir, Agents.JETSKI_CLI, getSuiteConfig().serving);
```
micahjo7: I recently added jetski cli trajectory parsing and creation of `trajectory_summary.json` to extract data and populate in the dashboard (model, guides/tools, tokens) in https://github.com/GoogleChrome/modern-web-guidance-src/pull/1042

this overwrites that file- could we merge that logic and update the collection methods in this file?

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,721 @@
...
+          }
+          const promptFound = extractInitialPromptFromLogs(Array.isArray(logData) ? logData : [], chatText);
+          if (promptFound) summary.initialPrompt = truncateMessage(promptFound, 1000);
+        }
+
+      const outPath = path.join(targetDir, 'trajectory_summary.json');
+      fs.writeFileSync(outPath, JSON.stringify(summary, null, 2), 'utf8');
+      console.log(`[TrajectoryParser] Generated trajectory summary: ${outPath}`);
+    } else {
+      console.warn(`[TrajectoryParser] No trajectory files found in ${targetDir} to summarize.`);
```
micahjo7: I don't think we store the token in the runtime.json file.

Also, I tihnk we can get the initail prompt from the evals.json file

harness/lib/trajectory-parser.ts
micahjo7: (with the option to extract shared logic between them into this file)

micahjo7 (commented):
Wondering what kind of validation you've done on this logic? Could you share examples of the generated `trajectory_summary.json` files? 

or, better would be to add tests (expand `trajectory-parsing.test.ts`, or split into separate tests `<agent>-trajectory-parsing.test.ts`) to make sure all the fields we want get populated correctly. It would be really great to actually invoke the agents with basic prompts and test the logic on those responses- then, we would catch any issues with model updates/variance of the produced trajectory files (which already happens quite often). This might be more overhead, could be added in the future if so.

micahjo7:
One of my concerns is that we sort of have similar logic between our (Codex and Claude) trajectory HTML generation and now this parsing `trajectory_summary` extraction logic.

A question I have about what the desired output is in the dashboard's `compare` view after all these PRs are completed. Do you have a screenshot of that view? 

Is it feasible to just directly use the trajectory HTML files that we already generate there?

paulirish:
> A question I have about what the desired output is in the dashboard's compare view after all these PRs are completed. Do you have a screenshot of that view?

i sent it in chat.

> Is it feasible to just directly use the trajectory HTML files that we already generate there?

Well it's probably like possible to bring the two into alignment more.   But... the cross-agent comparison wouldnt really work for this UI if we didnt rely on normalized step data.

paulirish:
@psiarkiewicz if its fine with you ill take over these branches :) really appreciate these contributions!

psiarkiewicz:
> @psiarkiewicz if its fine with you ill take over these branches :) really appreciate these contributions!

for sure!  Thanks!

harness/agents/claude-code-agent.ts
paulirish: yup fixed.

harness/lib/codex-trajectory-viewer.ts
paulirish: yup. unrelated drive-by improvements to Codex HTML formatting that don't 100% belong to this branch, topically.

harness/lib/trajectory-parser.ts
paulirish: Yeah, I agree. Pawel's coding agent probably moved them here because otherwise we end up with circular dependencies.  I know this because mine fought me pretty hard to avoid circular dependencies, but I agree the co-location is worth it and the cycle isn't a problem for us

harness/lib/trajectory-parser.ts
paulirish: Good call. cleaned this up

harness/lib/trajectory-parser.ts
paulirish: Done

harness/lib/trajectory-parser.ts
paulirish: Done

harness/lib/trajectory-parser.ts
paulirish: cleaned up routing and now rely on the enum.

 subagents are now handled.  that was one of the bigger parts of my edit.  We added subagent session parsing (subagent-*.jsonl / session-*-subagents-*.jsonl) and merged all subagent steps with monotonic timestamp ordering.

harness/agents/jetski-cli-agent.ts
paulirish: yup! fixed.

harness/lib/trajectory-parser.ts
paulirish: *   **Tokens**: You're right—`runtime.json` only tracks execution duration and status. We removed that dead fallback; token counts are parsed directly from raw agent session logs.
*   **Prompt**: We deleted the fragile chat-log regex scraper (`extractInitialPromptFromLogs`). Downstream consumers (dashboard and compare CLI) already have access to the exact task prompt via `evals.json` / `task.md`.

harness/lib/trajectory-parser.ts
paulirish: yup. now we route on the `Agents` enum.

paulirish:
btw I've consolidated the schema:

1.  Removed `schemaVersion`.
2.  `TrajectorySummary` now natively includes all the evaluation metrics (`model`, `tokenUsage`, `retrievedGuides`, `fileReadGuides`, `toolsUsed`) alongside `steps`.
3.  `parseJetskiTrajectory` now calls `parseJetskiCliSession` under the hood to preserve all metrics.
4.  `collection.ts` and `guidance_validation.ts` now read directly from `trajectory_summary.json` across all agents (with fallback to legacy extractors if missing).

harness/lib/collection.ts
```diff
@@ -15,7 +15,26 @@ function isTargetAppPresent(targetFile: string, targetPkgJson: string, targetPat
...
+function isNodeError(err: unknown): err is NodeJS.ErrnoException {
+  return err instanceof Error && 'code' in err;
+}
+
+function isEnoent(err: unknown): boolean {
+  return isNodeError(err) && err.code === 'ENOENT';
+}
+
 export function extractModelFromResults(resultsDir: string, agent: string): string {
+  const summaryPath = path.join(resultsDir, 'trajectory_summary.json');
```
micahjo7, line 14: for this method, we [pass](https://github.com/GoogleChrome/modern-web-guidance-src/blob/main/harness/evaluate.ts#L108) in the top level suite result dir (e.g. `nightly-2026-08-10_17-18-03-claude_code/`), which will not have the `trajectory_summary.json` in it (this method is designed to iterate over all the run dirs and collect the most used model across all trajectories)

it's alittle confusing currently because the other extraction methods (like `extractTokenUsageFromResults` below) pass in the specific run dir (e.g. `nightly-2026-08-10_17-18-03-claude_code/1/accessible-error-announcement/task/unguided/`) and only retrieve the data for that one run

i think what would be best is to update this method so that it also runs for the specific run dirs rather than suite dir. we would remove it from `evaluate.ts` and add it to `collectTaskRunEntry` in `collection.ts` where the other extraction methods run. then we have a model per trajectory, and in `metrics.ts` we add the logic to get the most used model and set it in the summary which we use to populate in `evals.json`

harness/agents/claude-code-agent.ts
micahjo7: continuation from [this comment](https://github.com/GoogleChrome/modern-web-guidance-src/pull/1186/changes#r3777710085), but if we made that refactoring, we would simply remove "true" here (and for all the other agent files) and can remove the `recursive` param/functionality from `getSessionFiles` completely.

harness/lib/collection.ts
```diff
@@ -15,7 +15,26 @@ function isTargetAppPresent(targetFile: string, targetPkgJson: string, targetPat
...
+  try {
+    const summary = JSON.parse(fs.readFileSync(summaryPath, 'utf8'));
+    if (summary.model && summary.model !== 'unknown') {
+      return summary.model;
+    }
+  } catch (err) {
+    if (!isEnoent(err) && !(err instanceof SyntaxError)) throw err;
+  }
+
+  // Legacy Fallback
```
micahjo7: i think we can remove this fallback here and below?

harness/lib/guidance_validation.ts
```diff
@@ -9,16 +9,42 @@ import { collectClaudeGuidesFromTrajectory, collectClaudeToolsFromTrajectory } f
...
+      return {
+        retrievedGuides: summary.retrievedGuides || [],
+        fileReadGuides: summary.fileReadGuides || []
+      };
+    }
+  } catch (err) {
+    if (!isEnoent(err) && !(err instanceof SyntaxError)) throw err;
+  }
+
+  // Legacy Fallback (when summary is missing or old format)
```
micahjo7: i think we can remove this fallback and below now

harness/lib/trajectory-parser.ts
micahjo7: not a big deal I think but these are still in the trajectory-parser file

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,965 @@
...
+      // Merge MCP logs if it is desktop Jetski
+      if (agentName === Agents.JETSKI) {
+        const mcpGuides = extractGuidesFromMcpLog(targetDir);
+        summary.retrievedGuides = [...new Set([...(summary.retrievedGuides || []), ...mcpGuides])];
+        if (!summary.toolsUsed) {
+          summary.toolsUsed = ['modern-web-guidance'];
+        } else if (!summary.toolsUsed.includes('modern-web-guidance')) {
+          summary.toolsUsed.push('modern-web-guidance');
+        }
+      }
```
micahjo7: don't think we need this

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,965 @@
...
+      try {
+        toolCalls.push(JSON.parse(line));
+      } catch {}
+    }
+  }
+  return toolCalls
+    .filter(call => call.tool === 'get_best_practices' && Array.isArray(call.result))
+    .flatMap(call => call.result.map((r: any) => r.id || ''))
+    .filter(Boolean);
+}
```
micahjo7: can remove

.gitignore
```diff
@@ -20,6 +20,7 @@ test-app-results/
 guides/**/test-results/
 guides/**/test-app-results/
 guides/**/.last-run.json
+tmp-eval-fixtures/
```
micahjo7: do we need this?

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,965 @@
...
+      stepNumber: 0,
+      thought: 'Completed task implementation and summarized changes',
+      action: {
+        type: 'other',
+        name: 'respond_to_user',
+        params: { response: truncateMessage(chatText, 300) }
+      },
+      outcome: { status: 'success' }
+    });
+  }
```
micahjo7: I don't think we need these fallbacks?

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,965 @@
...
+              }
+            });
+          } else if (obj.CommandLine) {
+            let actType: NonNullable<StandardizedStep['action']>['type'] = 'run_command';
+            let actName = 'run_command';
+            let params: any = { command: truncateMessage(obj.CommandLine, 150) };
+            if (/(?:modern-web-guidance|modern-web|\bgd\b)/.test(obj.CommandLine) && (obj.CommandLine.includes('search') || obj.CommandLine.includes('retrieve'))) {
+              actType = 'web_search';
+              actName = 'get_best_practices';
+              const qMatch = obj.CommandLine.match(/(?:search|retrieve)\s+["']?([^"'\n]+)["']?/i);
```
micahjo7: should we have this MWG specific logic here? isn't the logic here just to get the trajectory steps and normalize?

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,965 @@
...
+        const parts = stripped.split(/agent[-_]/);
+        const key = parts.length > 1 ? parts[parts.length - 1] : stripped.replace(/^(?:subagent-|session-)/, '');
+        subagentsMap[key] = Array.isArray(logData) ? logData : ((logData as any)?.messages || []);
+      } catch (e) {
+        console.warn(`[TrajectoryParser] Failed to parse subagent file ${file}:`, e);
+      }
+    }
+
+    if (agentName === Agents.JETSKI || agentName === Agents.JETSKI_CLI) {
+      summary = await parseJetskiTrajectory(targetDir, agentName);
```
micahjo7: we can probably just drop JETSKI and remove the `agentName` param from parseJetskiTrajectory

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,965 @@
...
+      action: {
+        type: 'other',
+        name: 'respond_to_user',
+        params: { response: truncateMessage(chatText, 300) }
+      },
+      outcome: { status: 'success' }
+    });
+  }
+
+  const legacy = parseJetskiCliSession(dirPath);
```
micahjo7: we could potentially combine logic from `parseJetskiCliSession` into this method so we don't do two separate parses of the db files

i guess the same applies to the other agents as well, but just a thought

harness/lib/collection.ts
```diff
@@ -15,7 +15,26 @@ function isTargetAppPresent(targetFile: string, targetPkgJson: string, targetPat
   return fs.existsSync(targetFile) || fs.existsSync(targetPkgJson) || (targetPatchFile ? fs.existsSync(targetPatchFile) : false);
 }
 
+function isNodeError(err: unknown): err is NodeJS.ErrnoException {
+  return err instanceof Error && 'code' in err;
+}
+
+function isEnoent(err: unknown): boolean {
+  return isNodeError(err) && err.code === 'ENOENT';
+}
```
micahjo7: nit: combine these into one method, and but in a shared helper?

the same methods exist in `trajectory-parser.ts` and `guidance-validation.ts`

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,965 @@
...
+        ]);
+        summary.retrievedGuides = guides.retrievedGuides;
+        summary.fileReadGuides = guides.fileReadGuides;
+        summary.toolsUsed = tools;
+        summary.model = model;
+        summary.tokenUsage = tokenUsage;
+      } else {
+        console.warn(`[TrajectoryParser] Warning: Unknown agent "${agentName}". Attempting generic Codex/standard trajectory parsing. To add another agent, register a parser in generateNormalizedTrajectory.`);
+        summary = parseCodexTrajectory(logData, subagentsMap);
+      }
```
micahjo7: nit: don't need a fallback here

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,965 @@
...
+    } else {
+      console.warn(`[TrajectoryParser] Warning: No session files found for non-Jetski agent "${agentName}". Cannot parse trajectory.`);
+    }
```
micahjo7: also don't need this?

harness/agents/claude-code-agent.ts
```diff
@@ -121,6 +122,7 @@ async function run() {
     }
 
     exportClaudeCodeTrajectories(workDir, targetDir);
+    await generateNormalizedTrajectory(targetDir, Agents.CLAUDE_CODE, userPrompt);
```
micahjo7: idea: move these into "finally" block so that if `runCliAgentCommand` fails, then we have more artifacts to debug and capture trajectory data until failure occurred

same for other agents if we want to make this change

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,965 @@
...
+    return 'other';
+  }
+
+  if (actionName.includes('retrieve') || (actionName.includes('get_best_practices') && actionParamsStr.includes('retrieve')) || actionParamsStr.includes('retrieve')) {
+    return 'guide_retrieval';
+  }
+
+  if (actionName.includes('search') || actionName.includes('get_best_practices') || actionName.includes('query_guidance') || actionParamsStr.includes('search')) {
+    return 'skill_search';
+  }
```
micahjo7: suggestion: move these below the `write` and `replace` if-block to prevent false positives (e.g. calling `replace_file_content` on a method called `retrieveUserData`)

harness/tests/trajectory-parser-normalization.test.ts
```diff
@@ -0,0 +1,490 @@
+import test from 'node:test';
```
micahjo7, line 1: add Codex CLI test?

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,965 @@
...
+ * truncated during normalization to keep `trajectory_summary.json` lightweight for fast
+ * UI timeline rendering and prompt budgeting in downstream comparison tools.
+ */
+export interface StandardizedStep {
+  stepNumber: number;
+  timestamp?: string;
+  subagentId?: string;
+  thought?: string;
+  action?: {
+    type: 'tool_call' | 'api_call' | 'web_search' | 'read_file' | 'write_file' | 'run_command' | 'other';
```
micahjo7: 'tool_call' | 'api_call' i think are unused types (not part of `mapToolType`)

harness/lib/trajectory-parser.ts
```diff
@@ -0,0 +1,965 @@
...
+        if (block && block.type === 'tool_result' && block.tool_use_id) {
+          const stepIdx = toolUseToStepMap.get(block.tool_use_id);
+          if (stepIdx !== undefined && steps[stepIdx]) {
+            const outText = typeof block.content === 'string' ? block.content : JSON.stringify(block.content);
+            steps[stepIdx].outcome = {
+              status: block.is_error ? 'error' : 'success',
+              message: truncateMessage(outText)
+            };
+
+            const match = typeof block.content === 'string' ? block.content.match(/agentId:\s*([a-zA-Z0-9_-]+)/) : null;
```
micahjo7: the `block` could be a string or array. we can instead use `outText` computed above:

`const match = outText.match(/agentId:\s*([a-zA-Z0-9_-]+)/);`

harness/lib/codex-trajectory-viewer.ts
micahjo7, line 1: we handle subagent trajectories for Codex in the `trajectory-parser` in this PR, but not for the traj HTML here. just an fyi, created b/546140187 to track that work

harness/lib/trajectory-parser.ts
paulirish: yeah totally agree, thanks. moved all agent-specific parsers directly into their respective agent files leaving trajectory-normalizer.ts as just the shared schema, dispatcher, and action classifier.

harness/lib/collection.ts
paulirish: makes total sense, thanks! moved this out of evaluate.ts and into collection.ts so it runs per task directory alongside the other extractors.

harness/agents/claude-code-agent.ts
paulirish: yup, removed `true` and cleaned up `getSessionFiles`.

harness/lib/collection.ts
paulirish: good catch, stripped out the legacy fallbacks here.

harness/lib/guidance_validation.ts
paulirish: removed! reads straight from `trajectory_summary.json` now.

harness/lib/trajectory-parser.ts
paulirish: thx done

.gitignore
paulirish: nope, removed it.

harness/lib/trajectory-parser.ts
paulirish: yeah totally agreeeeeee, having it in the shared normalizer felt out of place. 

I moved that logic into `jetski-cli-agent.ts`, so `trajectory-normalizer.ts` stays 100% generic while each agent file parses its own search/retrieve commands consistently.

harness/lib/trajectory-parser.ts
paulirish: done, dropped the unused enum and simplified the signature to just take the dir path.

harness/lib/trajectory-parser.ts
paulirish: combined these so we only do a single pass through the sqlite db now.   

tried to keep the diff minimal but....... also.. impossible. either way i ensured no test regressions.. hard to see at the PR level but i did it right with commit order.

harness/lib/collection.ts
paulirish: deduplicated and moved into agent-shared.ts.

harness/lib/trajectory-parser.ts
paulirish: cleaned up and removed the fallback guessing.

harness/agents/claude-code-agent.ts
paulirish: great idea, wrapped these in `finally` blocks across all the agent runners so we preserve artifacts on failures.

harness/lib/trajectory-parser.ts
paulirish: nice catch! moved code mutation checks above search/retrieve keywords and added unit tests covering methods like `retrieveUserData`.

harness/tests/trajectory-parser-normalization.test.ts
paulirish: added! created a dedicated codex-cli-parsing.test.ts with lots of coverage

harness/lib/trajectory-parser.ts
paulirish: cleaned up the unused action types.

harness/lib/trajectory-parser.ts
paulirish: good simplification, switched to `outText` directly.

harness/lib/codex-trajectory-viewer.ts
paulirish: appreciate that.

harness/lib/trajectory-parser.ts
paulirish: removed those fallbacks! we unified `parseJetskiTrajectory` with `parseJetskiCliSession` so steps and metadata are extracted directly from the sqlite database in a single pass without falling back to `chat_log.txt` or `modern-web.log`.

paulirish:
super appreciate the thorough review! pushed an update that addresses all of this:

*   co-located all 5 agent parsers directly in their respective agent files, keeping `trajectory-normalizer.ts` focused on schema, dispatching, and action classification.
*   moved `extractModelFromResults` out of `evaluate.ts` into `collection.ts` so it extracts per-run directory, and stripped out all legacy fallback paths.
*   moved `exportTrajectories` and `generateNormalizedTrajectory` into `finally` blocks across all agent runners so trajectory artifacts are preserved when runs fail.
*   reordered `categorizeAction` to check code mutations before search/retrieve keywords (preventing false positives on methods like `retrieveUserData`), and simplified the subagent parsing regexes.
*   split out tests into per-agent suites, including a new `codex-cli-parsing.test.ts` and test coverage for the strict `modern-web-guidance && retrieve` filter.

and handled the conflicts. oh so many conflicts.

paulirish:
Oh right and I added type declaration files for claude and codex... had to develop them myself since the internet doesnt have them yet.

but they're good. :) 

you're welcome, internet.

harness/lib/trajectory-normalizer.ts
```diff
@@ -0,0 +1,317 @@
...
+  loadPiLogs,
+  extractPiMetadata
+} from '../agents/pi-agent.ts';
+
+
+export const TRAJECTORY_SUMMARY_FILE = 'trajectory_summary.json';
+
+const TRAJECTORY_GLOB = 'session-*.{json,jsonl}';
+
+export function getSessionFiles(dir: string, recursive = false): string[] {
```
micahjo7: nit: we can remove recursive functionality now

harness/lib/guidance_validation.ts
```diff
@@ -1,94 +1,34 @@
...
-
-  console.warn(`Unknown agent ${agent} for skills collection`);
   return { retrievedGuides: [], fileReadGuides: [] };
 }
 
-export async function collectGuidanceToolsUsed(dir: string, serving: Serving, agent: string): Promise<string[]> {
-  if (serving === Serving.MCP || agent === Agents.JETSKI) {
-    if (fs.existsSync(path.join(dir, MODERN_WEB_LOG_FILE))) {
-      return ['modern-web-guidance'];
+export async function collectGuidanceToolsUsed(dir: string, _serving?: Serving, _agent?: string): Promise<string[]> {
```
micahjo7: we could remove `_serving` and `_agent` now? and from `collectGuidesUsed` above

and `_serving` can be removed from the agent files too (`collectCodexGuidesFromTrajectory` still uses it, but that should probably be modified to remove it too)

harness/lib/trajectory-normalizer.ts
```diff
@@ -0,0 +1,317 @@
...
+export interface SubagentMetadata {
+  id: string;
+  agent?: string;
+  purpose?: string;
+  totalSteps?: number;
+}
+
+export interface TrajectorySummary {
+  agent: string;
+  serving?: string;
```
micahjo7: nit: don't think we need serving here anymore either

harness/lib/collection.ts
```diff
@@ -2,40 +2,37 @@ import { glob } from "glob";
...
-    return extractGeminiCliModel(resultsDir);
-  } else if (agent === Agents.JETSKI_CLI) {
-    return extractJetskiCliModel(resultsDir);
-  } else if (agent === Agents.CLAUDE_CODE) {
-    return extractClaudeCodeModel(resultsDir);
-  } else if (agent === Agents.CODEX_CLI) {
-    return extractCodexCliModel(resultsDir);
-  } else if (agent === Agents.PI) {
-    return extractPiModel(resultsDir);
+export function extractModelFromResults(resultsDir: string, _agent?: string): string {
```
micahjo7: same comment, can remove `_agent` here and elsewhere in this file

micahjo7 (approved):
just some small comments on the unused params. with this update the tests and callers will likely need to be updated too.

paulirish:
+78, -127!   man that's a review from heaven.   thank you!

harness/tests/jetski-cli-parsing.test.ts
```diff
@@ -0,0 +1,231 @@
...
+  try {
+    const dbPath = path.join(tempDir, 'session-123.db');
+    const db = new DatabaseSync(dbPath);
+    db.exec(`
+      CREATE TABLE steps (idx INTEGER, step_type INTEGER, status INTEGER, metadata BLOB, step_payload BLOB);
+      CREATE TABLE gen_metadata (idx INTEGER, data BLOB);
+    `);
+
+    const mockActionJson = JSON.stringify({
+      CommandLine: 'gd retrieve "validate-input-after-interaction"',
```
micahjo7: nit just for clarity
```suggestion
      CommandLine: 'npx -y modern-web-guidance@latest retrieve "validate-input-after-interaction"',

File: harness/agents/jetski-cli-agent.ts [hunk 34]
```diff
@@ -218,31 +273,128 @@ export function parseJetskiCliSession(dirPath: string): TrajectorySummary {
...
+              const params: Record<string, any> = { command: obj.CommandLine || obj.toolAction };
+
+              if (/(?:modern-web-guidance|modern-web|\bgd\b)/.test(obj.CommandLine) && (obj.CommandLine.includes('search') || obj.CommandLine.includes('retrieve'))) {
+                actType = 'web_search';
+                actName = 'get_best_practices';
+                const qMatch = obj.CommandLine.match(/(?:search|retrieve)\s+["']?([^"'\n]+)["']?/i);
+                if (qMatch) {
+                  params.query = qMatch[1];
+                }
+              }
```
micahjo7: I think this is doing something that the other agents aren't doing. the other agents don't do this rename of the actType,name, and the  query. I think we can remove this

micahjo7 (commented):
sorry, a couple more comments I noticed while reviewing https://github.com/GoogleChrome/modern-web-guidance-src/pull/1187

harness/agents/codex-cli-agent.ts
```diff
@@ -156,115 +202,278 @@ export function extractCommandsFromCodexItem(obj: any): string[] {
...
+        const commands = extractCommandsFromCodexItem(entry);
+
+        let actionType: NonNullable<StandardizedStep['action']>['type'] = mapToolType(cmdName);
+        let actionName = cmdName;
+        let params: any = undefined;
+
+        if (commands.length > 0) {
+          actionType = 'run_command';
+          actionName = commands[0];
+          params = { cmd: commands[0] };
```
micahjo7: use `command` here for consistency with the other agents? instead of `cmd`

paulirish: done. also

im gonna do some further cleanup of the stanardized step params in 2/3

PR #1189, opened 2026-08-04
[#1026] Add web app origin migration use case

### LINKED ISSUE #1026: Web App Origin Migration
### web-feature-id

app-migration

### Feature description

The migrate_to and migrate_from web app manifest members move an installed app from one origin to another, within the same site. They preserve the user's installation settings on the device, like shortcuts.

### Expected release date

Chrome M150

### Coding agent assessment

_No response_

**Use cases:**
- [ ] #1203

jamesnw:
I am unsure of where this should go- HTML seemed least incorrect? Perhaps a new `web-app` folder?

### DIFF
--- a/guides/html/migrate-web-app-origin/demo.html
+++ b/guides/html/migrate-web-app-origin/demo.html
@@ -1,0 +1,10 @@
+This demo will not be functional, but will contain the correct files for migration:
+
+/demo.html > Site with <link> to manifest.json
+/manifest.json > Simple manifest with `migrate_to` as ordered map
+/.well-known/web-app-origin-association > Pointer to new site
+
+/new-site/demo.html > Site with <link> to new-site/manifest.json
+/new-site/manifest.json > Simple manifest with `migrate_from`
+
+Fallback will be to recommend a UI element instructing them to reinstall.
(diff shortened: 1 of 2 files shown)


guides/html/migrate-web-app-origin/demo.html
```diff
@@ -0,0 +1,10 @@
+This demo will not be functional, but will contain the correct files for migration:
```
jamesnw, line 1: @rviscomi  Will this cause testing issues? This doesn't work outside of a transition between same-site domains. Content is testable, but not behavior.

rviscomi: There are some changes to the eval pipeline coming, which would obsolete the need for demo files.

Could you summarize what we'd need to have to properly test this use case?

cc @micahjo7

micahjo7: we would only need the guide and expectations files

LeaVerou (approved):
Demo aside, the guide stub lgtm. 

I think `html` is fine for now. A `html/pwa` or `html/manifest` category would make sense too. No strong opinion either way!

guides/html/migrate-web-app-origin/demo.html
rviscomi: Ok let's merge this through and deal with the demo file later, if needed

PR #1191, opened 2026-08-04
[#1037] Local Network access restrictions use case

### LINKED ISSUE #1037: Local network access restrictions for WebTransport
### web-feature-id

local-network-access

### Feature description

Restricts the ability to make requests to the user's local network using WebTransport, gated behind a permission prompt.

A local network request is any request from a public website to a local IP address or loopback, or from a local website (e.g. intranet) to loopback. Gating the ability for websites to perform these requests behind a permission reduces the ability of sites to use these requests to fingerprint the user's local network.

This permission is restricted to secure contexts.

This work is adding to the Local Network Access Restrictions work here: https://chromestatus.com/feature/5152728072060928

### Expected release date

Chrome 144

### Coding agent assessment

_No response_

**Use cases:**
- [ ] #1495

jamesnw:
I don't think this warrants separate use cases for local vs. loopback. Since the previous behavior is that it works without requesting permissions, do we need to show a fallback that manually checks with the user first? Or view this as strictly a progressive enhancement?

### DIFF
--- a/guides/security/local-network-access/demo.html
+++ b/guides/security/local-network-access/demo.html
@@ -1,0 +1,5 @@
+This demo will have buttons that attempt to contact an address on a local network, triggering the `local-network` permission, and `localhost`, triggering the `loopback-network` permission.
+
+It will show how to handle when users deny the permission.
+
+The requirement for Secure Context and separate locally running servers may make this challenging to test.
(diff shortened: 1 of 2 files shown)


jamesnw:
@rowan-m Does this use case make sense? I'll note there's overlap with #1316, but I don't think we need to address IWAs with this.

rowan-m:
- Agreed on this being a single use case
- Agreed on not addressing IWAs here, completely separate and much more niche audience
- I think treat it like a progressive enhancement, i.e. the only active developer action I can see is the `targetAddressSpace` option for `fetch`, which would just be ignored otherwise.

rowan-m (approved):

PR #1194, opened 2026-08-04
Fix downgrade in workflow files

micahjo7:
Fixing accidental downgrade that happened in https://github.com/GoogleChrome/modern-web-guidance-src/pull/1181.

### DIFF
--- a/.github/workflows/guide-sync.yml
+++ b/.github/workflows/guide-sync.yml
@@ -38,7 +38,7 @@ jobs:
         uses: pnpm/action-setup@0977fd99725f1db4007ccb2928dbb4e90d06cc86 # v6.0.10
 
       - name: Setup Node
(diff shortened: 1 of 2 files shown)


.github/workflows/guide-sync.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         uses: pnpm/action-setup@0977fd99725f1db4007ccb2928dbb4e90d06cc86 # v6.0.10
 
       - name: Setup Node
-        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
```
paulirish, line 41: sounds good. thanks for fixing the downgrade

i guess the hashes are okay. didnt realize it was dependabot who did it first.  if its doing it then github i guess decided that's best practice.  sure.

paulirish (approved):

PR #1206, opened 2026-08-07
Remove content ATLs from codeowners

### LINKED ISSUE #1023: Can we automate codeowners regeneration for PRs?
## Problem

Since the ATL/codeowners work in #1006 (per #1007), a guide-adding PR often must run `node guides/generate-codeowners.ts` and commit the regenerated files (`CODEOWNERS` and `web-feature-groups.generated.json`), or CI fails on the sync test in `guides/guides-integrity.test.ts`. 

This is a lot of friction for content contributors. :/ 

I filed #1022 for the more straightforward improvement of not needing to regenerate `web-features-groups.generated.json` (which _also_ makes it smaller!). That should already reduce how many content PRs require this regeneration to only PRs that add **new** guides in areas with a separate codeowner. 

But I _think_ we can remove that friction too.

My understanding is that currently, two systems resolve the *same* ATL ownership from `category` + `web-feature-ids` + [`guides/atls.json`](../blob/main/guides/atls.json):

1. The [`atl-triage`](../blob/main/guides/atl-triage.ts) bot, which dynamically requests the correct feature-expert reviewer on guide PRs
2. The [`CODEOWNERS` build script](../blob/main/guides/generate-codeowners.ts) which materializes that same resolution into static per-guide lines in `CODEOWNERS`

## Proposed Fix

The main question is **can we rely on the `atl-triage` bot to auto-request PR _reviewers_, or we want feature area experts to be _required approvers_ too?**

If the answer is _yes_, we can **skip the CI check entirely**, rely on `atl-triage` to request the right reviewers, and regenerate automatically post-merge [^1]. This is **optimal UX for contributors**, but it all hinges on the answer to the question above.

[^1]: My agent claims there is precedent for this in this repo: it's the **same** mechanism as README in [`publish.yml`](../blob/main/.github/workflows/publish.yml), but I have not fact-checked this.

If we also need them to be **required approvers,** then it seems the only way to avoid the manual effort for contributors would be to **auto-commit to the PR**, only for PRs that need it. This doesn't work for fork PRs, but at least it would work for PRs from the main contributors. We could keep the CI check as a fallback for these cases (and for any other reason the auto-commit fails). I don't love this, but it does seem the lesser of two evils.

WDYT @paulirish @micahjo7 @rviscomi?

### LINKED ISSUE #1044: `CODEOWNERS` should have no codeowner 😅
_(Ok I think this is my favorite issue title I've ever filed 😅)_

Currently, @paulirish and @micahjo7 are listed as default codeowners of everything that has no more specific rule:

```
# Default owners for implementation and infra
* @paulirish @micahjo7
```

As a result, any PR that needs to regenerate codeowners also requests a review from @paulirish and @micahjo7, even if it is a purely content PR. 

Hopefully we can eventually fix #1023 so that no PR author _needs_ to regenerate `CODEOWNERS` but only one of the two proposed solutions would actually fix this (the one where we rely on triage bot, regenerating CODEOWNERS post-merge).

I believe it should be an easy fix: just add a blank entry for the `CODEOWNERS` file itself. Any other files that should be owner-less? E.g. `package.json`, the lockfile etc?

rviscomi:
Fixes #1044
Fixes #1023 

This change removes content ATLs from the codeowners file, so creating a new guide no longer has a dependency on updating it.

While I was in there, I also added support for transcluded feature guides, so a PR that modifies one will auto-assign the ATL for that feature and/or the categories in which that guidance is used.

I removed @paulirish and @micahjo7 as global codeowners, otherwise they would own all content files as well by default. They're still named on all eval files.


LeaVerou (approved):
LGTM!

Not sure the feature macros buy us anything, since those are typically already included in `web-feature-ids`, but it can't hurt either.

micahjo7 (approved):

PR #1256, opened 2026-08-10
Support temporary web-feature IDs

rviscomi:
Especially for an emerging web feature, an official feature ID may not be set by the time we're ready to start writing guidance for it. This causes a couple of problems:

- Guidance authors are forced to circumvent the CI check, which requires guides (even stubs) to have a `web-feature-ids` field, by setting it to an empty array.
- The corresponding issue for the feature is unable to be automatically linked to the guide, because there is no ID to connect them.

As a temporary workaround, this PR proposes a new `tmp-*` feature ID prefix, which will automatically pass the CI check, unless the back half of the feature ID is real. If we're using a tmp ID for a live feature, the CI check fails so that we can update it to the real thing.

Guide frontmatter and feature-level issues can temporarily refer to the same `tmp-*` ID until the official feature ID is minted.

This change was motivated by https://github.com/GoogleChrome/modern-web-guidance-src/pull/968 and https://github.com/GoogleChrome/modern-web-guidance-src/issues/668


LeaVerou:
I like this better than what I've been doing until now, which was a [YAML comment](https://raw.githubusercontent.com/GoogleChrome/modern-web-guidance-src/refs/heads/main/guides/ui-atoms/icons/guide.md).
But if it's completely arbitrary, I worry that we'll end up with numerous different aliases (e.g. `tmp-link-parameters`, `tmp-linked-parameters` etc), and it's a bit too easy to just come up with any number of these, completely speculatively.

Alternate proposal: What if we maintain a data structure of "approved" temporary web-feature ids in a file, and CI merges these with the official ones before proceeding?

Pros:
- We can even attach metadata to them and pilot them just like the real ones. We could even make it mandatory to link to the open issue for adding this feature to web-features to ensure we've done due diligence and aren't just using this system as an escape hatch
- Centralization ensures no divergence, everyone uses the same temporary feature id for the same thing
- We can use real web-feature ids, so no change needed when they become official

Cons:
- `tmp-*` makes it obvious that these are temp at the point of usage, no-one will be confused trying to look up nonexistent `web-feature ids.
- Increased friction for adding a new one (though this may be a feature, see above)

rviscomi:
Thanks, I like that idea.

> * `tmp-*` makes it obvious that these are temp at the point of usage, no-one will be confused trying to look up nonexistent `web-feature ids.

To address this, I merged your registry idea with the `tmp-` prefix, so it keeps the best of both worls.

paulirish (approved):

PR #1257, opened 2026-08-11
Use custom property values for interest delay properties

jakearchibald:
Reference the custom properties via var() instead of repeating the literal values, so the value only needs changing in one place.


guides/ui-behaviors/interest-triggered-action-previews/guide.md
```diff
@@ -77,8 +77,8 @@ When using the polyfill the CSS API changes slightly for the `:interest-source`
 /* Adjust the start and end delay for interest invokers */
 [interestfor] {
   --interest-delay-start: 0.2ms;
-  interest-delay-start: 0.2ms;
+  interest-delay-start: var(--interest-delay-start);
   --interest-delay-end: 0.1ms;
```
rviscomi, line 81: Not your change, but something I just noticed, should these units be `s` rather than `ms`?

jakearchibald: Oh, I didn't even notice. Almost certainly.

jakearchibald:
I couldn't find other instances of this pattern in the repo

rviscomi (approved):
LGTM. Thanks and welcome to the project @jakearchibald!

I've added @LeaVerou as the ATL for this category for visibility, but I know she's away and may not be able to approve. Since this is such a small fix I think it's safe to merge anyway.

LeaVerou:
This LGTM too fwiw! Thanks @jakearchibald!

PR #1263, opened 2026-08-12
Workflow to update web-features

### LINKED ISSUE #1211: Baseline status changes not reflected in MWG dist repo
In the dist repo's [content-based-styling.md](https://github.com/GoogleChrome/modern-web-guidance/blob/684ab9d7c6b78fc2cd5677912d874397cb2e5dfa/skills/modern-web-guidance/guides/css/content-based-styling.md#L77-L78) guide, it still says:

>Baseline status for :has(): Newly available. It's been Baseline since 2023-12-19.

However, this feature has been Widely available since web-features version 3.32.0, which was released on June 25.

After that date, I would have expected the next MWG release to update to version 3.32.0 and regenerate the Baseline status message to say that it's widely available.

I found that this is creating some confusion from an actual MWG user in https://github.com/timothyfroehlich/PinPoint/pull/1852:

>The same live tool reports `:has()` as **Newly available since 2023-12-19**. Thirty months from that date is 2026-06-19, which has passed — so either MWG's snapshot is stale or the tier really hasn't flipped. Every file in this PR lists `:has()` as Widely available. I did not change it: substituting a possibly-stale tool answer for a claim that is probably now correct would be trading one wrong assertion for another. Filed as a follow-up bead.

rviscomi:
Fixes #1211 

- pins package files to a specific `web-features` version instead of `"latest"`
- daily workflow to check for newer versions
- automatically opens a PR to update package files, if one doesn't already exist
- automatically leaves a comment in the PR to point out any outdated web feature IDs being used in the project (guides, feature-level docs, or ATL configs)
    - may be outdated if it's a `tmp-<id>` and `<id>` now exists, or
    - the feature ID was just renamed or split

---

_Example of a PR comment:_

⚠️ **Action Required**: This daily `@web-platform-dx/web-features` update introduced official feature IDs or platform record shifts (`moved` / `split`) affecting guidance in this repository.

### 1. Graduated Temporary Feature IDs (`tmp-*`)
- **`tmp-streaming-api` → `streaming-api`** (Primary feature available upstream)
  ↳ *Action:* Rename or delete `features/tmp-streaming-api.md`
  ↳ *Referenced in:*
    - `guides/user-experience/streaming-api/guide.md`
    - `guides/atls.json`
- **⚠️ `tmp-foo` was split upstream into:** `foo-a`, `foo-b` — inspect affected files to assign appropriate sub-feature ID(s)
  ↳ *Referenced in:*
    - `guides/ui-atoms/foo/guide.md`

### 2. Regular Feature Records Split or Moved Upstream
- **⚡ `scroll-snap` was split upstream into:** `scroll-snap-type`, `scroll-snap-align` — assign appropriate sub-feature ID(s).
  ↳ *Referenced in:*
    - `guides/css/scroll-snap/guide.md`
    - `guides/css/carousel/guide.md`
    - `guides/atls.json`
- **`old-prop` was moved upstream to:** `new-prop` — update reference to target ID.
  ↳ *Referenced in:*
    - `guides/css/layout/guide.md`

Before merging this PR, please resolve the items flagged above and re-run guide validation checks.


.github/workflows/update-web-features-daily.yml
```diff
@@ -0,0 +1,115 @@
+name: Daily web-features Update
+
+on:
+  schedule:
+    - cron: '0 5 * * *'
```
paulirish: Doing this outside of Dependabot gives me a little bit of concern, but I'm open to try it .


Since we only publish once a week, I think we can align with that schedule. So how about updating this every Thursday ?    If you still want it more than once a week, then how about twice a week ? :) 

At this slower pace, we don't need to worry as much about not making pull requests because ones already exist, etc.

paulirish (approved):

.github/workflows/update-web-features-daily.yml
rviscomi: SGTM, updated to Monday/Thursday

PR #1266, opened 2026-08-12
Create a registry for origin trial features

rviscomi:
Depends on #1263  (trying a PR stack for the first time, let's see how it works)

Creates `features/origin-trials.json`, a registry of features that are known to still be part of an origin trial. In a future PR, we can use this registry to make sure that guides that depend on these features can be omitted from MWG by default, unless the user opts in.

I initially considered adding a `origin-trial: true` field to the guide frontmatter (hence the branch name) but I didn't like how that required maintenance across all guides that touch the OT feature. Having a registry centralizes the OT status for all features.

This also incorporates a validation check in the newly added `guides/validate-after-web-feature-update.ts` script, which flags any features that we say are part of an active origin trial, but have actually just shipped to a stable browser. This is so that we don't hold back guidance for features any longer than necessary.

This PR doesn't do any validation to make sure that the OT is still active, but that was my reasoning for including `chromestatus_url` in the OT JSON, so that it could be used in the future. Maybe we'd query the Chrome Status API to get the OT milestone window and check that against the browser release dates provided by `web-features`. We could also hardcode the final milestone in the OT JSON, but I thought there'd be a chance of it going stale (more stale than chromestatus). Maybe it's a case of YAGNI.

### DIFF
--- a/features/origin-trials.json
+++ b/features/origin-trials.json
@@ -1,23 +1,11 @@
 {
   "canvas-html": {
-    "chromestatus_url": "https://chromestatus.com/feature/5172548013916160",
-    "milestones": {
-      "chrome_first": 148,
-      "chrome_last": 154
-    }
+    "chromestatus_url": "https://chromestatus.com/feature/5172548013916160"
   },
   "declarative-webmcp": {
-    "chromestatus_url": "https://chromestatus.com/feature/5117755740913664",
-    "milestones": {
-      "chrome_first": 149,
-      "chrome_last": 156
-    }
+    "chromestatus_url": "https://chromestatus.com/feature/5117755740913664"
   },
   "document-modelcontext": {
-    "chromestatus_url": "https://chromestatus.com/feature/5117755740913664",
-    "milestones": {
-      "chrome_first": 149,
-      "chrome_last": 156
-    }
+    "chromestatus_url": "https://chromestatus.com/feature/5117755740913664"
   }
 }
--- a/features/origin-trials.json
+++ b/features/origin-trials.json
@@ -1,0 +1,23 @@
+{
+  "canvas-html": {
+    "chromestatus_url": "https://chromestatus.com/feature/5172548013916160",
+    "milestones": {
+      "chrome_first": 148,
+      "chrome_last": 154
+    }
+  },
+  "declarative-webmcp": {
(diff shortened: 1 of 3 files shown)


features/origin-trials.json
```diff
@@ -0,0 +1,11 @@
+{
+  "canvas-html": {
```
paulirish, line 2: fyi months ago when this topic last came up i made this  https://github.com/paulirish/chromestatus

it relies on a combo of web-features + chromestatus + `content-chromeorigintrials-pa.googleapis.com/v1/trials`

and... while the chromestatus data is quite messy ..   i managed to get some decent truth out of it.

Here's the relevant query:

```
pnpm audit:ot-symbols

> @paulirish/chromestatus@0.1.25 audit:ot-symbols /Users/paulirish/code/chromestatus
> node scripts/print-active-ot-symbols.ts

==================================================================
       AUTHORITATIVE ACTIVE ORIGIN TRIAL INVENTORY
==================================================================

[Section 1]: Verified Mapped WebDX Symbols (9 unique identifiers mapped across 11 feature records):

[
  "speculation-rules",
  "digital-credentials",
  "declarative-webmcp",
  "local-network-access",
  "install",
  "canvas-html",
  "navigator-install",
  "opaquerange",
  "languagemodel"
]

------------------------------------------------------------------

[Section 2]: Unmapped Granular Platform Extensions (15 specific ChromeStatus proposals currently lacking dedicated WebDX shortcodes):

- Declarative Shadow DOM 'shadowrootadoptedstylesheets'
- Enhanced Canvas TextMetrics
- WebRTC Diagnostic Logging API
- Container Timing
- Speculative load measurement
- Prefetch activation beacon
- Autofill event
- Long Animation Frames style duration
- Connection Allowlists
- Permissions Policy: focus-without-user-activation
- CPU Performance API
- Email Verification Protocol
- Gamepad Event-Driven Input API
- WebAssembly Custom Descriptors
- Speculation Rules - moderate viewport heuristics controls
```

and i just updated the package so if you want to pull it in ... something like this would work

```ts
import otMapping from '@paulirish/chromestatus/data/ot-mapping.json' with { type: 'json' };
```

(but its always good to revalidate the list)  :)

paulirish (approved):

rviscomi:
@micahjo7 @paulirish is this PR still useful, given the updated direction for OT features? Or should I close it?

PR #1270, opened 2026-08-13
Make `tabindex=-1` on skip links a caveat

### LINKED ISSUE #874: Accessibility guide: Reconsider `tabindex="-1"` recommendation on skip link targets
The accessibility guide currently recommends adding `tabindex="-1"` to skip link targets in multiple places across section 1 (Content Navigability and Structure) and section 5 (Keyboard and Focus Management). 

There are two issues with this recommendation:

- It is likely no longer necessary in modern browsers. Manuel Matuzović [tested skip links without `tabindex="-1"` on targets](https://www.matuzo.at/blog/2026/skip-links-tabindex) across major browsers and screen readers on macOS, Windows and Android, concluding it is safe to remove in most setups, thanks to [sequential focus navigation starting point](https://html.spec.whatwg.org/multipage/interaction.html#sequential-focus-navigation).

- It introduces a usability regression. With `tabindex="-1"` on `main`, clicking on any non-interactive area inside `main` makes `main` itself the focus starting point. So, the next tab press sends focus to the first focusable element in `main`, not the next one after where the user clicked. The gov.uk team [removed `tabindex` from `main`](https://github.com/alphagov/govuk_elements/pull/534) for this reason.

@knowler noted in PR #597 that `tabindex="-1"` may still help with some screen magnifiers. This is why I feel a caveat in the same style as the Safari `role="list"` one in section 2 (Semantic HTML and ARIA) would be more accurate than the current unconditional advice.

knowler:
Resolves #874

Changes the guidance for skip links to not use `tabindex=-1` on the target. It instead provides a caveat for the screen magnifier case.

In my testing, `tabindex=-1` was not a consistent solution for screen magnifiers (i.e. they would often perform the same in both scenarios). The height of the document vs. the viewport also was a factor.

I will do a final pass of testing before I mark this as ready for review.

I could probably thin out the caveat a bit.

### DIFF
--- a/guides/accessibility/accessibility/guide.md
+++ b/guides/accessibility/accessibility/guide.md
@@ -20,7 +20,12 @@ Keep these principles in mind throughout:
 - **Place all content within landmarks**: Wrap each relevant part of the page in `<header>`, `<nav>`, `<main>`, `<aside>`, and `<footer>` elements so assistive-tech users can jump between regions.
 - **Structure main content with headings**: Use `<h1>`–`<h6>` sequentially (no jumping `<h1>` → `<h4>`) so screen-reader users get a navigable outline.
 - **Use lists for repeated, contiguous content**: `<ul>`/`<ol>` give assistive tech a count up front and let users skip the entire group.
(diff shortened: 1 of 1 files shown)


rviscomi:
@knowler is this ready for review?

Issue #1070, opened 2026-07-29
Focusgroup


shoaibdilawar:
### web-feature-id

focusgroup

### Feature description

Gives authors the ability to declaratively give composite widgets arrow key navigation, a guaranteed tab stop, and last-focused memory, replacing hand-coded roving tabindex scripts.
Example:
<div focusgroup="toolbar wrap" aria-label="Formatting">
  <button>Bold</button>
  <button>Italic</button>
  <button>Underline</button>
</div>

### Expected release date

Chrome 150

https://chromestatus.com/feature/5637601087193088

knowler:
Microsoft has [a polyfill for this](https://github.com/microsoft/polyfills/tree/main/packages/focusgroup). If it works well, we should use it and promote the use of `focusgroup` in cases where the feature works (e.g. toolbars, tablists, menus, etc.). Models know how to implement a roving tabindex/aria-activedescendant, but if we can guide it to focusgroup for most cases, I think that’d be better.

knowler:
There’s an issue when combining `focusgroup` with `popover`: the focusgroup’s semantics get clobbered by `popover`. Workaround is to nest the focusgroup inside of the popover. https://issues.chromium.org/issues/564673920

Impacts: #1244 #1435

PR #1275, opened 2026-08-13
[#1212] Button Group Guide

### LINKED ISSUE #1212: Create guide and evals for the button-group use case
Build a button group: a set of options (mutually exclusive or not) laid out horizontally and presented as visually connected buttons.

Affected web-feature IDs: [customizable-select](https://webstatus.dev/features/customizable-select)

Use case subdir: [guides/forms/button-group](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/forms/button-group)

Related features: #51

**Required files:**
- [x] demo.html
- [x] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

jamesnw:
Addresses #1212


guides/forms/button-group/guide.md
```diff
@@ -1,7 +1,138 @@
...
+  <div class="wrapper">
+    <!-- options here -->
+  </div>
+</select>
+<span id="multi-hint">Multiple selection allowed.</span>
+```
+
+### Fallbacks
+
+{{ FEATURE_FALLBACKS("customizable-select") }}
```
jamesnw: I'd welcome thoughts on what the fallback could be here- I don't think we can get a similar style and behavior falling back to non-customizable select. As is, the `.wrapper` is causing the listbox to show as empty. Perhaps we fallback to the native listbox UI, but remove the wrapper with JS?

guides/forms/button-group/guide.md
```diff
@@ -2,6 +2,129 @@
 name: button-group
 description: "Build a button group: a set of options (mutually exclusive or not) laid out horizontally and presented as visually connected buttons."
 web-feature-ids:
-  - customizable-select
-#  - mixins
+  - base-select
```
LeaVerou: https://web-platform-dx.github.io/web-features-explorer/features/customizable-select/ — works
https://web-platform-dx.github.io/web-features-explorer/features/base-select/ — 404

For mixins, my intent was that you frequently want these to share button styles so you'd define a mixin for button styles and apply it across all of these. But perhaps that's best split into a separate guide.

jamesnw: Gemini kept wanting to change `customizable-select` for some reason, and I missed it on my final check. Sorry! 

I agree this could be a good future use case for mixin, but given the current support, think a separate guide makes sense.

guides/forms/button-group/guide.md
LeaVerou: Darn, that's quite a problem. I'd try to avoid depending on JS for the wrapper if all other JS is PE. I wonder if we can work around the bug without a wrapper 🤔

LeaVerou:
I'm still reviewing but one potential issue is that left/right arrow keys do nothing. Should we use JS as PE so that these work? I haven't found a CSS way to influence the order of these (this is an even bigger problem with grid selects…)

jamesnw:
> I'm still reviewing but one potential issue is that left/right arrow keys do nothing. Should we use JS as PE so that these work? I haven't found a CSS way to influence the order of these (this is an even bigger problem with grid selects…)

I was wondering about this as well. Per the [Open UI explainer,](https://open-ui.org/components/customizable-select.explainer/#keyboard-behavior) the left and right arrow buttons intentionally do nothing. I also didn't find a CSS only solution, and agree on the JS as PE path.

guides/forms/button-group/guide.md
jamesnw: I did a bit more digging. I think this is a regression introduced in Firefox 149 (when customizable select was released behind a feature flag.) It is still there in 154. This may be a fairly big hole in the progressive enhancement story for customizable select.

Luckily, this specific use case looks like it doesn't actually require a wrapper. 

This is what it looks like currently without customizable select-
image

I don't think we can hack it to be horizontal without customizable select. I can clean it up so it looks good vertically, or fallback to an unstyled select. Neither quite seems to be a good fallback.

jamesnw:
I've removed the wrapper from this, although the guide will need to be updated to match- I wanted to wait until we see what happens with Firefox 149-154, where a wrapper breaks the select.

guides/forms/button-group/guide.md
jamesnw: A bit more testing-

- A wrapper is required in Chrome 148 (`display: flex` doesn't seem to apply to `<select>`s). It is not required in Chrome 151.
- Firefox 148 falls back correctly with a wrapper. Firefox 149-current can not have a wrapper. (Confirmed with OpenUI)

Given this, the demo has a wrapper, and shows how to remove it with JS.

I also simply removed all custom styling for the demo if `appearance: base-select` is not supported.

LeaVerou: I have a [hacky way](https://codepen.io/leaverou/pen/yyBLZwz) to make it horizontal in Firefox and older Chrome, but it doesn't work in Safari. And by "doesn't work" I mean the control is unreadable, so if we use this, we need a way to detect whether the technique would work and not apply any of it if it wouldn't, we can't just rely on PE. And I'm not sure if it's too hacky to use in general…

LeaVerou:
Given the current market share of customizable-select, I think we need a better fallback than just a listbox :( As it currently stands, this doesn't pass the test of "would I be happy if my agent did this?". 😞 
Not sure how to handle this. Perhaps JS that replaces it with a WC for Safari? 🤔 Any ideas welcome!

bramus:
> Given the current market share of customizable-select, I think we need a better fallback than just a listbox :(

One route we could take here is to feature detect support for `appearance: base-select` and then conditionally show the markup of either the customizable select or either the fallback markup _(that most likely uses radio buttons, or perhaps some buttons with manually controlled `aria-pressed` attributes)_. But like you said on the call: authors would end up shipping unnecessary markup + styles to all users :-/

Admittedly the markup would be limited in size (so perhaps it’s OK?). And maybe the styles can be shared across both by using similar classnames or writing selectors that target both markup variants.

guides/forms/button-group/guide.md
jamesnw: > * Firefox 148 falls back correctly with a wrapper. Firefox 149-current can not have a wrapper. (Confirmed with OpenUI)

Update from Keith (browser dev)- Only 153, 154, and 155 are impacted. 156 should be fixed.

jamesnw:
I evaluated https://github.com/josepharhar/customizable-select-polyfill, but don't think that makes sense here. 

So, I added a fallback to a group of buttons. There's little redundant styles, but there's a fair amount of redundant DOM. A reusable component would likely try to be smarter here, but that's too framework-dependent to make sense to prescribe here.

I'm also not entirely certain what to do with the wrapper element as it's fairly buggy. Earlier versions of Chrome have a bug that require it. Firefox 153-155 have a bug that can not have a wrapper element, but I think it's only if the feature flag is enabled. The per-the-spec and modern implementation does not require it, but it's not easily feature detectable.

LeaVerou:
The fallback is basically a whole separate implementation 😞 
I’m not sure what to do here. Perhaps the right call is to mark it as `draft` until we have a better solution? 🤔

LeaVerou:
Per today's discussion, I would suggest instead of a parallel implementation as a fallback, to say don't use this when both of the following are true:
1. It is essential that this looks like a button group (i.e. a listbox fallback would not be acceptable)
2. We must support browsers that don't support customizable-select

LeaVerou:
> In left-to-right, top-to-bottom writing modes, selects handle Up Arrow and Down Arrow keystrokes, but do not handle Left Arrow and Right Arrow keystrokes. Because the button group is horizontal, users will expect to be able to use the Left and Right Arrow keys to change the focus in the same way as the Up and Down arrows. Use JavaScript to manage this behavior.

This sounds like we're basically fighting against the writing mode, which makes me wonder if what I presented earlier as a hack is actually that hacky:

```css
select {
	writing-mode: vertical-lr;
}

select:dir(rtl) {
	writing-mode: vertical-rl;
}

option {
	writing-mode: horizontal-tb;
}
```

This not only makes it horizontal (without `display: flex`) in **_both_ Chrome and Firefox**, it also makes keyboard navigation work correctly out of the box without any JS! The downside is it's pretty broken in Safari (without the `@supports (appearance: base-select)`, which we'd remove to support FF), and I cannot think of a CSS-only way to detect that.
It does feel a little icky, but I wonder if it actually breaks something in some way I haven't thought of. Do you have any thoughts? I'll also reach out to some i18n folks I know and ask them too. I suspect it will turn out that there is a reason this feels wrong, but can't hurt to check!

LeaVerou:
>  I'll also reach out to some i18n folks I know and ask them too. I suspect it will turn out that there is a reason this feels wrong, but can't hurt to check!

@fantasai said:

> It’s hacky, but using writing-mode here is unlikely to break things. Seems less fragile than using JS, which is less likely to capture platform interaction patterns correctly.

(She also emphasized the caveat that she's an i18n expert, but not on input.)

jamesnw:
The further we've dug, the more I'm realizing that this is stretching the select element beyond what it is capable of doing. Does customizable select truly solve the problems that you note with radio button implementations (https://lea.verou.me/blog/2022/07/button-group/)? 

I'm wondering if we should be more specific about the use case we're solving (is this solely a presentational button group, a segmented control, or a toolbar)? Then, if the solution would be novel for LLMs, we would add it here.

bramus:
I agree with the sentiment that this might be overdoing things and that, perhaps, we should consider the `<select>` approach more as a “clever hack” right now instead of a best practice that we want to share with folks. We can always later revisit but at this very moment, it seems a bit too early …

LeaVerou:
I'm conflicted. On one hand I see the argument about being hacky, on the other hand, these are issues we'll face with any non-vertical layout, and I believe non-vertical layouts are absolutely in scope for `customizable-select`.
It does feel like laying out a listbox horizontally *should* be possible without too much ado. 
It does seem however that we've hit a platform limitation. It _should_ be possible to do this without either JS or `writing-mode`. Perhaps `focusgroup` v2 will help, though it still can't be applied via CSS AFAIK.

In terms of next steps, I don't want to leave this open forever, nor close it and throw away all this work by @jamesnw.
Now that #1027 is merged, one possible way forwards is to merge this with `draft: blocked` (or just `draft: true`) and add a note about the issues, linking to this PR.

jamesnw:
I added the `draft` key. (I don't view it as wasted effort- we learned a lot about this pattern and the limits of customizable selects through this).

LeaVerou (approved):

knowler:
The existing `focusgroup` feature (v1) that’s already in Chrome (and is almost spec — pending a PR to HTML) would help with the ARIA toolbar use case. That’s the only one that actually requires custom keyboard controls. The others should just be a part of the regular sequential focus navigation (reachable via tab/shift-tab).

```html
<!-- ARIA toolbar role is applied as minimum role -->
<div focusgroup="toolbar wrap" aria-label="Text formatting">
  <!-- These are navigable with arrow keys -->
  <button type=button>Bold</button>
  <button type=button>Italic</button>
  <button type=button>Underline</button>
</div>
```

The exclusive pressed/selected state part would be outside of what focusgroup does.

Cross-linking #1070

rviscomi:
Merging this through with "draft: blocked" set

PR #1276, opened 2026-08-13
Don't reassign an ATL who was unassigned

rviscomi:
If an ATL gets auto-assigned to an issue or PR, they should be able to remove their assignment without it automatically getting reassigned.

This change takes assignment removals into consideration before adding an ATL. Also cleaned up some weird tests.

cc @bramus


rviscomi:
Sorry, accidentally merged too soon! @paulirish LMK if you have any comments and I'll open a new PR

PR #1358, opened 2026-08-18
feat(guides): enforce character and line limit with baseline snapshot

TravenReese:
Implements 8,192 character (AGY CLI) and 256 line (Codex CLI) limit enforcement in guides-integrity.test.ts to prevent guidance truncation across agent harnesses. Existing oversized guides are recorded in a oversized-guides-baseline.json baseline file to allow them at current sizes while blocking any growth or new oversized guides.

### DIFF
--- a/guides/guides-integrity.test.ts
+++ b/guides/guides-integrity.test.ts
@@ -41,7 +41,7 @@ describe('Guides Validation (Single Source of Truth)', () => {
 
     it(`validates ${relativeDir}`, () => {
       const result = processGuideInventory([guide]);
-      
+
       if (result.hasError) {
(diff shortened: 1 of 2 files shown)


micahjo7:
From our investigation we found that both AGY and Codex find a way to retrieve even the longest guide we have (`stack-drill-down`) and there doesn't seem to be any negative effect on the evals.

AGY is chaining a series of `retrieve` calls, so that's not ideal, but I think the bug you filed and the directive to redirect to file should be enough (we should validate that the redirecting works). Enforcing these guide size limits might be unnecessarily limiting to the guides.

PR #1359, opened 2026-08-18
feat(scripts): add scan-guide-lengths script

TravenReese:
Adds a utility script (scripts/scan-guide-lengths.ts) to scan and display character and line count metrics across all guides in the repository. Supports filtering or inspecting specific guides by name or path, as well as an --update-baseline flag to automatically synchronize guides/oversized-guides-baseline.json. Stacked on top of #1358.

### DIFF
--- a/scripts/scan-guide-lengths.ts
+++ b/scripts/scan-guide-lengths.ts
@@ -1,0 +1,129 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import { parseArgs } from 'node:util';
+import matter from 'gray-matter';
+import { scanAllGuides, getGuideMarkdownPath } from '../lib/guide-validation.ts';
+import { rootDir } from '../lib/paths.ts';
+
+const MAX_RECOMMENDED_CHARS = 8192; // AGY CLI character limit
+const MAX_RECOMMENDED_LINES = 256;  // Codex CLI line limit
+
+const { values, positionals } = parseArgs({
+  args: process.argv.slice(2),
+  options: {
+    guide: {
+      type: 'string',
+      short: 'g',
+    },
+    oversized: {
+      type: 'boolean',
+      short: 'o',
+    },
+    'exceeds-limits': {
+      type: 'boolean',
+    },
+    help: {
+      type: 'boolean',
+      short: 'h',
+    },
+  },
+  allowPositionals: true,
+});
+
+if (values.help) {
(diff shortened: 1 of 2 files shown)


scripts/scan-guide-lengths.ts
```diff
@@ -0,0 +1,170 @@
...
+  console.log('\nOptions:');
+  console.log('  -g, --guide <path|name>  Scan a specific guide by path, name, or substring (default: all guides)');
+  console.log('  -o, --oversized          Only show guides exceeding limits (> 8192 chars or > 256 lines)');
+  console.log('      --exceeds-limits     Alias for --oversized');
+  console.log('      --update-baseline    Update guides/oversized-guides-baseline.json with current sizes of guides exceeding limits (> 8192 chars or > 256 lines)');
+  console.log('  -h, --help               Show help');
+  process.exit(0);
+}
+
+if (values['update-baseline']) {
```
micahjo7: let's remove the update-baseline flag references and functionality since this script is information and we don't need `guides/oversized-guides-baseline.json` anymore

scripts/scan-guide-lengths.ts
```diff
@@ -0,0 +1,170 @@
...
+const showOversizedOnly = values.oversized || values['exceeds-limits'];
+
+if (showOversizedOnly) {
+  displayResults = results.filter(
+    r => r.totalChars > MAX_RECOMMENDED_CHARS || r.lineCount > MAX_RECOMMENDED_LINES,
+  );
+}
+
+// Sort by character count descending
+displayResults.sort((a, b) => b.bodyChars - a.bodyChars);
```
micahjo7, line 122: sort by totalChars?

micahjo7 (commented):
since the other PR is closed, can you retarget this to `main`

also, we should add line to `scripts` in package.json which runs this script

scripts/scan-guide-lengths.ts
TravenReese: bodyChars represents the character count of the markdown body excluding the YAML frontmatter. Since frontmatter is stripped out during the build and retrieval step and isn't injected into the agent's context window, sorting by bodyChars gives us a more accurate reflection of the actual payload/token size delivered to the agent.

micahjo7: ah gotcha. then, should we update the check above in line 120 to also use `bodyChars`?

TravenReese: Yes, good catch done!

micahjo7 (approved):

PR #1365, opened 2026-08-21
[#1248] Scrollspy guide

### LINKED ISSUE #1248: Create guide and evals for the scrollspy use case
Highlight the currently visible section of a page in a navigation menu

Affected web-feature IDs: [scroll-marker-targets](https://webstatus.dev/features/scroll-marker-targets), [scroll-target-group](https://webstatus.dev/features/scroll-target-group)

Use case subdir: [guides/ui-components/scrollspy](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-components/scrollspy)

**Required files:**
- [x] demo.html
- [x] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

jamesnw:
Fixes #1248


guides/ui-components/scrollspy/expectations.md
```diff
@@ -0,0 +1,7 @@
+- Navigation links use fragment identifiers (e.g., `href="#section-id"`) that match the unique IDs of their target content sections.
+- The navigation container has `scroll-target-group: auto` applied to enable native browser tracking.
+- Navigation links use the `:target-current` pseudo-class for active state styling (preferably using `:where()` to manage specificity with fallbacks).
```
bramus: The recommendation to use `:where()` here (and in the code) seems a bit weird. Why would you not want `nav a:target-current` to be more specific than `nav a`?

If it’s for parsing reasons, then I’d say it’s for parsing reasons, but I would also, instead, use `:is()` to retain specificity.

guides/ui-components/scrollspy/guide.md
```diff
@@ -7,3 +7,128 @@ web-feature-ids:
...
+```html
+<nav>
+  <a href="#intro">Introduction</a>
+  <a href="#features">Features</a>
+  <a href="#pricing">Pricing</a>
+</nav>
+
+<main>
+  <section id="intro">
+    <h2>Introduction</h2>
```
bramus, line 56: Does this need a **MANDATORY** comment similar to the CSS, to flag that an `id` is mandatory here?

jamesnw: Added

guides/ui-components/scrollspy/expectations.md
jamesnw: It is for parsing, not specifity. Switched to `:is` - good catch, thanks!

bramus (approved):
Thanks for the changes!

guides/ui-components/scrollspy/guide.md
```diff
@@ -7,3 +7,129 @@ web-feature-ids:
...
+<script>
+  if (CSS.supports('scroll-target-group: auto')) {
+    const syncAriaCurrent = () => {
+      const currentLink = document.querySelector('nav a:target-current');
+      document.querySelectorAll('nav a').forEach(link => {
+        link.setAttribute('aria-current', link === currentLink ? 'true' : 'false');
+      });
+    };
+    // Update the aria on initial page load.
+    setTimeout(syncAriaCurrent, 10);
```
LeaVerou: (P2, as I don't think it makes a difference in most cases)
This is a bit arbitrary — many things may have not ended loading by then.
I'd suggest using `type=module` which automatically defers to DOMContantLoaded. Alternatively, attach to `DOMContantLoaded` and remove that listener if scrollend fires before then.

Another idea could also be to add an animation via `:target-current` and monitor `animationstart` — then it's guaranteed to fire whenever `:target-current` changes, and by in sync with it without us having to do much.

guides/ui-components/scrollspy/guide.md
```diff
@@ -7,3 +7,129 @@ web-feature-ids:
...
+</script>
+```
+
+## Best Practices
+
+- **DO** use `scroll-target-group: auto` on a dedicated container for navigation links: This ensures the browser knows which links should be tracked.
+- **DO** ensure targets have unique `id` attributes matching the links' `href` attributes.
+- **DO** provide enough vertical space for sections: `scroll-target-group` most clearly identifies the visible target when sections are large enough that only one or two are visible at a time.
+- **MANDATORY**: For accessibility, the visual state must be mirrored with `aria-current`. While `:target-current` handles the visual aspect, this is not exposed to screen readers.
+- **DO NOT** rely solely on color for the active state: Include other visual cues like font weight, an underline, or an indicator dot to ensure the state is accessible to users with color vision deficiencies.
```
LeaVerou: Is this needed? I'm thinking in most cases when using color for highlighting there is a very clear difference in lightness, which people with atypical color vision can generally perceive. Have you seen any inaccessible examples in the wild?

LeaVerou (approved):
Left a couple minor comments but LGTM overall!

guides/ui-components/scrollspy/guide.md
jamesnw: An example would be the Bluesky tabbar (different pattern, similar design)

image

Without the bottom blue bar, this would fail  [SC 1.4.1](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html), specifically the note that as long as there is a contrast between states of 3:1, color alone is fine. The demo colors are sufficient, but I do think it's at least worth mentioning. I made it less strong of a requirement though.

LeaVerou: Yes, but that's not an example of an inaccessible pattern, it's an example of a pattern that _would be_ inaccessible if we were to remove parts of it. I was curious if you had seen any _actively_ inaccessible cases in the wild. If these don't exist, that means LLMs haven't been trained on them!

PR #1371, opened 2026-08-24
Progress ring and Spinner Use cases

### LINKED ISSUE #1247: Create guide and evals for the progress-ring use case
Build a progress ring component that visually represents the completion status of a task or process, with support for content in the center and brand-consistent styling.

Affected web-feature IDs: [progress](https://webstatus.dev/features/progress), [conic-gradients](https://webstatus.dev/features/conic-gradients), [masks](https://webstatus.dev/features/masks), [registered-custom-properties](https://webstatus.dev/features/registered-custom-properties)

Use case subdir: [guides/ui-components/progress-ring](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-components/progress-ring)

Related features: #65

**Required files:**
- [x] demo.html
- [x] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

jamesnw:
Fixes #1247, #1250

Given that these link to each other and have overlap, I wanted to work on them together. 

I think there is a case for adding an indeterminate state to the progress ring, but opted to have them separate, as that pattern seemed to be common in design libraries.

Both have alternate markup possibilities, which I mention but opted not to expand on.


yesdhaniidan-web:
# Build with Google

Antigravity 2.0 makes it seamless to build with popular Google technology stacks. We have partnered with teams across Google to create curated bundles of primitives, including Skills, MCP servers, and Editor extensions, that are pre-configured for specific ecosystems.

Instead of searching for individual tools, you can enable these bundles to instantly empower your agents with deep knowledge of Google platforms.

### How to Enable

You can enable “Build with Google” integrations at two points:

*   **During Onboarding**: Select the checkboxes for the stacks you plan to use.
*   **In Settings**: Navigate to `Settings > Customizations > Build with Google Plugins` to add or remove integrations at any time.

### Available Bundles

### Modern Web Guidance

Keep your coding agent up to date with the latest web best practices.

*   **What’s Included**: Package of evergreen and expert-vetted skills for modern web
    *   **Key Capabilities**: Agent can build accessible, performant, and secure web experiences using injected guidance
    *   **Learn more**: [Read the Modern Web Guidance](http://goo.gle/modern-web-guidance)

### Firebase Bundle

Transform your AI coding agent into a specialized Firebase expert that can write code, configure Firebase Security Rules, and manage live resources.

*   **What’s Included**:
    *   Package of agent skills for core Firebase services, including Firestore, Authentication, App Hosting, and more
*   **Key Capabilities**:
    *   **Take action**: Do more than just write code. Your agent can initialize services, manage Authentication users, deploy new Firebase Security Rules, and work directly with your Cloud Firestore data.
    *   **Stay up-to-date**: Use official, version-aware prompts to guide your agent through setup tasks.
    *   **Improve accuracy**: Access your project’s environment and schemas to provide more relevant and accurate help.
*   **Learn More**: [Explore the Firebase Agent Skills Guide](https://firebase.google.com/docs/ai-assistance/agent-skills)

### Google Antigravity SDK

Using the Antigravity Python SDK to build AI agents

*   **What’s Included**:
    *   An agent skill containing architecture references, getting-started examples, and configuration guides for the Antigravity Python SDK (google-antigravity)
*   **Key Capabilities**:
    *   **Build agents**: Your agent can scaffold new Antigravity agents, configure models, register custom Python tools, and connect MCP servers using SDK best practices.
    *   **Stay safe**: Use official guidance to implement declarative safety policies, including deny-by-default templates and argument-level predicates.
    *   **Go deeper**: Access reference material for lifecycle hooks, multi-agent delegation, structured output, multimodal input, and token usage observability.
*   **Learn more**: [Visit the Antigravity SDK Repository](https://github.com/google-antigravity/antigravity-sdk-python)

### Android CLI

Core tools and knowledge required to develop for Android

*   **What’s included**: Android CLI and Android skills for core development tasks including making your UI edge-to-edge and adaptive, creating a testing framework, analyzing your R8 configuration to optimize performance, and more.
*   **Key capabilities**:
    *   Enable the agent of your choice to build high-quality Android apps
    *   Create new projects
    *   Deploy your app on an Android virtual device
    *   Test your app using natural language instructions
    *   Migrate to the latest versions of libraries and plugins
*   **Learn more**: [Explore the Android Agent Developer Tools](http://developer.android.com/tools/agents)

### Science

Curated collection of agent skills for science.

*   **What’s Included**: A high quality curated bundle of scientific skills
*   **Key Capabilities**: Improve agent capabilities on scientific workflows.
*   **Learn more**: [Review the DeepMind Science Skills Repository](https://github.com/google-deepmind/science-skills)

### Chrome DevTools

Reliable automation, in-depth debugging, and performance analysis in Chrome using Chrome DevTools and Puppeteer

*   **What’s Included**: Package of agent skills for Chrome DevTools and Puppeteer
*   **Key Capabilities**: Debugging accessibility (a11y) issues, auditing Core Web Vitals (LCP/INP), running visual automation tests, and diagnosing memory leaks.
*   **Learn more**: [Explore the Chrome DevTools Repository](https://github.com/ChromeDevTools/chrome-devtools-mcp)

### Dart and Flutter

The official Dart and Flutter plugin for building natively compiled, visually stunning applications for mobile, web, desktop, and embedded devices from a single codebase.

*   **What’s Included**: Package of agent skills for common Dart and Flutter workflows as well as Dart MCP server interact with your development environment and your running Dart or Flutter application.
*   **Key Capabilities**:
    *   **Enhance Developer experience**: Your agent can leverage deep static analysis tools, package management routines via pub, code formatting, and automated code-renaming capabilities
    *   **Multiplatform development**: Enable your agent to build high-quality apps across mobile, web, desktop with one codebase
    *   **Improve accuracy**: Provide instructions for agents to reliably complete common developer tasks like building adaptive layouts
*   **Learn more**: [Explore Flutter Agentic tools](https://docs.flutter.dev/ai/create-with-ai)

### Data Agent Kit Starter Pack

Specialized suite of skills and MCP tools for data engineers and database practitioners building end-to-end analytics, ETL, and AI workflows on Google Cloud.

*   **What’s Included**:
    *   Curated domain skills for BigQuery SQL & ML, dbt transformations, Apache Spark, Dataflow, Cloud Spanner, BigLake Iceberg catalogs, and workflow orchestration.
    *   Integration with Google Cloud data MCP servers for database introspection, query execution, and pipeline troubleshooting.
    *   Notebook, dashboard (Streamlit), and semantic modeling guidance.
*   **Key Capabilities**:
    *   **End-to-End Medallion Pipelines**: Prompt your agent in natural language to ingest raw data from Cloud Storage, transform it through bronze/silver/gold layers with dbt or Spark, and serve it via BigQuery or Spanner.
    *   **SQL & ML Code Generation**: Write optimized BigQuery SQL, train BigQuery ML models (XGBoost, K-Means), or build interactive LookML and Streamlit data apps.
    *   **Pipeline Orchestration & Debugging**: Automatically scaffold orchestration workflows and debug failed Dataflow or Dataproc jobs across the GCP data ecosystem.

### Google Maps Platform

Equip your AI coding agent with Google Maps specific knowledge and workflows to design, build, and deploy location-aware features across Web, Android, iOS, and Web Services.

*   **What’s Included**: Package of agent skills covering core Google Maps Platform APIs and SDKs - including Maps (2D/3D, Street View, Static, Styling, Clustering), Places, Address Validation, Geocoding, Routes & ETA (including eco-friendly routing), Geofencing, Heatmaps, and Environmental APIs (Air Quality, Pollen, Solar, Weather).
*   **Key Capabilities**:
    *   **Build location-rich experiences**: Enable your agent to write production-ready code for interactive mapping, place lookup, route optimization, geofencing, and environmental insights across mobile, web, and backend services.
    *   **Low-friction prototyping**: Get started using Maps Demo Key without needing immediate billing setup, with guided prompts to create account and restrict key when transitioning to production.
*   **Learn more**: [Explore the Google Maps Platform Agent Skills Repository](https://github.com/googlemaps/agent-skills)

guides/ui-components/progress-ring/guide.md
```diff
@@ -4,6 +4,168 @@ description: Build a progress ring component that visually represents the comple
...
+
+A progress ring (or circular progress bar) provides visual feedback on the status of a task. Unlike a linear progress bar, its circular shape is ideal for dashboards, card components, or anywhere space is constrained.
+
+This guide implements a progress ring by:
+
+- Using the native `<progress>` element as the semantic foundation to ensure the component is accessible to screen readers and keyboard users out-of-the-box.
+- Styling the component with `conic-gradient()` and `mask-image`, allowing for a fully responsive and themeable ring without the complexity of SVG path manipulation.
+- Leveraging CSS Custom Properties and `@property` to enable smooth, GPU-accelerated transitions of the progress fill.
+
+This approach is preferred over SVG-only solutions because it keeps the implementation entirely in CSS, making it easier to integrate with existing design systems and typography (especially for content in the center).
```
bramus: I’d argue that the SVG solution can also be entirely CSS controlled. To me, the advantage of using HTML as the source element here is A11Y-focused: the `<progress>` element has a `progressbar` role.

guides/ui-components/spinner/demo.html
```diff
@@ -0,0 +1,154 @@
...
+            transparent 25%,
+            var(--spinner-color)
+          );
+
+          mask-image: radial-gradient(
+            transparent calc(50% - var(--thickness)),
+            black calc(50% - var(--thickness) + 0.5px)
+          );
+
+          animation: spinner-rotate 0.8s linear infinite;
```
bramus: One thing I can see folks wanting to change is the duration. Perhaps also store this in a custom property?

guides/ui-components/spinner/guide.md
```diff
@@ -7,4 +7,97 @@ web-feature-ids:
...
+Alternatively, you may choose to omit the `<progress>` element, and add the `status` ARIA role to the `.loading-spinner` `<div>` element.
+
+### 2. Styles
+
+#### Spinner Ring and Trail
+
+The spinner uses a `conic-gradient` to create a fading trail effect. `mask-image` is used to create the ring shape.
+
+```css
+.loading-spinner {
```
bramus: Does this need to be safeguarded by a `:has(> progress:indeterminate)`?

bramus (changes requested):
Just some minor nits.

guides/ui-components/progress-ring/guide.md
```diff
@@ -4,6 +4,168 @@ description: Build a progress ring component that visually represents the comple
...
+
+Do not add a fallback value inside the `<progress>` element. It is not used by assistive technology and ignored by all modern browsers.
+
+#### Animation fallback
+
+{{ FEATURE_FALLBACKS("registered-custom-properties") }}
+
+If `@property` is not supported, the ring will jump to the new value instantly instead of transitioning smoothly. This does not break the functionality. For browsers without `@property`, you can achieve transitions using a JavaScript `requestAnimationFrame` loop to interpolate the `--value`, though the native CSS transition is preferred for performance.
+
+See the [Spinner guide](../spinner/guide.md) for handling indeterminate loading states.
```
rviscomi: This should use a `GUIDE_REF` macro and be moved outside of the fallback strategies section (I assume it's an alternative to the entire guide, not this specific fallback)

guides/ui-components/spinner/guide.md
```diff
@@ -7,4 +7,97 @@ web-feature-ids:
...
+
+A loading spinner (or activity indicator) informs users that a process is underway when the exact duration is unknown. Unlike a progress ring, a spinner is "indeterminate" and typically uses animation to signal activity.
+
+This guide implements a spinner by:
+
+- Using the native `<progress>` element as the semantic foundation. By omitting the `value` attribute, the browser treats it as an indeterminate progress bar, ensuring correct announcement by assistive technologies.
+- Styling the component with `conic-gradient()` to create a visual "trail" and `mask-image` to hollow out the center into a ring.
+- Animating the spinner efficiently using CSS transforms and respecting `prefers-reduced-motion` to ensure a comfortable experience for all users.
+
+See the [Progress Ring guide](../progress-ring/guide.md) for handling determinate tasks with a known duration.
```
rviscomi: Use `GUIDE_REF` macro

guides/ui-components/progress-ring/guide.md
```diff
@@ -4,6 +4,168 @@ description: Build a progress ring component that visually represents the comple
...
+    <progress value="75" max="100" aria-label="Task progress" class="visually-hidden"></progress>
+  </div>
+  <!-- Optional: Content to display in the center -->
+  <div class="progress-ring-content">
+    75%
+  </div>
+</div>
+```
+
+Alternatively, you can choose to apply the `progressbar` role to the `.progress-ring` `<div>`, in which case you must set the minimum, maximum and current values using Aria.
```
rviscomi: Is this ever preferred over the semantic `progress` element?

guides/ui-components/progress-ring/expectations.md
```diff
@@ -0,0 +1,14 @@
+# Expectations
```
rviscomi, line 1: Should this use case also have an expectation around `prefers-reduced-motion`?

guides/ui-components/spinner/demo.html
jamesnw: Good call. I also added the timing function as a custom property, as that is a common customization in existing libraries.

guides/ui-components/spinner/guide.md
jamesnw: I like it!

guides/ui-components/progress-ring/expectations.md
jamesnw: Added.

guides/ui-components/progress-ring/guide.md
jamesnw: I don't think so. I had included this as it is a common pattern in design systems, but I don't think that's a reason to include here. I'll remove.

guides/ui-components/progress-ring/guide.md
jamesnw: Updated

guides/ui-components/spinner/guide.md
jamesnw: Updated.

guides/ui-components/progress-ring/guide.md
jamesnw: I rephrased, but kept some of the intent- design tokens and typography _can_ be used in SVG, but it's not as straightforward.

guides/ui-components/spinner/guide.md
```diff
@@ -7,4 +7,99 @@ web-feature-ids:
...
+#### Respecting Motion Preferences
+
+Users with motion sensitivities may find fast-spinning elements disorienting. Always respect the `prefers-reduced-motion` media query.
+
+```css
+@media (prefers-reduced-motion: reduce) {
+  .loading-spinner {
+    /* Slow down the animation significantly rather than stopping it entirely,
+       so the user still knows that the process is active. */
+    animation-duration: 3s;
```
bramus: This now probably only needs to change `--spinner-duration`


```suggestion
    --spinner-duration: 3s;
```

guides/ui-components/spinner/guide.md
```diff
@@ -7,4 +7,99 @@ web-feature-ids:
...
+Alternatively, you may choose to omit the `<progress>` element, and add the `status` ARIA role to the `.loading-spinner` `<div>` element.
+
+### 2. Styles
+
+#### Spinner Ring and Trail
+
+The spinner uses a `conic-gradient` to create a fading trail effect. `mask-image` is used to create the ring shape. To ensure the loader is only shown when it is exposed to the accessibility tree, use `:has(> progress:indeterminate)`. 
+
+```css
+.loading-spinner:has(> progress:indeterminate) {
```
bramus: Now that I see this and also see the `prefers-reduced-motion` further down in the code: this increased the specificity from `(0,1,0)` to `(0,2,1)`, making it harder to override.

We could do this:

```suggestion
.loading-spinner:where(.loading-spinner:has(> progress:indeterminate)) {
```

… but that feels pretty ugly.

_(The extra `.loading-spinner` before `:has()` is necessary to prevent the `:has()` being matched against all elements)_

Alternative is to change the selector in `prefers-reduced-motion` to match the selector you currently have here.

jamesnw: It's a bit more complicated- we need to allow users to override the duration, but the prefers-reduced-motion media query needs to override the user's duration. I opted to have an internal property for the duration, and then use the `:where(:has)` pattern to allow users to still override on the `.loading-spinner` selector.

bramus: “Specificity and overrides dance” — On point commit message is very on point 😆

bramus (approved):
I have no further remarks. Thanks for the updates!

LeaVerou (changes requested):
## Implementation

- **Why 2 elements?** I might be missing something (e.g. some a11y reason), but why are we hiding the actual semantic element here? This seems to confer no benefit over the div + ARIA approach, it just adds boilerplate.
The idea of this guide was is to style the `<progress>` itself. 
- I'd use mask as a fallback and `background-clip: border-area` as the primary implementation, as it's more flexible and closer to user intent.
- This could also be a good use case for the new `accentcolor` system color keyword once it ships. Does it have a web-feature id? If not, we could use a `tmp-` (see #1256)
- Even generated content still works in that mode, so progress-ring could be a good use case for expanded `attr()` too! Make sure to apply a counter rotate. If that's janky, could switch to no rotate and doing it all in the start and end angles. 


## Visual design

- I'd avoid a gradient for the spinner. It looks dated, the current trend is flat line with either a rotate (e.g. [shadcn](https://ui.shadcn.com/docs/components/base/spinner)) or irregular animation (e.g. [WA](https://webawesome.com/docs/components/spinner)).
- Another thing that applies to both the progress-ring and the spinner is that we often want rounded line caps. I'm not convinced they matter tremendously in such a small size, but they can be done with `radial-gradient()` if need be. Might be worth a note.


I quickly Clauded this codepen to illustrate some of these: https://codepen.io/leaverou/pen/EaWaaGm

jamesnw:
> ## Implementation
> 
>     * **Why 2 elements?** I might be missing something (e.g. some a11y reason), but why are we hiding the actual semantic element here? This seems to confer no benefit over the div + ARIA approach, it just adds boilerplate.
>       The idea of this guide was is to style the `<progress>` itself.

There isn't an a11y issue. For the progress ring, we need a wrapper for the inner content anyways, and that carried over for the spinner. It was also to avoid the block that hides all the prefixed `progress-bar` and `progress-value`s. I moved the spinner to style `<progress>` directly- should I do that for progress-ring as well?

>     * I'd use mask as a fallback and `background-clip: border-area` as the primary implementation, as it's more flexible and closer to user intent.

Done. 
> 
>     * This could also be a good use case for the new `accentcolor` system color keyword once it ships. Does it have a web-feature id? If not, we could use a `tmp-` (see #1256)

Hmm, the `accentcolor` system color keyword is currently part of the `system-color` web feature, and hidden by a compute_from, so it appears to be widely available. It does appear to be [newly available](https://github.com/web-platform-dx/web-features/blob/db668cf582994291ae8ea8def787a8f4e60633d2/features/system-color.yml.dist#L41) - however, testing and MDN seem to point to it only working in installed PWAs? Is that correct?

> 
>     * Even generated content still works in that mode, so progress-ring could be a good use case for expanded `attr()` too! Make sure to apply a counter rotate. If that's janky, could switch to no rotate and doing it all in the start and end angles.

This was fairly straightforward to do with `attr()`, but I'm not sure what you mean by a "counter rotate". 

> 
> 
> ## Visual design
> 
>     * I'd avoid a gradient for the spinner. It looks dated, the current trend is flat line with either a rotate (e.g. [shadcn](https://ui.shadcn.com/docs/components/base/spinner)) or irregular animation (e.g. [WA](https://webawesome.com/docs/components/spinner)).

Removed the gradient. 

>     * Another thing that applies to both the progress-ring and the spinner is that we often want rounded line caps. I'm not convinced they matter tremendously in such a small size, but they can be done with `radial-gradient()` if need be. Might be worth a note.

I added this as solely a note in the guide, but didn't provide code for this- is that ok?

guides/ui-components/spinner/guide.md
```diff
@@ -7,4 +7,125 @@ web-feature-ids:
...
+  animation: spinner-rotate var(--_used-spinner-duration)
+    var(--spinner-timing) infinite;
+}
+
+@keyframes spinner-rotate {
+  to { transform: rotate(360deg); }
+}
+```
+
+You can also use a `radial-gradient` to make rounded end caps.
```
LeaVerou: It's ok not to include this in the demo for simplicity but we should provide a little more guidance: e.g. use trigonometry to position them at the ends of the radial gradient, how to size them, etc.

Also, I'd add that while this is barely visible in thinner rings, it starts making a significant difference in thicker ones, so it may not be the worst idea to just include it always. 

Image

guides/ui-components/spinner/demo.html
```diff
@@ -0,0 +1,173 @@
...
+              black var(--clip-boundary)
+            );
+            border: 0;
+          }
+
+          animation: spinner-rotate var(--_used-spinner-duration)
+            var(--spinner-timing) infinite;
+        }
+
+        @keyframes spinner-rotate {
```
LeaVerou: Even if the demo only includes a basic rotate, I think it would be useful to add a little guidance for how to implement the more irregular effect as I see it more frequently in polished UIs these days (I was actually surprised ShadCN only did a basic rotate). 

We could even do it by default, it's only marginally more complex and adds a lot of fidelity to the design.
It's basically something like:

```css
animation:
			progress-spin var(--speed) linear infinite,
			progress-dash 1.5s ease-in-out infinite;
```

with the animations being something like:

```css

@keyframes progress-spin {
	to { rotate: 1turn; }
}

@keyframes progress-dash {
	from { --arc-start: 0deg;   --arc-end: 3deg; }
	50%  { --arc-start: 100deg; --arc-end: 358deg; }
	to   { --arc-start: 355deg; --arc-end: 360deg; }
}
```

(and suitable registrations for the properties)

What do you think?

guides/ui-components/progress-ring/guide.md
```diff
@@ -4,6 +4,204 @@ description: Build a progress ring component that visually represents the comple
...
+    transparent calc(50% - var(--thickness)),
+    black calc(50% - var(--thickness) + 0.5px)
+  );
+  border: 0;
+}
+```
+
+{{ FEATURE_FALLBACKS("attr") }}
+
+For browsers that don't support the `attr()` CSS function for any property, manually set the `--value` CSS property.
```
LeaVerou: **P0**: This won't work well for a progress bar, as it's _meant_ to be updated dynamically. We need a MutationObserver, sadly.

guides/ui-components/spinner/guide.md
```diff
@@ -7,4 +7,125 @@ web-feature-ids:
...
+
+#### Respecting Motion Preferences
+
+Users with motion sensitivities may find fast-spinning elements disorienting. Always respect the `prefers-reduced-motion` media query. Set the internal `--_used-spinner-duration` property to override the user's `--spinner-duration` value.
+
+```css
+@media (prefers-reduced-motion: reduce) {
+  .loading-spinner {
+    /* Slow down the animation significantly rather than stopping it entirely,
+       so the user still knows that the process is active. */
```
LeaVerou, line 148: Yes!! This is something so many people get wrong.

guides/ui-components/progress-ring/guide.md
```diff
@@ -4,6 +4,204 @@ description: Build a progress ring component that visually represents the comple
...
+
+## Implementation
+
+### 1. Markup
+
+Use a wrapper to hold both the visual ring and the optional center content. The `<progress>` element remains the semantic source of truth. Use a utility class to visually hide the native progress bar while keeping it accessible.
+
+```html
+<div class="progress-ring-wrapper">
+  <progress value="75" max="100" aria-label="Task progress" class="progress-ring"></progress>
```
LeaVerou, line 33: Nit, totally optional: Do we need `progress-` in the class name when it's a `<progress>` element? `progress.ring` reads a lot more naturally.

LeaVerou:
> There isn't an a11y issue. For the progress ring, we need a wrapper for the inner content anyways, and that carried over for the spinner. 
> [...]
> I moved the spinner to style `<progress>` directly- should I do that for progress-ring as well?

I was going to say we don't if it's just a matter of showing a simple unformatted percentage, but I had forgotten generated content on `<progress>` still doesn't work in Gecko. Bummer.

In that case, do we actually need the extra `<progress>` element? The way I see it, it adds complexity (we now have two elements, and we need to make sure they are aligned etc) for little discernible benefit over the ARIA pattern.

One potential way forwards is: use `<progress>` if you don't need content, use the ARIA pattern if you do. We can also use a `<progress>` instead of `::before` as a fallback when `background-clip: border-area` isn't available (masking the whole element would also mask its content). But fine with using `<progress>` in both cases if you think that's better, no strong opinion either way.

> It was also to avoid the block that hides all the prefixed `progress-bar` and `progress-value`s. 

I think that's ok, we'll have a similar issue in slider too. 

We could frame the prefixed ones as fallbacks (in the fallbacks section), and only use the actual standard pseudo-elements (`::fill`, `::track` etc) in the main guidance, though that feels a little disingenuous given no browser implements the standard pseudo-elements.


> > ```
> > * This could also be a good use case for the new ``accentcolor`` system color keyword once it ships. Does it have a web-feature id? If not, we could use a ``tmp-`` (see #1256)
> > ```
> 
> Hmm, the `accentcolor` system color keyword is currently part of the `system-color` web feature, and hidden by a compute_from, so it appears to be widely available. It does appear to be [newly available](https://github.com/web-platform-dx/web-features/blob/db668cf582994291ae8ea8def787a8f4e60633d2/features/system-color.yml.dist#L41) - however, testing and MDN seem to point to it only working in installed PWAs? Is that correct?

There are two components of this: 
1. Is `accentcolor` supported as a system color keyword, and does it resolve to a reasonable value? (even if not the system one for fingerprinting reasons)
2. Does it actually resolve to the page accent color if set via the `accent-color` property?

I think even 1 alone makes it a better fallback than the hardcoded blue we currently have, 2 just improves things further. 

They should probably have separate web-feature ids. I'll file an issue in the web-features repo. Meanwhile, that's the textbook case for temporary web-feature ids!

> > ```
> > * Even generated content still works in that mode, so progress-ring could be a good use case for expanded ``attr()`` too! Make sure to apply a counter rotate. If that's janky, could switch to no rotate and doing it all in the start and end angles.
> > ```
> 
> This was fairly straightforward to do with `attr()`, 

We do need a MO in the fallback to update it though otherwise the control will fail at this most basic function as a progress bar 😀 
See https://github.com/GoogleChrome/modern-web-guidance-src/pull/1371/changes#r3889155946

> but I'm not sure what you mean by a "counter rotate".

NVM, there is no content in the indeterminate case and no rotation in the determinate case so this doesn't apply. Also gencontent doesn't work reliably, so just ignore this, sorry!

> > ## Visual design

> > ```
> > * Another thing that applies to both the progress-ring and the spinner is that we often want rounded line caps. I'm not convinced they matter tremendously in such a small size, but they can be done with ``radial-gradient()`` if need be. Might be worth a note.
> > ```
> 
> I added this as solely a note in the guide, but didn't provide code for this- is that ok?

See https://github.com/GoogleChrome/modern-web-guidance-src/pull/1371/changes#r3889139078

LeaVerou (changes requested):
See comments. Only critical one is using a MO for the `attr()` fallback, the rest are mostly suggestions.

guides/ui-components/progress-ring/guide.md
jamesnw: Removed

guides/ui-components/progress-ring/guide.md
jamesnw: Added

guides/ui-components/spinner/demo.html
jamesnw: I added this. I don't love the fallback for registered properties  (`!('registerProperty' in CSS')` in JS setting the animation duration for `progress-dash` to 0), perhaps there's a better way?

guides/ui-components/spinner/guide.md
jamesnw: I didn't get to this in the initial revision pass, but think it makes sense to include.

guides/ui-components/progress-ring/guide.md
```diff
@@ -1,5 +1,5 @@
 ---
-name: progress-ring
+name: ring
```
LeaVerou: Accidental find & replace?

guides/ui-components/progress-ring/guide.md
```diff
@@ -29,10 +29,10 @@ See the {{ GUIDE_REF("spinner") }} for handling indeterminate loading states.
 Use a wrapper to hold both the visual ring and the optional center content. The `<progress>` element remains the semantic source of truth. Use a utility class to visually hide the native progress bar while keeping it accessible.
 
 ```html
-<div class="progress-ring-wrapper">
-  <progress value="75" max="100" aria-label="Task progress" class="progress-ring"></progress>
+<div class="ring-wrapper">
+  <progress value="75" max="100" aria-label="Task progress" class="ring"></progress>
   <!-- Optional: Content to display in the center -->
-  <div class="progress-ring-content">
+  <div class="ring-content">
```
LeaVerou, line 35: Is this intentional? `ring` by itself doesn't quite make sense when used on a div, what I suggested was dropping `progress-` from the class when it was being used on `progress`. If we have divs using it, we need the `progress-` part too. Not that it matters, it's not like the LLM will just adopt whatever names we use…

guides/ui-components/progress-ring/guide.md
```diff
@@ -1,9 +1,212 @@
...
+}
+progress.loading-spinner:indeterminate::slider-fill {
+  display: none;
+  background: none;
+}
+```
+
+#### Container and Ring
+
+The wrapper provides the positioning context. The `<progress>` element handles the visual gradient and mask.
```
LeaVerou:
```suggestion
The wrapper provides the positioning context. The `<progress>` element handles the visual gradient.
```

(We're no longer using a mask as part of the core guidance.)

guides/ui-components/progress-ring/guide.md
```diff
@@ -1,9 +1,212 @@
...
+  range.addEventListener('input', (e) => {
+    const newValue = e.target.value;
+    
+    progress.value = newValue;
+    
+    // Update optional center content
+    content.textContent = `${newValue}%`;
+  });
+</script>
+```
```
LeaVerou:
```suggestion
Update the `value` attribute on the `<progress>` element and the text content of `.ring-content` (if displaying a percentage) whenever the value changes.
```

(Agents know how to use basic DOM methods.)

guides/ui-components/progress-ring/guide.md
```diff
@@ -1,9 +1,212 @@
...
+### 4. Optional Success State
+
+You can use the CSS attribute selector to automatically update the ring's appearance (e.g., changing the color to green) when the task reaches completion.
+
+```css
+/* Change the fill color to green when the progress reaches 100% */
+progress.ring[value="100"] {
+  --fill-color: #10b981;
+}
+```
```
LeaVerou, line 144: Ooh I love this!
Would it make sense to mention using `+` to e.g. hide the percentage and show a checkmark? (without actually showing code for them) or is it obvious?

guides/ui-components/progress-ring/guide.md
```diff
@@ -1,9 +1,212 @@
...
+Users with motion sensitivities may find the transition between values disorienting. Respect the `prefers-reduced-motion` media query by transitioning immediately.
+
+
+```css
+@media (prefers-reduced-motion: reduce) {
+  progress.ring {
+    transition-duration: 0s;
+  }
+}
+```
```
LeaVerou: I know undoing is more common, but why not just conditionally apply the transition only when `prefers-reduced-motion` is not `reduce` (using nesting, as it's now Widely Available)? 
Currently this contradicts the advice from the CSS skill to avoid applying things and then unapplying them.

guides/ui-components/progress-ring/guide.md
```diff
@@ -1,9 +1,212 @@
...
+
+The core components of this implementation — `<progress>` and `conic-gradient()` are Baseline Widely available.
+
+Do not add a fallback value inside the `<progress>` element. It is not used by assistive technology and ignored by all modern browsers.
+
+#### Animation fallback
+
+{{ FEATURE_FALLBACKS("registered-custom-properties") }}
+
+If `@property` is not supported, the ring will jump to the new value instantly instead of transitioning smoothly. This does not break the functionality. For browsers without `@property`, you can achieve transitions using a JavaScript `requestAnimationFrame` loop to interpolate the `--value`, though the native CSS transition is preferred for performance.
```
LeaVerou:
```suggestion
If `@property` is not supported, the ring will jump to the new value instantly instead of transitioning smoothly. In most cases this is fine, and does not break any functionality.

If the transition is absolutely necessary, you can check for `@property` support and use a `requestAnimationFrame()` loop to interpolate `--value` in older browsers.
```

(Rewrote to discourage the rAF workaround a bit more)

guides/ui-components/progress-ring/guide.md
```diff
@@ -1,9 +1,212 @@
...
+
+{{ FEATURE_FALLBACKS("background-clip-border-area") }}
+
+For browsers that don't yet support `background-clip: border-area`, fall back to a `mask-image` to hollow out the center.
```
LeaVerou:
```suggestion
For browsers that don't yet support `background-clip: border-area`, fall back to a `mask-image` to hollow out the center of the `<progress>` element.
```

guides/ui-components/spinner/guide.md
```diff
@@ -7,4 +7,179 @@ web-feature-ids:
...
 
-<!-- Link to progress-ring for determinate state -->
+## Overview
+
+A loading spinner (or activity indicator) informs users that a process is underway when the exact duration is unknown. Unlike a progress ring, a spinner is "indeterminate" and typically uses animation to signal activity.
+
+This guide implements a spinner by:
+
+- Using the native `<progress>` element as the semantic foundation. By omitting the `value` attribute, the browser treats it as an indeterminate progress bar, ensuring correct announcement by assistive technologies.
+- Styling the component with `conic-gradient()` to create a visual "trail" and `mask-image` to hollow out the center into a ring.
```
LeaVerou:
```suggestion
- Styling the component with `conic-gradient()` to create a visual "trail" and `background-clip: border-area` to hollow out the center into a ring.
```

LeaVerou (approved):
LGTM, see comments and suggestions, but none of them is a blocker.

guides/ui-components/spinner/demo.html
LeaVerou: Just saw this. Agreed I don't like the JS fallback. Hmmm.

One way to do it purely in CSS is to define a property that contains `, progress-dash 1.5s ease-in-out infinite` as the initial value, then use it like `var(--progress-dash-animation, )`. No property registration, the property would be empty and you'd just get blank.
It does add a pointless registered property, but we're already registering properties so…

guides/ui-components/progress-ring/guide.md
jamesnw: Hmm- good point. I think this is in tension with wanting a consistent pattern- it isn't be clear that `.ring` and `.progress-ring` are part of the same component's markup. I think I'll revert to `.progress-ring` everywhere.

guides/ui-components/progress-ring/guide.md
jamesnw: Good call, and simplified the guide a bit.

guides/ui-components/progress-ring/guide.md
jamesnw: I think that's probably obvious, and perhaps better handled in the script anyways? That way content changes are happening in one place.

guides/ui-components/spinner/demo.html
jamesnw: I like that more, conceptually. VSCode's CSS parser doesn't like the preceeding comma, but that's something to look into another day...

guides/ui-components/progress-ring/guide.md
LeaVerou: Yup, agreed.

guides/ui-components/spinner/guide.md
```diff
@@ -7,4 +7,148 @@ web-feature-ids:
...
+@property --arc-end {
+  syntax: "<angle>";
+  inherits: false;
+  initial-value: 0deg;
+}
+/* Use a custom property to conditionally include the dash animation */
+@property --progress-dash-animation {
+  syntax: "*";
+  inherits: false;
+  initial-value: , progress-dash 3s ease-in-out infinite;
```
LeaVerou, line 76: The actual `@keyframes` animation for `progress-dash` seems to have been deleted, was that intentional?

jamesnw: No, that was unintentional. Re-added.

LeaVerou:
Not sure why it's failing CI, but fine to merge by me!

guides/ui-components/spinner/guide.md
```diff
@@ -7,4 +7,169 @@ web-feature-ids:
   - progress
 ---
 
-<!-- Link to progress-ring for determinate state -->
+# Overview
```
rviscomi: Looks like guides-integrity.test.ts doesn't like this heading, calls it "vague"

We could use the guide name instead:

```suggestion
# Loading spinner
```

guides/ui-components/progress-ring/guide.md
```diff
@@ -4,6 +4,183 @@ description: Build a progress ring component that visually represents the comple
 web-feature-ids:
   - progress
   - conic-gradients
+  - masks
+  - registered-custom-properties
 ---
 
-<!-- Link to spinner for indeterminate state -->
+# Overview
```
rviscomi:
```suggestion
# Progress ring
```

rviscomi (commented):
Addresses the failing CI check

PR #1372, opened 2026-08-24
Auto-generate release notes

rviscomi:
Fixes an open TODO to generate release notes, using the Gemini API in the publish workflow. Falls back to statically generated release notes if the Gemini API fails all 3 attempts for whatever reason.

Also updates the outdated `serving/skills-cli/RELEASING.md` doc.

For testing, you can generate a dry run of the release notes with this command:

```sh
pnpm --filter serving run publish-skills --dry-run
```

Or to test/backfill previous releases, pass the tag ID to the script directly:

```sh
node --env-file=.env --experimental-strip-types serving/skills-cli/generate-release-notes.ts v0.0.185
```

Omit `--env-file=.env` to test the static fallback.

Example of release notes generated for this week's 0.0.185 release:

```md
# Release Notes: `v0.0.185`

## 🔄 Updated Guides

- Updated the **Translator API** guide to mandate the global `Translator` interface over the deprecated `window.ai.translator` namespace, clarify the four availability statuses, and emphasize the strict user gesture requirement when initiating or attaching to model downloads via `Translator.create()`.

## 🌐 Browser Support Updates

* **sibling-count() and sibling-index()**: Now **Baseline Newly available** across 2 guides (`dynamic-sibling-styling`, `dynamic-sibling-animations`).

## 📊 Benchmark Evaluations

| Agent + Model | Tasks / Assertions | Unguided → Guided Pass Rate | Uplift |
| :--- | :---: | :---: | :---: |
| **codex_cli** (gpt-5.6-sol) | 130 / 1112 | 60% → **83%** | **+23pp** |
| **claude_code** (claude-opus-5) | 130 / 1033 | 58% → **92%** | **+34pp** |
| **antigravity** (gemini-launch-candidate) | 130 / 1112 | 64% → **90%** | **+26pp** |

---
**Full Changelog**: https://github.com/GoogleChrome/modern-web-guidance/compare/v0.0.184...v0.0.185
```

Another example using a recent release that added new use cases:

```md
# Release Notes: `v0.0.183`

## 🆕 New Guides

- Introduced a new guide for **State-Aware Sticky Headers** detailing how to use CSS Scroll State Queries to natively style elements when they become stuck, eliminating the need for JavaScript scroll listeners while addressing layout flickering and fallback strategies.

## 🔄 Updated Guides

* Updated the **Chrome Extensions Skill** guide to include critical requirements for preserving user gestures during `chrome.permissions.request()` calls in service workers and clarified `activeTab` permission limitations for side panels.
* Expanded the **Language Model** guide with a comprehensive "Dos and Don'ts" section covering session pre-warming, safe streaming response sanitization, and UX best practices for AI-driven interfaces.
* Enhanced the **Prompt API** guide with detailed best practices for session cloning, using `initialPrompts` during initialization, and leveraging structured output schemas for predictable results.
* Updated the **Interest-Triggered Action Previews** guide to correct CSS unit values and property syntax within the interest invoker polyfill examples.

## 🌐 Browser Support Updates

* **:has()**: Now **Baseline Widely available** across 2 guides (`child-state-based-styling`, `content-based-styling`).
* **CSS Masks**: Now **Baseline Widely available** across 5 guides (`interactive-content-reveal`, `complex-shapes`, `shaped-cutouts`, `soft-edge-content-fade`, `visually-texture-content`).
* **linear() easing**: Now **Baseline Widely available** in `physics-based-easing`.
* **field-sizing**: Now **Baseline Newly available** in `form-fields-automatically-fit-contents`.

## 🔌 Plugins

* Added support for the **Kimi** plugin marketplace.

## 📊 Benchmark Evaluations

| Agent + Model | Tasks / Assertions | Unguided → Guided Pass Rate | Uplift |
| :--- | :---: | :---: | :---: |
| **codex_cli** (gpt-5.6-sol) | 130 / 1112 | 60% → **83%** | **+23pp** |
| **claude_code** (claude-opus-5) | 130 / 1033 | 58% → **92%** | **+34pp** |
| **antigravity** (gemini-launch-candidate) | 130 / 1112 | 64% → **90%** | **+26pp** |

---
**Full Changelog**: https://github.com/GoogleChrome/modern-web-guidance/compare/v0.0.179...v0.0.183

```


serving/skills-cli/generate-release-notes.ts
```diff
@@ -0,0 +1,474 @@
...
+): string {
+  const guideFiles = changedFiles.filter(f => f.includes('guide.md') || f.includes('/guides/'));
+  const pluginFiles = changedFiles.filter(f => f.includes('-plugin') || f.includes('plugin.json'));
+
+  const sections: string[] = [`# Release Notes: \`v${newVersion}\`\n`];
+
+  if (guideFiles.length > 0) {
+    sections.push('### 📖 Guidance & Web Platform Updates\n');
+    for (const file of guideFiles) {
+      const guideName = path.basename(path.dirname(file)) || path.basename(file, '.md');
```
micahjo7: I think the `file` here would be something like:

skills/modern-web-guidance/guides/category/guide-name.md

so `guideName` would actually be the category iiuc

could we add a test for the fallback release notes? (and for the generated release notes too, to make any necessary sanity checks before publishing them)

serving/skills-cli/generate-release-notes.ts
```diff
@@ -0,0 +1,474 @@
...
+
+    const sections: string[] = [];
+    const addedGuides: string[] = [];
+    const modifiedPatches: string[] = [];
+    const changedFiles: string[] = [];
+
+    for (const line of rawDiff.trim().split('\n')) {
+      if (!line.trim()) continue;
+      const [status, ...fileParts] = line.trim().split(/\s+/);
+      const targetFile = fileParts[fileParts.length - 1];
```
micahjo7: nit: `git diff --name-status` uses tabs to delimit

```suggestion
const tabIndex = line.indexOf('\t');
const status = line.slice(0, tabIndex).trim();
const targetFile = line.slice(tabIndex + 1).trim();

File: serving/skills-cli/generate-release-notes.ts [hunk 3]
```diff
@@ -0,0 +1,474 @@
...
+    const data = (await response.json()) as any;
+    const text = data.candidates?.[0]?.content?.parts?.[0]?.text?.trim();
+
+    if (!text) {
+      console.warn('Empty response from Gemini API. Falling back to default generator.');
+      return generateFallbackReleaseNotes(previousTag, newVersion, evalSummary, changedFiles);
+    }
+
+    // Clean any wrapping markdown code blocks if the model returned ```markdown ... ```
+    const cleanedText = text.replace(/^```markdown\n/, '').replace(/\n```$/, '').trim();
```
micahjo7: to cover couple more edge cases

```suggestion
const cleanedText = text
  .replace(/^```(?:markdown|md)?\r?\n/, '')
  .replace(/\r?\n```$/, '')
  .trim();

File: serving/skills-cli/publish-skills.ts [hunk 4]
```diff
@@ -150,6 +174,8 @@ async function main() {
   const publishCliDir = path.join(DIST_DIR, "skills-cli");
 
   if (isDryRun) {
+    const { featuresCount, useCasesCount } = updateReadmeWithFeaturesAndUseCases(publishCliDir);
```
micahjo7: this will update the actual readmes on dry run which we probably don't want, maybe we can decouple the count logic in this method from the actual updating and call that method here instead

serving/skills-cli/publish-skills.ts
```diff
@@ -103,13 +103,36 @@ async function publishToDistributionRepo(publishCliDir: string, newVersion: stri
...
+      target: `v${newVersion}`,
+    });
+  } catch (err) {
+    console.warn('Could not generate release notes from published tag. Falling back to local payload diff:', err);
+    releaseNotes = await generateReleaseNotes({
+      previousTag: latestTag,
+      newVersion,
+      publishCliDir,
+    });
+  }
```
micahjo7: could we just call with the publishCliDir primarily?

```suggestion
const releaseNotes = await generateReleaseNotes({
  previousTag: latestTag,
  newVersion,
  publishCliDir,
});

File: serving/skills-cli/generate-release-notes.ts [hunk 6]
```diff
@@ -0,0 +1,474 @@
...
+      }
+      if (modifiedPatches.length > 0) {
+        sections.push(`### 🔄 Modified Files & Content Diff:\n${modifiedPatches.join('\n\n')}`);
+      }
+
+      guideDiff = sections.join('\n\n');
+    }
+  } catch (err) {
+    console.warn(`Warning: Could not fetch GitHub compare between ${previousTag} and ${targetTag}:`, err);
+  }
```
micahjo7: if we get a gh api error here, will that cause issues? should we throw here or remove the try/catch?

serving/skills-cli/generate-release-notes.ts
```diff
@@ -0,0 +1,474 @@
...
+  '**/*',
+  '!**/.cache/**',
+  '!**/tfjs_model_minilm/**',
+  '!**/*.{js,mjs,ts,bin,map,gz}',
+  '!**/skill-version.txt',
+  '!THIRD_PARTY_NOTICES',
+  '!skills/modern-web-guidance/package.json',
+];
+
+export function getPreviousTag(targetTag: string): string {
```
micahjo7: im wondering if we need this. do we need to run generate-release-notes manually?

micahjo7 (commented):
what do you think about automating more of the notes creation (like the benchmark table and links/heading/structure of the notes) and keep the responsibility of the llm to just the summaries (to prevent variance)?

serving/skills-cli/generate-release-notes.ts
rviscomi: It's useful for testing. I ran some manual tests of previous releases just to make sure that the notes are correct, but we could also write some automated tests against a specific release so the expected output would be relatively deterministic.

serving/skills-cli/generate-release-notes.ts
rviscomi: Good callout, yeah let's remove the try/catch so it fails loudly

serving/skills-cli/generate-release-notes.ts
rviscomi: Good catch, fixed and added tests

serving/skills-cli/publish-skills.ts
rviscomi: Yeah good call

rviscomi:
> what do you think about automating more of the notes creation (like the benchmark table and links/heading/structure of the notes) and keep the responsibility of the llm to just the summaries (to prevent variance)?

Thanks @micahjo7, I've addressed this and your inline comments. The file grew to 1000+ LOC so I also split it up into a few smaller files.

Updated the example output in the PR description with the latest changes.

@paulirish FYI I also updated the `serving/skills-cli/RELEASING.md` doc to capture these changes and also removed some obsolete info.

serving/skills-cli/release-notes-gemini.ts
```diff
@@ -0,0 +1,277 @@
...
+    validate: (bullets) => bullets.length === guideNames.length,
+    validationErrorMsg: (count) =>
+      `Gemini updated guide summary validation failed: expected ${guideNames.length} bullets, received ${count}.`,
+  });
+}
+
+/**
+ * Uses Gemini to generate bullet summaries for changed guides.
+ */
+export async function generateGuideSummariesWithGemini(opts: {
```
micahjo7: I don't think this method is used anymore

serving/skills-cli/release-notes-gemini.ts
```diff
@@ -0,0 +1,277 @@
...
+    }
+  }
+
+  return null;
+}
+
+/**
+ * Uses Gemini to generate bullet summaries for newly added guides.
+ */
+export async function generateNewGuideSummariesWithGemini(opts: {
```
micahjo7: nit: we could use a shared helper/const for some of the duplicate text between this method and `generateUpdatedGuideSummariesWithGemini`

serving/skills-cli/release-notes-diff.ts
```diff
@@ -0,0 +1,613 @@
...
+      }
+
+      records.push({ relPath, oldPath, status, patch, isLink });
+    }
+
+    const classified = classifyChanges(records);
+    return {
+      ...classified,
+      evalSummary: getLatestEvalResultsSummary(),
+    };
```
micahjo7: we can include evalSummary within classifyChanges and update here and in `getConsumerFacingDiff`

serving/skills-cli/release-notes-markdown.ts
```diff
@@ -0,0 +1,228 @@
...
+    }
+    const guidesList = uniqueGuides.map(g => `\`${g}\``).join(', ');
+    return `* **${entry.featureName}**: ${entry.statusDescription} across ${uniqueGuides.length} guides (${guidesList}).`;
+  });
+}
+
+export interface BuildReleaseNotesMarkdownOptions {
+  previousTag: string;
+  newVersion: string;
+  guideBullets?: string[];
```
micahjo7: could we remove guideBullets and it's usage now? and rely on the new/updated/removed

serving/skills-cli/release-notes-diff.ts
```diff
@@ -0,0 +1,613 @@
...
+  const tags = output.split('\n').map(t => t.trim()).filter(Boolean);
+  const targetIndex = tags.indexOf(targetTag);
+  if (targetIndex !== -1 && targetIndex + 1 < tags.length) {
+    return tags[targetIndex + 1];
+  }
+  if (tags.length > 1) {
+    return tags[1];
+  }
+  throw new Error(`Could not determine previous tag for ${targetTag}`);
+}
```
micahjo7, line 102: suggestion from my agent

A few edge cases in `getPreviousTag`:
1. If `targetTag` is not in `tags` (e.g. `HEAD` or an upcoming unreleased version), returning `tags[1]` skips the latest tag (`tags[0]`).
2. `--merged ${targetTag}` throws if `targetTag` hasn't been created locally yet.
3. `gh api .../tags` defaults to 30 items without `--paginate`.

this impl would address those^
```suggestion
export function getPreviousTag(targetTag: string): string {
  let tags: string[] = [];

  try {
    const output = execSync('git tag -l "v*.*.*" --sort=-v:refname', {
      encoding: 'utf8',
      cwd: rootDir,
      stdio: ['pipe', 'pipe', 'ignore'],
    }).trim();
    tags = output.split('\n').map(t => t.trim()).filter(Boolean);
  } catch {}

  if (tags.length === 0) {
    try {
      const ghOutput = execSync(
        `gh api --paginate repos/GoogleChrome/modern-web-guidance/tags --jq '.[].name'`,
        { encoding: 'utf8', cwd: rootDir, stdio: ['pipe', 'pipe', 'ignore'] }
      ).trim();
      tags = ghOutput.split('\n').map(t => t.trim()).filter(Boolean);
    } catch {}
  }

  const targetIndex = tags.indexOf(targetTag);
  if (targetIndex !== -1 && targetIndex + 1 < tags.length) {
    return tags[targetIndex + 1];
  }
  if (tags.length > 0) {
    return tags[0];
  }

  throw new Error(`Could not determine previous tag for ${targetTag}`);
}

File: serving/skills-cli/release-notes-diff.ts [hunk 13]
```diff
@@ -0,0 +1,613 @@
...
+export interface RawChangeRecord {
+  relPath: string;
+  status: string; // 'A', 'M', 'D', 'R', 'added', 'modified', 'removed', 'renamed'
+  oldPath?: string;
+  patch?: string;
+  isLink?: boolean;
+}
+
+export interface ClassifiedChanges {
+  guideDiff: string;
```
micahjo7: remove `guideDiff` in favor of the specific diff objs added below

serving/skills-cli/release-notes-gemini.ts
rviscomi: Done

serving/skills-cli/release-notes-gemini.ts
rviscomi: Good catch

serving/skills-cli/release-notes-markdown.ts
rviscomi: Done

serving/skills-cli/release-notes-diff.ts
rviscomi: Done

serving/skills-cli/release-notes-diff.ts
rviscomi: Replaced with a new `classifyChangesWithEvals` wrapper function

micahjo7 (approved):

PR #1375, opened 2026-08-25
feat(guides): Add file redirection instructions for large retrieve outputs to modern-web-guidance skill

TravenReese:
This commit updates the modern-web-guidance skill instructions to guide AI agents on handling large retrieval responses that might exceed terminal output limits. It advises agents to redirect the output to a temporary file (such as .temp-guide.md) so they can read the full guide and its fallback rules completely before generating code.

Screenshot 2026-08-26 at 12 40 09 AM

### DIFF
--- a/guides/modern-web-guidance/SKILL.md
+++ b/guides/modern-web-guidance/SKILL.md
@@ -75,6 +75,13 @@ Once you have a relevant `id` from the search results, call this script using th
 npx -y modern-web-guidance@latest retrieve "<id>"
 ```
 
+> **Handling Large Output / Truncation**:
+> Guides contain complete code samples and browser fallbacks that may exceed terminal output limits. If the output is truncated, redirect to a temporary file and read it with your file viewing tool:
+> ```sh
+> npx -y modern-web-guidance@latest retrieve "<id>" > .temp-guide.md
+> ```
+> Read the complete file before generating code, and delete the temporary file when finished.
+
 **Example Output**:
 `The markdown content of the guide describing implementation steps...`
 


guides/modern-web-guidance/SKILL.md
```diff
@@ -75,6 +75,12 @@ Once you have a relevant `id` from the search results, call this script using th
 npx -y modern-web-guidance@latest retrieve "<id>"
 ```
 
+> **Handling Large Output / Truncation**:
+> Guides contain complete code samples and browser fallbacks that may exceed terminal output limits. If the output is truncated, redirect to a temporary file and read it with your file viewing tool:
+> ```sh
+> npx -y modern-web-guidance@latest retrieve "<id>" > .temp-guide.md
+> ```
+> Read the complete file before generating code, and delete the temporary file when finished.
```
micahjo7, line 83: lgtm, but I'm curious, is there a specific reason you added this example, compared to what we have [here](https://source.corp.google.com/piper///depot/google3/third_party/jetski/build_with_google/modern-web-guidance-plugin/skills/modern-web-guidance/SKILL.md;l=88)?

micahjo7 (approved):
since we don't have specific tests for this, could you add the validation you saw from the trajectory that the agent followed this (into the PR description)?

let's get @paulirish sign-off on this too before merging

guides/modern-web-guidance/SKILL.md
TravenReese: I looked into some prompt engineering patterns and it’s generally recommended to be as explicit as possible rather than leaving the execution details up to the model. While stronger models can usually figure out the redirection and file reading on their own, this skill will be used across a bunch of different agents and model tiers. For smaller or more literal agents, having the concrete command (> .temp-guide.md) removes ambiguity and prevents failure modes like guessing filenames or misinterpreting the redirect step.

paulirish: Yeah, I favor the more minimal approach here.  (And again, this actually is only a problem for jetski and AGY..... so... well.... models :) )

PR #1383, opened 2026-08-27
Link to guides and features in release notes

rviscomi:
Follow up to #1372 

- References to guides are now linked to the markdown source. The link goes to the MWG dist repo, version-pinned to the tag number of the release.
- References to web features are now linked to their page on webstatus.dev.

Example diff from the v0.0.185 release:

```diff
## 🔄 Updated Guides

-* Updated the **translator** guide to mandate using the global `Translator` interface instead of the deprecated `window.ai.translator` namespace, require user gestures for model downloads, and ensure identical language options are passed to both availability and creation methods.
+* Updated the **[translator](https://github.com/GoogleChrome/modern-web-guidance/blob/v0.0.185/skills/modern-web-guidance/guides/built-in-ai/translator.md)** guide to mandate using the global `Translator` interface instead of the deprecated `window.ai.translator` namespace, require user gestures for model downloads, and ensure identical language options are passed to both availability and creation methods.

## 🌐 Browser Support Updates

-* **sibling-count() and sibling-index()**: Now **Baseline Newly available** across 2 guides (`dynamic-sibling-styling`, `dynamic-sibling-animations`).
+* **[sibling-count() and sibling-index()](https://webstatus.dev/features/sibling-count)**: Now **Baseline Newly available** across 2 guides ([`dynamic-sibling-styling`](https://github.com/GoogleChrome/modern-web-guidance/blob/v0.0.185/skills/modern-web-guidance/guides/css/dynamic-sibling-styling.md), [`dynamic-sibling-animations`](https://github.com/GoogleChrome/modern-web-guidance/blob/v0.0.185/skills/modern-web-guidance/guides/ui-behaviors/dynamic-sibling-animations.md)).
```

### DIFF
--- a/serving/skills-cli/release-notes-diff.ts
+++ b/serving/skills-cli/release-notes-diff.ts
@@ -4,6 +4,7 @@ import path from 'node:path';
 import { execSync } from 'node:child_process';
 import matter from 'gray-matter';
 import { minimatch } from 'minimatch';
+import { features } from 'web-features';
 import { rootDir } from '../../lib/paths.ts';
 export interface EvalSummaryItem {
   testId?: string;
@@ -21,6 +22,7 @@ export interface EvalSummaryItem {
 
 export interface BaselineUpdateInfo {
   featureName: string;
+  featureId?: string;
   statusRank: number; // 1: Widely, 2: Newly, 3: Limited
   statusDescription: string;
   guideName: string;
@@ -182,15 +184,15 @@ export function stripBaselineLinesFromPatch(patch?: string): string {
 }
 
 /**
- * Extracts structured Baseline / browser support update info from a patch.
+ * Extracts structured Baseline / browser support update info from a single hunk or patch block.
  */
-export function parseBaselineUpdateFromPatch(guideName: string, patch: string): BaselineUpdateInfo {
-  const addedLines = patch
+export function parseBaselineUpdateFromHunk(guideName: string, hunk: string): BaselineUpdateInfo | null {
+  const addedLines = hunk
     .split('\n')
     .filter(l => l.startsWith('+') && !l.startsWith('+++'))
     .map(l => l.slice(1).trim());
 
-  const removedLines = patch
+  const removedLines = hunk
     .split('\n')
     .filter(l => l.startsWith('-') && !l.startsWith('---'))
     .map(l => l.slice(1).trim());
@@ -199,6 +201,7 @@ export function parseBaselineUpdateFromPatch(guideName: string, patch: string): 
   let statusRank = 3;
   let statusDescription = 'Updated browser engine support';
 
+  // 1. Check added lines for a status transition
   for (const line of addedLines) {
     const match = line.match(/Baseline status for\s+(.+?):\s*(Widely available|Newly available|Limited availability)/i);
     if (match) {
@@ -218,10 +221,17 @@ export function parseBaselineUpdateFromPatch(guideName: string, patch: string): 
     }
   }
 
-  if (featureName.toLowerCase() === 'masks') {
-    featureName = 'CSS Masks';
-  } else if (!featureName) {
-    featureName = guideName;
+  // 2. If status line didn't change (e.g. engine update in Limited status), extract featureName from hunk context
+  if (!featureName) {
+    const match = hunk.match(/Baseline status for\s+(.+?):/i);
+    if (match) {
+      featureName = match[1].trim();
+    }
+  }
+
+  // If the feature name cannot be extracted, omit this update
+  if (!featureName) {
+    return null;
   }
 
   if (statusRank === 3) {
@@ -237,12 +247,36 @@ export function parseBaselineUpdateFromPatch(guideName: string, patch: string): 
     }
   }
 
+  const featureId = resolveWebFeatureId(featureName);
+
   return {
     featureName,
+    featureId,
     statusRank,
     statusDescription,
     guideName,
   };
+}
+
+/**
+ * Extracts all Baseline / browser support update infos from a patch (supporting multi-hunk diffs).
+ */
+export function parseBaselineUpdatesFromPatch(guideName: string, patch: string): BaselineUpdateInfo[] {
+  const hunks = patch.split(/(?=@@ -\d+)/g).filter(h => h.trim());
+  const updates: BaselineUpdateInfo[] = [];
+  for (const hunk of hunks) {
+    if (hasBaselineUpdateInPatch(hunk)) {
+      const update = parseBaselineUpdateFromHunk(guideName, hunk);
+      if (update) {
+        updates.push(update);
+      }
+    }
+  }
+  return updates;
+}
+
+export function parseBaselineUpdateFromPatch(guideName: string, patch: string): BaselineUpdateInfo | null {
+  return parseBaselineUpdatesFromPatch(guideName, patch)[0] || null;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


serving/skills-cli/release-notes-diff.ts
```diff
@@ -319,14 +353,112 @@ export function getGuideDescription(relPath: string): string | undefined {
...
 
-/**
- * Extracts unique guide identifiers from changed file paths.
- */
 export function getUniqueGuideNames(changedFiles: string[]): string[] {
   const guideFiles = changedFiles.filter(isGuideFile);
   return Array.from(new Set(guideFiles.map(getGuideName)));
 }
 
+export const GITHUB_REPO_URL = 'https://github.com/GoogleChrome/modern-web-guidance';
```
micahjo7, line 414: nit: we can use this in release-notes-markdown.ts as well

serving/skills-cli/release-notes-diff.ts
```diff
@@ -218,10 +221,17 @@ export function parseBaselineUpdateFromPatch(guideName: string, patch: string):
     }
   }
 
-  if (featureName.toLowerCase() === 'masks') {
-    featureName = 'CSS Masks';
-  } else if (!featureName) {
-    featureName = guideName;
+  // 2. If status line didn't change (e.g. engine update in Limited status), extract featureName from hunk context
+  if (!featureName) {
+    const match = hunk.match(/Baseline status for\s+(.+?):/i);
```
micahjo7: In `serving/lib/baseline.ts`, limited availability is output as `<feature> has limited availability.` rather than `Baseline status for <feature>:`. 
As written, `match(/Baseline status for\s+(.+?):/i)` won't catch limited availability features or their browser engine updates.

```suggestion
    const match = hunk.match(/(?:Baseline status for\s+(.+?):|(.+?)\s+has limited availability)/i);
```

also, we can remove `|Limited availability` from line 206 above, and  `generate-release-notes.test.ts` test where it says "Baseline status for Masks: Limited availability." should be updated

serving/skills-cli/release-notes-markdown.ts
```diff
@@ -1,10 +1,57 @@
...
+
+    for (const guideName of prioritizedGuides) {
+      const url = getGuideGithubUrl(guideName, ref);
+      if (!url) continue;
+
+      // Skip if this URL is already linked in the bullet
+      if (result.includes(`](${url})`)) continue;
+
+      // Replace bold markdown: **guideName**
+      const exactBoldRegex = new RegExp(`\\*\\*${escapeRegExp(guideName)}\\*\\*`, 'gi');
```
micahjo7: If Gemini formats an existing link with brackets on the outside (e.g. `[**guide**](url)` or `[`guide`](url)`), these regexes will match inside the brackets and create invalid nested markdown like `[**[guide](url)**](url)`. Suggestion so that text that is already inside markdown links is ignored:

`const exactBoldRegex = new RegExp(`(?<!\\[)\\*\\*${escapeRegExp(guideName)}\\*\\*(?!\\]\\()`, 'gi');`

and

`const codeRegex = new RegExp(`(?<!\\[)\`${escapeRegExp(guideName)}\`(?!\\]\\()`, 'g');` below

serving/skills-cli/release-notes-diff.ts
```diff
@@ -319,14 +353,112 @@ export function getGuideDescription(relPath: string): string | undefined {
...
+  return undefined;
+}
+
+/**
+ * Returns the direct GitHub URL for a guide or skill in GoogleChrome/modern-web-guidance.
+ */
+export function getGuideGithubUrl(guideName: string, ref = 'main'): string | undefined {
+  const relPath = getGuidePathInDistribution(guideName);
+  if (!relPath) return undefined;
+  const normalizedRef = ref.startsWith('v') || ref === 'main' || ref.startsWith('refs/') ? ref : `v${ref}`;
```
micahjo7: we can simplify this so we add it if it starts with a digit 

```suggestion
const normalizedRef = /^\d+\./.test(ref) ? `v${ref}` : ref;

File: serving/skills-cli/release-notes-diff.ts [hunk 5]
```diff
@@ -319,14 +353,112 @@ export function getGuideDescription(relPath: string): string | undefined {
...
+ */
+export function formatGuideBoldLink(guideName: string, ref = 'main'): string {
+  const url = getGuideGithubUrl(guideName, ref);
+  return url ? `**[${guideName}](${url})**` : `**${guideName}**`;
+}
+
+// Lookup mapping lowercase feature IDs and display names to canonical feature IDs
+const featureNameToIdMap = new Map<string, string>();
+for (const [id, f] of Object.entries(features)) {
+  featureNameToIdMap.set(id.toLowerCase(), id);
```
micahjo7: In `web-features`, a few features have display names that match another feature's canonical ID. For example, `overflow-shorthand` has display name `"overflow"`, which currently overwrites the canonical `"overflow"` ID in this map (same for `lang` and `selection`).

We can prevent overwriting existing canonical IDs by adding a `!has()` check:

```suggestion
  if (f.name && !featureNameToIdMap.has(f.name.toLowerCase())) {

File: serving/skills-cli/release-notes-diff.ts (hunk 2 above)
Comment (rviscomi) [reply serving/skills-cli/release-notes-diff.ts]: Great catch! This actually uncovered a few other buggy support statements:

- a browser removing support
- going from mobile-only support to full support
- edits to the supported browser version number

Fixed those as well.

File: serving/skills-cli/release-notes-diff.ts (hunk 1 above)
Comment (rviscomi) [reply serving/skills-cli/release-notes-diff.ts:414]: Done

File: serving/skills-cli/release-notes-diff.ts (hunk 5 above)
Comment (rviscomi) [reply serving/skills-cli/release-notes-diff.ts]: Actually, I think this was a bit over-engineered. When we write browser support statements, we always use the feature name. The code for that does have a fallback to the feature ID, but that was more about type safety as there's not a single case in the web-features dataset of a feature (`kind === "feature"`) missing a name, so the ID is never used.

So to avoid this problem, I've updated this to always map feature names to IDs

File: serving/skills-cli/release-notes-markdown.ts (hunk 3 above)
Comment (rviscomi) [reply serving/skills-cli/release-notes-markdown.ts]: Made the change as suggested just in case anything goes wrong, but I decided to also remove the prompt for Gemini to add the links in the first place. As long as we have this function to deterministically linkify a guide, it's silly to have Gemini do it with less reliability.

File: serving/skills-cli/release-notes-gemini.ts [hunk 6]
```diff
@@ -110,6 +110,7 @@ interface BuildGuideSummaryPromptOptions {
   type: 'new' | 'updated';
   guideNames: string[];
   guideDiff: string;
+  ref?: string;
```
micahjo7: we can remove this now, here and in `generateNewGuideSummariesWithGemini`, `generateUpdatedGuideSummariesWithGemini`

serving/skills-cli/release-notes-diff.ts
```diff
@@ -319,14 +406,111 @@ export function getGuideDescription(relPath: string): string | undefined {
...
+  return undefined;
+}
+
+/**
+ * Returns the direct GitHub URL for a guide or skill in GoogleChrome/modern-web-guidance.
+ */
+export function getGuideGithubUrl(guideName: string, ref = 'main'): string | undefined {
+  const relPath = getGuidePathInDistribution(guideName);
+  if (!relPath) return undefined;
+const normalizedRef = /^\d+\./.test(ref) ? `v${ref}` : ref;
```
micahjo7: nit: indentation

micahjo7 (approved):
Looks great, thanks Rick!

serving/skills-cli/release-notes-gemini.ts
rviscomi: Fixed

serving/skills-cli/release-notes-diff.ts
rviscomi: Fixed

rviscomi:
The [broken CI / basics check](https://github.com/GoogleChrome/modern-web-guidance-src/actions/runs/33418092505/job/99573488214?pr=1383) is unrelated. See https://github.com/GoogleChrome/modern-web-guidance-src/pull/1354#issuecomment-5481850627

PR #1386, opened 2026-08-28
Add Progress bar guidance

### LINKED ISSUE #1246: Create guide and evals for the progress use case
Build a styled progress bar that communicates completion of a task or process, in both determinate and indeterminate states.

Affected web-feature IDs: [accent-color](https://webstatus.dev/features/accent-color), [progress](https://webstatus.dev/features/progress)

Use case subdir: [guides/ui-components/progress](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-components/progress)

Related features: #37

**Required files:**
- [ ] demo.html
- [ ] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

dvdherron:
Fixes #1246 

Open to feedback on fleshing out the Fallbacks section. Should the fully custom approach to styling be considered a fallback? Also, feedback on some of the accessibility approaches here would be great.


guides/ui-components/progress/guide.md
```diff
@@ -4,6 +4,316 @@ description: Build a styled progress bar that communicates completion of a task
...
-<!-- Link to progress-ring and spinner for circular styling -->
+# Build a Styled Progress Bar
+
+The `<progress>` element is the semantic way to represent the completion progress of a task, like downloading or uploading information or completing part of a multi-step process. Using the native element helps with accessibility, as screen readers have an implicit `role="progressbar"` and can announce values.
+
+## How to implement
+
+1.  **Use the native `<progress>` element**: It provides built-in accessibility and platform-consistent behavior.
+2.  **Define the state**:
+    *   **Determinate**: Set the `max` and `value` attributes. Note that `<progress>` does **not** support a `min` attribute; the minimum is always `0`.
```
LeaVerou: We don't need to set the `max` attribute if we're using the default 0-1 scale (though it may still be better for clarity).
I'm not sure we need to mention that it doesn't support a `min`, `<progress>` is old enough that this is known.

guides/ui-components/progress/guide.md
```diff
@@ -4,6 +4,316 @@ description: Build a styled progress bar that communicates completion of a task
...
+    *   **Indeterminate**: Omit the `value` attribute for tasks of unknown duration to set the progress bar into an indeterminate state. You can also use the `:indeterminate` pseudo-class to explicitly put the progress bar in an indeterminate state.
```
LeaVerou: The `:indeterminate` (or any other) pseudo-class cannot affect state, it just matches only when the progress bar is already in an indeterminate state.

guides/ui-components/progress/guide.md
```diff
@@ -4,6 +4,316 @@ description: Build a styled progress bar that communicates completion of a task
...
+
+The `<progress>` element is the semantic way to represent the completion progress of a task, like downloading or uploading information or completing part of a multi-step process. Using the native element helps with accessibility, as screen readers have an implicit `role="progressbar"` and can announce values.
+
+## How to implement
+
+1.  **Use the native `<progress>` element**: It provides built-in accessibility and platform-consistent behavior.
+2.  **Define the state**:
+    *   **Determinate**: Set the `max` and `value` attributes. Note that `<progress>` does **not** support a `min` attribute; the minimum is always `0`.
+    *   **Indeterminate**: Omit the `value` attribute for tasks of unknown duration to set the progress bar into an indeterminate state. You can also use the `:indeterminate` pseudo-class to explicitly put the progress bar in an indeterminate state. 
+    *   **Mandatory:** If a progress bar needs to change from determinate to indeterminate, you must remove the value attribute (`element.removeAttribute('value')`);
```
LeaVerou: This is known.

guides/ui-components/progress/guide.md
```diff
@@ -4,6 +4,316 @@ description: Build a styled progress bar that communicates completion of a task
...
+3.  **Standardize Styles**: Use `appearance: none` and `border: unset` to normalize the progress bar across different browsers before applying more customized styles.
```
LeaVerou: Why `border: unset`? I'd just set whatever border we need (or `border: 0` if we don't need one). Reset styles makes sense more broadly, here we're only dealing with a single element, and we know how we want to style it.

guides/ui-components/progress/guide.md
```diff
@@ -4,6 +4,316 @@ description: Build a styled progress bar that communicates completion of a task
...
+4.  **Ensure Accessibility**:
```
LeaVerou: Nice. Might want to make this a separate sub-section.

ping @knowler to review this too

guides/ui-components/progress/guide.md
```diff
@@ -4,6 +4,316 @@ description: Build a styled progress bar that communicates completion of a task
...
+    *   **Associate a label**: Use `<label for="...">`, `aria-labelledby`, or `aria-label`. 
+    *   **Contextual state**: Use `aria-describedby` to reference the loading progress of a section of a page. Use `aria-busy="true"` on the container being updated. Set `aria-busy` to `"false"` when the task is complete. 
+    *   **Forced Announcements**: Set `tabindex="-1"` and call `.focus()` on the element in JavaScript when significant updates occur to force screen readers to announce the new progress.
```
LeaVerou: Might want to clarify that "the element" here means "the `<progress>` element".

guides/ui-components/progress/guide.md
```diff
@@ -4,6 +4,316 @@ description: Build a styled progress bar that communicates completion of a task
 web-feature-ids:
   - accent-color
   - progress
+  - color-scheme
 ---
 
-<!-- Link to progress-ring and spinner for circular styling -->
+# Build a Styled Progress Bar
+
+The `<progress>` element is the semantic way to represent the completion progress of a task, like downloading or uploading information or completing part of a multi-step process. Using the native element helps with accessibility, as screen readers have an implicit `role="progressbar"` and can announce values.
```
LeaVerou: Once #1371 is merged, we should add `GUIDE_REF` references to them here. This is for a horizontal styled progress bar, whereas these are for a circular one.

guides/ui-components/progress/guide.md
```diff
@@ -4,6 +4,316 @@ description: Build a styled progress bar that communicates completion of a task
...
+@keyframes shimmer {
+  0% { background-position: 200% 0; }
+  100% { background-position: -200% 0; }
+}
+```
+
+## Best Practices
+
+*   **DO** use `<progress>` for task completion. Use the `<meter>` element for scalar measurements.
+*   **DO** use `accent-color` for simple branding and `color-scheme` for dark mode support.
```
LeaVerou: Note that this will stop working once we apply our own styles (with `appearance: none`). We can use `accentcolor` if supported to bring it back, but there's no web-feature for it currently. So this is more of an alternative path rather than complementary. Still worth mentioning though, but I'd move it to the intro, i.e. "if all you need is to customize the colors, you might want to just do that. This guide is for deeper customizations"

guides/ui-components/progress/guide.md
```diff
@@ -4,6 +4,316 @@ description: Build a styled progress bar that communicates completion of a task
...
+*   **DO** normalize with `appearance: none` and `border: unset` if you need a specific custom look.
+*   **DO** manage accessibility by ensuring labels are present and using `tabindex="-1"` + `.focus()` for dynamic updates.
+*   **Mandatory:** **DO NOT** add a fallback value inside the `<progress>` element. It is not used by assistive technology and ignored by all modern browsers. 
+*   **DO NOT** use `<progress>` for scroll position indicators; use scroll-driven animations instead.
+*   **DO** respect `prefers-reduced-motion` if you apply custom animations.
+*   **DO** ensure proper contrast between the progress bar and track when adden custom styles.
```
LeaVerou: Typo: "adden"

Should we move the last two bullets to the a11y section?

guides/ui-components/progress/guide.md
```diff
@@ -4,6 +4,316 @@ description: Build a styled progress bar that communicates completion of a task
...
+  <progress id="p" value="20" max="100">20%</progress>
+  <button id="update-btn">Add 10%</button>
+
+<style>
+  :root {
+    --brand-color: teal;
+  
+    color-scheme: light dark;
+    /* Set accent color for progress and other built-in UI elements */
+    accent-color: var(--brand-color, AccentColor); /* Use system accent color as a fallback */
```
LeaVerou: Please note that `AccentColor` is not supported in every browser, and there is no web-feature id for it so we can't just use it.

(We can however detect support and apply it conditionally — or even extract that into a `--default-accent-color` variable)

LeaVerou (changes requested):
See comments for some more specific things. Bigger picture feedback is that the bulk of this guide is several large code examples, and it's not always clear what the agent is meant to extract from each of these.

In general, we try to avoid large code examples as they mix important bits with boilerplate, and thus can be wasteful in terms of tokens and less clear which bits are the actual guidance vs scaffolding. Instead, extract the meaningful guidance into short prose, and use (short!) code snippets for those things that are best explained in code (complementing the prose).

guides/ui-components/progress/guide.md
dvdherron: Updated this so that it's clear that `max` is not required.

guides/ui-components/progress/guide.md
dvdherron: Adjusted this so that the guidance is clear this should only be used for styling the `progress`.

guides/ui-components/progress/guide.md
dvdherron: removed

guides/ui-components/progress/guide.md
dvdherron: Removed

guides/ui-components/progress/guide.md
dvdherron: Added an Accessibility Considerations section and gathered related guidance there. (And requested a review from Nathan)

guides/ui-components/progress/guide.md
dvdherron: Clarified here.

guides/ui-components/progress/guide.md
dvdherron: Added `GUIDE_REF` references to the` project-ring` and `spinner guides`.

guides/ui-components/progress/guide.md
dvdherron: Expanded a bit on the use of `accent-color` for simple styling. And moved this higher up into the intro.

guides/ui-components/progress/guide.md
dvdherron: Fixed the typo and added these points to the new a11y section

guides/ui-components/progress/guide.md
dvdherron: I removed the use of `AccentColor` to keep this simple.

knowler (changes requested):
I think the upload example is a bit of a mismatch for the loading pattern described by the ARIA spec. [What the ARIA spec describes for the `progressbar` role about using `aria-busy` and `aria-describedby`](https://w3c.github.io/aria/#progressbar:~:text=updated%2E-,If,loading%2E) would be applicable for a separate progress bar that referencing the loading state of another region of the page (i.e. the progressbar is not within the loading region) and it shouldn’t be used for cases like this where the contents of the section are already loaded and there’s just some task that’s been triggered within that. [For screen readers that support `aria-busy`](https://a11ysupport.io/tech/aria/aria-busy_attribute) (not many), it can be expected to prevent updates from that part of the page and that was the case in one Windows screen reader when I tested: it didn’t report the upload progress since that was within the busy content.

As far as screen reader updates for a task like this, I think [a multi-function button](https://adrianroselli.com/2021/01/multi-function-button.html) might be a better choice as it keeps the focus on the button, avoids disabling it (the focus loss can be disorienting and error prone), and then any separate visual progress indicator could be presentational (or at least not cause additional live region updates). That doesn’t necessarily help for this use case though.

I would need to look into this more before I give suggestions, but I just wanted to flag what I’ve reviewed so far. I have a feeling it might be best to avoid the pattern described in the ARIA spec as `aria-busy` does not have great support and `aria-hidden` often achieves the same effect.

As for making the progress bar report its state, I think that could require some context for where and how its being used. Live regions can be annoying, so not every progress bar should report its state when it’s being updated, especially if the user can perform other tasks while the progress bar is loading (i.e. you could choose to report when it’s done instead of as it changes to reduce noise but announce the important state change).

There also _could_ be some animation considerations for indeterminate progress bars as they relate to [WCAG 2.2 SC 2.2.2 Pause, Stop, Hide](https://www.w3.org/WAI/WCAG22/quickref/#pause-stop-hide) (e.g. an indeterminate progress bar displayed in parallel to other content could be considered distracting for some users).

guides/ui-components/progress/guide.md
```diff
@@ -9,303 +9,148 @@ web-feature-ids:
...
-    *   **Contextual state**: Use `aria-describedby` to reference the loading progress of a section of a page. Use `aria-busy="true"` on the container being updated. Set `aria-busy` to `"false"` when the task is complete. 
-    *   **Forced Announcements**: Set `tabindex="-1"` and call `.focus()` on the element in JavaScript when significant updates occur to force screen readers to announce the new progress.
-
-## Example code: Determinate state
+    *   **Determinate**: Set the `value` attribute. Setting the `max` value is optional unless a scale different from the default 0-1 scale is needed. 
+    *   **Indeterminate**: Omit the `value` attribute for tasks of unknown duration to set the progress bar into an indeterminate state. 
+3. **Tint the progress bar**: Use `accent-color` for simply changing the color of the progress bar. More extensive cross-browser customizations require resetting default browser styles with `appearance: none` and the use of targeted pseudo selectors.
+4.  **Standardize Styles**: Use `appearance: none` to normalize the progress bar across different browsers before applying more customized styles.
```
LeaVerou, line 23: These are largely different approaches: either you're ok with just `accent-color` or you want full styling. Currently they are described as steps of one process.

guides/ui-components/progress/guide.md
```diff
@@ -4,6 +4,161 @@ description: Build a styled progress bar that communicates completion of a task
...
+  appearance: none;
+}
+```
+
+2. **Tint the progress bar:** Use browser targeted pseudo selectors to apply color to the progress bar and track. 
+
+```css
+/* **Mandatory**: -moz- and -webkit- pseudo selectors cannot be grouped. If grouped one of the selectors will be ignored. */
+/* Apply color for custom progress track (background) */
+progress,
```
LeaVerou, line 76: Why are we grouping the base `progress` with the WebKit selectors?
Can't we just style `progress` with the track color and then use the prefixed selectors to make that background transparent so that the progress bg shows through?
Currently, applying a background to `progress` (e.g. somewhere else in the codebase) would produce no result. It's not a major thing, but if we can use the standard CSS property, that's generally better.

guides/ui-components/progress/guide.md
```diff
@@ -4,6 +4,161 @@ description: Build a styled progress bar that communicates completion of a task
...
+}
+```
+
+3. **Add a custom animation for the indeterminate state:** Use `:not([value])` or the `:indeterminate` pseudo class to target the progress element in its indeterminate state to add a custom animation. 
+
+```css
+/* Indeterminate state animation */
+/* Optional: Use the :indeterminate psueudo-class as part of a selector (progress:indeterminate) */
+progress:not([value]) {
+  background: linear-gradient(90deg, transparent, var(--progress-color, currentColor), transparent);
```
LeaVerou, line 99: I would suggest using hard stops to create an animated flat area (similar to the UA default, possibly with a more interesting animation), as a gradient is harder to integrate with arbitrary design systems.

guides/ui-components/progress/guide.md
```diff
@@ -4,6 +4,161 @@ description: Build a styled progress bar that communicates completion of a task
...
+}
+
+@keyframes loading {
+  from { background-position: 200% 0; }
+  to { background-position: -200% 0; }
+}
+
+/* Respect prefers-reduced-motion */
+@media (prefers-reduced-motion: reduce) {
+  progress:not([value]) {
```
LeaVerou, line 111: Use `:indeterminate` rather than `:not([value])`

guides/ui-components/progress/guide.md
```diff
@@ -4,6 +4,161 @@ description: Build a styled progress bar that communicates completion of a task
...
+    /* Slow down the animation significantly and use a subtler visual change */
+    animation-duration: 10s;
+    background: linear-gradient(90deg, var(--color-surface), var(--accent-color), var(--color-surface));
```
LeaVerou, line 114: Nit: a keyword like `to right` is easier to read.
Also, why did we use transparent above and an actual surface color here? 
And why are we repeating `background-size`? If it's because we used `background`, we can just use `background-image` instead.

LeaVerou (changes requested):
Hi David, thanks for the changes! Much better. Left a few comments, but most are minor.

PR #1391, opened 2026-08-31
Revert "gd dev output for faster-spa-view-transitions"

micahjo7:
Reverts GoogleChrome/modern-web-guidance-src#1354

This broke the build because https://github.com/GoogleChrome/modern-web-guidance-src/pull/1378 removes the methods used in this grader. I will create a new PR for the generated evals after this gets reverted.


micahjo7:
Regenerated in https://github.com/GoogleChrome/modern-web-guidance-src/pull/1401

PR #1392, opened 2026-08-31
skills: recommend pnpx runner and scoped permission rules

paulirish:
### Summary

- **Runner preference**: Advises agents to prefer `pnpx` over `npx` when `pnpm` is available for faster execution and better local caching (noting that `pnpx` does not use `-y`).
- **Permission scoping**: Instructs agents to scope command approval rules specifically to `npx -y modern-web-guidance@latest *` or `pnpx modern-web-guidance@latest *` rather than over-permissive bare `npx *` / `pnpx *`.
- **Node test resilience**: Safely checks `typeof localStorage.getItem === 'function'` in `eval-view/utils.js` for test runs under Node.

### DIFF
--- a/guides/modern-web-guidance/SKILL.md
+++ b/guides/modern-web-guidance/SKILL.md
@@ -86,8 +86,10 @@ When generating or modifying code, cross-check the implementation against the re
 - **Applicable Guidance & Fallbacks**: Ensure the relevant modern patterns and necessary fallback strategies from the guide are correctly applied, without forcing unrequested features.
 - **Task Fulfillment**: Confirm that the implementation fully satisfies the user's request.
 
-## Using npx
+## Using npx / pnx
 
+-   Prefer `pnx` over `npx` if `pnpm` is available.
+-   When requesting tool permissions, allowlist `npx -y modern-web-guidance@latest *` specifically (or the `pnx` equivalent), never bare `npx *`.
 -   IMPORTANT: on Windows, using `npx` may fail. Use `npx.cmd ...` instead.
 -   Network access is required for fetching npm packages needed by the task.
 -   If the `npx -y modern-web-guidance…` command hangs, you may be offline. Try running again in offline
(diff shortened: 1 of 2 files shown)


guides/modern-web-guidance/SKILL.md
```diff
@@ -86,8 +86,10 @@ When generating or modifying code, cross-check the implementation against the re
 - **Applicable Guidance & Fallbacks**: Ensure the relevant modern patterns and necessary fallback strategies from the guide are correctly applied, without forcing unrequested features.
 - **Task Fulfillment**: Confirm that the implementation fully satisfies the user's request.
 
-## Using npx
+## Using npx / pnx
```
Snugug:
```suggestion
## Using npx / pnpx
```

guides/modern-web-guidance/SKILL.md
```diff
@@ -86,8 +86,10 @@ When generating or modifying code, cross-check the implementation against the re
...
 
+-   Prefer `pnx` over `npx` if `pnpm` is available.
+-   When requesting tool permissions, allowlist `npx -y modern-web-guidance@latest *` specifically (or the `pnx` equivalent), never bare `npx *`.
```
Snugug:
```suggestion
-   Prefer `pnpx` over `npx` if `pnpm` is available.
-   When requesting tool permissions, allowlist `npx -y modern-web-guidance@latest *` specifically (or the `pnpx` equivalent), never bare `npx *`.
```

`pnpx` doesn't have a `-y` flag and throws an error when trying to use it with one. We should instruct the agent appropriately

guides/modern-web-guidance/SKILL.md
Snugug: Suggest using `pnpx` instead of `pnx` as `pnx` was introduced in 11+ but `pnpx` goes back further and has an alias for `pnx` in newer versions.

Snugug (changes requested):

guides/modern-web-guidance/SKILL.md
paulirish: yeah.. i was thinking we use the canonical name .. but... pnpx is gonna be whats in the training data.. so... we can go with that.

guides/modern-web-guidance/SKILL.md
paulirish: yah i noticed that distinction too. but yeah ill include it.

Snugug (approved):

PR #1399, opened 2026-09-01
Fix layout issues, don't clip them (fixes #1396)

### LINKED ISSUE #1396: Canvas -> display: block vs overflow: hidden
A pattern I see often is something like this

```html
<style>
html, body {
  margin: 0;
  height: 100%;
  overflow: hidden;   /* wrong! */
}

canvas {
  width: 100%;
  height: 100%;
}
</style>
<body>
  <canvas></canvas>
</body>
```

What that code does is declare a canvas with gap after it (canvas defaults to `display: inline`). The gap ends up creating a scrollbar. The user then covers over the fact that they asked for a layout with a scrollbar by turning the scrollbar off with `overflow: hidden`.

The correct practice is not to ask for a canvas with a gap after it by declaring the canvas to be `display: block;`

```html
<style>
html, body {
  margin: 0;
  height: 100%; /* or 100dhv etc... */
}

canvas {
  width: 100%;
  height: 100%;
  display: block;
}
</style>
<body>
  <canvas></canvas>
</body>
```

Asking for the wrong thing and then hiding the mistake is arguably not a best practice. I'm not sure that the prompt for that would be. Something like

> Do not use `overflow: hidden` to mask layout issues. Fix the layout issues.
> For example `<canvas>`, ``, `<svg>`, `<video>`, `<iframe>` by default use 
> `display: inline` which adds a gap. If you do not want the gap then remove it
> by styling the element as `display: block`

LeaVerou:
(also removed a pointless "Do")

### DIFF
--- a/guides/css/css-layout/guide.md
+++ b/guides/css/css-layout/guide.md
@@ -35,12 +35,12 @@ Walk the decision tree top-to-bottom and stop at the first match. Note that layo
 
 ### 1.2 Working principles
 
-**Do:**
-
 - Use logical properties (`inline-size`, `block-size`, `margin-inline`, `padding-block`, `inset-inline-start`) for layout dimensions and spacing — see {{ GUIDE_REF("css") }} for full coverage.
 - Apply the content-first vs layout-first mental model: flexbox when items dictate flow, grid when you define the skeleton first.
 - Use the `place-*` shorthands (`place-content`, `place-items`, `place-self`) to align across both axes in one declaration.
 - Reach for intrinsic sizing (`min-content`, `max-content`, `fit-content()`) and flexible tracks (`fr`, `minmax()`) before fixed `width`/`height` — fewer media queries, more resilient layouts.
+- Do NOT use `overflow: hidden` to mask layout issues — fix them at the source.
+For example certain replaced elements like `<canvas>`, `<img>`, `<svg>`, `<video>`, `<iframe>` use `display: inline` by default, which can add a gap in certain layouts. Instead of clipping the gap with `overflow`, remove it by applying a block display-outside value (e.g. `display: block`, `grid`, `flex` etc).
 - Use `aspect-ratio` to reserve space for media and prevent layout shift before assets load.
 
 ```css


guides/css/css-layout/guide.md
```diff
@@ -35,12 +35,12 @@ Walk the decision tree top-to-bottom and stop at the first match. Note that layo
...
 ### 1.2 Working principles
 
-**Do:**
-
 - Use logical properties (`inline-size`, `block-size`, `margin-inline`, `padding-block`, `inset-inline-start`) for layout dimensions and spacing — see {{ GUIDE_REF("css") }} for full coverage.
 - Apply the content-first vs layout-first mental model: flexbox when items dictate flow, grid when you define the skeleton first.
 - Use the `place-*` shorthands (`place-content`, `place-items`, `place-self`) to align across both axes in one declaration.
 - Reach for intrinsic sizing (`min-content`, `max-content`, `fit-content()`) and flexible tracks (`fr`, `minmax()`) before fixed `width`/`height` — fewer media queries, more resilient layouts.
+- Do NOT use `overflow: hidden` to mask layout issues — fix them at the source.
+For example certain replaced elements like `<canvas>`, `<img>`, `<svg>`, `<video>`, `<iframe>` use `display: inline` by default, which can add a gap in certain layouts. Instead of clipping the gap with `overflow`, remove it by applying a block display-outside value (e.g. `display: block`, `grid`, `flex` etc).
```
jamesnw, line 43: Will "block display-outside value" be clear to an LLM? Or does the spec enum of `display-outside` (block, inline, run-in) confuse this with display values that implicitly use the `block` `display-outside` value?

Also, could we provide a reason to select one over another?

jamesnw (approved):
A couple questions, but nothing blocking here.

guides/css/css-layout/guide.md
LeaVerou: It was in @greggman's testing!
I intentionally didn't recommend a specific mode, since that's so use case dependent.

PR #1402, opened 2026-09-02
Guide for installing web apps use case

captainbrosset:
This is my first guide. I'm opening this as a draft PR for now. Before going further, I'd love some early feedback on whether this is the right general direction.


rviscomi:
If you haven't already, I'd recommend running your guide through the [validation skill](https://github.com/GoogleChrome/modern-web-guidance-src/blob/main/.agents/skills/project-guide-validation/SKILL.md), which can catch some common mistakes.

guides/pwa/install-web-app/guide.md
```diff
@@ -0,0 +1,263 @@
...
+
+# Installing web applications
+
+Use the Web Install API to give users an explicit, in-page way to install a web
+application. Installation is always completed through browser-controlled consent
+UI; websites cannot silently install an app.
+
+Use the declarative `<install>` element when the browser-provided install control
+fits the page. Use `navigator.install()` when the application needs a custom
+button or needs to respond to the result.
```
captainbrosset, line 19: @LiaHiscock @tomayac curious to hear your thoughts about this part of the guide, which is pretty key. When should we recommend using `<install>` and `navigator.install()`?

tomayac: If both are available, I'd personally rather use `navigator.install()` so I get to style my own button. `<install>` is the second-best, but even then I'd maybe still be tempted to use `beforeinstallprompt`. I think the styling aspects can be the deciding factor for most brands.

PR #1408, opened 2026-09-03
harness: preserve pi agent sessions by default and clean up boolean env parsing

paulirish:
Fixes an issue where Pi ran with `--no-session` by default during evals (due to `PI_NO_SESSION !== 'false'`), so no session files were ever written or exported to results. Also cleans up related env var boolean checks across the repo.

- **Pi session persistence**: Pi now keeps sessions by default so trajectories actually get captured and graded. Only runs ephemeral if `PI_NO_SESSION=true|1`.
- **Nested session export**: `exportPiTrajectories()` now searches recursively and adds the `session-` prefix so trajectory normalizers pick them up.
- **Boolean env parsing**: Added a small `parseBooleanEnv()` helper (`1`/`0`, `true`/`false`, `yes`/`no`) to replace brittle string checks for `PI_NO_SESSION`, `NO_OPEN`, and `DRY_RUN`.
- **`sync-use-cases`**: Added `--write` / `--live` flags while keeping dry-run as default.
- **Codex parsing**: Safely stringifies non-string function call outputs to avoid crashes during normalization.

### DIFF
--- a/eval-view/server.js
+++ b/eval-view/server.js
@@ -8,9 +8,10 @@ import os from 'os';
 import { exec, spawn } from 'child_process';
 import { runAllManifests } from './generate-manifests.js';
 import { extractSuiteSummary } from './summary-extractor.js';
+import { parseBooleanEnv } from '../lib/env.ts';
 
 const PORT = process.env.PORT || 8081;
-const STATIC = process.env.STATIC === 'true';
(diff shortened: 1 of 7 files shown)


paulirish:
ran into this because our pi/qwen eval run didnt save any session.jsonl files which.. seems unintentional.  @connorjclark  yah?

connorjclark (approved):

PR #1410, opened 2026-09-06
fix(navigation-drawer): guard the viewport width used for the observer threshold

### LINKED ISSUE #1409: navigation-drawer: 1 / window.innerWidth throws when innerWidth is 0, leaving the drawer unable to close
## Summary

The `navigation-drawer` guide computes the IntersectionObserver threshold as a fraction of the viewport width:

https://github.com/GoogleChrome/modern-web-guidance-src/blob/HEAD/guides/ui-components/navigation-drawer/guide.md#L263

```js
const visibleThreshold = 1 / window.innerWidth;
```

When `window.innerWidth` is `0`, this evaluates to `Infinity`, and `IntersectionObserver` **throws from the constructor** rather than clamping or ignoring the value:

```
Failed to construct 'IntersectionObserver': Failed to read the 'threshold' property
from 'IntersectionObserverInit': The provided double value is non-finite.
```

The same line is in the demo: `guides/ui-components/navigation-drawer/demo.html#L233`.

## Why it matters more than a stray console error

In the guide's architecture the observer is the single source of truth for the drawer's state — it is what calls `onDrawerOpened()` / `onDrawerClosed()`. It is also constructed *before* the trigger and dismissal handlers are wired up (step 4 precedes step 5).

So the exception takes the rest of the initialisation with it, including every `addEventListener`. The resulting symptom is confusing, because the drawer is not dead: `popover="manual"` still promotes the panel to the top layer, so **the menu opens and then cannot be closed** — not by the close button, not by tapping the scrim, not with Escape. On a phone, with no browser chrome in a PWA, that traps the user on the page.

## Reproduction

```js
// Any browser console:
new IntersectionObserver(() => {}, { threshold: [1 / 0, 1] });
// Uncaught TypeError: Failed to construct 'IntersectionObserver': ...non-finite.
```

I hit this for real in an automated browser where the page had not been composited yet and `window.innerWidth` was `0`. I would expect the same path in any zero-width layout — a display:none or unsized iframe, and plausibly a background or prerendered tab that runs scripts before first layout — though I have only verified the automated-browser case directly.

## Suggested fix

The width is only there to express "one pixel", so any sane fallback preserves the intent:

```js
const viewportWidth =
  window.innerWidth || document.documentElement.clientWidth || 360;
const visibleThreshold = 1 / viewportWidth;
```

With a real width the value is unchanged (`375` → `0.00267`, `1280` → `0.00078`), so this is purely defensive.

Happy to send a PR for `guide.md` and `demo.html` if that helps.

basekvk:
Fixes #1409.

## The bug

`navigation-drawer` derives the IntersectionObserver threshold from the viewport width:

```js
const visibleThreshold = 1 / window.innerWidth;
```

When `window.innerWidth` is `0` — a page that has not been laid out yet — this is `Infinity`, and `IntersectionObserver` throws from the constructor rather than clamping:

```
Failed to construct 'IntersectionObserver': Failed to read the 'threshold' property
from 'IntersectionObserverInit': The provided double value is non-finite.
```

Because step 4 builds the observer *before* step 5 attaches the trigger and dismissal handlers, the exception takes the rest of the script with it. The drawer is not inert, though: `popover="manual"` still promotes the panel to the top layer, so **the menu opens and then cannot be closed** — not by the close button, not by tapping the scrim, not with Escape.

## Reproduced on this repo's own demo

Same browser, same page, `window.innerWidth === 0`:

| File | Console |
| --- | --- |
| `demo.html` at `7de9677` (unchanged) | `Uncaught TypeError: ...non-finite` at line 272 |
| `demo.html` with this patch | clean |

## The change

```js
const viewportWidth =
  window.innerWidth || document.documentElement.clientWidth || 360;
const visibleThreshold = 1 / viewportWidth;
```

The width only exists to express "one pixel", so any sane fallback preserves the intent. With a real viewport the value is unchanged — `375` → `0.00267`, `1280` → `0.00078` — so this is purely defensive and does not alter behaviour anywhere it already worked.

Applied to both `guide.md` and `demo.html`, with a short comment explaining why the guard is there so it does not get simplified away later.

## Note on the CLA

I have not signed the Google CLA yet. Happy to do so — flagging it up front so this is not blocked silently.

### DIFF
--- a/guides/ui-components/navigation-drawer/demo.html
+++ b/guides/ui-components/navigation-drawer/demo.html
@@ -230,7 +230,12 @@
     }
 
     function observeOpenAndClosedState() {
-      const visibleThreshold = 1 / window.innerWidth; // At least 1px on screen.
+      // At least 1px on screen. The width is guarded because
+      // `window.innerWidth` is 0 before first layout, and a non-finite
+      // threshold makes the IntersectionObserver constructor throw.
+      const viewportWidth =
+        window.innerWidth || document.documentElement.clientWidth || 360;
+      const visibleThreshold = 1 / viewportWidth;
       const observer = new IntersectionObserver(
         (entries) => {
           const entry = entries.at(-1); // Only the last entry is relevant.
--- a/guides/ui-components/navigation-drawer/guide.md
+++ b/guides/ui-components/navigation-drawer/guide.md
@@ -260,7 +260,14 @@ function onDrawerClosed() {
 // Treat "any pixel of the sheet visible inside the popover root" as
 // "open enough to count as not closed". This threshold is intentionally
 // tiny so the closed callback only fires once the sheet is truly gone.
-const visibleThreshold = 1 / window.innerWidth;
+//
+// Guard the width: `window.innerWidth` is 0 in a page that has not been
+// laid out yet, and `1 / 0` is Infinity, which makes the
+// IntersectionObserver constructor throw. That would take the rest of
+// this script with it, leaving a drawer that opens and can never close.
+const viewportWidth =
+  window.innerWidth || document.documentElement.clientWidth || 360;
+const visibleThreshold = 1 / viewportWidth;
 
 const observer = new IntersectionObserver(
   (entries) => {


LeaVerou (commented):
At first glance, this seems to be treating the symptom, not the cause. Why would this code be running before the page is laid out? Perhaps the right fix would be to make sure it doesn't?

PR #1411, opened 2026-09-06
Create atrule-support-conditionals guide

LeaVerou:
Supersedes #757 see https://github.com/GoogleChrome/modern-web-guidance-src/pull/757#issuecomment-5559167626

### DIFF
--- a/guides/css/atrule-support-conditionals/guide.md
+++ b/guides/css/atrule-support-conditionals/guide.md
@@ -5,7 +5,9 @@ web-feature-ids:
   - supports-at-rule
 ---
 
-Use  `@supports at-rule(<at-rule name>) {…}` to conditionally apply certain CSS rules or declarations only if a certain at-rule is supported by the browser.
+# Conditionally Apply CSS Based on At-Rule Support
+
+Use `@supports at-rule(@<at-rule name>) {…}` to conditionally apply certain CSS rules or declarations only if a certain at-rule is supported by the browser.
 This can contain entire rules, or be used in conjunction with CSS nesting to conditionally apply certain declarations.
 
 ```css
--- a/guides/css/atrule-support-conditionals/guide.md
+++ b/guides/css/atrule-support-conditionals/guide.md
@@ -3,7 +3,6 @@ name: atrule-support-conditionals
 description: Conditionally apply certain CSS rules or declarations only if a certain at-rule is supported by the browser.
 web-feature-ids:
   - supports-at-rule
-draft: true
 ---
 
 Use  `@supports at-rule(<at-rule name>) {…}` to conditionally apply certain CSS rules or declarations only if a certain at-rule is supported by the browser.
--- a/guides/css/atrule-support-conditionals/guide.md
+++ b/guides/css/atrule-support-conditionals/guide.md
@@ -47,7 +47,7 @@ For example, instead of `@supports at-rule(@function)` to detect supports for cu
 
 In other cases there may be pseudo-classes or pseudo-elements that are only available when a certain at-rule is supported, which can be used to detect support.
 
-### Detecting container query types
+### Detecting container query types { #supports-container-types }
 
 `at-rule()` ONLY allows detection of at-rules by **name**.
 It does NOT allow detection of other syntax or descriptors.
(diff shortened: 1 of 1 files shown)


rviscomi:
@jamesnw could you take this one?

guides/css/atrule-support-conditionals/guide.md
```diff
@@ -0,0 +1,156 @@
...
+@supports at-rule(@starting-style) {
+  @starting-style {
+    /* ... */
+  }
+}
+```
+
+Regular CSS graceful parsing will already ignore rules the browser doesn't understand, so this is unnecessary.
+
+
```
jamesnw, line 37:
```suggestion
```

guides/css/atrule-support-conditionals/guide.md
```diff
@@ -0,0 +1,156 @@
...
+## Detecting support for at-rules without `@supports at-rule()` { #detecting-at-rules-without-supports-at-rule }
```
jamesnw, line 38: Perhaps this is intentional, but it seems like this should be the `Fallback Strategies` section, and then the current content of the `Fallback Strategies` section should be moved to the main body. I know this is a weird one where there are not yet at-rules that are newer than this.

guides/css/atrule-support-conditionals/guide.md
```diff
@@ -0,0 +1,156 @@
...
+  animation:
+    progress-spin var(--_used-spinner-duration) linear infinite
+    var(--progress-dash-animation, );
+}
+```
+
+
+For a more general-purpose solution, you can register a non-inheriting property and set it to ` ` (a space) on the root element:
+
+<!-- Testcase: https://codepen.io/leaverou/pen/KwWzpVr -->
```
jamesnw, line 102: Should this be removed?

guides/css/atrule-support-conditionals/guide.md
```diff
@@ -0,0 +1,156 @@
...
+### Try properties, values, or selectors first
+
+`@supports` can already detect properties, values, and selectors (via `@supports selector(...)`).
+In many cases, you can use these to detect support for a feature without needing to check for the at-rule itself.
+
+For example, instead of `@supports at-rule(@function)` to detect supports for custom functions, you can use `@supports (color: --foo())`.
+
+In other cases there may be pseudo-classes or pseudo-elements that are only available when a certain at-rule is supported, which can be used to detect support.
+
+### Detecting container query types { #supports-container-types }
```
jamesnw, line 51: I think this shouldn't be under the `Detecting support for at-rules without `@supports at-rule()` header, as it's a limitation of `@supports at-rule()`, even if it is supported.

guides/css/atrule-support-conditionals/guide.md
```diff
@@ -0,0 +1,156 @@
...
+
+```css
+@supports at-rule(@starting-style) {
+  @starting-style {
+    /* ... */
+  }
+}
+```
+
+Regular CSS graceful parsing will already ignore rules the browser doesn't understand, so this is unnecessary.
```
jamesnw, line 35:
```suggestion

```css
/* BAD: CSS already ignores rules the browser doesn't understand, so this is unnecessary */
@supports at-rule(@starting-style) {
  @starting-style {
    /* ... */
  }
}
```
```

File: guides/css/atrule-support-conditionals/guide.md [hunk 6]
```diff
@@ -0,0 +1,156 @@
...
+```js
+const sheet = new CSSStyleSheet();
+sheet.replaceSync(`
+  @page {
+    @top-left {
+      content: "test";
+    }
+  }
+`);
+const SUPPORTS_PAGE_MARGINS = Boolean(sheet.cssRules[0]?.cssRules[0]);
```
jamesnw, line 140:
```suggestion
const SUPPORTS_PAGE_MARGINS = Boolean(sheet.cssRules[0]?.cssRules?.[0]);
```

Otherwise this throws in non-supporting browsers.

guides/css/atrule-support-conditionals/guide.md
```diff
@@ -0,0 +1,156 @@
...
+Notes:
+- Both of these only work if whitespace is an acceptable alternative. For other cases, you will need to use JS as described below.
+- The general-purpose solution depends on non-inheritance, so it will not work for declarations on the root element.
+
+### Using JavaScript as a last resort
+
+For entire at-rules, check for the presence of certain interfaces in JavaScript.
+For example, if `globalThis.CSSFunctionRule` is defined, then `@function` is supported.
+
+If you don’t know the interface name, or you need to test more deeply (e.g. for certain preludes, nested rules, descriptors), you can set up a test and check how it was parsed:
```
jamesnw, line 129: I'm cautious about prompting an agent to do this type of exploration- it irks me when I see an agent do something like this to answer a question I could find the answer to much more efficiently (or already know the answer to). 

I see agents jumping to this to soon, or this enabling overengineered patterns. Is this a scenario a user is likely to hit?

guides/css/atrule-support-conditionals/guide.md
```diff
@@ -0,0 +1,156 @@
...
+    @top-left {
+      content: "test";
+    }
+  }
+`);
+const SUPPORTS_PAGE_MARGINS = Boolean(sheet.cssRules[0]?.cssRules[0]);
+```
+
+IMPORTANT:
+- When using JS to detect support for CSS features, apply a class to the root element and branch off that instead of applying the styles from JavaScript directly.
```
jamesnw, line 144:
```suggestion
After determining support with JavaScript, apply a class to the root element, and use that class in your CSS to apply the fallback. Do not apply the styles from JavaScript directly.

IMPORTANT:
```

I didn't think this fits with the other items in the`Important` header, and I reworded for (I hope) clarity. It might also be worth having a code snippet for this?

jamesnw (changes requested):
Good guidance, with lots of great fallbacks which are very needed for this.

PR #1414, opened 2026-09-07
[#1243] Hovercard use case

### LINKED ISSUE #1243: Create guide and evals for the hovercard use case
Build a hovercard component that displays an interactive structured preview of the page a certain type of link (e.g. a user profile) will navigate to when hovered over.

Affected web-feature IDs: [popover](https://webstatus.dev/features/popover), [anchor-positioning](https://webstatus.dev/features/anchor-positioning), [interest-invokers](https://webstatus.dev/features/interest-invokers), [cross-document-view-transitions](https://webstatus.dev/features/cross-document-view-transitions)

Use case subdir: [guides/ui-components/hovercard](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-components/hovercard)

Related features: #32 #39 #59 #50

**Required files:**
- [ ] demo.html
- [ ] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

jamesnw:
Addresses #1243 

1. Hovering a link (uses interest invokers) opens a popover, positioned with anchor positioning
2. Clicking the original link while the popover is opened uses view transitions to make the popover seamlessly expand into the next page
3. The headline inside the popover isn't a link. (When using a screen reader, when reading the headline in the popover, clicking enter triggers the link that opened the popover, so no redundant link is needed)
4. Clicking while the popover is closed does a default crossfade transition
5. Going back from the article to the main page does a default crossfade, since the popover is no longer open


guides/ui-components/hovercard/guide.md
```diff
@@ -11,3 +12,122 @@ guides:
...
+When you use `interestfor`, the browser handles the assistive-technology wiring:
+- It implicitly associates the source element with the target via `aria-describedby` when the target is plaintext, or via `aria-details` when it contains interactive content.
+- **DO NOT** add `aria-describedby` or `aria-details` manually; `interestfor` manages this state dynamically.
+
+## 2. Positioning with Anchor Positioning
+
+Position the hovercard relative to its trigger using **CSS Anchor Positioning**. When using `interestfor`, the trigger becomes an implicit anchor for the popover.
+
+```css
+[popover] {
```
LeaVerou: Isn't this a little too general?

guides/ui-components/hovercard/guide.md
```diff
@@ -11,3 +12,122 @@ guides:
...
+
+See the {{ GUIDE_REF("animate-to-from-top-layer")}} guide to see how to use `transition` with `allow-discrete` and `@starting-style` to animate the hovercard's appearance and disappearance from the top layer.
+
+Always respect reduced motion.
+
+```css
+/* MANDATORY: Respect user preference for reduced motion by disabling transitions. */
+@media (prefers-reduced-motion: reduce) {
+  [popover] {
+    transition: none;
```
LeaVerou, line 75: I'll defer to @knowler on this one, but this doesn't seem the type of motion that is triggering for people with `prefers-reduced-motion`? If anything, it seems the abrupt loading might be worse?

LeaVerou (changes requested):
This looks great! 

A couple small issues:
- The entry transition of the hovercard should ideally grow from the target, not its center
- I’m not convinced an exit transition is useful, it seems to get in the way.
- When pressing Back you get the exit transition (even when you directly load the URL, i.e. _not_ by pressing the hovercard link!), which seems a bit weird.
- How does it work when there are multiple links pointing to the same page?
- Some web feature ids are missing (e.g. `<link rel=expect>`
- Instead of duplicating content for the hovercard, would it make sense to fetch the HTML the first time the link is hovered and read it from there? That seems more maintainable and scalable, and this is PE anyway.

See also the couple comments I left. 

Nit: Can we rename `index.html` to `demo.html` for consistency?

jamesnw:
> This looks great!
> 
> A couple small issues:
> 
>     * The entry transition of the hovercard should ideally grow from the target, not its center
Updated

>     * I’m not convinced an exit transition is useful, it seems to get in the way.
>     * When pressing Back you get the exit transition (even when you directly load the URL, i.e. _not_ by pressing the hovercard link!), which seems a bit weird.

I think you're referencing the default view transition between pages, not the popover exit transition? Should we maybe have a separate guide describing how to disable root animation if you just want specific animations? 

>     * How does it work when there are multiple links pointing to the same page?

Just fine- since we're assigning `view-transition-name` to the popover content, there only ever is 1 instance of any of the names on the index page.

>     * Some web feature ids are missing (e.g. `<link rel=expect>`
Added `blocking-render` and `link-rel-expect`

>     * Instead of duplicating content for the hovercard, would it make sense to fetch the HTML the first time the link is hovered and read it from there? That seems more maintainable and scalable, and this is PE anyway.

This really depends on the use case. The hovercard will very rarely load the exact content of its destination- in most cases it will only have a snippet (as is the case in the demo) rather than the full article, or an alternate layout. (I'm by happenstance working on an article addressing the many ways to do this based on the use case).

> See also the couple comments I left.
> 
> Nit: Can we rename `index.html` to `demo.html` for consistency?

Done.

guides/ui-components/hovercard/guide.md
jamesnw: Added a class

guides/ui-components/hovercard/guide.md
jamesnw: [WCAG guidance](https://www.w3.org/WAI/WCAG21/Understanding/animation-from-interactions.html) is to disable "non-essential" animations, and I haven't seen recommendations to keep some types of animations. Happy to change this if there's other guidance.

bramus (changes requested):
~~There is a web feature called interest invokers that was designed to cover this usecase. You can try it in https://codepen.io/editor/bramus/pen/01a09116-9ed5-79c0-8b5b-0cc4c73f6807 (hover the “Rewrite” button, or tab into it).~~ There is also a polyfill available https://github.com/mfreed7/interestfor.

Was this considered?

_Edit: LOL, the code **is** using `interestfor` … was looking a bit to much at the View Transitions code. Question about the polyfill remains._

Furthermore, View Transitions seem a bit out of place for communicating the core message of this feature. I don’t think we want all sites that want to show a hovercard suddenly to also start using View Transitions.

guides/ui-components/hovercard/article-a.html
```diff
@@ -0,0 +1,89 @@
...
+        <p class="byline">By Delphi Aguilar</p>
+        <img src="https://images.unsplash.com/photo-1520552159191-e28a1d9f0d7e?q=80&w=400&auto=format&fit=crop" alt="A green hummingbird with a vibrant blue throat and a curved bill seated on a branch.">
+   		 <p>Heron egret pelican cormorant swan goose duck mallard teal kingfisher. Heron egret pelican cormorant swan goose duck
+  mallard teal kingfisher heron egret pelican cormorant swan goose duck mallard teal kingfisher. Heron egret pelican
+  cormorant swan goose duck mallard teal kingfisher heron egret pelican cormorant swan goose duck mallard teal kingfisher.</p>
+       Photo by <a href="https://unsplash.com/@licole?utm_source=unsplash&utm_medium=referral&utm_content=creditCopyText">Chris Charles</a> on <a href="https://unsplash.com/photos/blue-and-green-bird-on-top-of-brown-branch-during-daytime-9APFPoNb9iw?utm_source=unsplash&utm_medium=referral&utm_content=creditCopyText">Unsplash</a>
+   	 </article>
+    </main>
+    <a href="#"></a>
+    <script src="./script.js"></script>
```
bramus, line 87: Did you forget to commit the script?

guides/ui-components/hovercard/guide.md
```diff
@@ -3,11 +3,133 @@ name: hovercard
...
+       ensuring the view transition snapshot is taken at the right time. -->
+  <link rel="expect" href="#main-content" blocking="render">
+</head>
+<body>
+  <main id="main-content">
+    <h1 style="view-transition-name: --title">The buzz on hummingbirds</h1>
+    <img style="view-transition-name: --image" src="bird.jpg">
+  </main>
+</body>
+```
```
bramus, line 121: As mentioned in my previous comment, this seems non-essential to the pattern.

guides/ui-components/hovercard/guide.md
```diff
@@ -3,11 +3,133 @@ name: hovercard
...
+
+## Fallback strategies
+
+This pattern is a progressive enhancement and does not require a fallback. Users with browsers that do not support these patterns will see a standard link that takes them to the next page when they click.
```
bramus, line 125: As mentioned in my previous commment, there is a polyfill: https://github.com/mfreed7/interestfor

guides/ui-components/hovercard/guide.md
```diff
@@ -3,11 +3,133 @@ name: hovercard
...
+
+```css
+[popover].hovercard {
+  /* Position the card below the trigger and aligned with the inline end.
+     This assumes the trigger is the implicit anchor. */
+  position-area: block-end span-inline-end;
+}
+```
+
+## 3. Animating entry and exit
```
bramus, line 65: The CSS `interest-delay` property can be used to control the timings. I think it should be covered in this section. You declare this on the `[interestfor]` element.

bramus (changes requested):

guides/ui-components/hovercard/article-a.html
jamesnw: There isn't actually a script- I'm still getting used to the new CodePen 2.0 boilerplate 😆

bramus: I thought it looked familiar 😀

guides/ui-components/hovercard/guide.md
LeaVerou: +1. Should we tweak the guidance to mention both options, or always recommend a polyfill? Leaning towards the former.

LeaVerou:
> > ```
> > * I’m not convinced an exit transition is useful, it seems to get in the way.

With this I meant, when you hover away from the link, the fade out seems to get in the way. I think it might produce better UX to keep the transition when hovering, but hide abruptly when hovering away, since the user's attention is now elsewhere and the transition serves no purpose other than decoration.
But it's not a strong opinion. @bramus, any thoughts?

> > * When pressing Back you get the exit transition (even when you directly load the URL, i.e. _not_ by pressing the hovercard link!), which seems a bit weird.
> > ```
> 
> I think you're referencing the default view transition between pages, not the popover exit transition? Should we maybe have a separate guide describing how to disable root animation if you just want specific animations?

Previously, I was seeing the popover fade out even when I loaded a page via the URL directly. I can no longer reproduce this, no nvm!

> > ```
> > * Instead of duplicating content for the hovercard, would it make sense to fetch the HTML the first time the link is hovered and read it from there? That seems more maintainable and scalable, and this is PE anyway.
> > ```
> 
> This really depends on the use case. The hovercard will very rarely load the exact content of its destination- in most cases it will only have a snippet (as is the case in the demo) rather than the full article, or an alternate layout. (I'm by happenstance working on an article addressing the many ways to do this based on the use case).

I was thinking of reading the page metadata, but yes, it totally depends on the case (and how many there are). Perhaps we could mention it as an option?

bramus:
> > > ```
> > > * I’m not convinced an exit transition is useful, it seems to get in the way.
> > > ```
> 
> With this I meant, when you hover away from the link, the fade out seems to get in the way. I think it might produce better UX to keep the transition when hovering, but hide abruptly when hovering away, since the user's attention is now elsewhere and the transition serves no purpose other than decoration. But it's not a strong opinion. @bramus, any thoughts?

Yes, +1 to that. 

I also gave the feedback to Blink Engineering a few times already to change the default UA durations. They are way too long! Values that work for me are `0.20s 0.1s`.

One trick I also apply is to not have a start-delay when the keyboard was used to focus the `[interestfor]` element.

Full code:

```css
[interestfor] {
  interest-delay: 0.20s 0.1s;

  &:focus-visible {
    interest-delay: 0s;
  }
}
```

LeaVerou:
> Furthermore, View Transitions seem a bit out of place for communicating the core message of this feature. I don’t think we want all sites that want to show a hovercard suddenly to also start using View Transitions.

I think animating a preview of a page to the page itself is one of the primary use cases for view transitions!
But I do agree with @bramus that having to enable cross-doc view transitions *in general* to use a hovercard is weird.
Ideally, we'd only enable them when the hovercard is clicked, unless they are _separately_ desirable for navigation in general.
I believe we can do that by essentially making them a no-op unless the hovercard popover is open (or use JS to disable them "properly", but I'm not sure that's warranted).

----

Btw when I tried out the demo the morphing seems to be very temperamental and seems to very often fall back to a fade. I only got it to work in 1 out of 4 navigations below:


https://github.com/user-attachments/assets/497b5955-407a-4fa4-84e2-00b6dc3610ef

There is also currently discussion and imminent prototyping about `::interest-button` being added by default as a _sibling_. How should we handle this here?

bramus:
> There is also currently discussion and imminent prototyping about `::interest-button` being added by default as a _sibling_. How should we handle this here?

As it’s still in development and will be opt-in _(thanks to you! 🙏 )_ perhaps it’s fine to omit it for now?

LeaVerou (changes requested):
See comments. I do think this is in good shape, but we should figure out why the VTs don't work correctly most of the time, and not enable them for all navigations just to use a hovercard.

Also to disable the popover exit transition, since @bramus agreed with that too. I quite liked @bramus' idea of using different delays for keyboard navigation, and this is exactly the kind of trick agents won't do on their own!

LeaVerou:
> > There is also currently discussion and imminent prototyping about `::interest-button` being added by default as a _sibling_. How should we handle this here?
> 
> As it’s still in development and will be opt-in _(thanks to you! 🙏 )_ perhaps it’s fine to omit it for now?

Oh thank goodness! I saw that in passing and thought they had gone back to opt-out. 😮‍💨
Yes, totally fine to omit it then.

guides/ui-components/hovercard/guide.md
jamesnw: I adjusted to say it's PE, but you can fallback. I also created an `interest-invoker` feature to consolidate the fallback guidance.

guides/ui-components/hovercard/guide.md
jamesnw: Added as part of the new interest-invoker feature, and linked here.

jamesnw:
(Sorry for editing your comment when I was trying to quote it)
> One trick I also apply is to not have a start-delay when the keyboard was used to focus the `[interestfor]` element.
> 
> Full code:
> 
> ```css
> [interestfor] {
>   interest-delay: 0.20s 0.1s;
> 
>   &:focus-visible {
>     interest-delay: 0s;
>   }
> }
> ```
> 
> _(EDIT: Perhaps the interest-delay-start needs a very small value on `:focus-visible` to prevent the popover from appearing all over the place when tabbing quickly through a document)_

I prefer the  pattern of having a  start delay that's consistent across input method, and then setting the start to 0s if an interest-triggered popover is already open-

```css
:root:has(:interest-source) [interestfor] {
  interest-delay-start: 0s;
}
```

I added that in 3c5e5ee

bramus:
> I added that in [3c5e5ee](https://github.com/GoogleChrome/modern-web-guidance-src/commit/3c5e5eed754bf8f5579043c77fe97ab07497528a)

Can you also add it to the demo?

Also see my edit int he comment: perhaps it should _not_ be `0s` to allow quick tabbing through links without having the popover show erratically while you do so.

jamesnw:
I think I've addressed the transitions, but I may have misunderstood something. My technique for the conditional transition is this, but is there a better way? 

```
::view-transition-old(*),
::view-transition-new(*) {
  animation-duration: 0s;
}

:root:has([popover]:popover-open) {
  ::view-transition-old(*),
  ::view-transition-new(*) {
    animation-duration: 0.4s;
  }
}
```
> Can you also add it to the demo?
> 
> Also see my edit int he comment: perhaps it should _not_ be `0s` to allow quick tabbing through links without having the popover show erratically while you do so.

I added it to the demo, and give it a test. I think the 0s works, but am open to other thoughts. The specific case here is that you are tabbing through a list of items with interest-triggered popovers, so as soon as you tab to something without an interest-triggered popover, the delay-start reverts. If you tab TO something with an interest-triggered popover, but never stay long enough on one to trigger interest, no popovers don't open. I think that prevents unexpected behaviors.

bramus:
> > Can you also add it to the demo?
> > Also see my edit int he comment: perhaps it should _not_ be `0s` to allow quick tabbing through links without having the popover show erratically while you do so.
> 
> I added it to the demo, and give it a test. I think the 0s works, but am open to other thoughts.

Oh right, I was talking about the `&:focus-visible` selector I suggested, whereas your code was about `:root:has(:interest-source) [interestfor] { … }`. Didn’t notice 🤦 

The `:root:has(:interest-source) [interestfor]` is definitely valid. The thing I suggested is to tweak the initial delay when tabbing through things. But with the current shorter `delay-start` of `0.2s` it already seems fine :)

---

> I think I've addressed the transitions, but I may have misunderstood something. My technique for the conditional transition is this, but is there a better way?
> 
> ```
> ::view-transition-old(*),
> ::view-transition-new(*) {
>   animation-duration: 0s;
> }
> 
> :root:has([popover]:popover-open) {
>   ::view-transition-old(*),
>   ::view-transition-new(*) {
>     animation-duration: 0.4s;
>   }
> }
> ```

The problem here is that you need to target the groups, as they — with your code — still have a default `animation-duration`. So, like this:

```css
:root:has([popover]:popover-open) {
  ::view-transition-group(*) {
    animation-duration: 0.4s;
  }
}
```

(The `animation-duration` set on the `-group` will inherit onto the `-image-pair`, `-old`, and `-new` thanks to some UA stylesheet magic that sets `animation-duration: inherits` on those elements).

That said, I think we should really remove it from this demo. I don’t see how this is relevant to the “hovercard” pattern on its own. Like, Wikipedia has hovercards … without any View Transitions.

LeaVerou:
> That said, I think we should really remove it from this demo. I don’t see how this is relevant to the “hovercard” pattern on its own. Like, Wikipedia has hovercards … without any View Transitions.

Do you think VTs hurt the user experience in some way, or is this on the basis that most hovercards don't use them?

bramus:
> Do you think VTs hurt the user experience in some way, or is this on the basis that most hovercards don't use them?

I think adding VTs deviate too much from the question at hand here. Like, we could also add speculation rules to these pages, but that is adjacent to the core ask.

jamesnw:
> > Do you think VTs hurt the user experience in some way, or is this on the basis that most hovercards don't use them?
> 
> I think adding VTs deviate too much from the question at hand here. Like, we could also add speculation rules to these pages, but that is adjacent to the core ask.

Would it be sufficient to keep VTs, but as optional? Perhaps we could also more clearly scope that this use case is "Link Preview Hovercards"?

guides/ui-components/hovercard/guide.md
knowler: I personally avoid grouping all animation into reduced motion, since I do think that some animation can be beneficial for user experience. (Aside, another opinion: I also like layer on motion animations as a progressive enhancement using `prefers-reduced-motion: no-preference`)

Some reading:

- https://alistapart.com/article/designing-safer-web-animation-for-motion-sensitivity/#section7
- https://css-tricks.com/nuking-motion-with-prefers-reduced-motion/
- https://css-tricks.com/revisiting-prefers-reduced-motion/#animation-isnt-unnecessary

LeaVerou (commented):
@bramus 
> > Do you think VTs hurt the user experience in some way, or is this on the basis that most hovercards don't use them?
> 
> I think adding VTs deviate too much from the question at hand here. Like, we could also add speculation rules to these pages, but that is adjacent to the core ask.

In your opinion when are VTs warranted, in general? Because I can't think of a use case where they are _actually_ essential, they are always an enhancement, right?

That said, I agree with @jamesnw that making them optional makes sense, but then we should probably include some guidance about when to reach for them. Any thoughts on what that might be, @bramus? I cannot personally think of any case where they could be _harmful_, but you know VTs more deeply.

If we remove them entirely the guide is practically empty. All it would be saying is basically "here's how to make a tooltip with interest invokers". Which we already have another guide for…
Unless we have additional guidance that isn't VTs and is specific to hovercards, another way to proceed might be to specialize the use case so that the VT is part of the description and agents pick up this guide only when that's desirable. 🤔

PR #1421, opened 2026-09-09
[#1236] checkbox use case

sturobson:
this effects #1236

### DIFF
--- a/guides/ui-components/checkbox/guide.md
+++ b/guides/ui-components/checkbox/guide.md
@@ -9,16 +9,10 @@ web-feature-ids:
 
 # Styling Checkboxes
 
-Checkboxes are a core interactive element of web forms. Previous attempts to
-style checkboxes often involved hiding the native `<input>` element entirely and
-rendering a complex tree of custom wrapper `<div>`s or `<span>`s with ARIA
-roles. This legacy approach is highly error-prone, which can break native keyboard
-navigation, effect screen reader announcements, abd break native form validation, and standard
-form submission.
+Checkboxes are a core interactive element of web forms. Previous attempts to style checkboxes often involved hiding the native `<input>` element entirely and rendering a complex tree of custom wrapper `<div>`s or `<span>`s with ARIA roles. This legacy approach is highly error-prone, which can break native keyboard navigation, affect screen reader announcements, break native form validation, and disrupt standard form submission.
 
-Modern CSS allows us to style checkboxes directly—either by lightly customizing the native appearance using the **`accent-color`** property, or by overriding default presentation entirely using **`appearance: none`** directly on the `<input type="checkbox">` element.
+Modern CSS allows you to style checkboxes directly, either by customizing the native appearance using the **`accent-color`** property, or by overriding default presentation entirely using **`appearance: none`** directly on the `<input type="checkbox">` element.
 
----
 
 ## 1. The Bare-Bones HTML Structure
 
@@ -36,7 +30,6 @@ Always use semantic, accessible HTML. Associate the checkbox with its label by n
 <label for="terms-input">I accept the terms and conditions</label>
 ```
 
----
 
 ## 2. Implementation
 
@@ -46,7 +39,6 @@ For quick, brand-consistent styling that preserves 100% of native rendering and 
 
 ```css
 .checkbox-native {
-  /* Set the checkbox background tint for checked states */
   accent-color: #1a73e8;
   
   /* Sizing is controlled via relative units */
@@ -56,32 +48,26 @@ For quick, brand-consistent styling that preserves 100% of native rendering and 
 }
 ```
 
----
 
 ### Method B: Fully Custom Styling (`appearance: none`)
 
 To build bespoke checkmark shapes, borders, and animations, use `appearance: none` directly on the `<input>` element. This strips the native browser styling while leaving the `<input>` in the DOM as the interactive target.
 
-We can render a crisp vector checkmark with **CSS masking** on a pseudo-element. This keeps our markup entirely flat and lets us style the checkmark color using `currentColor` (matching the parent's text color).
+You can render a crisp vector checkmark with **CSS masking** on a pseudo-element. This keeps your markup entirely flat and lets you style the checkmark color using `currentColor` (matching the parent's text color).
 
 ```css
 .checkbox-custom {
   /* Remove default browser visual box */
   appearance: none;
-  -webkit-appearance: none;
+  -webkit-appearance: none; /* Legacy support */
   
-  /* Create custom design shell */
+  /* Create custom grid layout shell */
   display: inline-grid;
   place-content: center;
   width: 1.25em;
   height: 1.25em;
-  border: 2px solid #74777f;
-  border-radius: 4px;
-  background-color: #ffffff;
+  border: 2px solid currentColor;
   cursor: pointer;
-  
-  /* Transitions for interactions */
-  transition: border-color 0.15s ease, background-color 0.15s ease;
 }
 
 /* Custom Checkmark Icon via ::before */
@@ -96,19 +82,7 @@ We can render a crisp vector checkmark with **CSS masking** on a pseudo-element.
   
   /* Scale to 0 (hidden) by default to animate check/uncheck */
   transform: scale(0);
-  transition: transform 0.15s cubic-bezier(0.12, 0.8, 0.32, 1.2);
-}
-
-/* Hover state styling */
-.checkbox-custom:hover:not(:disabled) {
-  border-color: #43474e;
-}
-
-/* Checked state color changes */
-.checkbox-custom:checked {
-  background-color: #1a73e8;
-  border-color: #1a73e8;
-  color: #ffffff; /* Color of checkmark background-color (currentColor) */
+  transition: transform 0.15s ease;
 }
 
 /* Animate checked scale */
@@ -118,14 +92,11 @@ We can render a crisp vector checkmark with **CSS masking** on a pseudo-element.
 
 /* Disabled State */
 .checkbox-custom:disabled {
-  background-color: #e1e2e4;
-  border-color: #8e9196;
-  color: #8e9196;
   cursor: not-allowed;
+  opacity: 0.5;
 }
 ```
 
----
 
 ## 3. Styling the Indeterminate State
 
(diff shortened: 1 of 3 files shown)


guides/ui-components/checkbox/guide.md
```diff
@@ -5,3 +5,146 @@ web-feature-ids:
...
+
+/* Custom Checkmark Icon via ::before */
+.checkbox-custom::before {
+  content: "";
+  width: 0.7em;
+  height: 0.7em;
+  background-color: currentColor;
+  
+  /* Render a custom checkmark SVG via CSS mask */
+  mask: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none" stroke="black" stroke-width="3.5" stroke-linecap="round" stroke-linejoin="round"><polyline points="20 6 9 17 4 12"></polyline></svg>') no-repeat center / contain;
```
LeaVerou: We should store this in a token (e.g. `--icon-checkmark`) and reference it. Ideally, once the icons guide is ready, we could even reference that.

guides/ui-components/checkbox/guide.md
```diff
@@ -5,3 +5,146 @@ web-feature-ids:
...
+  
+  /* Scale to 0 (hidden) by default to animate check/uncheck */
+  transform: scale(0);
```
LeaVerou: Why not use `scale` instead? Then you can also add `individual-transforms` in the frontmatter as a web feature.

guides/ui-components/checkbox/guide.md
```diff
@@ -5,3 +5,146 @@ web-feature-ids:
...
+The indeterminate state represents a checkbox that is neither fully checked nor unchecked (commonly used in tree-views or nested categories). 
+
+In HTML, the `indeterminate` state cannot be set via an attribute; it **must be configured programmatically via JavaScript**. You style it using the `:indeterminate` pseudo-class in CSS:
+
+```javascript
+// Programmatically set the indeterminate property to true
+document.getElementById('my-checkbox').indeterminate = true;
+```
+
+For native checkboxes, the browser automatically styles the indeterminate state (including matching the custom color set via `accent-color`).
```
LeaVerou: This could be condensed _a lot_. Agents know about this already. The purpose of this section is to make sure they don't forget `:indeterminate`, not to teach them what it does and how it works.

guides/ui-components/checkbox/guide.md
```diff
@@ -5,3 +5,146 @@ web-feature-ids:
...
+/* Styling native indeterminate checkbox accent-color */
+.checkbox-native:indeterminate {
+  accent-color: #1a73e8;
+}
+
+/* Styling custom indeterminate checkbox */
+.checkbox-custom:indeterminate::before {
+  transform: scale(1);
+  /* Render a custom horizontal dash SVG for indeterminate states */
+  mask: url('data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none" stroke="black" stroke-width="4.5" stroke-linecap="round" stroke-linejoin="round"><line x1="5" y1="12" x2="19" y2="12"></line></svg>') no-repeat center / contain;
```
LeaVerou: Same comment as above. Perhaps `--icon-dash`?

guides/ui-components/checkbox/guide.md
sturobson: 👍🏼 I missed it here, but thought it would be good to use a custom property in this way for the sub menus in the `<menu>` guide. Will update this accordingly.

guides/ui-components/checkbox/guide.md
sturobson: some of it's muscle memory 😆

guides/ui-components/checkbox/guide.md
```diff
@@ -5,3 +5,146 @@ web-feature-ids:
...
+  }
+  ```
+- **DO NOT** use `outline: none` or hide focus indicators on checkboxes.
+- **DO** ensure the checkbox or the label has an active click/touch target of at least `44px x 44px` to comply with mobile touch guidelines.
+- **DO** provide a `:disabled` style variant that lowers opacity, alters background, and sets `cursor: not-allowed` to convey active states clearly.
+
+
+## Fallback strategies
+
+For browsers that don't support the standard CSS Masking, custom checkbox checkmarks will fall back to stacked text indicators or can leverage standard SVG background images inside `@supports` checks.
```
LeaVerou: [Masks are Baseline Widely Available](https://web-platform-dx.github.io/web-features-explorer/features/masks/) so we don't need this. The only feature that's not BWA here is `accent-color`.

LeaVerou (changes requested):
This is shaping up really nicely! See comments for a few changes, but nothing major.

sturobson:
@LeaVerou I've hopefully addressed the points raised. I'll add details to each comment.

guides/ui-components/checkbox/guide.md
sturobson: I've now added this with a custom property. I've not included all the things from the icons guide (`@property` etc) as I thought it would over complicate matters for this use case.

guides/ui-components/checkbox/guide.md
sturobson: updated to `scale: Nn` and `transition: scale X Y Z` where appropriate.

guides/ui-components/checkbox/guide.md
sturobson: I've reduced this down

guides/ui-components/checkbox/guide.md
sturobson: as per the other comment above - resolved using custom properties

guides/ui-components/checkbox/guide.md
sturobson: removed this

guides/ui-components/checkbox/guide.md
```diff
@@ -4,4 +4,154 @@ description: Style checkboxes, including custom icons, layout, colors, check/unc
 web-feature-ids:
   - accent-color
   - indeterminate
+  - individual-transforms
+  - masks
 ---
+
+# Styling Checkboxes
+
+Checkboxes are a core interactive element of web forms. Previous attempts to style checkboxes often involved hiding the native `<input>` element entirely and rendering a complex tree of custom wrapper `<div>`s or `<span>`s with ARIA roles. This legacy approach is highly error-prone, which can break native keyboard navigation, affect screen reader announcements, break native form validation, and disrupt standard form submission.
```
LeaVerou:
```suggestion
```

I think the guide works fine without this paragraph.

guides/ui-components/checkbox/guide.md
```diff
@@ -4,4 +4,154 @@ description: Style checkboxes, including custom icons, layout, colors, check/unc
...
+<!-- Nested Label Pattern (Implicit Association) -->
+<label>
+  <input type="checkbox" name="subscribe" class="checkbox-native" />
+  Subscribe to newsletter
+</label>
+
+<!-- Explicit Association Pattern -->
+<input type="checkbox" id="terms-input" name="terms" class="checkbox-custom" />
+<label for="terms-input">I accept the terms and conditions</label>
+```
```
LeaVerou: Agents know how checkboxes and labels work. I'd just condense this to one sentence telling them to use regular checkboxes, associated with labels as usual.

guides/ui-components/checkbox/guide.md
```diff
@@ -4,4 +4,154 @@ description: Style checkboxes, including custom icons, layout, colors, check/unc
...
+
+
+## 2. Implementation
+
+### Method A: Native Customization (`accent-color`)
+
+For quick, brand-consistent styling that preserves 100% of native rendering and operating system focus indicators, use the `accent-color` property. This colors the checkbox background when checked/active, and the browser automatically selects a high-contrast checkmark color.
```
LeaVerou: Note that this is not how Safari works currently. Set the accent-color to `yellow` and observe the difference.

Chrome: 

Image

Safari:

Image

Safari *changes* the accent-color instead of finding a suitable foreground color.

I'd suggest recommending against light colors for this approach and say if they need to use light accent color, to prefer the second approach.

guides/ui-components/checkbox/guide.md
```diff
@@ -4,4 +4,154 @@ description: Style checkboxes, including custom icons, layout, colors, check/unc
...
+/* Custom Checkmark Icon via ::before */
+.checkbox-custom::before {
+  content: "";
+  width: 0.7em;
+  height: 0.7em;
+  background-color: currentColor;
+  
+  /* Apply the mask utilizing the checkbox-icon custom property */
+  mask: var(--checkbox-icon) no-repeat center / contain;
+  -webkit-mask: var(--checkbox-icon) no-repeat center / contain; /* Legacy support */
```
LeaVerou:
```suggestion
```

No legacy in the core guidance, that belongs to the fallbacks section (and I believe is not needed there either, since unprefixed masks are now BWA).

Also, even when using vendor-prefixed properties, they should _never_ be last!

guides/ui-components/checkbox/guide.md
```diff
@@ -4,4 +4,154 @@ description: Style checkboxes, including custom icons, layout, colors, check/unc
...
+.checkbox-custom {
+  /* Remove default browser visual box */
+  appearance: none;
+  -webkit-appearance: none; /* Legacy support */
+  
+  /* Create custom grid layout shell */
+  display: inline-grid;
+  place-content: center;
+  width: 1.25em;
+  height: 1.25em;
```
LeaVerou: `aspect-ratio: 1` would help reduce repetition here and below.
But also, `1.25em` and `0.7em` kind of magic numbers. What do they represent?  What is the relationship between them? 

We have three lengths:
- a) Checkbox dimensions
- b) Icon dimensions
- c) Spacing around the icon

c = a - b

We’re currently specifying a and b. To adjust c, you need to do the math and adjust one of the others. Is that optimal? 
If it would be more convenient to adjust c (spacing) separately and have a or b adapt, we can make `padding` work (which currently does nothing).

If we keep it this way, it could be convenient to have an `--icon-size` variable that can be specified on the checkbox directly, so that consumers don't need to be tweaking pseudo-elements.

LeaVerou (changes requested):
This looks great. See comments for some, mostly minor, issues and suggestions for tighetning. 

I think the most important ones to address are:
- Removing the `-webkit-mask`
- The Safari accent-color issue

The rest are more suggestions than blockers.

guides/ui-components/checkbox/guide.md
rviscomi: @LeaVerou could we merge the accent-color guidance into https://github.com/GoogleChrome/modern-web-guidance-src/pull/900 and reuse that here?

PR #1425, opened 2026-09-10
[ATL triage] Set new `content`and `needs-atl` labels

rviscomi:
I wanted an easier way to do a couple of routine maintenance tasks:

- see which open PRs are content-related as opposed to infra work
- see which content issues or PRs don't already have a dedicated ATL

This PR updates the ATL triage script to auto-add these labels as needed.

While testing this, I needed to make a few related changes/fixes:

- added @LeaVerou as the ATL for the entire CSS web feature group, so she'd be auto-assigned to more new-feature issues (otherwise I'd need to manually label the issue with `css`)
- fixed how the triage workflow didn't automatically run on forked PRs by switching to `pull_request_target`
- added the ability to run triage dry runs
- removed keyword matching as a means of triaging issues (eg DONT assign Bramus just because the description uses the word "transitions")
- stopped updating labels or assignments on draft PRs

<details>
<summary>Note: zizmor flagged pull_request_target as dangerous. I suppressed that warning. Here's why it's safe...</summary>

#### 1. Untrusted PR Code Execution
* **Zizmor Warning**: The most common vulnerability with `pull_request_target` is checking out untrusted code from the fork PR (`ref: ${{ github.event.pull_request.head.sha }}`) and running `npm test`, linters, or build scripts.
* **Our Defense**: We do **not** check out the PR's branch. `actions/checkout` runs with default settings, which checks out `main` (the target repository's trusted code). The triage script that executes is 100% trusted repository code.

#### 2. Shell & Argument Injection (e.g., `xargs`, shell interpolation)
* **Zizmor Warning**: An attacker can craft malicious branch names or file paths (e.g. `guides/css/test"; rm -rf /; ".md`) that break out of shell commands if interpolated into command strings.
* **Our Hardening**: 
  - We refactored all CLI calls in [`guides/atl-triage.ts`](file:///Users/rviscomi/git/guidance/guides/atl-triage.ts#L304-L452) from `child_process.execSync` to **`child_process.execFileSync('gh', [...args])`**. 
  - `execFileSync` bypasses `/bin/sh` entirely and passes the argument array directly to the OS kernel (`execve`). Even if a filename contains quotes, semicolons, backticks, or `$()`, it is treated strictly as a literal string argument to `gh`.
  - Added strict path validation: any path containing directory traversal (`..`) or not starting with `guides/` is immediately rejected before querying GitHub.

#### 3. Environment Injection (`GITHUB_ENV`, `GITHUB_PATH`, `LD_PRELOAD`)
* **Zizmor Warning**: Untrusted data written to `$GITHUB_ENV` or `$GITHUB_PATH` can hijack executables or load malicious shared libraries.
* **Our Defense**: Neither the workflow YAML nor `atl-triage.ts` writes to `$GITHUB_ENV` or `$GITHUB_PATH`.

#### 4. Template Injection in `run:` Steps
* **Zizmor Warning**: Using `${{ github.event... }}` expressions inside inline bash scripts allows attackers to inject arbitrary bash commands.
* **Our Defense**: The workflow's `run:` block contains **zero** template expressions:
  ```yaml
  run: |
    node --experimental-strip-types guides/atl-triage.ts
  ```

---

### Additional Hardening from Zizmor Recommendations

In [`.github/workflows/atl-triage.yml`](file:///Users/rviscomi/git/guidance/.github/workflows/atl-triage.yml#L6-L19), we implemented all the hardening measures recommended by Zizmor for workflows using `pull_request_target`:

```yaml
  # zizmor: ignore[dangerous-triggers] Trusted checkout of main; no PR code executed; queries API via execFileSync
  pull_request_target:
    types: [opened, synchronize, labeled, ready_for_review]
    branches: [main]

permissions:
  issues: write
  pull-requests: write
  contents: read

jobs:
  triage:
    runs-on: ubuntu-latest
    if: github.repository == 'GoogleChrome/modern-web-guidance-src'
```

1. **Branch filter (`branches: [main]`)**: Restricts the workflow to PRs targeting `main`, preventing execution against stale or abandoned branches.
2. **Repository guard (`if: github.repository == ...`)**: Ensures the workflow never runs with privileges inside an external fork if a user enables Actions on their fork.
3. **Least-privilege permissions**: Permissions are explicitly scoped to only what the triage script requires (`issues: write`, `pull-requests: write`, `contents: read`).

</details>


micahjo7 (commented):
would you mind also updating `gd-dev-eval` to just `eval` for consistency?

guides/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "type": "module",
   "scripts": {
-    "test": "node --experimental-strip-types --test --test-reporter ${TEST_REPORTER:-dot} **/*.test.ts",
+    "test": "GH_TOKEN=fake-test-token GITHUB_TOKEN=fake-test-token node --experimental-strip-types --test --test-reporter ${TEST_REPORTER:-dot} **/*.test.ts",
```
micahjo7: I see these values are also being set in `atl-triage.test.ts`. Do we need them in both places?

rviscomi: No, removed from here

rviscomi (commented):
Actually, I reverted the `gd-dev-content` change. I didn't realize how much automation relied on that label, which we wouldn't want to also apply to anything tagged with the broader `content` label. So I left `gd-dev-eval` as-is.

guides/atl-triage.ts
```diff
@@ -420,30 +834,52 @@ export function handlePR(
       const category = parts[1];
       const guideName = parts[2];
       const filename = parts[parts.length - 1];
-      if (contentFilenames.has(filename) || isContentLabelled) {
+      if (SME_CONTENT_FILENAMES.has(filename) || isContentLabelled) {
+        hasEvaluatedContent = true;
+        const relativeGuidePath = `guides/${category}/${guideName}/${GUIDE_FILE}`;
```
micahjo7, line 849: nit: if `file` is `guides/modern-web-guidance/SKILL.md`, then this will be `guides/modern-web-guidance/SKILL.md/guide.md`

we should also avoid potential errors for changes in lib (or anything not in "category" folders?)

guides/atl-triage.ts
```diff
@@ -408,9 +819,12 @@ export function handlePR(
...
   console.log(`PR modified ${files.length} files.`);
   
-  const contentFilenames = new Set([GUIDE_FILE, EXPECTATIONS_FILE, SKILL_FILE]);
-  const isContentLabelled = labels.includes('gd-dev-content');
+  const isContentLabelled = labels.includes('content') || labels.includes('gd-dev-content');
   const matchedAtls = new Set<string>();
+  const allResolvedAtls = new Set<string>();
+  let hasEvaluatedContent = false;
+
+  const touchesSmeContent = files.some(isSmeContentFile);
```
micahjo7, line 830: should `features/*.md` files be included in this? right now, changes to those files will be add the `needs-atl` but not the `content` label

guides/atl-triage.ts
```diff
@@ -194,8 +303,33 @@ export function getAtlsFromDescription(description: string, atlConfig: AtlConfig
...
-    const eventsOutput = child_process.execSync(
-      `gh api repos/{owner}/{repo}/issues/${issueNumber}/events --paginate --jq '.[] | select(.event == "unassigned") | .assignee.login'`,
+    const eventsOutput = child_process.execFileSync(
+      'gh',
+      [
+        'api',
+        `repos/{owner}/{repo}/issues/${issueNumber}/events`,
+        '--paginate',
+        '--jq',
+        '.[] | select(.event == "unassigned" and (.actor.type != "Bot" and (.actor.login | endswith("[bot]") | not))) | .assignee.login'
```
micahjo7: nit: If an event was performed by a deleted/ghost account, `.actor` or `.actor.login` can be null. JQ's endswith() fails with exit code 5 when passed a non-string input, which causes execFileSync to throw.

```suggestion
        '.[] | select(.event == "unassigned" and (.actor and .actor.type != "Bot" and ((.actor.login // "") | endswith("[bot]") | not))) | .assignee.login'
```

similarly for lines 331, 476, and 494

guides/atl-triage.ts
```diff
@@ -287,37 +590,120 @@ export function handleIssue(
...
+        console.warn(`Failed to fetch event history for issue #${issueNumber}:`, err);
+      }
+
+      if (!wasPreviouslyUnlabeled) {
+        issueLabelsToAdd.push('needs-atl');
+      } else {
+        console.log(`Skipping adding 'needs-atl' to issue #${issueNumber} because it was previously removed.`);
+      }
+    }
+  } else if (hasAtl) {
```
micahjo7: edge case for if a PR originally has content changes, but on subsequent commits removes them

```suggestion
  } else if (hasAtl || !isContent) {
```

and same thing in line 940 (check also `|| !hasEvaluatedContent`)

guides/atl-triage.ts
rviscomi: Done

guides/atl-triage.ts
rviscomi: Done

guides/atl-triage.ts
rviscomi: Done

guides/atl-triage.ts
rviscomi: Done

rviscomi (commented):
Thanks! All suggestions applied

micahjo7 (approved):

PR #1426, opened 2026-09-10
[#1234] card use case

sturobson:
addresses #1234


LeaVerou (changes requested):
I need to review this more carefully, but from a first look one thing I noticed is that it's repeating guidance from the other guides. It should either link to them (via `GUIDE_REF`) or transclude the relevant sections (if absolutely essential).

sturobson:
I've made some changes to this to reduce the amount of guidance whilst trying to ensure it stays focused on the guide it is meant to be.

PR #1428, opened 2026-09-10
[#1223] Responsive Table use case

### LINKED ISSUE #1223: Create guide and evals for the responsive-table use case
Build a data table that presents tabular data with proper header semantics, and remains readable and navigable on small screens and with assistive technologies.

Affected web-feature IDs: 

Use case subdir: [guides/ui-atoms/responsive-table](https://github.com/GoogleChrome/modern-web-guidance-src/tree/main/guides/ui-atoms/responsive-table)

**Required files:**
- [ ] demo.html
- [ ] expectations.md
- [ ] tasks/task.md
- [ ] negative-demo.html
- [ ] grader.ts

jamesnw:
Addresses #1223

### DIFF
--- a/guides/ui-atoms/responsive-table/demo.html
+++ b/guides/ui-atoms/responsive-table/demo.html
@@ -1,0 +1,145 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Responsive Table Demo</title>
+    <style>
+      body {
+        font-family:
+          system-ui,
+          -apple-system,
+          sans-serif;
+        margin: 2rem;
+        --z-row-header: 5;
+        --z-col-header: 10;
+        --z-corner: 15;
+      }
+      @media (width >= 600px) {
+        table {
+          border-collapse: separate; /* Required for sticky borders to behave better in some browsers */
+          border-spacing: 0;
+          width: 100%;
+        }
+        th,
+        td {
+          border: 1px solid #ccc;
+          padding: 0.5rem;
+          text-align: left;
+          white-space: nowrap;
+          background: white;
+        }
+        thead {
+          position: sticky;
+          top: 0;
+          z-index: var(--z-col-header);
+        }
+        thead th {
+          background: #eee;
+        }
+        /* Sticky row headers - must stay on TH to stick horizontally within the row */
+        th[scope="row"] {
+          position: sticky;
+          background: #f9f9f9;
+          z-index: var(--z-row-header);
+        }
+        /* First row header column */
+        th:nth-child(1) {
+          left: 0;
+        }
+
+        .table-wrapper {
+          overflow: scroll;
+          max-inline-size: 100cqi;
+          max-block-size: 100cqb;
+        }
+      }
+      @media (width < 600px) {
+        table {
+          width: 100%;
+        }
+        thead {
+          /* Visually hidden, display none can break screen readers */
+          position: absolute !important;
+          clip-path: inset(50%) !important;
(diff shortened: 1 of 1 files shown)


guides/ui-atoms/responsive-table/guide.md
```diff
@@ -1,5 +1,137 @@
...
+
+# Responsive tables
+
+Large data tables often become unreadable on small screens as columns overflow or shrink beyond legibility. This guide demonstrates how to use **sticky positioning** to keep headers visible during scrolling and how to transform the table into a mobile-friendly "stacked" layout when space is limited.
+
+## Recommended Approach
+
+The core of a responsive table is maintaining the relationship between data cells and their headers.
+
+1.  **Semantic Foundation**: Use standard `<table>` elements with `<thead>`, `<tbody>`, and `<th>` elements. Use `scope="col"` and `scope="row"` to ensure assistive technologies can map data correctly.
```
LeaVerou: Isn't this only necessary when `auto` would infer the wrong scope? It will get very noisy if every single `<th>` has `scope`.

LeaVerou (changes requested):
This still repeats the header on every single `<td>` which is _extremely_ unwieldy.
I haven't seen evals, but my hunch is that as written, this is something an agent could do on their own.
The purpose of this guide is to improve on that a bit.

Idea: what about using attributes on the table itself, then `attr()` + custom property registration to pass them down as a resolved `<string>` so it can be added via `::before`?
If that doesn't work, then just CSS variables on the `<table>`, at least those would only be specified once.
Also, the demo is formatted like a data grid. For that sort of table, scrolling on mobile is probably fine since it scrolls even on desktop. This vertical treatment is better suited for tables that are fine on desktop, but not on narrow viewports.

I also wonder if we may want to wrap in a `<details>` with its `<summary>` positioned below it, so that we can show a limited height on mobile at first, and then users can expand the full table. This would also likely pull in `::details-content`, and `interpolate-size` as PE. On large viewports, we can apply `display: contents` to the `<details>`, `<summary>` and `::details-content` so that there is no collapsing. @knowler would that pattern be ok for a11y? If not, any other?

guides/ui-atoms/responsive-table/guide.md
jamesnw: Correct- this was lingering from an earlier more complicated example. Removed.

jamesnw:
> This still repeats the header on every single `<td>` which is _extremely_ unwieldy. I haven't seen evals, but my hunch is that as written, this is something an agent could do on their own. The purpose of this guide is to improve on that a bit.
> 
> Idea: what about using attributes on the table itself, then `attr()` + custom property registration to pass them down as a resolved `<string>` so it can be added via `::before`? If that doesn't work, then just CSS variables on the `<table>`, at least those would only be specified once. 

Whoops- it actually does use the "CSS variables on the `<table>`" strategy, I just forgot to remove the attributes from each cell. I tried out the " `attr()` + custom property" strategy, but it just seemed like a few unneeded layers. 

>Also, the demo is formatted like a data grid. For that sort of table, scrolling on mobile is probably fine since it scrolls even on desktop. This vertical treatment is better suited for tables that are fine on desktop, but not on narrow viewports.

Simplified to fewer columns and more realistic data.

> I also wonder if we may want to wrap in a `<details>` with its `<summary>` positioned below it, so that we can show a limited height on mobile at first, and then users can expand the full table. This would also likely pull in `::details-content`, and `interpolate-size` as PE. On large viewports, we can apply `display: contents` to the `<details>`, `<summary>` and `::details-content` so that there is no collapsing. @knowler would that pattern be ok for a11y? If not, any other?

It seems this is outside of the prompt- I'd be surprised as a developer if my table was partially hidden while trying to make it responsive.

PR #1432, opened 2026-09-14
[#1224] drag use cass 

sturobson:
addresses #1224

### DIFF
--- a/guides/ui-behaviors/drag/guide.md
+++ b/guides/ui-behaviors/drag/guide.md
@@ -4,3 +4,134 @@ description: Support dragging to move a certain overlay element (e.g. a dialog) 
 web-feature-ids:
   - user-select
 ---
+
+# Implementing Draggable Overlay Elements (Dialogs, Panels, Widgets)
+
+## Purpose
+This guide helps you implement a highly interactive "drag to move" behavior for overlay elements (dialogs, modals, floating panels, widgets) positioned freely on a page. It covers required techniques, common accessibility pitfalls, and fallback support.
+
+## Core Concept
+There is no native browser API for freely repositioning an element by dragging it around the viewport. The HTML Drag and Drop API (`draggable="true"`) is designed for drag-and-drop data transfer (such as reordering lists or uploading files), not continuous visual x/y repositioning. You must implement repositioning manually using Pointer, Mouse, or Touch events.
+
+---
+
+## Basic Implementation
+
+### 1. HTML Markup Structure
+Keep your markup semantic and clean. Designate a header inside the overlay element to behave as the drag handle:
+
+```html
+<div class="dialog" id="dialog">
+  <!-- The header bar acts as the drag handle -->
+  <div class="dialog-header" id="dialogHeader">Drag me</div>
+  <div class="dialog-body">
+    <p>Dialog content.</p>
+    <!-- Explicitly selectable section -->
+    <div class="selectable">This text remains copyable.</div>
+  </div>
+</div>
+```
+
+### 2. Styles (CSS)
+Apply `user-select: none` only to the drag handle, and use `touch-action: none` to prevent touch-screen scrolling gestures from interfering with the drag:
+
+```css
+.dialog {
+  position: absolute;
+  /* Prevent touch gestures like scrolling from competing with the drag */
+  touch-action: none; 
+}
+
+.dialog-header {
+  cursor: grab;
+  /* REQUIRED: prevents text-highlighting artifacts during a drag gesture */
+  user-select: none;
+  -webkit-user-select: none; /* Legacy support */
+  -ms-user-select: none;     /* Legacy support */
+}
+
+.dialog-header.dragging {
+  cursor: grabbing;
+}
+
+.selectable {
+  /* Enforces text selection inside the content body */
+  user-select: text;
+  -webkit-user-select: text;
+}
+```
+
+### 3. Repositioning Script (JS / Pointer Events)
+Using standard **Pointer Events** (`pointerdown`, `pointermove`, `pointerup`) is the preferred approach because they automatically unify mouse, touch, and stylus actions into a single event model.
+
+```javascript
+const dialog = document.getElementById('dialog');
+const header = document.getElementById('dialogHeader');
+let isDragging = false, offsetX = 0, offsetY = 0;
+
+// Track coordinate offsets to prevent the dialog from "jumping" on grab
+header.addEventListener('pointerdown', (e) => {
+  if (e.target.closest('button')) return; // ignore close buttons
+  e.preventDefault();
+  
+  isDragging = true;
+  header.classList.add('dragging');
+  
+  const rect = dialog.getBoundingClientRect();
+  offsetX = e.clientX - rect.left;
+  offsetY = e.clientY - rect.top;
+});
+
+// Bind move and up to the document so dragging continues even if the cursor leaves the handle
+document.addEventListener('pointermove', (e) => {
+  if (!isDragging) return;
+  
+  let left = e.clientX - offsetX;
+  let top = e.clientY - offsetY;
+  
+  // Bounds checking: constrain dialog coordinates within the viewport limits
+  const maxLeft = window.innerWidth - dialog.offsetWidth;
+  const maxTop = window.innerHeight - dialog.offsetHeight;
+  
+  left = Math.max(0, Math.min(left, maxLeft));
+  top = Math.max(0, Math.min(top, maxTop));
+  
+  dialog.style.left = `${left}px`;
+  dialog.style.top = `${top}px`;
+});
+
+document.addEventListener('pointerup', () => {
+  isDragging = false;
+  header.classList.remove('dragging');
+});
+```
+
+---
+
+## Strategic Implementation & Best Practices
+
+*   **DO use `user-select: none` strictly on the drag handle**, not the entire dialog. Ensure users can highlight and copy content inside the dialog body.
+*   **DO calculate cursor coordinate offsets.** Capturing the delta between pointer coordinates and the dialog's top-left corner on grab prevents the dialog from "jumping" or snapping.
+*   **DO attach move and release events to the `document`**, not the drag handle. Fast cursor movements can temporarily outrun the handle, resulting in a frozen drag if listeners are bound only to the header.
+*   **DO perform bounds-checking.** Always constrain positioning using `Math.max` and `Math.min` against the window dimensions so users cannot drag dialogs off-screen.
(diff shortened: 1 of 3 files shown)


guides/ui-behaviors/drag/guide.md
```diff
@@ -4,3 +4,252 @@ description: Support dragging to move a certain overlay element (e.g. a dialog)
...
+     │ Modal (showModal())     │     │ Non-Modal (show())      │
+     │ ─────────────────────── │     │ ─────────────────────── │
+     │ • Has native ::backdrop │     │ • No backdrop           │
+     │ • Focus is trapped      │     │ • Background stays active│
+     │ • Native Escape close   │     │ • Manual Esc close needed│
+     └─────────────────────────┘     └─────────────────────────┘
+```
+
+* **Modal (via `dialog.showModal()`)**: Best for focused tasks. The browser automatically traps focus and closes the dialog on `Escape` keypresses without any custom JavaScript.
+* **Non-Modal (via `dialog.show()`)**: Best for floating utility panels. Keeps the underlying document fully editable while the user rearranges or drags the dialog. **Note**: Modeless dialogs do not natively close on `Escape`; a minor `keydown` listener must be added manually (see the Variation section below).
```
LeaVerou: Most of this is about dialog UX, and should move to the dialog guide (if not already there). The only relevant parts here are those around dragging. But from the perspective of this guide, we already _have_ a modal or non-modal dialog and we are implementing dragging on it, we are not trying to decide what paradigm to follow.

Also, while dialogs are a common use case, the guide is meant to be about dragging **in general**.

guides/ui-behaviors/drag/guide.md
```diff
@@ -4,3 +4,252 @@ description: Support dragging to move a certain overlay element (e.g. a dialog)
...
+2. **Override Default Centering**: You **MUST** override the default browser centering style (`margin: auto`) with `margin: 0` in CSS so manual `style.left` and `style.top` coordinate changes are respected without layout conflicts.
+3. **Designate a Keyboard-Focusable Handle**: Apply `user-select: none` to your header handle to prevent text-selection highlights. Add `tabindex="0"` to the handle and listen for arrow keys to nudge the dialog.
+4. **Track Coordinated Offsets**: Store the offset delta between the cursor's coordinates and the dialog's top-left corner on grab start to prevent the dialog from "jumping" on click.
+5. **Bind Listeners to the Document**: Attach `pointermove` and `pointerup` event listeners to the `document` rather than the drag handle itself. This keeps the drag continuous if the user moves their mouse rapidly.
+6. **Enforce Viewport Boundaries**: Clamp coordinates during movement against the window bounds to keep the dialog fully on-screen.
+
+## Standard Modal Code Implementation
+
+This is the standard, 100% complete implementation of a draggable **Modal Dialog**.
```
LeaVerou: Same comment as for the drag-and-drop guide. We do not need a complete implementation. That just wastes tokens telling agents things they already know. We _only_ need the delta from what agents would have done by default.

guides/ui-behaviors/drag/guide.md
```diff
@@ -4,3 +4,252 @@ description: Support dragging to move a certain overlay element (e.g. a dialog)
...
+* **DO** ensure modeless overlays can be closed via keyboard. Add manual listeners to handle the `Escape` key on modeless (non-modal) dialogs, as browsers do not provide native Escape closing for modeless elements.
+* **DO** calculate cursor offsets relative to the element's top-left corner on grab start. This prevents the element from snapping or jumping suddenly to align its top-left corner directly under the pointer.
+* **DO** attach `pointermove` and `pointerup` event listeners to the `document` or `window`. This ensures the drag sequence remains continuous even if the user moves their mouse rapidly and temporarily slides off the drag handle.
+* **DO** perform strict viewport bounds-checking. Clamping coordinates with `Math.max` and `Math.min` ensures elements cannot be accidentally dragged off-screen.
+* **DO** declare `touch-action: none` on the draggable container. This tells the browser on mobile devices to suppress standard scrolling or pinch-to-zoom gestures that would otherwise compete with dragging.
+* **DO** treat dragging as a progressive enhancement. If JavaScript is disabled or fails to load, ensure the element remains centered, visible, readable, and functional.
```
LeaVerou: This is use case dependent. In some cases dragging can be PE, in others it's crucial to the experience (e.g. a game).

guides/ui-behaviors/drag/guide.md
```diff
@@ -4,3 +4,252 @@ description: Support dragging to move a certain overlay element (e.g. a dialog)
...
+// Add this under the accessibility keydown listener block:
+dialog.addEventListener('keydown', (e) => {
+  if (e.key === 'Escape' && dialog.open && !dialog.matches(':modal')) {
+    e.preventDefault();
+    dialog.close();
+  }
+});
+```
+
+## Best Practices
```
LeaVerou: Most of these are great! That's exactly the type of guidance needed.

LeaVerou (changes requested):
See individual comments. On a high level:
- This should not be so narrowly focused on dialogs. It's a guide about implementing dragging **in general**. It should _include_ the relevant bits for dragging dialogs, but as a "and if by the element you're dragging happens to be a dialog…", not as the sole use case.
- Same as the dnd guide: we do not need a complete implementation, just the delta. The exercise I suggested in the dnd guide would help here too. What do agents do? How would you correct it?

sturobson:
I've updated this following the feedback.

guides/ui-behaviors/drag/guide.md
```diff
@@ -1,6 +1,76 @@
...
+Do not apply `role="application"` to the handle. It unnecessarily changes assistive-technology interaction.
+
+Apply these styles to the handle, not to the entire draggable element, so text outside the handle remains selectable and touch gestures outside it keep their usual behavior:
+
+```css
+.drag-handle {
+  cursor: grab;
+  touch-action: none; /* Prevent a touch drag from becoming page scrolling. */
+  user-select: none;
+  -webkit-user-select: none;
```
LeaVerou: Prefixed property should come before the non-prefixed one

guides/ui-behaviors/drag/guide.md
```diff
@@ -1,6 +1,76 @@
 ---
 name: drag
-description: Support dragging to move a certain overlay element (e.g. a dialog) around the page.
+description: Support dragging to move a floating element around the page.
 web-feature-ids:
   - user-select
```
LeaVerou:
```suggestion
  - pointer-events-api
  - touch-action
  - user-select
```

guides/ui-behaviors/drag/guide.md
```diff
@@ -1,6 +1,76 @@
...
+
+## Provide keyboard movement
+
+When the handle has focus, move the element with the arrow keys and call `preventDefault()` for those keys. Use the same clamped positioning helper as pointer movement. A 15px increment is a reasonable starting point; adapt it to the component's size and precision needs.
+
+## Dialog-specific note
+
+A `<dialog>` can use this mechanism with its header as the handle. Set `margin: 0` before assigning `left` and `top`, because the browser's default dialog centering margins interfere with manual positioning. For dialog opening, closing, focus, and dismissal behavior, see {{ GUIDE_REF("declarative-dialog-popover-control") }} and {{ GUIDE_REF("platform-controls-dismiss-dialog") }}.
+
+## Decide on fallback behavior
```
LeaVerou: Nit: While MWG is not very consistent, we do tend to use a handful of specific headings for this, like `Fallbacks` or `Fallback Strategies` (we should probably be more consistent…)

guides/ui-behaviors/drag/guide.md
```diff
@@ -1,6 +1,76 @@
...
+
+Treat dragging as progressive enhancement when it only repositions a component: without JavaScript, keep the element, its content, and its controls visible, readable, and functional in a sensible default position.
```
LeaVerou: I would perhaps add to not show a handle if JS doesn't run

guides/ui-behaviors/drag/guide.md
```diff
@@ -1,6 +1,76 @@
...
+
+When dragging is essential to the experience, such as moving a game piece, do not present a non-draggable fallback as equivalent behavior. Ensure the required pointer and keyboard interactions are available instead.
```
LeaVerou: In which case does this apply? If no JS, pointer & keyboard interactions won't work either. And for any features here that are not BWA there are fallbacks to use. So in what case would dragging not be available entirely?

LeaVerou (changes requested):
I just asked some of the cheapest models (ChatGPT free tier, Haiku, Gemini Flash-Lite) to "write me some code to drag an element on a web page". All of them knew:
- `cursor: grab` / `grabbing`
- `user-select: none`
- Recording the pointer offset on grab so the element doesn't jump
- Keeping the drag alive after the pointer leaves the element

Now this doesn't mean we should remove these from the guide entirely.
But I'd condense it *a lot*, possibly even 1-2 bullets.

What at least one of them didn't do (and thus, yay that we have it 🎉):
- Use pointer events, not mouse events
- `setPointerCapture()`
- `touch-action: none`
- Use a dedicated handle instead of making the whole element draggable
- Don't start a drag from interactive descendants
- Keyboard movement
- Clamp to the viewport
- `position: fixed` (in `expectations.md` but not in the guide)
- `-webkit-user-select` and not just `user-select`
- `margin: 0` on `<dialog>`

Here is the complete code they produced: https://gist.github.com/LeaVerou/93ade27b8f584a9d0269a2f19dc4da9d

## Fixes

- A right-click (or any non-primary button) must never start a drag. It does in the demo today: check `event.button === 0` on `pointerdown`.
- the demo calls `preventDefault()` on `pointerdown` but the guide doesn't. I'd drop it from the demo; it also blocks click-to-focus on the handle.
- `margin: 0` also applies to `[popover]` as it has the same UA styles

## Potential improvements

These are not blockers, but rather suggestions for enhancement.

### Improve UX: Drag threshold

Don't enter the dragging state until the pointer has moved a few pixels (~4px) from `pointerdown`. Otherwise every click on the handle shows `grabbing`, and in richer UIs a lot more than the cursor could change when a drag starts. Set a `.dragging` class only once the threshold is crossed, hang the cursor and any other drag styling off that, and remove it on `lostpointercapture`, which covers both `pointerup` and `pointercancel`.

### For heavy elements: `translate` during the drag?

Writing `left`/`top` on every `pointermove` relayouts the element and its contents each frame. That's fine for a dialog, but I wonder if for certain  layout-heavy cases it may be worth using `translate` while dragging (compositor only, no layout), then on release read `getBoundingClientRect()`, write `left`/`top`, and clear the `translate`.

Definitely not something to focus on, might be worth a bullet perhaps.

PR #1433, opened 2026-09-14
[#1225] drag-and-drop use case

sturobson:
addresses #1225


sturobson:
I've added some additional JS that can help keyboard users drag and drop, this might need some a11y testing.

LeaVerou (changes requested):
This is _excessively_ long and includes way _way_ too much code.

Remember the goal is not completeness. We are not trying to provide a JS library to the agent. The goal is to correct incorrect assumptions, using whatever means best communicates that (prose or short code snippets). 

I'd suggest stripping the demo back from the d&d behavior and asking all three agents to implement d&d on it. What did they get wrong? How would you correct them? The union of that is the guide.

It may also help to do the `drag` guide first, since this builds on it (I'll review that next).

sturobson:
> the goal is not completeness

I need this as an A1 poster for a reminder.

sturobson:
updated wit ha stripped back version

knowler (changes requested):
This approach works well in VoiceOver, but with NVDA it’s a little tricky since in order to use the arrow keys when the reordering mode is enabled, the user must know how to manually enter NVDA’s “forms mode” (i.e. <kbd>NVDA</kbd> + <kbd>space</kbd>), which passes the actual key presses to the page instead of using a virtual cursor (i.e. the virtual cursor is operated by the arrow keys). Typically NVDA does this automatically for relevant controls, but `<button>` is not one of those controls, so it’s not something every user would be acquainted with. I would expect JAWS to have the same issue.

It also adds steps for these users: once to get into the reordering mode, potentially another to get into forms mode, then the actual movement, then steps to get out of the reordering mode, another to potentially out of forms mode, and do the same for other elements.

The user also doesn’t know to use the arrow keys, so that extra instructions you’ve included are necessary and that takes up space (it is visually relevant, since a sighted keyboard user could need the instructions as well).

Instead we could use buttons for the individual movements (e.g. “move up” or “move down” and even other movements “move to top” or “move to bottom”). This would avoid extra steps for keyboard/screen reader users.

This approach comes with a trade-off that now there are extra buttons in the UI. Those buttons could be a part of an actions menu to help reduce clutter (though perhaps for the sake of keeping the guidance/demo simple, that could just be mentioned and not demonstrated).

sturobson:
Thanks for the feedback @knowler ❤️. I've taken a look and had a bit of a refactor. 

**Modified & Refactored**
- `guide.md` & `demo.html`: Shifted keyboard controls from a complex "grab-and-arrow" model to direct Up/Down buttons, moving task content first in the DOM for natural tab flow.
- `guide.md`: Added guidance on visual-to-DOM mismatches and the experimental CSS `reading-flow` property.
- `expectations.md`: Rewrote test criteria to target direct movement buttons, boundary states, focus retention, and live announcements.

**Added**
- CSS Icon Engine: Integrated `@property` `--icon-start` variables, `@container style(...)` queries, SVG variables, and `@supports` fallbacks into the demo stylesheet to eliminate HTML SVGs.
- Metadata: Added `reading-flow`, `masks`, `container-style-queries`, and `registered-custom-properties` to the guide's frontmatter.

**Removed**
- Boilerplate & Inline SVGs: Removed repetitive SVG paths from HTML and over 80 lines of auxiliary JS logic (grabbing state, cloning previews, and Escape-cancel logic) from the guide.


note: @LeaVerou this relies and is (I guess) blocked(?) by #1222 as I have used the icon technique here.

guides/ui-behaviors/drag-and-drop/guide.md
```diff
@@ -1,9 +1,133 @@
...
+
+# Rearrange Items with Drag and Drop
+
+Add reordering to an existing list, grid, or other collection. Reorder the real children in the DOM so visual, keyboard, and reading order stay aligned. For why CSS `order` is not a substitute for reordering interactive content, see {{ GUIDE_REF("css-layout") }}.
+
+For choosing a drag handle, preserving the pointer offset, and applying handle-scoped `touch-action` and `user-select`, see {{ GUIDE_REF("drag") }}.
+
+## Keep movement active while reordering
+
+A drag implementation that moves the dragged item in the DOM can lose pointer capture when the item is removed and reinserted. Register `pointermove`, `pointerup`, and `pointercancel` on `document` for the active drag, and remove them when it ends.
```
LeaVerou: This is a workaround for when `moveBefore()` is not supported.
Workarounds should go in the Fallback Strategies section. The core guidance is generally written assuming all relevant web features are supported.

guides/ui-behaviors/drag-and-drop/guide.md
```diff
@@ -1,9 +1,133 @@
...
+    }
+  }
+}
+```
+
+## Support keyboard reordering
+
+Avoid using a "grab-and-drag" keyboard mode (e.g., Space to grab, arrows to move). Screen readers like NVDA and JAWS operate in **Browse Mode** by default, using the arrow keys to navigate text. Because a standard `<button>` does not trigger automatic mode switching into Forms Mode, arrow key presses are intercepted by the screen reader and never reach the page.
+
+Instead, provide explicit, standard movement controls for each item. This completely avoids virtual cursor traps. To keep the UI clean, declare vector icons as CSS custom variables and use the **CSS-driven icon engine** to style your handles and controls; see {{ GUIDE_REF("icons") }}.
```
LeaVerou: How you create icon buttons is entirely orthogonal to the purpose of this guide. Just describe the desired UI.

LeaVerou: But also, are visible up/down buttons really necessary? I wonder if arrow keys when the handle is focused could work? Obviously then the handle wouldn't be `aria-hidden`, and would communicate its functionality to SR users.

Otherwise we'd end up needing a separate mode, because that's _way_ too much visual noise to always display by default. And modes have their own UX issues… https://en.wikipedia.org/wiki/Mode_(user_interface)#Avoid_when_possible

guides/ui-behaviors/drag-and-drop/guide.md
```diff
@@ -1,9 +1,133 @@
...
+
+### Align reading order with visual layout
+Structure your DOM row naturally so the focusable context (the label or input) precedes the action controls. Visually place your drag handle on the left, but mark it with `aria-hidden="true"` so keyboard and screen-reader users completely bypass it:
```
LeaVerou: Why on the left? Does that apply in RTL too?

guides/ui-behaviors/drag-and-drop/guide.md
```diff
@@ -1,9 +1,133 @@
...
   - drag
   - move-dom-element-without-losing-state
+  - icons
 ---
+
+# Rearrange Items with Drag and Drop
+
+Add reordering to an existing list, grid, or other collection. Reorder the real children in the DOM so visual, keyboard, and reading order stay aligned. For why CSS `order` is not a substitute for reordering interactive content, see {{ GUIDE_REF("css-layout") }}.
+
+For choosing a drag handle, preserving the pointer offset, and applying handle-scoped `touch-action` and `user-select`, see {{ GUIDE_REF("drag") }}.
```
LeaVerou: I'm not sure we can make a handle mandatory. There are cases where the best UI is indeed dragging the elements themselves. But we can definitely _recommend_ a handle!

guides/ui-behaviors/drag-and-drop/guide.md
```diff
@@ -1,9 +1,133 @@
...
+      <span class="icon" style="--icon: var(--icon-chevron-up);" aria-hidden="true"></span>
+    </button>
+    <button class="move-down-button" type="button" aria-label="Move Outline project down" title="Move down">
+      <span class="icon" style="--icon: var(--icon-chevron-down);" aria-hidden="true"></span>
+    </button>
+  </div>
+</li>
+```
+
+For complex layouts where visual and DOM order cannot naturally align, use the experimental **`reading-flow`** CSS property inside grid/flex containers to instruct the browser's tab-order to follow visual coordinates rather than DOM source code:
```
LeaVerou: Drop "experimental" which can bias agents, that's why we print out `BASELINE_STATUS` later, so the agent has the necessary info.

guides/ui-behaviors/drag-and-drop/guide.md
```diff
@@ -1,9 +1,133 @@
...
+  display: grid;
+  grid-template-columns: auto 1fr auto;
+  reading-flow: grid-visual; /* Syncs visual layout and tab flow */
+}
+```
+
+### Disable boundaries and safeguard focus
+Disable the "Move up" button on the first item and the "Move down" button on the last item. 
+
+When an item moves and causes a focused button to become disabled, the browser resets focus to the `body`. To prevent focus loss, programmatically shift focus to the sibling movement button of the moved item:
```
LeaVerou: I believe the code example below does not add anything the agent cannot infer by reading this sentence. It knows how to programmatically set focus to a sibling! If there's something in particular about this that you're worried it may get wrong, point _that_ out.

guides/ui-behaviors/drag-and-drop/guide.md
```diff
@@ -1,9 +1,133 @@
...
+    activeBtn?.focus();
+  }
+
+  announce(`${item.dataset.label} moved ${direction}.`);
+}
+```
+
+## Decide on fallback behavior
+
+When reordering is supplementary, keep the collection readable and usable in its default DOM order without JavaScript. When reordering is essential to the experience, provide the required pointer and keyboard interactions instead of presenting the static order as equivalent behavior.
```
LeaVerou: Same question I asked in the `drag` guide. When does the second sentence apply?

guides/ui-behaviors/drag-and-drop/guide.md
```diff
@@ -1,9 +1,133 @@
...
+  <div class="task-controls">
+    <button class="move-up-button" type="button" aria-label="Move Outline project up" title="Move up">
+      <span class="icon" style="--icon: var(--icon-chevron-up);" aria-hidden="true"></span>
+    </button>
+    <button class="move-down-button" type="button" aria-label="Move Outline project down" title="Move down">
+      <span class="icon" style="--icon: var(--icon-chevron-down);" aria-hidden="true"></span>
+    </button>
+  </div>
+</li>
+```
```
LeaVerou: Please correct me if I'm mistaken, but I believe this code snippet does not add anything that the prose hasn't already told the avent.

guides/ui-behaviors/drag-and-drop/guide.md
```diff
@@ -1,9 +1,133 @@
...
+
+For complex layouts where visual and DOM order cannot naturally align, use the experimental **`reading-flow`** CSS property inside grid/flex containers to instruct the browser's tab-order to follow visual coordinates rather than DOM source code:
+```css
+.task {
+  display: grid;
+  grid-template-columns: auto 1fr auto;
+  reading-flow: grid-visual; /* Syncs visual layout and tab flow */
```
LeaVerou: Not a valid value: https://drafts.csswg.org/css-display-4/#reading-flow

guides/ui-behaviors/drag-and-drop/demo.html
```diff
@@ -0,0 +1,368 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, navigator-flow=true, initial-scale=1.0">
```
LeaVerou: I cannot find anything on `navigator-flow=true`

LeaVerou (changes requested):
High-level comments:

- I left more specific comments, but overall: A big difference between technical writing for humans and technical writing for agents: humans need the same things explained multiple times and demonstrated in a number of different ways. Here, that would just burn tokens. We do not need complete code examples for things. Use (short) code snippets to demonstrate things that are best explained that way, not to further illustrate the same points already made in the prose or for completeness. Also related: reframe this as we are **course correcting**. We don't need to mention everything there is to do and handhold the agent every step of the way. We just need to mention enough to course correct them, and ideally no more than that (because it costs).
- The icon stuff is _completely_ out of scope. Just describe the intended UI, and the agent can create icon buttons however it creates icon buttons (which may involve looking up the icon guide when we have it, but this does not concern us here!). That's another reason to avoid long code blocks: they force you to show irrelevant detail, that the agent could inadvertently latch on, whereas prose keeps it high-level. 
- Might be worth mentioning that the clone is only necessary if we actually want to show a faded copy, which may not be the best UX for every case.

PR #1435, opened 2026-09-15
[#1244] menu use case

sturobson:
addresses #1244


sturobson:
I missed –

```markdown
<!-- Baseline best practices to be referenced by any guide that involves a popup menu (dropdown menus, context menus, overflow menus, etc)

  Include:                                                                                                             
    - subgrid for icons
    - proper keyboard nav
    - safe triangles for submenus
    - converting to drawer on small screens?
-->     
```

when working on this, so trying to address what I've missed here (safe triangles and narrow viewports)

LeaVerou (commented):
@sturobson would it be useful for me to review now or should I wait?

sturobson:
> @sturobson would it be useful for me to review now or should I wait?

I'm just trying to figure out how best to "workaround" (code or documentation) for a limitation I've found using popover and scrolling containers in webkit/safari. 

I guess, hold off for now? And I'll message / update when it's ready

--- 

In future -- would you rather I move the PR to draft if I find something that I missed / needs fixing or correcting?

sturobson:
@LeaVerou as what was commented on in [this PR](https://github.com/GoogleChrome/modern-web-guidance-src/pull/1433). 

I think this one might need a re-think and pairing down of information as I think I've gone a bit too deep.

sturobson:
Since the last commit, and following feedback, I have -

- Reduced the menu guide to be more focused guidance.
- Removed the narrow-viewport drawer. (U think the sub menu stuff might be better in its own guide?) 
- Removed quite a bit of code from the demo, and guide that was more faff than needed.
- Removed conflicting guidance around whether submenus.

- Retained all web features in the frontmatter and made each relevant to the use case:
  - Popover menus
  - CSS Anchor Positioning
  - semantic menu behavior
  - Subgrid command-row alignment
  - sticky positioning
  - scroll-state query stuck-state styling

- Added concise guidance for:
  - choosing menu semantics rather than navigation semantics;
  - opening and anchor-positioning a popover menu;
  - required arrow-key, Home, End, and Escape behavior for `role="menu"`;
  - opening, closing, and returning focus from nested submenus;
  - optional subgrid alignment for labels, shortcuts, and submenu indicators; and
  - optional sticky scroll-state visual enhancement.

- Added `GUIDE_REF` references to existing guidance for resilient context menus and nested dropdown behavior rather than repeating its broader positioning and fallback guidance.
- Reworked `expectations.md` to test observable command-menu outcomes, with optional and conditional expectations for the visual enhancements.
- Rebuilt `demo.html` as a small, working File command menu with a Preferences submenu, anchored popovers, keyboard navigation, focus return, subgrid-aligned rows, and a sticky scroll-state enhancement.

knowler (commented):
This is a very good use case for `focusgroup` which I think would be better than adding the roving tabindex behaviour ourselves and semantics ourselves. We can use Microsoft’s focusgroup polyfill as a fallback: https://github.com/microsoft/polyfills/tree/main/packages/focusgroup

See the explainer (the spec PR hasn’t quite landed): https://open-ui.org/components/scoped-focusgroup.explainer/

Probably more helpful, the behaviours: https://open-ui.org/components/scoped-focusgroup.explainer/#supported-behaviors

```html
<button type="button" popovertarget="actions" popovertargetaction="toggle">Actions</button>
<!-- this receives a minimum ARIA role of menu -->
<div id="actions" focusgroup="menu" popover="auto">
  <!-- these receive minimum ARIA roles of menuitem -->
  <button type="button">Save</button>
  <button type="button">Save duplicate</button>
  <button type="button">Delete</button>
  <button type="button">Share</button>
</div>
```

What this doesn’t handle is any sort of _selection_ behaviour. It just handles the focus related behaviour and setting the minimum semantics for that behaviour.

sturobson:
ok, fingers crossed for this. I've tested in MacOS on Safari and Chrome using VoiceOver. 

To accomodate the suggestion I've done some things - 

- HTML Refactored to nest `focusgroup`: Nested the `<div focusgroup="menu">` container inside the `<div popover="auto">` wrapper.
  - Why: Workaround for Chromium Bug #564673920 where popovers clobber focusgroup semantics.
- Roving Tabindex JS Replaced with `focusgroup` Event Handlers: Removed custom roving-tabindex navigation JS, leaving only opening, submenu transition, and dismissal event listeners.
  - Why: The native/polyfilled focusgroup now handles focus movement natively, but JS is still required to bridge trigger openings and focus transition.
- Tethering and Stuck State CSS Added: Added CSS anchor positioning for triggers/menus and container scroll-state queries for sticky headers.
  - Why: To automatically layout menus relative to buttons and add a shadow to the header only when stuck at the viewport top.
- Guide & Demos Synchronised: Updated code snippets in `guide.md` and both demos to feature complete code blocks.

bramus:
Some feedback when trying this out:

- When I `TAB` onto “File” and activate it using `SPACE` the menu opens. Then hitting  `↓` scrolls the page.
  - I would have expected the first item in the menulist to be auto-focused, and that hitting `↓``↑` would cycle through the options.
- When I use `↓` to cycle from “New file” to “Preferences” I can open the sub menu using `SPACE`
  - I would have expected that opening the menu would auto-focus the first option in the submenu.
  - I would have expected `→` to also open the menu
- When I use `↓` to cycle from “New file” to “Preferences” and then hit `ESC` to close the menu, the menu correctly close. When then re-activating “File” again (be it using `ENTER` or `SPACE`) and then hitting `TAB`, I see that “Preferences” is selected again
  - I would have expected that opening the menu would auto-focus the first option in the submenu
  - I would have expected that opening the menu would not remember the last focused option, but start from scratch again.

Other keyboard interactions (such as `ESC` closing one level at a time) map to what I would expect to happen.

I cannot speak for the A11Y part of this PR, but a quick look at the source tells me there is no usage of [`role=menu`](https://w3c.github.io/aria/#menu) nor [`role=menubar`](https://w3c.github.io/aria/#menubar) nor [`role=menuitem`](https://w3c.github.io/aria/#menuitem) … although it’s mentioned in this thread.

bramus (changes requested):
_(See previous comment)_

sturobson:
@bramus Thanks for taking a look and providing some great feedback ❤️

Some of the decisions that led to these "counter-intuitive keyboard behaviours" were a direct byproduct of experimenting with the new browser-native `focusgroup` attribute across browsers. 

A brief context on what happened, and how I'm trying to fix it:

 1. State Retention (Why the menu "remembers" its position)
 `focusgroup` seems to natively manage a "roving tabindex" under the hood. When the menu closed, the browser retained tabindex="0" on the last-focused item, causing focus to land back there upon re-open.

 2. Arrow Key Page Scrolling
 I initially avoided binding custom arrow key listeners to prevent clashing with native `focusgroup` behaviours, which let default page scrolling slip through.

 3. Submenu Focus Shifts and Race Conditions
 Moving focus to submenus synchronously caused VoiceOver to double-announce parent trigger states (collapsed/expanded) and child states. To bypass this, we left focus on the parent trigger, which led to the lack of auto-focus.
 
 4. ArrowRight/Left and Escape Mapping
 I agree completely. I'll map ArrowRight to open submenus (focusing the first item), ArrowLeft to close them (returning focus to parent), and Escape to close one level at a time.
 
For 1,2,3 I'll also see what can be done. 

Again, thanks for the helpful, constructive, thoughtful feedback

sturobson:
Addressed the keyboard and accessibility feedback, tested with VoiceOver, and documented the expectation that NVDA should provide equivalent results:

- Added explicit menubar, menu, and menuitem semantics. 
- Opening a menu with Space, Enter, ArrowDown, pointer activation, or ArrowRight now focuses its first enabled item.
- Added ArrowUp/ArrowDown cycling, including wrapping, without allowing the page to scroll.
- Added submenu focus reset, ArrowRight opening, and level-by-level Escape/ArrowLeft handling.
- Reopening a menu now starts at its first item rather than restoring the previously focused option.
- Removed redundant menu labels that caused unnecessary VoiceOver announcements.

I've also added a 'reasoning.md' document to the folder for guide authors of the future to understand architecture decisions and workarounds that may not be obvious by looking at the demos and guide.

bramus:
Thanks, seems much better now. One final nit: There is a significant delay between opening a menu and moving the focus to the first item of the opened menu. Ideally that should be instant.

bramus (changes requested):
_(See previous comment about removing the focus-move delay)_

As for the A11Y part: I am not the A11Y expert that can verify whether the A11Y is correctly approached here.

guides/ui-components/menu/guide.md
```diff
@@ -1,22 +1,209 @@
...
+```html
+<!-- Trigger button -->
+<button id="file-trigger" type="button" popovertarget="file-menu" popovertargetaction="toggle" aria-haspopup="menu">
+  File
+</button>
+
+<!-- Popover container -->
+<div id="file-menu" popover="auto" aria-labelledby="file-trigger">
+  <!-- WORKAROUND: Nest focusgroup inside popover to preserve semantics -->
+  <div focusgroup="menu" aria-label="File commands">
```
knowler: We can skip naming the popover since its context value would be redundant (i.e. the menu already has file in its name). Also, I think we should suggest using the name of the popover trigger as the name for the menu (probably read as “File menu” as context for the first menuitem that’s focused).

```suggestion
<div id="file-menu" popover="auto">
  <!-- WORKAROUND: Nest focusgroup inside popover to preserve semantics -->
  <div focusgroup="menu" aria-labelledby="file-trigger">
```

I need to test in screen readers, but I’d even be tempted to set `role=presentation` on the `popover` element to avoid any group semantics announcement. Ideally, once the clobbering bug gets fixed, the menu semantics will take precendance and we won’t need the nesting.

guides/ui-components/menu/basic-demo.html
```diff
@@ -0,0 +1,82 @@
...
+      }, focusAnnouncementDelay));
+    };
+    const closeSub = (returnFocus = true) => { subMenu.hidePopover(); subTrigger.setAttribute('aria-expanded', 'false'); if (returnFocus) subTrigger.focus(); };
+    const closeRoot = () => { closeSub(false); rootMenu.hidePopover(); rootTrigger.setAttribute('aria-expanded', 'false'); rootTrigger.focus(); };
+    const activate = item => { if (item === subTrigger) { subTrigger.setAttribute('aria-expanded', 'true'); subMenu.showPopover(); focusFirst(subMenu); } else { status.textContent = `${item.textContent.trim()} selected`; closeRoot(); } };
+    rootTrigger.addEventListener('keydown', event => { if (event.key === 'ArrowDown' || event.key === 'Enter' || event.key === ' ') { event.preventDefault(); if (!rootMenu.matches(':popover-open')) rootMenu.showPopover(); focusFirst(rootMenu); } });
+    [rootMenu, subMenu].forEach(menu => {
+      menu.addEventListener('toggle', () => { const open = menu.matches(':popover-open'); const trigger = menu === rootMenu ? rootTrigger : subTrigger; trigger.setAttribute('aria-expanded', String(open)); if (open) focusFirst(menu); });
```
knowler: Curious if `beforetoggle` might help with initial focus slowness? Might be too early.

knowler:
Accessibility-wise, `focusgroup` should apply the relevant ARIA roles for each pattern it supports on both the `focusgroup` bearing element and for the element with their focus mananged (unless something else clobbers it like `role` or as we’ve seen `popover` — so we should still be testing). Then the `focusgroup` itself would include the sufficient keyboard behaviour related to focus for those roles. Selection/activation is left to the author.

As far as the “remembering the last focused item” — that’s a `focusgroup` feature that can be disabled if we include the `nomemory` behaviour to the `focusgroup` attribute (i.e. `focusgroup="menu nomemory"`). There are similar features for wrapping and direction.

knowler (commented):
We don’t need to set `aria-expanded` on the popover triggers since [they already have that state implicitly](https://w3c.github.io/aria/html-aam/#att-command-popovers), though we might be able to use setting it explicitly when closed to our advantage because I don’t think we should prioritize the expanded/collapse state being announced, so I’m curious if we left `aria-expanded=false` long enough for the popover open and menuitem to be focused that when we removed `aria-expanded` it would avoid the announcement for the trigger?

Update: deferring it doesn’t seem to help. Really curious how Primer is doing it for their ActionMenu component.

Update 2: if we set `aria-hidden=true` and then in the next frame remove it, we can prevent the announcement on the button as the focus moves to the first item. I hate it, but it works.

knowler:
I think `popovertarget` and `popovertargetaction` are doing too much [under the hood accessibility-wise](https://developer.mozilla.org/docs/Web/API/Popover_API/Using#popover_accessibility_features) to be a good match here, which is a shame (likewise for `command=toggle-popover`). The implicit `aria-expanded` it sets can’t be overridden and it also sets `aria-details` implicitly which is not necessary for the pattern at hand.

I think we need to make sure the popover only gets triggered imperatively, so that it doesn’t get counted as the source. That way we don’t need to worry about the race condition for `aria-expanded` since it’s not being set by the browser (we will need to now manage setting it). We can just immediately move the focus to the first menu item.

PR #1437, opened 2026-09-15
fix(ci): make doc update commit before tagging publish version

connorjclark:
the tag check doesn't play nice with the commit that the publish job does. the job checks if HEAD has a tag, and if so does nothing. But then when a publish happens it tags HEAD, then it makes a new commit (for eval results). Next time it runs, it sees that commit has no tag, then does a useless publish.

this bug resulted in a useless publish: https://github.com/GoogleChrome/modern-web-guidance/commit/16ed22b43fa56157adadfb25f18a354197d497b9


paulirish (approved):

paulirish:
i have a revision i'll pitch to you.

PR #1439, opened 2026-09-15
ci: harden publish pipeline against empty releases and race conditions

paulirish:
### Net Impact

Eliminates empty version bump releases (such as `v0.0.189`), prevents split-brain state between `modern-web-guidance-src` and `modern-web-guidance`, and ensures nightly evals telemetry continuously updates without creating empty packages.

### Changes

- **Substantive Distribution Change Gate**: Replaces fragile `git describe --exact-match` (which failed scheduled runs with exit code 1) with `getExactDistributionDiff(latestTag, publishCliDir)` & `hasSubstantiveChanges()`. Scheduled runs with zero substantive changes exit cleanly with code 0 instead of triggering failure alert emails or cutting empty versions.
- **Decouple Evals & Documentation Updates**: Automatically commits and pushes updated `README.md` and benchmark telemetry (`eval-results-summary.json`) with `[skip ci]` on every run, even when a release is skipped.
- **Gate NPM Publishing**: Signals `published=true/false` via `$GITHUB_OUTPUT` and gates `npm publish --provenance`, preventing orphan npm releases on non-substantive runs.
- **Atomic Source Consensus**: Consolidates branch commit updates and tag creation into `git push --atomic origin HEAD:"$ref" v$newVersion`. The public distribution repository and npm are only touched after source consensus succeeds.
- **Cleanup & Tests**: Co-locates `hasSubstantiveChanges` in `release-notes-diff.ts` and adds unit tests in `generate-release-notes.test.ts`.


paulirish:
meh.

serving/skills-cli/publish-skills.ts
```diff
@@ -162,6 +157,52 @@ async function main() {
...
+  if (!isDryRun) {
+    // Always update README with features/evals for both distribution bundle and source repo
+    console.log('Checking automated documentation updates to source repo...');
+    updateReadmeWithFeaturesAndUseCases([ROOT_DIR, publishCliDir]);
+
+    let hasDocDiff = false;
+    try {
+      execSync('git diff --quiet README.md serving/skills-cli/eval-results-summary.json', { cwd: ROOT_DIR });
+    } catch {
+      hasDocDiff = true;
```
connorjclark, line 173: why eat the error and assume a diff?

if execSync fails to run the command I'd prefer the script to error.

serving/skills-cli/publish-skills.ts
```diff
@@ -203,27 +237,25 @@ async function main() {
...
 
-    // Update both the distribution bundle README and the source repo README
-    const { featuresCount, useCasesCount } = updateReadmeWithFeaturesAndUseCases([ROOT_DIR, publishCliDir]);
+    const { allFeatureIds, readyGuides } = getFeaturesAndUseCases();
+    const featuresCount = allFeatureIds.size;
+    const useCasesCount = readyGuides.length;
+
+    // 1. Tag locally on the source repository
+    console.log(`Creating local Git tag v${newVersion}...`);
+    execSync(`git tag v${newVersion}`, { stdio: 'inherit', cwd: ROOT_DIR });
```
connorjclark, line 246: if we used annotated tags, we could add a message containing the publish diff.

PR #1440, opened 2026-09-16
[#1231] spatial navigation 

sturobson:
addresses #1231

I've tried to be as focused as possible on this, keeping an eye on the feedback in this issue, hopefully it's nearly there rather than wildly off-the-mark


sturobson:
thought of another contained use case over lunch - the large 'image gallery' you sometimes get when expanding an image on amazon. I found myself trying to navigate with cursors keys

guides/ui-behaviors/spatial-navigation/guide.md
```diff
@@ -1,9 +1,188 @@
...
+  };
+
+  let best = null;
+  let minDistance = Infinity;
+  const weight = 2.5; // Favor on-axis targets over diagonal targets
+
+  for (const candidate of candidates) {
+    if (candidate === currentEl) continue;
+
+    const candidateRect = candidate.getBoundingClientRect();
```
bramus: This can be optimized by caching the bounding client rects. It would also require some clever strategy to reevaluate things when necessary … perhaps a `ResizeObserver` on the parent container?

bramus (commented):
I can’t speak for the A11Y parts, but testing it out I found the demo to be working as I would expect it to work.

Added one optimization remark.

guides/ui-behaviors/spatial-navigation/demo.html
```diff
@@ -0,0 +1,359 @@
...
+    // Visual 2D Distance Heuristic Algorithm
+    function findNextSpatialFocus(currentEl, direction, candidates) {
+      // Cache all candidate bounding client rects up front.
+      // Fetching rects sequentially in a single pass allows the browser to batch
+      // layout queries, completely eliminating potential layout thrashing during the loop.
+      const rects = new Map();
+      rects.set(currentEl, currentEl.getBoundingClientRect());
+      for (const cand of candidates) {
+        rects.set(cand, cand.getBoundingClientRect());
+      }
```
bramus, line 252: This will still re-get all bounding client rects every time `findNextSpatialFocus` is called. The `Map` should be stored globally.

bramus (changes requested):

sturobson:
Here's a summary of what's been done:

1. Persistent Global Cache with Event-driven Invalidation
2. High-Frequency Debouncing via requestAnimationFrame
3. Tightening & Streamlining the Guide

guides/ui-behaviors/spatial-navigation/demo.html
sturobson: hopefully I've got this sussed now - let me know if there's something I've completely missed :)

guides/ui-behaviors/spatial-navigation/guide.md
sturobson: this should be, hopefully, resolved now - let me know if there's something I've completed missed in the implementation :)

guides/ui-behaviors/spatial-navigation/guide.md
```diff
@@ -1,9 +1,181 @@
...
+
+To find the best target in a direction (`up`, `down`, `left`, `right`), filter candidates by their center coordinates, then rank them by combining **projection distance** (distance along the navigation axis) and **orthogonal distance** (distance along the perpendicular axis), applying a penalty weight ($w \ge 2$) on the orthogonal axis to favor on-axis movement.
+
+```javascript
+// Global cache for candidate bounding client rects to prevent layout thrashing
+let rectCache = null;
+let invalidationFrameId = null;
+
+function getRectCache(currentEl, candidates) {
+  if (!rectCache) {
```
bramus: This check seems a bit too blunt, as the cache might have _some_ entries, but not _all_ entries _(e.g. after a DOM mutation that inserted an extra item in the parent, without that action affecting layout)_

I’d rework this to check whether the `rectCache` has a certain entry stored or not. If not, it’d call `getCBR()`, store it, and return it.

bramus (changes requested):

guides/ui-behaviors/spatial-navigation/guide.md
```diff
@@ -1,9 +1,181 @@
 ---
 name: spatial-navigation
 description: Implement accessible spatial navigation using arrow keys that works correctly in different layout modes (grid etc)
-web-feature-ids: []
+web-feature-ids:
+  - scroll-into-view
```
LeaVerou: Would focusgroup help here? If so:

```suggestion
  - scroll-into-view
#  - focusgroup
```

(just to remember to get back to it)

guides/ui-behaviors/spatial-navigation/guide.md
```diff
@@ -1,9 +1,181 @@
...
+* All inactive items have `tabindex="-1"`.
+* When navigating, programmatically move focus by shifting the `tabindex="0"` attribute and calling `focus()`.
+
+---
+
+### Visual Candidate Selection
+
+In asymmetrical, wrapped, or grid layouts, index-based mapping fails. Instead, query live visual bounding boxes and calculate 2D distance.
+
+To find the best target in a direction (`up`, `down`, `left`, `right`), filter candidates by their center coordinates, then rank them by combining **projection distance** (distance along the navigation axis) and **orthogonal distance** (distance along the perpendicular axis), applying a penalty weight ($w \ge 2$) on the orthogonal axis to favor on-axis movement.
```
LeaVerou: Not sure inline LaTeX works, I'd just use <= or ≤

guides/ui-behaviors/spatial-navigation/guide.md
```diff
@@ -1,9 +1,181 @@
...
+  }
+  return best;
+}
+```
+
+---
+
+### Keypress Integration & Scroll Cooperation
+
+Your event router must prevent standard viewport scrolling **only** when a valid spatial focus transition occurs.
```
LeaVerou: Does the code example add something or just illustrates this sentence? If the latter, cut.

LeaVerou (changes requested):
Thanks for iterating on this!

## Algorithm bugs

I got Claude to run every arrow from every card in the demo (1024px viewport, the demo's own `findNextSpatialFocus`, spot-checked with real key presses) and we found a few bugs with the current algorithm.
`—` = no target, native scroll. Bold = wrong.

| From | Left | Right | Up | Down |
|---|---|---|---|---|
| 1-1 | **3-1** | 1-2 | — | 2-1 |
| 1-2 | 1-1 | 1-3 | — | **3-2** |
| 1-3 | 1-2 | **3-4** | — | 2-2 |
| 2-1 | **1-1** | 2-2 | 1-1 | 3-1 |
| 2-2 | 2-1 | **1-3** | 1-3 | 3-3 |
| 3-1 | — | 3-2 | **1-1** | — |
| 3-2 | 3-1 | 3-3 | 2-1 | — |
| 3-3 | 3-2 | 3-4 | 2-2 | — |
| 3-4 | 3-3 | — | **1-3** | — |

I did not go through the algorithm in detail, but that was its assessment on the main culprits and how to fix them, which seems reasonable:

> **1. The direction test compares centers.** Left from 1-1 has no same-row candidate, so 3-1 (two rows down, center 8px to the left) counts as "left" and wins. Same for Right from 1-3 → 3-4, Left from 2-1 → 1-1, Right from 2-2 → 1-3. A candidate should only count as being in a direction if its *near edge* is at or past the current element's *far edge* (e.g. for Right: `candidate.left >= current.right - tolerance`).
>
> **2. Weighted center distance skips rows.** Up from 3-1: card 2-1 sits directly above it, but it's wide so its center is 102px off. Card 1-1 is a whole row further but better centered, and wins by 4px. Same for Up from 3-4 → 1-3 and Down from 1-2 → 3-2. Fix: prefer candidates whose extent on the perpendicular axis *overlaps* the current element, and only fall back to the weighted distance when nothing overlaps. That's essentially what the CSS Spatial Navigation spec and Chrome's built-in spatnav do.

## Is the rect cache worth it?

The cache gets invalidated after every move that scrolls the page anyway, so whenever the widget is taller than the viewport, the cache rarely survives a keypress. And the staleness surface includes too many possible things to monitor them all.

Since keydown is a user-paced event, reading a handful of rects on each press is cheap _anyway_. But also, the browser only recomputes layout when something is dirty. If nothing has changed since the last layout, `getBoundingClientRect()` just reads geometry the engine already has. And when something *has* changed, our cache would be stale anyway. So the only work a JS cache can ever skip is the cheap read.

To test if my assumption is correct, I got Claude to measure this. In Chrome, with the demo page padded to ~500 cards, reading every rect with a clean layout took ~120 µs, i.e. ~0.25 µs per element, so ~2 µs for the demo's 9 cards. A forced relayout after a real mutation cost ~185 µs and is paid once per keypress regardless of how many rects are read, and the cache can't avoid it since it's stale at that point.

Also, principle 4 currently says "use live geometries" and "cache them" in the same bullet, which seem to be at odds.

## Describe the crux of the algorithm instead of providing a 90 loc code block

Instead of providing a long code block and expecting agents to extract the crux from it, we should _give them_ the crux directly — they can write the code!

That would also fix some of the current issues with the existing code, such as assuming a single widget per page, and could present tradeoffs in a much better format than code.
E.g. right now the code presents `2.5` as if it were a constant, but prose could explain when to tune it and why.

The core algorithm fits in a few bullets. Perhaps something like:

- **Candidates:** the focusable descendants of the widget container that pass `checkVisibility()`. Hidden elements return a zero rect at (0,0) and would otherwise become "up/left" targets.
- **Geometry:** read `getBoundingClientRect()` for the current element and all candidates when handling the keypress
- **Direction filter:** a candidate is in the pressed direction only if its near edge is at or past the current element's far edge on that axis (allow a few px tolerance). Never compare centers.
- **Aligned first:** among filtered candidates, prefer those whose extent on the perpendicular axis overlaps the current element's extent. Pick the one with the smallest gap along the navigation axis; break ties by smallest center offset.
- **Fallback scoring:** if no candidate overlaps, score each as `gap + w × perpendicularCenterOffset` with `w ≥ 2` (example value, tune per layout) and pick the minimum.
- **No candidate:** return nothing and do *not* `preventDefault()`, so the browser scrolls natively.
- **Candidate found:** `preventDefault()`, move the roving `tabindex`, call `focus()`, then `scrollIntoView({ block: 'nearest', inline: 'nearest' })`.
- **Don't intercept** when a modifier key is held (Alt/Cmd+Arrow are browser shortcuts), when the event was already handled (`e.defaultPrevented`), or when the innermost target consumes arrow keys itself (see below).
- Only act when the focused element is itself a candidate. If items contain their own focusable controls, either make those candidates too or leave them alone; there's no need to resolve a parent with `closest()`.
- **Scope the listener** to the container, not `document`, so multiple widgets can coexist.

### How to test what elements consume arrow keys?

It's not perfect, but perhaps something like:
- Use `e.composedPath()[0]` for the target, since `e.target` is retargeted to the host across shadow boundaries
- Test with `matches(':read-write, select, input[type=range], [role=slider]')`.
- Closed shadow roots can't be inspected, so content inside them can't be protected. :(


## On the Opt-out Switch for arrow keys

I'm no a11y SME, but I’m thinking that with roving tabindex only one item is tab-reachable. If users turn arrow navigation off how would they reach the rest of the widget?

I believe that in a composite widget (ARIA grid, listbox, toolbar) arrow keys are the expected interaction anyway, so a toggle would work against the pattern.

And I'm not sure WCAG asks for it either: [2.1.4](https://www.w3.org/WAI/WCAG21/Understanding/character-key-shortcuts.html) covers single-character keys, and arrows aren't character keys. The concern behind @knowler's comment, losing viewport scrolling, is already addressed by not calling `preventDefault()` when there's no candidate.

Unless I'm missing something here, I'd suggest dropping the "Customizable Settings & WCAG Compliance" section and keeping at most one line under the principles: avoid single-character shortcuts like h/j/k/l for spatial navigation.

## Demo

Btw, it would be helpful for testing if the demo included an example with a real CSS grid with `order`, `span`, and gaps (the case from the original issue), not just a flex container.

bramus:
> ## Is the rect cache worth it?
> The cache gets invalidated after every move that scrolls the page anyway, so whenever the widget is taller than the viewport, the cache rarely survives a keypress. And the staleness surface includes too many possible things to monitor them all.
> 
> Since keydown is a user-paced event, reading a handful of rects on each press is cheap _anyway_. But also, the browser only recomputes layout when something is dirty. If nothing has changed since the last layout, `getBoundingClientRect()` just reads geometry the engine already has. And when something _has_ changed, our cache would be stale anyway. So the only work a JS cache can ever skip is the cheap read.
> 
> To test if my assumption is correct, I got Claude to measure this. In Chrome, with the demo page padded to ~500 cards, reading every rect with a clean layout took ~120 µs, i.e. ~0.25 µs per element, so ~2 µs for the demo's 9 cards. A forced relayout after a real mutation cost ~185 µs and is paid once per

Great research!

PR #1445, opened 2026-09-16
feat(evals): add expectations for privacy discipline guide

TravenReese:
Adds the expectations.md test specification for the privacy discipline guide as a base for eval implementation


rowan-m (commented):
This is a valid list, but I want to better understand how this fits in the overall scheme of things. The point about data collection is good as a high-level guide for a whole system, but the points around social sharing are extremely specific. Is that expected here? Or should these all be at a similar thematic level?

TravenReese:
> This is a valid list, but I want to better understand how this fits in the overall scheme of things. The point about data collection is good as a high-level guide for a whole system, but the points around social sharing are extremely specific. Is that expected here? Or should these all be at a similar thematic level?

Makes sense. Since this is a discipline eval covering a broad area, I agree that keeping it at a higher thematic level works much better. It doesn't need to capture every specific detail from the guide, just the core practices generally applicable across most apps. I'll update the PR to broaden those points.

rowan-m (approved):
LGTM!

PR #1449, opened 2026-09-16
[#1086] JPEG XL

### LINKED ISSUE #1086: JPEG XL decoding support (image/jxl) in blink
Adds support for decoding JPEG XL (image/jxl) images in Blink using jxl-rs, a memory-safe pure Rust decoder.

  JPEG XL is a modern image format standardized as ISO/IEC 18181 that offers:
  - Progressive decoding for improved perceived loading performance
  - Support for wide color gamut, HDR, and high bit depth
  - Animation support

This implementation uses jxl-rs (https://github.com/libjxl/jxl-rs) instead of the C++ libjxl reference decoder to meet Chromium's memory safety requirements. The decoder is gated behind the enable-jxl-image-format flag and enable_jxl_decoder build flag.

CL: https://chromium-review.googlesource.com/c/chromium/src/+/7184969

---
**Web Feature ID**: jpegxl
**Chrome Releases**: Chrome 145

https://chromestatus.com/feature/5114042131808256

jamesnw:
I opted not to include JPEG XL in the `Comprehensive Responsive Image Component`, as putting it after webp would mean it would never be used, and it is unlikely we want to preference JPEG XL above webp (or avif).


guides/performance/performance/guide.md
```diff
@@ -192,7 +192,7 @@ Rendering involves Layout, Style, Paint, and Compositing calculations. CSS Conta
 Images typically represent the largest payload on a given web page. Optimization requires format negotiation, responsive sizing, and layout stabilization.
 
 ### DOs
-*   **DO serve modern formats (AVIF / WebP)**: Use the `<picture>` element to offer AVIF (best compression), falling back to WebP, and finally JPEG/PNG for legacy browsers.
+*   **DO serve modern formats (AVIF / WebP / JPEG XL)**: Use the `<picture>` element to offer AVIF (best compression), falling back to WebP, JPEG XL, and finally JPEG/PNG for legacy browsers.
```
tunetheweb: I think this wording may get some people's heckles up, while this is still under debate.

Maybe we should be less opinionated and go with something like this:

```suggestion
*   **DO serve modern formats (AVIF / WebP / JPEG XL)**: Use the `<picture>` element to offer AVIF or JPEG XL (best compression), falling back to WebP, and finally JPEG/PNG for legacy browsers.
```

tunetheweb (approved):
> I opted not to include JPEG XL in the `Comprehensive Responsive Image Component`, as putting it after webp would mean it would never be used, and it is unlikely we want to preference JPEG XL above webp (or avif).

I think that's fine. It's an example.

guides/performance/performance/guide.md
jamesnw: Good call- I went back and forth on that, and am happy to go with this.

PR #1452, opened 2026-09-16
feat(evals): add expectations for css-layout discipline guide

TravenReese:
Adds the expectations.md test specification for the css-layout discipline guide as a base for eval implementation


castastrophe (changes requested):
Sorry! Hit submit too soon. Review incoming...

guides/css/css-layout/expectations.md
```diff
@@ -0,0 +1,38 @@
...
+- The implementation MUST set `min-inline-size: 0` (or `min-width: 0`) on flex items that contain long unbreakable content such as URLs or code strings.
+- The implementation MUST NOT declare both `flex-basis` and `inline-size` / `width` on the same flex item.
+- The implementation MUST NOT use `order` or a `*-reverse` value of `flex-direction` to reorder interactive content away from its DOM order.
+- The implementation MUST use `display: grid` for two-dimensional layouts that define both rows and columns.
+- The implementation MUST use `grid-template-areas` with named regions for page-level shell layouts, and assign children to those regions via `grid-area`.
+- The implementation MUST use `repeat(auto-fit, minmax(<min>, 1fr))` or `repeat(auto-fill, minmax(<min>, 1fr))` for responsive card grids with an unknown item count.
+- The implementation MUST use `fr` units or `minmax()` for flexible track sizing instead of fixed pixel track widths.
+- The implementation MUST use `subgrid` (`grid-template-rows: subgrid` or `grid-template-columns: subgrid`) so that internal card elements align across sibling cards.
```
castastrophe:
```suggestion
- When aligning card elements across sibling cards that are laid out with grid, the implementation MUST use `subgrid` (`grid-template-rows: subgrid` or `grid-template-columns: subgrid`) so that internal card elements align.
```

guides/css/css-layout/expectations.md
```diff
@@ -0,0 +1,38 @@
...
+- The implementation MUST use safe alignment (such as `align-items: safe center`) on flex containers whose content can exceed the container size.
+- The implementation MUST set `min-inline-size: 0` (or `min-width: 0`) on flex items that contain long unbreakable content such as URLs or code strings.
+- The implementation MUST NOT declare both `flex-basis` and `inline-size` / `width` on the same flex item.
+- The implementation MUST NOT use `order` or a `*-reverse` value of `flex-direction` to reorder interactive content away from its DOM order.
+- The implementation MUST use `display: grid` for two-dimensional layouts that define both rows and columns.
+- The implementation MUST use `grid-template-areas` with named regions for page-level shell layouts, and assign children to those regions via `grid-area`.
+- The implementation MUST use `repeat(auto-fit, minmax(<min>, 1fr))` or `repeat(auto-fill, minmax(<min>, 1fr))` for responsive card grids with an unknown item count.
+- The implementation MUST use `fr` units or `minmax()` for flexible track sizing instead of fixed pixel track widths.
```
castastrophe: This does seem to contradict with the guide's example code which shows `grid-template-columns: 200px 1fr`. Parts of a track can be flexible without preventing other parts of it from being fixed.

guides/css/css-layout/expectations.md
```diff
@@ -0,0 +1,38 @@
...
+- The implementation MUST use `subgrid` (`grid-template-rows: subgrid` or `grid-template-columns: subgrid`) so that internal card elements align across sibling cards.
+- The implementation MUST declare an explicit `grid-template-rows` or `grid-template-columns` fallback immediately before the corresponding `subgrid` declaration in the same rule.
+- The implementation MUST NOT use `grid-auto-flow: dense` on grids containing interactive or focusable content.
```
castastrophe: This could potentially conflict with item 38 in that this says not to use dense on interactive content but masonries are often galleries with links. Something to consider.

guides/css/css-layout/expectations.md
```diff
@@ -0,0 +1,38 @@
+- The implementation MUST use `display: flex` for one-dimensional row or column arrangements such as navigation bars, toolbars, and action rows.
+- The implementation MUST use `gap` (or `row-gap` / `column-gap`) for spacing between flex and grid children instead of margins on the children.
+- The implementation MUST use the `flex` shorthand (e.g., `flex: 1 1 250px`) on flex items rather than declaring `flex-grow`, `flex-shrink`, and `flex-basis` separately.
+- The implementation MUST apply `flex-wrap: wrap` (or the `flex-flow` shorthand) on flex containers whose items can overflow the inline axis.
+- The implementation MUST use an auto margin (such as `margin-inline-start: auto`) to push a single flex item to the end of the main axis rather than `justify-self`.
+- The implementation MUST use safe alignment (such as `align-items: safe center`) on flex containers whose content can exceed the container size.
+- The implementation MUST set `min-inline-size: 0` (or `min-width: 0`) on flex items that contain long unbreakable content such as URLs or code strings.
+- The implementation MUST NOT declare both `flex-basis` and `inline-size` / `width` on the same flex item.
```
castastrophe: I think this guidance is too broad to be helpful. A common pattern is `flex: none; width: 200px;` for example and that sets the basis to auto by default. Maybe instead:

```suggestion
- The implementation MUST NOT declare both a non-auto `flex-basis` and `inline-size` / `width` on the same flex item.
```

guides/css/css-layout/expectations.md
```diff
@@ -0,0 +1,38 @@
...
+- The implementation MUST apply `scrollbar-gutter: stable` to scrollable containers to prevent layout shift when scrollbars appear.
+- The implementation MUST apply `overscroll-behavior: contain` (or `none`) to scrollable containers to prevent scroll chaining into the page.
+- The implementation MUST implement multi-line truncation using `display: -webkit-box` with `-webkit-box-orient: vertical` and `-webkit-line-clamp`, paired with the unprefixed `line-clamp` declaration.
+- The implementation MUST NOT use `overflow: hidden` to mask layout gaps caused by inline replaced elements, and MUST instead set a block display-outside value (such as `display: block`) on those elements.
+- The implementation MUST use `aspect-ratio` on media containers to reserve space and prevent Cumulative Layout Shift before assets load.
+- The implementation MUST use intrinsic sizing keywords (`min-content`, `max-content`, or `fit-content`) instead of fixed dimensions where content should determine the size.
+- The implementation MUST use CSS logical properties (`inline-size`, `block-size`, `margin-inline`, `padding-block`, `inset-inline-start`) for layout dimensions and spacing.
```
castastrophe:
```suggestion
- The implementation SHOULD prefer the use CSS logical properties (`inline-size`, `block-size`, `margin-inline`, `padding-block`, `inset-inline-start`) for layout dimensions and spacing.
```

We show several examples even here of using width/height so perhaps less strict wording here is beneficial?

guides/css/css-layout/expectations.md
```diff
@@ -0,0 +1,38 @@
...
+- The implementation MUST NOT invoke `.showModal()` on an element that also declares a `popover` attribute.
+- The implementation MUST tether overlays to their triggers using `anchor-name` on the trigger and `position-anchor` on the overlay rather than manual JavaScript coordinate math.
+- The implementation MUST use `position-area` or `anchor()` to place anchored overlays and `position-try-fallbacks` (such as `flip-block`) to reposition them on viewport overflow.
+- The implementation MUST feature-detect anchor positioning with `@supports (anchor-name: --x)` and provide an absolutely positioned fallback.
+- The implementation MUST use `overflow: auto` (not `overflow: scroll`) on scrollable regions so scrollbars appear only when content overflows.
+- The implementation MUST use `overflow: clip` rather than `overflow: hidden` when content should be clipped without creating a scroll container.
+- The implementation MUST apply `scrollbar-gutter: stable` to scrollable containers to prevent layout shift when scrollbars appear.
+- The implementation MUST apply `overscroll-behavior: contain` (or `none`) to scrollable containers to prevent scroll chaining into the page.
```
castastrophe: This might be too broad - scrollable containers could also apply to the main content scroller. Maybe we specify `nested scroll regions like sidebars, modals, and lists`?

guides/css/css-layout/expectations.md
```diff
@@ -0,0 +1,38 @@
...
+- The implementation MUST NOT use `overflow: hidden` to mask layout gaps caused by inline replaced elements, and MUST instead set a block display-outside value (such as `display: block`) on those elements.
+- The implementation MUST use `aspect-ratio` on media containers to reserve space and prevent Cumulative Layout Shift before assets load.
+- The implementation MUST use intrinsic sizing keywords (`min-content`, `max-content`, or `fit-content`) instead of fixed dimensions where content should determine the size.
+- The implementation MUST use CSS logical properties (`inline-size`, `block-size`, `margin-inline`, `padding-block`, `inset-inline-start`) for layout dimensions and spacing.
+- The implementation MUST use dynamic viewport units (`dvh`, `dvw`, `dvb`, or `dvi`) for full-height mobile layout containers instead of `vh`.
+- The implementation MUST NOT use `100vw` for full-width layout containers.
+- The implementation MUST implement masonry-style galleries using multi-column (`columns`) with `break-inside: avoid`, or grid with `grid-auto-flow: dense`, and MUST gate any `grid-template-rows: masonry` usage behind an `@supports` block.
```
castastrophe: The guide says the agreed name is `display: grid-lanes`, but this entry (and the guide's code sample as well) only put `grid-template-rows: masonry` behind @supports. Should we not also be gating on display: grid-lanes as well or instead?

guides/css/css-layout/expectations.md
```diff
@@ -0,0 +1,38 @@
...
+- The implementation MUST use `position-area` or `anchor()` to place anchored overlays and `position-try-fallbacks` (such as `flip-block`) to reposition them on viewport overflow.
+- The implementation MUST feature-detect anchor positioning with `@supports (anchor-name: --x)` and provide an absolutely positioned fallback.
+- The implementation MUST use `overflow: auto` (not `overflow: scroll`) on scrollable regions so scrollbars appear only when content overflows.
+- The implementation MUST use `overflow: clip` rather than `overflow: hidden` when content should be clipped without creating a scroll container.
+- The implementation MUST apply `scrollbar-gutter: stable` to scrollable containers to prevent layout shift when scrollbars appear.
+- The implementation MUST apply `overscroll-behavior: contain` (or `none`) to scrollable containers to prevent scroll chaining into the page.
+- The implementation MUST implement multi-line truncation using `display: -webkit-box` with `-webkit-box-orient: vertical` and `-webkit-line-clamp`, paired with the unprefixed `line-clamp` declaration.
```
castastrophe: This leaves off the overflow: hidden / clip requirements though doesn't it Without one of those, the extra lines still render, so an implementation could pass this test and visibly fail.

guides/css/css-layout/expectations.md
```diff
@@ -0,0 +1,38 @@
...
+- The implementation MUST NOT declare both `flex-basis` and `inline-size` / `width` on the same flex item.
+- The implementation MUST NOT use `order` or a `*-reverse` value of `flex-direction` to reorder interactive content away from its DOM order.
+- The implementation MUST use `display: grid` for two-dimensional layouts that define both rows and columns.
+- The implementation MUST use `grid-template-areas` with named regions for page-level shell layouts, and assign children to those regions via `grid-area`.
+- The implementation MUST use `repeat(auto-fit, minmax(<min>, 1fr))` or `repeat(auto-fill, minmax(<min>, 1fr))` for responsive card grids with an unknown item count.
+- The implementation MUST use `fr` units or `minmax()` for flexible track sizing instead of fixed pixel track widths.
+- The implementation MUST use `subgrid` (`grid-template-rows: subgrid` or `grid-template-columns: subgrid`) so that internal card elements align across sibling cards.
+- The implementation MUST declare an explicit `grid-template-rows` or `grid-template-columns` fallback immediately before the corresponding `subgrid` declaration in the same rule.
```
castastrophe:
```suggestion
- The implementation MUST declare an explicit `grid-template-rows` or `grid-template-columns` fallback before the corresponding `subgrid` declaration in the same rule.
```

I removed "immediately" because it's much stricter than the guide which just says it needs to be somewhere earlier.

Also, subgrid is baseline now so we might not need the fallback anymore anyway.

guides/css/css-layout/expectations.md
```diff
@@ -0,0 +1,38 @@
...
+- The implementation MUST establish a containment context with `container-type: inline-size` (or `size`) on a wrapper element before querying it.
+- The implementation MUST use `@container` queries for component-level responsive behavior instead of viewport `@media` queries.
+- The implementation MUST use container query units (`cqi`, `cqb`, `cqw`, `cqh`, `cqmin`, or `cqmax`) inside `clamp()` for typography or spacing that scales with the container.
+- The implementation MUST reserve `@media` queries for page-level layout and user preference conditions (such as `prefers-color-scheme` or `prefers-reduced-motion`).
+- The implementation MUST NOT use `block-size` as a `container-type` value.
+- The implementation MUST use the `popover` attribute for transient non-modal overlays and the native `<dialog>` element with `.showModal()` for modal interactions.
+- The implementation MUST NOT invoke `.showModal()` on an element that also declares a `popover` attribute.
+- The implementation MUST tether overlays to their triggers using `anchor-name` on the trigger and `position-anchor` on the overlay rather than manual JavaScript coordinate math.
```
castastrophe: The popover's invoker `popovertarget` can act as an implicit anchor. Chrome supports this; I haven't checked other browsers. A correct implementation using it would fail this anchor-name/position-anchor MUST however.

guides/css/css-layout/expectations.md
```diff
@@ -0,0 +1,38 @@
...
+- The implementation MUST use `overflow: auto` (not `overflow: scroll`) on scrollable regions so scrollbars appear only when content overflows.
+- The implementation MUST use `overflow: clip` rather than `overflow: hidden` when content should be clipped without creating a scroll container.
+- The implementation MUST apply `scrollbar-gutter: stable` to scrollable containers to prevent layout shift when scrollbars appear.
+- The implementation MUST apply `overscroll-behavior: contain` (or `none`) to scrollable containers to prevent scroll chaining into the page.
+- The implementation MUST implement multi-line truncation using `display: -webkit-box` with `-webkit-box-orient: vertical` and `-webkit-line-clamp`, paired with the unprefixed `line-clamp` declaration.
+- The implementation MUST NOT use `overflow: hidden` to mask layout gaps caused by inline replaced elements, and MUST instead set a block display-outside value (such as `display: block`) on those elements.
+- The implementation MUST use `aspect-ratio` on media containers to reserve space and prevent Cumulative Layout Shift before assets load.
```
castastrophe: width/height attributes on  also reserve space effectively without the aspect-ratio -- I believe this would reject a solution however that used that but not aspect-ratio.

guides/css/css-layout/expectations.md
```diff
@@ -0,0 +1,38 @@
...
+- The implementation MUST use intrinsic sizing keywords (`min-content`, `max-content`, or `fit-content`) instead of fixed dimensions where content should determine the size.
```
castastrophe: `where content should determine the size.` Can a grader apply this rule consistently or is it too subjective?

castastrophe (changes requested):
As this guide is rather hefty to begin with, I think the current format of this guide could cause issues. No single eval prompt will need subgrid, popover, <dialog>, anchor positioning, line-clamp, masonry and fluid type all at once. Graded as written, almost every run fails. My suggestion therefore would be to prefix each entry with the topic-specific item. A precedence was set for this in the `content-based-styling` guides ("If..." and "When...").

I also think we might be missing a few worth bullets for:
- container-type: size needs a definite block-size
- don't mix physical and logical keywords in position-area
- use subgrid on one axis only when the number of children varies
- auto-fit vs. auto-fill

micahjo7:
Thanks for the PR @TravenReese and for the review @castastrophe!

The suggestion to prefix each expectation sounds great. On the eval side, for these discipline guides, we will have to instruct the grader generation agent to generate test cases only for the applicable expectations (partial coverage is better than none). It's good for the expectations themselves to be comprehensive, though. In the future, we can add more apps to test against until we have full grader coverage of all the expectations.

guides/css/css-layout/expectations.md
TravenReese: Yeh I think this is bit to subjective looking back at it, so I'll remove this one.

LeaVerou (approved):
Idea: For discipline guides (and this includes the other PRs), we could write the expectations in such a way that if the guidance is not relevant, it will always pass. Then, instead of applying them to their own separate use cases, we could run them against all other existing use cases and no subsetting is necessary. Eventually they could even be used against arbitrary websites, as a modern web lighthouse of sorts 😁 

Many of them are already worded that way (by clarifying when they are relevant), but I've often seen that type of nuance get lost in translation (to the grader script).
In general, since these must be translated to deterministic unit tests, human-interpretable nuance would be lost. I saw some untestable ones in the other two PRs, I'll leave a comment there.

Additionally, the discipline guides largely contain design principles (i.e. rules of thumb) and I worry this turns them into prescriptive hard guidelines. 

But with those caveats, this LGTM!

PR #1461, opened 2026-09-16
chore: remove non-CLI serving approaches and the Jetski GUI agent

micahjo7:
## Summary

Two related cleanups that both amount to deleting code paths nothing uses anymore.

**1. Serving approaches.** Skills CLI is the only serving approach we ship, so the machinery for *selecting* between approaches (MCP server, raw Skills, megaskill) is dead weight.

**2. Jetski GUI agent.** The Puppeteer-driven IDE agent has been superseded by Jetski CLI.

Net: **39 files changed, +96 / −1671**.

> [!IMPORTANT]
> **Jetski CLI is untouched.** Only the GUI/IDE agent is removed. `JETSKI_CLI_BIN`, `JETSKI_MODEL`, `jetski-cli-agent.ts`, the `.gemini/jetski/skills` skill destination, and `gd dev`'s default agent all behave exactly as before.

## ⚠️ Action required for contributors

The root `config.ts` is gitignored, so this PR cannot update yours. The old `config.ts.example` instructed importing the now-deleted `Serving` enum:

```ts
import { Agents, Serving, type SuiteConfig } from './harness/config.ts';
```

If your local `config.ts` still has that import, **every `gd` command and `pnpm typecheck` will fail** after pulling this change:

```
SyntaxError: The requested module './harness/config.ts'
does not provide an export named 'Serving'
```

**Fix:** drop `Serving` from the import and delete the `serving:` property. See the updated `config.ts.example`.

A leftover `serving:` key with no import is harmless — it merges through inertly and is ignored.

## Serving approaches

- Delete `serving/mcp-server/` and the megaskill build path
- Remove the `Serving` enum, `SuiteConfig.serving`, `mcpServersToEnable`, `updateMcpConfig()`, `modernWebServerPath`, `mcpApiKey`, and the legacy `enableSkills` compat shim
- Drop the `@modelcontextprotocol/sdk` and `zod` dependencies
- Reports now emit a constant `serving: 'skills_cli'`
- Rename the root `build:mcp` script to `build:guides`

`skillsToEnable` is **intentionally retained** — that functionality stays as-is.

## Jetski GUI agent

- Delete `harness/agents/jetski-agent.ts` and the `Agents.JETSKI` enum member
- Remove `JETSKI_BIN` / `JETSKI_DEBUG_PORT` / `JETSKI_PROFILE_DIR`
- Remove the orphaned `jetskiDir` config and `JETSKI_DIR` env var, plus the dead `validatePaths()` that was its only reader
- Drop the direct `puppeteer-core` dependency from the harness (it remains transitively via `@google/gemini-cli-core`)
- Remove the Jetski button from the eval-view launch form

### Two behavioral changes worth a closer look

1. **Agent-script fallback.** Both resolvers in `run_suite.ts` now fall back to `gemini-cli-agent.ts` instead of `jetski-agent.ts`. This matches `defaultSuiteConfig.agent`, so an unrecognized agent string now lands on the documented default rather than the GUI agent.

2. **Concurrency.** The Jetski-only `--workspace-concurrency 1` override is gone, so *all* agents now honor `suiteConfig.workerCount`. No remaining agent forces serial execution.

## What was deliberately left alone

Historical dashboard rendering is preserved so archived GUI runs still display correctly:

| Location | Why kept |
|---|---|
| `landing.js` agent badge (`name.includes('jetski')`) | Keeps the ✦ badge on archived runs |
| `api.js` → `jetski_info.json` read path | Only the deleted agent ever *wrote* this file; the read is guarded by `try/catch` + a null check, so it degrades silently while still surfacing version metadata for old suites |
| `collect-eval-results.ts` / `investigate.ts` folder parsing | Parses GCS paths going back months |

## Testing

| Check | Result |
|---|---|
| `pnpm preflight` (build + typecheck + lint + tests) | ✅ exit 0, re-run on the final commit |
| `pnpm --filter eval-view run test:e2e` | ✅ 9 passed — removed DOM node broke nothing |
| Clean rebuild of `dist/skills-cli` | ✅ 143 use cases, all 3 esbuild bundles |
| Built CLI `search` / `list` / `retrieve` / `--help` / `--version` | ✅ correct output |
| `publish-skills --dry-run` (ci.yml:45) | ✅ generates release notes + benchmark table |
| `serving build` (ci.yml:52) | ✅ zero git drift in `serving/build` |
| `quick-smoke.ts jetski` | ✅ fails cleanly: `Unknown agent: jetski` |
| `git status` after build | ✅ clean |

Three `FULL=1` tests (`install-claude`, `install-gemini`, `install-skills`) fail on this branch — these are **pre-existing** and unrelated. Verified by stashing all work and reproducing byte-identical failures at clean `HEAD`. They spawn a real LLM and assert it *chose* to call the search tool, which is inherently non-deterministic. Notably `install-skills` fails on `searchCalled` while `skillActivated` passes, confirming the dist installs and loads correctly.


paulirish (approved):
Image

so jealous. Nice one.

PR #1463, opened 2026-09-17
[#1220] switch (checkbox) guide / use case

sturobson:
addresses #1220


guides/forms/switch/guide.md
```diff
@@ -1,11 +1,136 @@
...
+---
+
+## Implementation steps
+
+1.  **Declare the Checkbox Control**: Use a standard `<input type="checkbox">` as the semantic foundation for your switch toggle.
+2.  **Apply the Switch Attribute**: Add the standard `switch` attribute directly to the checkbox `<input>` element: `<input type="checkbox" switch>`. This opts-in to the browser's native switch rendering and gesture mechanics.
+3.  **Establish Accessible Labeling**: Nest the switch control inside a `<label>` element or link it using explicit `id` and `for` attributes. This ensures assistive technologies correctly map the switch toggle's description.
+4.  **Style with Accent Color**: Use the standard `accent-color` CSS property on the checkbox to style its active/track state to match your design tokens.
+5.  **Ensure Focus-Visible Contrast**: Define custom `:focus-visible` outline styles to provide keyboard navigation users with a distinct, high-contrast visual indicator when the switch is focused.
+6.  **Trigger Actions via Event Listeners**: Bind standard `change` event listeners to the input to handle state mutations immediately, and leverage native attributes like `checked` to query or restore the state.
```
LeaVerou: A lot of this is basically saying "use a standard checkbox in the standard ways". The only new bit is the `switch` attribute. The rest is (or should be!) covered by other guides about how to use checkboxes properly or is stuff agents already know. _Maybe_ a single sentence would be warranted that tells agents to use a regular checkbox, with regular labels, outline styles, etc.
We could also have another sentence to explain how `accent-color` works with native switches specifically.

guides/forms/switch/guide.md
```diff
@@ -1,11 +1,136 @@
...
+  const isEnabled = e.target.checked;
+  
+  // Trigger immediate UI state updates.
+  // This provides immediate feedback, matching the UX expectations of a toggle switch.
+  document.documentElement.dataset.theme = isEnabled ? 'dark' : 'light';
+});
+```
+
+For implementing a robust, production-ready website theme toggle with Storage state preservation (`localStorage`), system theme synchronization (`matchMedia`), and FOUC (Flash of Unstyled Content) prevention, see {{ GUIDE_REF("dark-mode") }}.
+
```
LeaVerou: This example is way, _way_ too long for the value it adds. A lot of it is basically explaining to agents how to use checkboxes and labels and there is very little in it that is specific to switches. I'd condense significantly to only the switch-specific stuff (agents don't need a full working example to grok things, they are not humans), or even remove it entirely.

LeaVerou (changes requested):
See comments.

But most importantly: The meat of this guide should be in the fallbacks section! It currently just recommends graceful degradation, which is not realistic for many (possibly most) projects. Switch is still Safari-only, so following this guidance right now if an agent user asks for a switch is almost guaranteed to piss off the user and possibly get MWG uninstalled. There are _very_ few cases where it is acceptable to show a checkbox to [85% of users](https://caniuse.com/wf-switch-control) where a switch was intended.

It needs to cover feature detection, and how you can load a stylesheet that polyfills it using regular checkbox styling (e.g. https://blog.tomayac.com/2026/01/12/a-polyfill-for-the-html-switch-element/ but there are others too, perhaps do some comparative research and see which one is best?).
Unfortunately feature detection is JS-only. I've done a lot of research into a CSS-only method and could not find anything.

And yes, mention PE for the cases where this can be PE. But not as the only solution!

PR #1464, opened 2026-09-17
docs(webmcp): update agentic-javascript-tools guide

Kulikowski:
## Summary

Updates [`guides/webmcp/agentic-javascript-tools/guide.md`](file:///usr/local/google/home/kaskulikowski/Documents/workspaces/google-workspaces/modern-web-guidance-src/guides/webmcp/agentic-javascript-tools/guide.md) to reflect recent WebMCP spec updates, clarifies the producer/consumer architecture, and documents recent breaking changes in Chromium versions 153 through 155.

## What Changed

- **Architecture clarity (Producer vs Consumer)**:
  - Clarified that web pages only need to call `document.modelContext.registerTool()` to expose tools. Pages do not need `getTools()` or `executeTool()`.
  - Documented that browser agents, browser extensions, DevTools, and system-level assistants discover and run tools out-of-band.
  - Documented the purpose of `getTools()` and `executeTool()` specifically for in-page JavaScript agents (chat widgets, `<iframe>`s) and automated test harnesses (Playwright, Vitest).

- **Cross-origin tool sharing (`exposedTo` & `fromOrigins`)**:
  - Added documentation and code examples for the `exposedTo` registration option.
  - Explained the mutual two-way handshake required to share tools between frames in the document tree (`exposedTo` on registration + `fromOrigins` on discovery).

- **Dynamic tool discovery**:
  - Added code examples for listening to the `toolchange` event on `document.modelContext` to track tools added or removed at runtime.

- **Breaking changes by Chromium version & migration fallbacks**:
  - **Chromium `153.0.8009.0`**: `execute(inputObject, { signal })` receives per-execution cancellation signal; aborting the registration signal no longer cancels active in-flight runs.
  - **Chromium `154.0.8017.0`**: Added `consequentialHint` to `ToolAnnotations` for high-stakes actions requiring user confirmation.
  - **Chromium `155.0.8051.0`**: `RegisteredTool.inputSchema` from `getTools()` returns a JavaScript object instead of a JSON string.
  - **Chromium `155.0.8052.0`**: `executeTool()` accepts an optional plain JavaScript object directly instead of requiring stringified JSON.
  - Added backwards-compatibility shims for older Chromium runtimes in the fallbacks section.


guides/webmcp/agentic-javascript-tools/guide.md
```diff
@@ -35,6 +42,34 @@ await document.modelContext.registerTool({
 controller.abort();
 ```
 
+Since Chrome `153.0.8009.0`, aborting the registration signal only unregisters the tool — it does not cancel executions that are already running. Tools that start long-running work must also honor the per-execution signal described below, or unregistering will leave orphaned requests and stale UI updates behind.
```
rviscomi, line 45: I see this in a few places, so this comment applies generally.

The exact Chrome version is probably TMI. This is an evergreen guide so agents should always treat it as the latest best practices without needing to know the change log.

The inverse might be more useful: discouraging what the old behavior was so that agents can migrate prior implementations.

guides/webmcp/agentic-javascript-tools/guide.md
```diff
@@ -118,8 +301,58 @@ export function createInventoryTool(inventoryManager) {
...
 
 ```javascript
-const modelContext = document.modelContext || navigator.modelContext;
-if (modelContext && 'registerTool' in modelContext) {
+if ('modelContext' in document) {
   // Register tools
 }
 ```
+
+### Supporting older Chromium versions
```
rviscomi, line 309: Rather than documenting the change log, I'd suggest embedding these defensive strategies directly into the code examples in the body of the guide. It seems like good practice to do in all cases, especially if there are high risks of runtime errors. WDYT?

rviscomi (changes requested):
Left a couple of suggestions.

@micahjo7 it also looks like the grader for this has drifted out of alignment with the expectations file. We should think about a better process for marking when eval files need to be regenerated.

micahjo7:
@rviscomi do the expectations need to change on this guide as well? Should new automation notify us that graders should be regenerated whenever the guide changes, expectations change, or both?

rviscomi:
> @rviscomi do the expectations need to change on this guide as well? Should new automation notify us that graders should be regenerated whenever the guide changes, expectations change, or both?

No expectation changes needed in this PR AFAICT. Notifying on changes to the expectations seems like a good first step.

guides/webmcp/agentic-javascript-tools/guide.md
Kulikowski: Fixed, please take a look

guides/webmcp/agentic-javascript-tools/guide.md
Kulikowski: Removed this from PR, we will have to remove some content when 153 is no longer Stable

beaufortfrancois (approved):

guides/webmcp/agentic-javascript-tools/guide.md
Kulikowski: Removed 153 mentions while keeping the guidance, 154 is going live soon

guides/webmcp/agentic-javascript-tools/guide.md
```diff
@@ -145,7 +145,7 @@ await document.modelContext.registerTool({
     },
     required: ["url"],
   },
-  execute: async ({ url, priority }, { signal } = {}) => {
+  execute: async ({ url, priority }, { signal }) => {
```
rviscomi, line 148: I'm curious why this was removed. Wasn't it a back-compat best practice to avoid TypeErrors?

beaufortfrancois: Chrome 153 has shipped ModelContextTool execute AbortSignal. We don't need this check anymore since we're targeting Chrome stable.

rviscomi: Oh I thought we we did want backwards compatibility, at least ensuring that we don't break experiences of users on versions older than 153. Or maybe I'm misunderstanding and this isn't actually a breaking change? It depends on whether the function is called with 1 or 2 parameters.

Kulikowski: We decided to keep the latest state, because coding agents can compare what-is-in-IDE vs what-is-in-guidance and make decisions based on this. "Migration" is also not multi-step complex workflow, it's basically doing a diff

Kulikowski: @rviscomi if that works from your perspective I would leave it as is. WDYT

rviscomi (approved):

PR #1472, opened 2026-09-17
Keep use cases with draft guides in "Needs guidance" status

rviscomi:
### Summary

When a guide has `draft: true` (or a truthy value like `draft: blocked`), it is withheld from publishing/distribution because the guidance is not yet completed.

This PR updates:
1. **`getStatusName()` in `lib/guide-validation.ts`**: Accepts `isDraft`. If truthy, the guide retains the `ProjectStatus.NeedsGuidance` status instead of prematurely advancing to `Needs evals`.
2. **`buildRequiredFilesChecklist()` in `guides/sync-use-cases.ts`**: Only marks `- [x] Use case guidance (guide.md)` when `inv.hasGuide && !inv.draft`.
3. **Unit Tests**: Adds test coverage for draft handling in `getStatusName` and `buildRequiredFilesChecklist`.

### Validation
- `pnpm test` passed across all workspace packages.
- `pnpm typecheck` and `pnpm lint` clean.


micahjo7:
To advance to "needs evals", should we also require the `expectations.md` to exist?

micahjo7 (approved):

rviscomi:
> To advance to "needs evals", should we also require the `expectations.md` to exist?

Yes, I think so. Updating

PR #1476, opened 2026-09-18
Clarify the status of Safari's bug in the IME-submit guide

rviscomi:
Safari 27 fixed the `isComposing` bug this week, so the BCD key is now "Newly Available" but the IME Safe Enter to Submit guide's fallback strategy section was still referring to the bug as if it's still occurring. I don't have evidence of it, but I suspect that it might confuse coding agents to see conflicting info about Safari supporting the BCD key while the prose describes it as an active issue in the present tense.

To avoid that confusion, I've updated how the guide talks about the bug so that it's in the past tense and clarified that the bug affected older versions of Safari.


rviscomi:
Low risk change, so merging this through for the 190 release.

PR #1477, opened 2026-09-18
Recommend the Baseline extension for VS Code to contributors

rviscomi:
(Disclaimer: this is my extension)

This PR adds `rviscomi.baseline-vscode` to VS Code's recommended extensions config.

https://open-vsx.org/extension/rviscomi/baseline-vscode
https://marketplace.visualstudio.com/items?itemName=rviscomi.baseline-vscode

Here are some of the features that would be useful to content contributors:

It shows the Baseline status of web features inline in the editor:

image

It provides additional feature info in a hovercard:

image

And it shows immediate feedback if the feature ID is invalid:

image

It also supports the `BASELINE_STATUS` and `FEATURE_FALLBACKS` macros:

image

You can also Cmd+Click on the feature ID to navigate to the feature's webstatus.dev page, or use the command palette to search for features by name.


rviscomi:
I can also imagine there being value in an even more repo-specific extension that also does things like expand GUIDE_REF or BASELINE_STATUS macros into their full text, so there are no surprises when it's rendered in the final md file. If we go that route, I'd remove the macro support from this extension.

Issue #1385, opened 2026-08-27
Adding "Using C++ on the web" guides under `guides/wasm`


sbc100:
See https://b/496092782

We would like to add guidance for how to use existing C++ code/libraries on the web.   There is a the initial outline of of skill added back in #506.   IIUC the next step would be to move that into the guides directory where it can be tested against the various agents.

I have developed several task over in personal repo: https://github.com/sbc100/emscripten-guidance-evals/tree/main/tests and have been evaluating them there. These are:

- audio-processor
- file-compressor
- fractal-flames
- image-transcoder

I am proposing to somehow add these under `guides/wasm/...`. 

A few open questions:

1. Can we have multiple tasks for the same guide?  (all these tasks requires essentially the same common guidance).
2. Is there some way to use the `empty-app` for my tasks?  `SUPPORTED_BASE_APPS` seems to be hardcoded to `['daily-grind', 'devtools-times']`, but these apps are maybe not good starting points the evals above.
3. These are long run complex task that need to download tools and source code as part of their execution.   Some of them can take up to 20m+.    Is this going to work in the current framework?  (Both the downloading part and the execution time part).

PR #1479, opened 2026-09-18
Add an initial C/C++ guide under guides/wasm/

sbc100:
This change deletes `skills-src/cpp-on-the-web` and replaces it with an
initial C/C++ usage guide.

The initial task I have created is to port and use the flam3 fractal
library on the page. Since generating the fractal can take a significant
amount of time the guided agent should run the WebAssembly program in a
background worker.

Graders generated using `gd dev guides/wasm/cpp-on-the-web --gen-grader`

See: #1385


micahjo7:
We can remove the grader.ts and task.md files that were generated.

sbc100:
> We can remove the grader.ts and task.md files that were generated.

The grader.ts was auto-generated but the task.md files were hand crafted.   Is that not allowed/recommeded?

guides/wasm/cpp-on-the-web/targets/daily-grind/task.md
```diff
@@ -0,0 +1 @@
+- Update index.html to include a fractal flame image along with a button to regenerate a new one at random.  Use the open-source C/C++ flam3 library (https://github.com/scottdraves/flam3) to generate these images client-side.
```
micahjo7, line 1: just want to make sure that the `flam3` isn't something that should be included in the guide itself. Do we need this here?

micahjo7:
@sbc100 all of our current tasks are generated (no author has really written them so far). But it doesn't hurt if you want to write them yourself. Left a comment on the task file.

guides/wasm/cpp-on-the-web/targets/daily-grind/task.md
sbc100: Exactly, flam3 should not be mentioned in the guide itself I believe.    The guide includes information about porting C/C++ libraries in general but I don't think it can/should mention any specific library such as this one, right?

sbc100:
> @sbc100 all of our current tasks are generated (no author has really written them so far). But it doesn't hurt if you want to write them yourself. Left a comment on the task file.

I think this reason that authoring the task makes sense is in this case is that we have a specific task in mind (i.e. porting a specific library to Wasm), or is the idea that we can encode all these specifics into the expectations.md and then leave the task as auto-generated?

I'm happy to give that a shot if that makes sense.   I would like to be specific about which information we do not want to include in the task if possible though.   For example, in this case I don't want the task to including anything about offloading work to a worker or avoiding blocking the main thread.  The idea is that the (guided) agent should do that unprompted.

micahjo7:
Yep, that makes sense! To give you some more context, here is the prompt we give to the agent for task generation. You can see that it keeps it very simple and tries to focus on guide activation through the guide's description only: https://github.com/GoogleChrome/modern-web-guidance-src/blob/main/guides/gd-dev-prompts.ts#L159.

I would say, if it is important that the `flam3` library is included in the prompt, then you should keep your tasks instead of letting the agent generate.

sbc100:
Should I still remove the grader?   I think the CI check told me that I had to include it (it wasn't part of my original upload)

micahjo7:
Ah I see, we do have a CI check which requires the grader file to be present if the task is also present. In that case, you can leave the graders. I will create an issue for follow up which will regenerate the evals after this is checked-in.

sbc100:
Thanks.  Since I wasn't able to do a full `gd dev` or `gd eval` locally I'm curious to see how well these agents to and how well the auto-generated grader for this works.   I suspect that this task might be a so much larger than other existing onces that we might exceed various limits of the system... 

I'm hoping I don't need to be in business of writing the grader myself, and that the auto-generated graders can actually judge the expectations here.

micahjo7 (approved):
Tasks LGTM from eng side. Will regenerate evals after check-in.

rviscomi (approved):

sbc100:
Thanks guys.  Lets see what happens when you try to re-generate the evals.  I suspect we will run into more issues, but we can solve them as they show up.    I know this task it a lot bigger / different than existing ones so I do expect some teething problems.

micahjo7 (approved):

PR #1480, opened 2026-09-19
Ignore renamed features in release notes if no change in browser support

rviscomi:
A dry run of the 0.190.0 release notes shows two entries for features that were renamed in the latest version of web-features, but haven't actually changed browser support:

>* **[Long animation frames performance entries](https://webstatus.dev/features/long-animation-frames)**: Updated browser engine support in [`identify-heavy-scripts`](https://github.com/GoogleChrome/modern-web-guidance/blob/v0.0.190/skills/modern-web-guidance/guides/performance/identify-heavy-scripts.md).
>* **[Page visibility state performance entries](https://webstatus.dev/features/page-visibility-state)**: Updated browser engine support across 2 guides ([`calculate-total-foreground-time`](https://github.com/GoogleChrome/modern-web-guidance/blob/v0.0.190/skills/modern-web-guidance/guides/performance/calculate-total-foreground-time.md), [`detect-initial-visibility-state`](https://github.com/GoogleChrome/modern-web-guidance/blob/v0.0.190/skills/modern-web-guidance/guides/performance/detect-initial-visibility-state.md)).

The fallback text used "Updated browser engine support" doesn't make sense here. These have been removed from the release notes since there's no impact on implementations.

Masonry was also renamed to grid-lanes, but it did have a change in browser support:

>* **[Grid lanes](https://webstatus.dev/features/grid-lanes)**: Added **Safari 26.4** support in [`css-layout`](https://github.com/GoogleChrome/modern-web-guidance/blob/v0.0.190/skills/modern-web-guidance/guides/css/css-layout.md).

This is ok to keep because the Safari support is new.


serving/skills-cli/release-notes-diff.ts
micahjo7, line 253: Renames still leak through for Newly/Widely features: step 1 never checks the removed line's status, so `-Baseline status for X: Widely available / +... for Y: Widely available`reads as a fresh promotion.

```suggestion
      // An upstream rename keeps the same status on both sides of the diff.
      const removedStatus = removedLines
        .map(l => l.match(/Baseline status for\s+.+?:\s*(Widely available|Newly available)/i))
        .find(Boolean)?.[1]
        ?.toLowerCase();
      if (removedStatus === status) break;

File: serving/skills-cli/release-notes-diff.ts [hunk 1]
```diff
@@ -286,52 +286,56 @@ export function parseBaselineUpdateFromHunk(guideName: string, hunk: string): Ba
...
-
-      for (const [key, info] of addedMap.entries()) {
-        if (!removedMap.has(key)) {
-          brandNew.push(info.raw);
-        } else if (removedMap.get(key)!.version !== info.version) {
-          versionUpdated.push(info.name);
-        }
+    if (!addedSupported) {
+      return null;
+    }
```
micahjo7, line 312: When a feature loses all support, Supported by: is replaced by "not natively supported by any major browser yet" — no added Supported by: line, so this early-returns and the regression goes unreported.

```suggestion
    if (!addedSupported) {
      if (!removedSupported) return null;
      const dropped = [...parseEngineMap(removedSupported).values()].map(i => i.name);
      if (dropped.length === 0) return null;
      return {
        featureName,
        featureId: resolveWebFeatureId(featureName),
        statusRank: 3,
        statusDescription: `Removed **${formatList(dropped)}** support`,
        guideName,
      };
    }

File: serving/skills-cli/release-notes-diff.ts [hunk 2]
```diff
@@ -244,14 +244,23 @@ export function parseBaselineUpdateFromHunk(guideName: string, hunk: string): Ba
...
     if (match) {
-      featureName = match[1].trim();
       const status = match[2].trim().toLowerCase();
+      // An upstream rename keeps the same status on both sides of the diff.
+      const removedStatus = removedLines
+        .map(l => l.match(/Baseline status for\s+.+?:\s*(Widely available|Newly available)/i))
+        .find(Boolean)?.[1]
+        ?.toLowerCase();
+      if (removedStatus === status) {
+        return null;
```
micahjo7, line 260: return null also drops renames that coincide with a real support change — formatStatusMessage emits Supported by: for Widely/Newly too, so this is reachable. That's the Masonry → grid-lanes case you called out as worth keeping. break falls through to the engine comparison instead, and step 2 re-extracts featureName. 

```suggestion
        break;

micahjo7 (approved):

serving/skills-cli/release-notes-diff.ts
rviscomi: Sorry not sure I understand. The Masonry rename coincided with a support change, but not a status change (it's still limited availability) so it wouldn't run in this code path.

The only way `return null` is executed is if the feature was Newly/Widely and still is Newly/Widely,  in which case it's impossible for there to be any change in browser support, by definition.

micahjo7: ah thanks, you're right. resolving

PR #1482, opened 2026-09-21
fix: describe Claude marketplace template

austdx:
Add the missing Claude marketplace description to the maintained distribution template. This fixes the validator warning at its source rather than editing the generated modern-web-guidance repository.

Validation: `claude plugin validate --strict serving/skills-cli/template/.claude-plugin/marketplace.json` passes without warnings; `git diff --check` passes. The distribution builder copies this template and updates plugin versions while preserving the description.

Validation is limited to this static metadata change. I did not run the full workspace preflight or acquire/run Gemini tooling. No generated distribution files are changed.


micahjo7 (approved):

micahjo7:
Thank you for the contribution!

